PR #31, opened 2021-06-15
fix typo in directory name

stv0g:


mlsmaycon (approved):

mlsmaycon:
Thanks for noticing and fixing it

PR #44, opened 2021-07-02
feature: basic management service implementation

braginini:


management/store.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Store represents an account storage
+type Store struct {
+	Accounts map[string]*Account
+
+	// mutex to synchronise Store read/write operations
+	mux       sync.Mutex `json:"-"`
+	storeFile string     `json:"-"`
+}
+
```
mlsmaycon, line 45: Missing doc

management/server.go
```diff
@@ -0,0 +1,37 @@
...
+func NewServer(dataDir string) (*Server, error) {
+	store, err := NewStore(dataDir)
+	if err != nil {
+		return nil, err
+	}
+	return &Server{
+		Store: store,
+	}, nil
+}
+
```
mlsmaycon, line 24: Missing doc

management/server.go
```diff
@@ -0,0 +1,37 @@
...
+func (s *Server) RegisterPeer(ctx context.Context, req *proto.RegisterPeerRequest) (*proto.RegisterPeerResponse, error) {
+
+	err := s.Store.AddPeer(req.SetupKey, req.Key)
+	if err != nil {
+		return &proto.RegisterPeerResponse{}, status.Errorf(404, "provided setup key doesn't exists")
+	}
+
+	return &proto.RegisterPeerResponse{}, nil
+}
+
```
mlsmaycon, line 35: Missing doc

cmd/management.go
```diff
@@ -0,0 +1,61 @@
...
+			SetupCloseHandler()
+			<-stopCh
+			log.Println("Receive signal to stop running Management server")
+		},
+	}
+)
+
+func init() {
+	mgmtCmd.PersistentFlags().IntVar(&mgmtPort, "port", 33073, "Server port to listen on (e.g. 33073)")
+	mgmtCmd.PersistentFlags().StringVar(&mgmtDataDir, "datadir", "/data", "Server data directory location (e.g. /data")
```
mlsmaycon, line 59: /data works well for docker. But account for the possibility of having the server running on non-docker env, we can set the default to /var/lib/wiretrustee. Or maybe follow the config check we got in the client to account for Windows systems, see https://github.com/wiretrustee/wiretrustee/blob/main/cmd/root.go#L34

management/store.go
mlsmaycon: We are also missing handling file creation/existance here or at the command level

mlsmaycon (changes requested):

cmd/management.go
mlsmaycon: As the data dir is an initial implementation, we will skip moving to a different folder for now.

mlsmaycon (approved):

Issue #40, opened 2021-07-01
Set up basic peer-management gRpc server


braginini:
a separate management gRpc server with a basic protocol definition and a health check endpoint

- [x] healthcheck endpoint
- [x] healthcheck test
- [x] protobuf servcie descption 
- [x] command in cmd/ to start the server (similar to signal)
- [x] docker image

Issue #39, opened 2021-07-01
Peer registration endpoint (simplified version)


braginini:
Endpoint to register a  new peer.

The request should contain:

- [ ] Protocol version
- [x] Peer Wireguard Public Key
- [x] Setup key (pre-authorized key, the one generated on the server)
- [x] local file storage

Please, consider the extension of the protocol (new fields, encryption, etc)

PR #47, opened 2021-07-09
adding peer add endpoint

casonadams:
I would like to add something like this.  I am not sure the best approach.  It would be nice to have an option to make the peer get written to the config also.

This command would then add a peer while the application is running, not requiring a restart.
```bash
curl -d '{"WgPubKey": "pubkey", "WgAllowedIps": "fd00::0/8"}' http://localhost:7777/peer
```

### DIFF
--- a/connection/engine.go
+++ b/connection/engine.go
@@ -1,7 +1,12 @@
 package connection
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+
 	"github.com/cenkalti/backoff/v4"
 	ice "github.com/pion/ice/v2"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


mlsmaycon:
Hi @casonadams , thanks for opening the PR. 

We are working on adding a management layer to the system to improve peer management. Part of this layer will be to improve this aspect of adding new peers without needing to restart the service, but the way we invasion that will be different than what you proposed as it won't introduce a point of concern that is having a local HTTP endpoint on each client, which can increase the complexity of managing many peers as it might introduce security risk and the system is not managing firewalls and ACLs. Our initial idea is to keep file support without the need to restart and introduce a management system.

You can follow #39, #40 and #44 to see progress on that.

casonadams:
@mlsmaycon sounds great!  Any "first issues" I can take a stab at here?  I'd love to help this project along.

braginini:
hey @casonadams 
You could have a look at this one https://github.com/wiretrustee/wiretrustee/issues/43
We are enabling TLS for the gRpc services and Signal is one of them.
Yesterday we finished implementing Letsencrypt for the Management service (see https://github.com/wiretrustee/wiretrustee/pull/52 ) so you can use it as an example.  

Thank you very much for being active :)

Issue #49, opened 2021-07-14
Support live peer list update


mlsmaycon:
Support live update of peers without requiring a restart. 

Relates to #19

PR #51, opened 2021-07-17
feat: Support live peer list update

mlsmaycon:
Support live update of peers without requiring a restart.

Closes to #49


connection/engine.go
```diff
@@ -125,7 +115,7 @@ func (e *Engine) InitializePeer(wgPort int, myKey wgtypes.Key, peer Peer) {
 func (e *Engine) ClosePeerConnection(peer Peer) error {
```
braginini: I think we should rename it to something like RemovePeer because we are not only closing the connection but also removing a peer completely
```suggestion
func (e *Engine) RemovePeer(peer Peer) error {
```

braginini:
Another point @mlsmaycon 
What happens in the following case:
1. connection to a peer is being established (not yet established, e.g. the other side is offline)
2. connection is blocked until timeout
3. someone calls ClosePeerConnection
4. will the attempt of establishing the connection (1) be terminated, or wait until timeout?

connection/engine.go
mlsmaycon: Renamed to RemovePeerConnection, removing Peer from a config file or management client will be the job of another process

mlsmaycon:
> Another point @mlsmaycon
> What happens in the following case:
> 
> 1. peer tries to connect to the other one (connection not yet established, e.g. the other side is offline)
> 2. connection is blocked until timeout or the other side responds
> 3. Engine.ClosePeerConnection is called
> 
> **Q:** will the attempt of establishing the connection (1) be terminated, or wait until timeout?

it will be terminated and wouldn't initiate a new one, but it was waiting until the timeout to terminate the go routine. So I've added another case in the Open function to terminate the function in case the ```conn.closeCond.C``` channel is closed.

braginini (approved):

PR #55, opened 2021-07-18
Extend Management to support peer changes distribution

braginini:
Implement a backbone of changes distribution

- [x] peer registration
- [x] send updates to other peers under the same account once a new peer has been registered

see:
https://github.com/wiretrustee/wiretrustee/pull/55#issuecomment-884054921


braginini:
I recommend not going through the changes (there many, including refactors) but rather checking the core logic and tests:
* service definition management/management.proto
* management/server.go (Sync and RegisterPeer methods)
* management/store.go (GetPeersForAPeer)
* management/management_test.go

andpar83 (approved):

PR #66, opened 2021-07-22
add config properties to the SyncResponse of the management gRpc service

braginini:
- [x] extend SyncResponse with Wiretrustee peer config properties management/management.proto
- [x] generate new gRpc code


management/proto/management.proto
```diff
@@ -31,13 +31,19 @@ message EncryptedMessage {
...
-message SyncRequest {
-
-}
+message SyncRequest {}
 
+// SyncResponse represents a state that should be applied to the local peer (e.g. Wiretrustee servers config as well as local peer and remote peers configs)
 message SyncResponse {
-  // A list of peers available
-  repeated string peers = 1;
+  // Global config
```
andpar83, line 38: Yes, I wanted to tell you that we need this config. Strange a little bit that it's in a SyncResponse, not in Register but it's ok too.

management/proto/management.proto
```diff
@@ -59,6 +63,47 @@ message ServerKeyResponse {
...
+message WiretrusteeConfig {
+  // a list of STUN servers
+  repeated HostConfig stuns = 1;
+  // a list of TURN servers
+  repeated HostConfig turns = 2;
+
+  // a Signal server config
+  HostConfig signal = 3;
+
+  // a Management server config
```
andpar83: Is it needed? Since client was able to request this data, it should know where management server lives

management/proto/management.proto
```diff
@@ -59,6 +63,47 @@ message ServerKeyResponse {
...
+  HostConfig management = 4;
+}
+
+// HostConfig describes connection properties of some server (e.g. STUN, TURN, Signal, Management)
+// User and Password are mostly used for TURN servers.
```
andpar83: To reduce confusion, I'd add 2 separate structs, like AnonymousHostConfig and LoggedInHostConfig (with better names, not these)

mlsmaycon: I think we will need a protocol type to describe the connection with TURN/STUN servers (TLS, DTLS or PLAIN)

management/proto/management.proto
braginini: In case we will need to change it.
Not likely to happen, also we could change via DNS.
I can remove it

management/proto/management.proto
braginini: Both done

management/proto/management.proto
```diff
@@ -59,6 +63,55 @@ message ServerKeyResponse {
...
+
+  // a Signal server config
+  HostConfig signal = 3;
+}
+
+// HostConfig describes connection properties of some server (e.g. STUN, Signal, Management)
+message HostConfig {
+  string host = 1;
+  int32 port = 2;
+  Protocol protocol = 5;
```
mlsmaycon: just a field id no following the sequence :)

mlsmaycon (approved):

PR #67, opened 2021-07-23
Move management server to a separate directory

andpar83:
It still requires build and documentation update.


management/impl/file_store.go
```diff
@@ -1,4 +1,4 @@
-package management
+package impl
```
braginini: @andpar83 
I'm not sure about having a separate package "impl" that has everything inside.
This would also mean that everything that is imported from impl package will have such an import path:
`import "github.com/wiretrustee/wiretrustee/management/impl/..."`

maybe we could have the following structure instead (see the key differences below the structure)? 

```
management
   ├── cmd
   │   ├── management.go
   │   └── root.go
   ├── proto
   │   ├── management_grpc.pb.go
   │   ├── management.pb.go
   │   └── management.proto
   ├── server
   │   ├── testdata
   │   │   └── store.json
   │   ├── management_suite_test.go
   │   ├── management_test.go
   │   └── server.go
   ├── store
   │   ├── file_store.go
   │   └── store.go
   ├── Dockerfile
   ├── main.go
   └── README.md

```

**The key differences are:**

-  a separate `cmd/ `package having all the command line operations inside.
Also in main.go it is clearly visible that only cmd is imported but not the whole `impl` package
`import cmd "github.com/wiretrustee/wiretrustee/management/cmd"`

- a separate `store/` package to keep everything related to the storage

- a separate `server/` package to keep the core files (server.go, main.go, etc), `testdata/` package 
-  main and Readme are at the root of the `management/` folder

I think we will also add a `client/` package later that will be imported by other parts of the repo (e.g. the Wiretrustee client app) 

What do u think @andpar83 
[patch.zip](https://github.com/wiretrustee/wiretrustee/files/6868611/patch.zip)

andpar83: Yes, this looks better. Thank you!

braginini (approved):

PR #71, opened 2021-07-25
Add client's interaction with management service

andpar83:
Also, added used creation on the server when seupkey is not provided to make it work somehow


cmd/up.go
```diff
@@ -28,6 +37,50 @@ var (
 				os.Exit(ExitSetupFailed)
 			}
 
+			log.Printf("Connecting to management server %s", config.ManagementAddr)
```
braginini: Not sure if it is too early for this PR and maybe you have this planned, but just a couple of ideas:

1. I think we could use `mgmCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)` to throw an error if it takes a long time to connect to the server (will be easy to catch with a retry later)
2. Detect whether the service has TLS enabled (probably take this from the config file) and create a client with credentials. The client creation of the service with TLS could look like this:

```
tlsConf := &tls.Config{} //will load system defaults
mgmCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
mgmConn, err := grpc.DialContext(  
		mgmCtx,  
		config.ManagementAddr,  
		grpc.WithTransportCredentials(credentials.NewTLS(tlsConf)),
		grpc.WithBlock(),  
		grpc.WithKeepaliveParams(keepalive.ClientParameters{  
			Time:    3 * time.Second,  
			Timeout: 2 * time.Second,  
		}))
```

andpar83: 1. Done.
2. Better create a separate PR for this.

braginini: Soudns good!

braginini (approved):

Issue #79, opened 2021-08-08
Self contained signal


mlsmaycon:
In order to ensure that different pieces of the code is self contained, we need to move signal client and server to their own folders
- [x] separate client and server
- [ ] move cmd to signal
- [ ] update builds

PR #80, opened 2021-08-08
Refactor: Move Signal server and client

mlsmaycon:
Moved Signal's server and client code to separated directories. 

Relates to #79


braginini:
What about cmd? Don't you wanna move it as well inside the project ? @mlsmaycon

braginini (approved):

mlsmaycon:
There will be another PR for the CMD

PR #81, opened 2021-08-11
peer management HTTP API

braginini:


management/server/account.go
```diff
@@ -129,20 +206,28 @@ func (manager *AccountManager) AddPeer(setupKey string, peerKey string) (*Peer,
...
 
-	accountId := uuid.New().String()
 	setupKeyId := uuid.New().String()
 	setupKeys := make(map[string]*SetupKey)
 	setupKey := &SetupKey{Key: setupKeyId}
 	setupKeys[setupKeyId] = setupKey
-	network := &Network{Id: uuid.New().String(), Net: net.IPNet{}, Dns: ""}
+	network := &Network{
+		Id:  uuid.New().String(),
+		Net: net.IPNet{IP: net.ParseIP("100.64.0.0"), Mask: net.IPMask{255, 192, 0, 0}},
```
mlsmaycon, line 220: This Netmask will allow for a larger range, but that may overlap lots of other 10.X networks, we can keep a /16 range (255.255.0.0)

This will allow:

HostMin:   10.64.0.1
HostMax:   10.64.255.254
Hosts/Net: 65534

management/server/http/middleware/jwtmiddleware.go
```diff
@@ -0,0 +1,237 @@
...
+		m.logf("  Error: No credentials found (CredentialsOptional=false)")
+		return fmt.Errorf(errorMsg)
+	}
+
+	// Now parse the token
+	parsedToken, err := jwt.Parse(token, m.Options.ValidationKeyGetter)
+
+	// Check if there was an error in parsing...
+	if err != nil {
+		m.logf("rrror parsing token: %v", err)
```
mlsmaycon: small typo here

management/server/account.go
braginini: I was guided by rfc6598:
https://datatracker.ietf.org/doc/html/rfc6598

See also description here:
https://en.m.wikipedia.org/wiki/Reserved_IP_addresses

It is a reserved address for service providers.
And we won't overlap with 10.x

mlsmaycon: You are right, I've overlooked the network address and thought that it was a 10.64, not a 100.64.

mlsmaycon (approved):

PR #82, opened 2021-08-11
Self contained signal cmd build

### LINKED ISSUE #79: Self contained signal

mlsmaycon:
- [x] move cmd to signal
 - [x]  update builds

Closes #79

### DIFF
--- a/.github/workflows/golang-test.yml
+++ b/.github/workflows/golang-test.yml
@@ -46,5 +46,13 @@ jobs:
       - name: Install modules
         run: go mod tidy
 
-      - name: run build
+      - name: run build cli
         run: GOOS=${{ matrix.os }} go build .
+
+      - name: run build management
+        run: GOOS=${{ matrix.os }} go build .
+        working-directory: management
+
+      - name: run build signal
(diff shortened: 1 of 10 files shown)


mlsmaycon:
We can consider renaming the subcommand signal to run or up to be less redundant. @braginini  @andpar83 what do you think?

`wiretrustee-signal run` or  `wiretrustee-signal up` either one looks better

braginini:
> We can consider renaming the subcommand signal to run or up to be less redundant. @braginini @andpar83 what do you think?
> 
> `wiretrustee-signal run` or `wiretrustee-signal up` either one looks better

Yeah, I think we should do what you suggest.
Same for the management.

But I think we should use these concepts to name cmds
https://github.com/spf13/cobra#concepts

mlsmaycon:
Ok, I will update it to run and open another PR for management command

braginini (approved):

PR #90, opened 2021-08-17
add wiretrustee LOGIN command

braginini:


client/cmd/login.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return loginResp, nil
+}
+
+// registerPeer checks whether setupKey was provided via cmd line and if not then it prompts user to enter a key.
+// Otherwise tries to register with the provided setupKey via command line.
+func registerPeer(serverPublicKey wgtypes.Key, client *mgm.Client, setupKey string) (*mgmProto.LoginResponse, error) {
+
+	var err error
+	if setupKey == "" {
```
mlsmaycon, line 108: it might be better to validate the setupKey format: https://pkg.go.dev/github.com/google/uuid#Parse

client/internal/config.go
```diff
@@ -24,11 +38,17 @@ func createNewConfig(managementURL string, configPath string) (*Config, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		config.ManagementURL = URL
 	} else {
-		config.ManagementURL = ManagementAddrDefault
+		config.ManagementURL = managementURLDefault
 	}
 
+	config.IFaceBlackList = []string{iface.WgInterfaceDefault, "tun0"}
```
mlsmaycon, line 50: didn't get the reason for hardcoding blacklisting of tun0?

braginini: Some users had issues when using OpenVPN alongside Wiretrustee. So I decided to exclude the default one. 
I think we could do a better job of discovering the interfaces we want for ICE without hardcoding but for now, should be fine. @mlsmaycon

mlsmaycon: Got it. thanks

mlsmaycon (approved):

PR #94, opened 2021-08-21
Extend peer http endpoint

braginini:
Add new HTTP API endpoints:
- [x] update peer
- [x] delete peer
- [x] get peer

### DIFF
--- a/management/server/http/server.go
+++ b/management/server/http/server.go
@@ -62,11 +62,10 @@ func (s *Server) Start() error {
 	peersHandler := handler.NewPeers(s.accountManager)
 	keysHandler := handler.NewSetupKeysHandler(s.accountManager)
 	r.HandleFunc("/api/peers", peersHandler.GetPeers).Methods("GET", "OPTIONS")
-	r.HandleFunc("/api/peers/{id}", peersHandler.HandlePeer).Methods("GET", "POST", "DELETE", "OPTIONS")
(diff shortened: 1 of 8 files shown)


management/server/http/server.go
```diff
@@ -62,10 +62,10 @@ func (s *Server) Start() error {
 	peersHandler := handler.NewPeers(s.accountManager)
 	keysHandler := handler.NewSetupKeysHandler(s.accountManager)
 	r.HandleFunc("/api/peers", peersHandler.GetPeers).Methods("GET", "OPTIONS")
+	r.HandleFunc("/api/peers/{id}", peersHandler.HandlePeer).Methods("GET", "PUT", "DELETE", "OPTIONS")
 
 	r.HandleFunc("/api/setup-keys", keysHandler.GetKeys).Methods("GET", "OPTIONS")
-	r.HandleFunc("/api/setup-keys", keysHandler.CreateKey).Methods("PUT")
-	r.HandleFunc("/api/setup-keys/{id}", keysHandler.HandleKey).Methods("GET", "POST", "OPTIONS")
+	r.HandleFunc("/api/setup-keys/{id}", keysHandler.HandleKey).Methods("GET", "PUT", "POST", "OPTIONS")
```
mlsmaycon: Just not sure about the behavior for the POST method with /api/setup-keys/{id}. Also, from the doc, we should use the method only for /api/setup-keys as we can't provide the {id} when creating the key and it is not in use in the code.

mlsmaycon (approved):

PR #112, opened 2021-09-06
feature: add logging to a file

braginini:


signal/cmd/root.go
```diff
@@ -52,11 +63,29 @@ func SetupCloseHandler() {
...
 // InitLog parses and sets log-level input
-func InitLog(logLevel string) {
+func InitLog(logLevel string, logPath string) {
 	level, err := log.ParseLevel(logLevel)
 	if err != nil {
 		log.Errorf("Failed parsing log-level %s: %s", logLevel, err)
 		os.Exit(ExitSetupFailed)
 	}
+
+	if logPath != "" {
```
mlsmaycon: Tested by adding stdout path as `--log-file /proc/self/fd/1`, but I got this warning when triggering the log rotation:

```time="2021-09-06T21:48:47Z" level=info msg="can't rename log file: rename /proc/self/fd/1 /proc/self/fd/1-2021-09-06T21-48-47.313: no such file or directory"```

Also tested passing an empty string with ("") and that works, but is not intuitive and for Docker, brings complexity to scape characters and avoid making "" nonempty but the file name.

For log file, makes sense to add a keyword that can be used to indicate console log outputs, allowing small changes to our Dockerfile. Maybe, even using `console` as --log-file would be an option.

File: signal/cmd/root.go [hunk 2]
```diff
@@ -33,10 +40,14 @@ func Execute() error {
 func init() {
 
 	stopCh = make(chan int)
+	defaultLogFile = "/var/log/wiretrustee/signal.log"
+	if runtime.GOOS == "windows" {
+		defaultLogFile = os.Getenv("PROGRAMDATA") + "\\Wiretrustee\\" + "signal.log"
+	}
 
 	rootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "info", "")
+	rootCmd.PersistentFlags().StringVar(&logFile, "log-file", defaultLogFile, "sets Wiretrustee log path")
```
mlsmaycon: Please update the Readme for Signal and Management with command output

signal/cmd/root.go
mlsmaycon: if you go with the key, we also need to update the docker-compose in the infrastructure_files

mlsmaycon (changes requested):

mlsmaycon (approved):
just comment the log volume in docker-compose and we are good to go

PR #144, opened 2021-11-01
fix: client app retry logic

braginini:


management/client/client.go
```diff
@@ -97,7 +97,7 @@ func (c *Client) Sync(msgHandler func(msg *proto.SyncResponse) error) error {
 			log.Errorf("failed to open Management Service stream: %s", err)
 			return err
 		}
-
+		backOff.Reset()
```
mlsmaycon: This won't have the desired effect if we block the connection to receive events for more than the max elapsed time

mlsmaycon: perhaps we should move/add it to line 106

signal/client/client.go
```diff
@@ -101,14 +101,19 @@ func (c *Client) Receive(msgHandler func(msg *proto.Message) error) {
...
-			err := c.connect(c.key.PublicKey().String(), msgHandler)
+			stream, err := c.connect(c.key.PublicKey().String())
 			if err != nil {
 				log.Warnf("disconnected from the Signal Exchange due to an error: %v", err)
 				c.connWg.Add(1)
 				return err
 			}
 
 			backOff.Reset()
+			err = c.receive(stream, msgHandler)
```
mlsmaycon, line 111: Same comment as for management client, if we reset things before blocking we risk facing the same issue. We can add backOff.Reset() in the error treatment as well

management/client/client.go
braginini: done

signal/client/client.go
braginini: done

mlsmaycon (approved):

PR #160, opened 2021-11-20
feature: add Wireguard preshared-key support

braginini:
See https://github.com/wiretrustee/wiretrustee/issues/139#issuecomment-952561984

### DIFF
--- a/client/cmd/login.go
+++ b/client/cmd/login.go
@@ -30,10 +30,9 @@ var (
 				return err
 			}
 
-			config, err := internal.GetConfig(managementURL, configPath)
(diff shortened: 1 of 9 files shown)


braginini:
`wiretrustee up --setup-key <SETUP KEY> --preshared-key <PRESHARED KEY>`

**Note:** Once set for a peer, all other peers will need it set as well in order to connect.

Can be also added manually to the config file `/etc/wiretrustee/config.json`, e.g.:
`"PreSharedKey": "iDglzWMp6quPoaQNnH2uqpljYHnjVpl5Jn1vYOWXNEo=",`

mlsmaycon (approved):
LGTM

PR #183, opened 2022-01-07
Fix concurrency on the client

braginini:


client/cmd/up_test.go
```diff
@@ -36,7 +36,7 @@ func TestUp_Start(t *testing.T) {
 
 func TestUp(t *testing.T) {
 
-	defer iface.Close()
+	defer iface.Close("wt0")
```
mlsmaycon, line 39: hardcoded wt0 will limit tests to non-macOS machines because macOS accepts utunX

client/internal/engine.go
```diff
@@ -112,6 +109,11 @@ func (e *Engine) Stop() error {
 // Connections to remote peers are not established here.
 // However, they will be established once an event with a list of peers to connect to will be received from Management Service
 func (e *Engine) Start() error {
+	/*go func() {
```
mlsmaycon: left over?

client/internal/engine_test.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, err
+	}
+
+	resp, err := mgmtClient.Register(*publicKey, setupKey)
+	if err != nil {
+		return nil, err
+	}
+
+	conf := &EngineConfig{
+		WgIface:      fmt.Sprintf("wt%d", i),
```
mlsmaycon: wt as iface name will fail on macOS, same comment as up test

mlsmaycon: you can do something like:

```GO
	if runtime.GOOS == "darwin" {
		ifaceName = fmt.Sprintf("utun1%d", i)
	} else {
		ifaceName = fmt.Sprintf("wt%d", i)
	}
```
`utun1` to avoid collision with some default interfaces

mlsmaycon (changes requested):

client/cmd/up_test.go
braginini: This value will be ignore for Mac and Windows.
Mac and Windows use tun shared variable of a userspace go.

This package has to be rewritten I think because it is misleading a bit

client/internal/engine.go
braginini: Yeah, will remove

client/internal/engine_test.go
braginini: Right, will change

mlsmaycon (approved):

PR #184, opened 2022-01-08
Link account id with the external user store

### LINKED ISSUE #173: Link account id with the external user store
When a user is created or is accessing the system for the first time after the initial multitenancy implementation, we should generate or retrieve its account ID and link it to the user id in the external access management provider.

mlsmaycon:
* idpmanager interface for managing external IDP providers
* auth0manager for managing Auth0 
* check if wt_account_id attribute was provided within the JWT token
* if no wt_account_id was provided, associate it to the user id in auth0
* new configuration options for idpmanager
* updated tests

closes #173


management/server/account.go
```diff
@@ -155,6 +155,42 @@ func (am *AccountManager) GetAccount(accountId string) (*Account, error) {
...
 }
 
+//GetAccountByUserOrAccountId look for an account by user or account Id, if no account is provided and
+// user id doesn't have an account associated with it, one account is created
+func (am *AccountManager) GetAccountByUserOrAccountId(userId, accountId string) (*Account, error) {
+	am.mux.Lock()
+	defer am.mux.Unlock()
+	switch {
+
+	case accountId != "":
```
braginini: I think it is better to use if instead of switch.

management/server/account.go
```diff
@@ -155,6 +155,42 @@ func (am *AccountManager) GetAccount(accountId string) (*Account, error) {
...
+				}
+			} else {
+				// other error
+				return nil, err
+			}
+		}
+		return account, nil
+	}
+
+	return nil, status.Errorf(codes.NotFound, "no valid user or account Id provided")
```
braginini, line 186: The response error code is NotFound but message says that the parameters are invalid.
I'd say these have to match

management/server/auth0idp/auth0idp.go
```diff
@@ -0,0 +1,138 @@
+package auth0idp
```
braginini: Maybe having a package idp is better?
And then, inside, auth0.go

management/server/auth0idp/auth0idp.go
```diff
@@ -0,0 +1,138 @@
...
+type JWTToken struct {
+	AccessToken   string `json:"access_token"`
+	ExpiresIn     int    `json:"expires_in"`
+	expiresInTime time.Time
+	Scope         string `json:"scope"`
+	TokenType     string `json:"token_type"`
+}
+
+// getJWTToken retrieves access token to use the Auth0 Management API
+func (am *Auth0Manager) getJWTToken() error {
```
braginini: If we have multiple management instances.
Requesting the token on each of them shouldn't be an issue, right?

braginini (approved):

management/server/auth0idp/auth0idp.go
mlsmaycon: different token per instance and as we only use one account manager, there should be only one instance

management/server/account.go
mlsmaycon: missed that and we can fix it with the email update

PR #191, opened 2022-01-14
Change Management Sync protocol to support incremental (serial) network changes

braginini:
Only Management Service changes + Protocol.
Client will be done in the following PR


management/server/management_proto_test.go
```diff
@@ -0,0 +1,344 @@
...
+const (
+	TestValidSetupKey = "A2C8E62B-38F5-4553-B31E-DD66C696CEBB"
+)
+
+// registerPeers registers peersNum peers on the management service and returns their Wireguard keys
+func registerPeers(peersNum int, client mgmtProto.ManagementServiceClient) ([]*wgtypes.Key, error) {
+
+	var peers = []*wgtypes.Key{}
+	for i := 0; i < peersNum; i++ {
+		key, err := wgtypes.GenerateKey()
```
mlsmaycon: should use GeneratePrivateKey()

mlsmaycon (changes requested):

management/server/management_proto_test.go
braginini: done

mlsmaycon (approved):

PR #200, opened 2022-01-23
Fix detection of Wireguard kernel module

stv0g:
The old code seemed rather complex, slow and was in fact broken for a Raspberry Pi running Debian Buster which caused it to fallback to the userspace Wireguard implementation.

Closes #200


stv0g:
I just realized that this might actually not work as we dont want to check weather the module is loaded but rather weather it could be loaded.

Is Linux loading the module actually on-demand during link creation?
I couldnt find any code in wt which does it..

braginini:
> I just realized that this might actually not work as we dont want to check weather the module is loaded but rather weather it could be loaded.
> 
> Is Linux loading the module actually on-demand during link creation? I couldnt find any code in wt which does it..

You are right, the code should try loading the module.
So far it just checks whether it has been loaded.

Not sure about the in-demand loading. We will test it.

stv0g:
@braginini I spend some time on it.

Actually there is a way for the kernel to automatically load the module via a call to `request_module()`: https://github.com/torvalds/linux/blob/dd81e1c7d5fb126e5fbc5c9e334d7b3ec29a16a0/net/core/rtnetlink.c#L3430

This will in fact cause the kernel to invoke the userspace modprobe program.

I spend some time thinking about what would be the best and most robust approach to detect if a module is loadable or not. I ended up, that its the best when we actually attempt to load it.

At the same time, we dont want to call `modprobe` directly as this could break if `modprobe` is not available (e.g. in containers) or located on a non-standard path.

My solution now try to create a dummy Wireguard interface with an invalid MTU attribute.
The kernel can only validate this MTU attribute if it knows what kind of interface should be created.
This will trigger the kernel to load the module.

We now can detect the presence of the Wireguard module by checking the error when creating this dummy interface.
It will fail in any case. But if Wireguard exists it fails with `EINVAL` (caused by the MTU) otherwise it fails with `EOPNOTSUPP`

braginini:
Neat solution @stv0g ! It seems to be quite reliable.as well.

FYI: there is a test failure which is unrelated to your changes. In case your are trying to figure out whether you have caused it - you haven't :) So just ignore it.

We will fix it, there is some inconsistency in the tests.
https://github.com/stv0g/wiretrustee/runs/4913869170?check_suite_focus=true

mlsmaycon (approved):
Thank you @stv0g it looks really good. 

@braginini as notes for us:
- add PR testing
- add  test for unloaded module scenario (`sudo modprobe -r wireguard` might help)

PR #217, opened 2022-02-02
Store domain information

mlsmaycon:
* retrieve domain from jwt token
* store domain when creating new account
* store domain for existing accounts


mlsmaycon:
@braginini  As we discussed, will be better to make room for multitenancy on the storage schema first

braginini (approved):

braginini (changes requested):
need account_test that checks domain

braginini (approved):

PR #224, opened 2022-02-09
feature(ice): use single socket to gather connections candidates

gigovich:


braginini:
solved in https://github.com/wiretrustee/wiretrustee/pull/232

PR #230, opened 2022-02-15
extracted accountmanager to interface

shatoboar:
Accountmanager has been extracted to an interface, and the previous implementation renamed to DefaultAccountManager

### DIFF
--- a/management/server/grpcserver.go
+++ b/management/server/grpcserver.go
@@ -16,7 +16,7 @@ import (
 
 // Server an instance of a Management server
 type Server struct {
-	accountManager *AccountManager
+	accountManager *DefaultAccountManager
 	wgKey          wgtypes.Key
 	proto.UnimplementedManagementServiceServer
 	peersUpdateManager     *PeersUpdateManager
@@ -28,7 +28,7 @@ type Server struct {
 const AllowedIPsFormat = "%s/32"
 
 // NewServer creates a new Management server
-func NewServer(config *Config, accountManager *AccountManager, peersUpdateManager *PeersUpdateManager, turnCredentialsManager TURNCredentialsManager) (*Server, error) {
+func NewServer(config *Config, accountManager *DefaultAccountManager, peersUpdateManager *PeersUpdateManager, turnCredentialsManager TURNCredentialsManager) (*Server, error) {
 	key, err := wgtypes.GeneratePrivateKey()
 	if err != nil {
 		return nil, err
@@ -158,7 +158,7 @@ func (s *Server) registerPeer(peerKey wgtypes.Key, req *proto.LoginRequest) (*Pe
 		return nil, status.Errorf(codes.NotFound, "provided setup key doesn't exists")
 	}
 
-	//todo move to AccountManager the code below
+	//todo move to DefaultAccountManager the code below
 	networkMap, err := s.accountManager.GetNetworkMap(peer.Key)
 	if err != nil {
 		return nil, status.Error(codes.Internal, "internal server error")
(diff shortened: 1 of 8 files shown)


management/server/grpcserver.go
```diff
@@ -16,7 +16,7 @@ import (
 
 // Server an instance of a Management server
 type Server struct {
-	accountManager *AccountManager
+	accountManager *DefaultAccountManager
```
braginini: I suggest keeping the AccountManager interface here, then we can easily replace it with some other implementation if needed. 
It will also be easier to mock in tests
That is the whole purpose of interfaces :)

management/server/grpcserver.go
```diff
@@ -28,7 +28,7 @@ type Server struct {
 const AllowedIPsFormat = "%s/32"
 
 // NewServer creates a new Management server
-func NewServer(config *Config, accountManager *AccountManager, peersUpdateManager *PeersUpdateManager, turnCredentialsManager TURNCredentialsManager) (*Server, error) {
+func NewServer(config *Config, accountManager *DefaultAccountManager, peersUpdateManager *PeersUpdateManager, turnCredentialsManager TURNCredentialsManager) (*Server, error) {
```
braginini: I suggest keeping the AccountManager interface here, then we can easily replace it with some other implementation if needed.
It will also be easier to mock in tests
That is the whole purpose of interfaces :)

management/server/http/handler/setupkeys.go
```diff
@@ -41,7 +41,7 @@ type SetupKeyRequest struct {
 	Revoked   bool
 }
 
-func NewSetupKeysHandler(accountManager *server.AccountManager, authAudience string) *SetupKeys {
+func NewSetupKeysHandler(accountManager *server.DefaultAccountManager, authAudience string) *SetupKeys {
```
braginini: I suggest keeping the AccountManager interface here, then we can easily replace it with some other implementation if needed.
It will also be easier to mock in tests
That is the whole purpose of interfaces :)

management/server/http/server.go
```diff
@@ -19,12 +19,12 @@ type Server struct {
...
 	certManager    *autocert.Manager
 	tlsConfig      *tls.Config
-	accountManager *s.AccountManager
+	accountManager *s.DefaultAccountManager
 }
 
 // NewHttpsServer creates a new HTTPs server (with HTTPS support) and a certManager that is responsible for generating and renewing Let's Encrypt certificate
 // The listening address will be :443 no matter what was specified in s.HttpServerConfig.Address
-func NewHttpsServer(config *s.HttpServerConfig, certManager *autocert.Manager, accountManager *s.AccountManager) *Server {
+func NewHttpsServer(config *s.HttpServerConfig, certManager *autocert.Manager, accountManager *s.DefaultAccountManager) *Server {
```
braginini: I suggest keeping the AccountManager interface here, then we can easily replace it with some other implementation if needed.
It will also be easier to mock in tests
That is the whole purpose of interfaces :)

management/server/peer.go
```diff
@@ -56,7 +56,7 @@ func (p *Peer) Copy() *Peer {
 }
 
 //GetPeer returns a peer from a Store
-func (am *AccountManager) GetPeer(peerKey string) (*Peer, error) {
+func (am *DefaultAccountManager) GetPeer(peerKey string) (*Peer, error) {
```
braginini, line 59: I suggest keeping the AccountManager interface here, then we can easily replace it with some other implementation if needed.
It will also be easier to mock in tests
That is the whole purpose of interfaces :)

braginini (changes requested):

braginini (approved):

PR #232, opened 2022-02-16
single socket ice

braginini:


client/internal/engine.go
```diff
@@ -22,25 +23,37 @@ import (
...
 	// WgPrivateKey is a Wireguard private key of our peer (it MUST never leave the machine)
 	WgPrivateKey wgtypes.Key
+
 	// IFaceBlackList is a list of network interfaces to ignore when discovering connection candidates (ICE related)
 	IFaceBlackList map[string]struct{}
 
 	PreSharedKey *wgtypes.Key
+
+	// UDPMuxPort default value 55050
+	UDPMuxPort *int
```
mlsmaycon: I don't see the use case for setting this port, but we could keep it and ditch the pointer here, as it will default to 0. 

It should help with tests too, as it will generate random ports

client/internal/engine.go
```diff
@@ -134,6 +179,29 @@ func (e *Engine) Start() error {
 		return err
 	}
 
+	muxConnPort := 55050
+	if e.config.UDPMuxPort != nil {
```
mlsmaycon: if UDPMuxPort defaults to 0, we can simply accept its value and we will have either a random port from 0 or wherever we set as UDPMuxPort's value.

mlsmaycon (changes requested):

client/internal/engine.go
braginini: Done. I kept explicit definition on port=0 to be more clear

mlsmaycon (approved):

Issue #246, opened 2022-03-02
"TypeError: Failed to fetch" upon log in. 


WGandy:
I just set this up on ubuntu 20.04.4 with docker compose for self hosting.  
I think everything went according to plan. There isn't anything too weird in the logs, as far as I can tell. But I get this type error line twice just above the dotted box for peers. And, if I click on the link to add a peer, I can't select the key type (default or one shot). 

![image](https://user-images.githubusercontent.com/81717633/156309667-6bc74118-4a69-4c1c-bf8b-3bc1ab834a9b.png)

WGandy:
Yes, I referenced that guide and watched the youtube video. 

The one issue I had while installing was a problem with the dns for my domain. Let's Encrypt failed on the first try and it seems to be because that domain/subdomain was involved with cloudflare proxy. I switched to an entirely different domain (and simple dns provided by the registrar), updated the setup.env file and auth0 and ran the ./configure.sh script again. That time it made the certs correctly. But perhaps running the configure.sh a second time caused a second problem? 

I'm planning to wipe out the entire wiretrustee directory and start over.  Unless you have some other suggestions.

mlsmaycon:
Hello @WGandy, we've updated the templates to use a shared certificate for both, dashboard and management API.

Can you please verify ?

WGandy:
Just to follow up, Maycon did a great job of digging into this and the update seems to solve the issue. 
There is a detail that I'm still concerned might snag someone. Step 3 of the "Add Peer" instructions on the management dashboard needs to be updated to include an additional argument. 
`--management-url [https://](https://$WIRETRUSTEE_DOMAIN:33073/)` 
I disregarded Maycon's instructions (forgot) and just followed the instructions presented on the Add Peer page which got me into trouble.  

I suspect that this argument may be baked into the client code in the future and then won't be necessary??

mlsmaycon:
Thank you @WGandy, I've opened the ticket https://github.com/wiretrustee/wiretrustee-dashboard/issues/18

Issue #244, opened 2022-02-27
using the Self-hosting Guide without success


chmis8000:
1:management.json:
![management json](https://user-images.githubusercontent.com/19686229/155884822-6d684879-462a-40fe-aa23-3bbb3d906c41.png)

2:docker-compose.yml:
![2022-02-27_214029](https://user-images.githubusercontent.com/19686229/155884766-dd0d7b1e-dcea-46e5-bda3-de73af10c07d.png)

3:[Coturn_log.txt](https://github.com/wiretrustee/wiretrustee/files/8148437/Coturn_log.txt)

4:https://github.com/wiretrustee/wiretrustee/blob/main/docs/self-hosting.md
vs
[README.md](https://github.com/wiretrustee/wiretrustee#readme) start with ""Running Dashboard, Management, Signal and Coturn"
The descriptions are inconsistent.

In addition to modifying setup.env, what else do I need to modify?
Can the documentation be improved? Thank you.

Neptunespace:
Same here

mlsmaycon:
Hello @chmis8000 @Neptunespace, sorry to hear that you are facing issues. I will review the guide and update it according to the latest changes.

mlsmaycon:
Hello @chmis8000 and @Neptunespace,

We've updated the documentation and templates to reflect the latest changes and to offer a better support for turn configuration as well.

chmis8000:
@mlsmaycon  it's not ok, signal,management,dashboard no error,coturn has many errors.
Change from centos8 system to ubuntu20.4 system/centos7.6.
![a](https://user-images.githubusercontent.com/19686229/158062291-8617794e-00a5-48fd-ac9d-dcc3314555c4.png)
![b](https://user-images.githubusercontent.com/19686229/158062338-2adff076-2318-46e5-a16b-2a0091f63b2c.png)
![c](https://user-images.githubusercontent.com/19686229/158062347-5b5e36eb-aac6-4f3a-baae-cc5589c9464f.png)
[error.txt](https://github.com/wiretrustee/wiretrustee/files/8239776/error.txt)

mlsmaycon:
hello @chmis8000 most of the errors and warnings from Coturn are caused by default configuration settings. These are ok to ignore as they don't affect Coturn's functionality. 

The only settings that can be improved from our end are the certificate files path, that we can comment them out or point to the same files used by dashboard and management. Let me evaluate if they will work as expected or if we need further changes for that.

[...]

PR #249, opened 2022-03-05
Updated self-hosted scripts and documentation

mlsmaycon:
Added more variables to setup.env and Updated the documentation.

We are now configuring the turn server with the template as well.


mlsmaycon:
Relates to #246 and #244

braginini (approved):

PR #255, opened 2022-03-09
feat(ui): refactor main module, embed icons

gigovich:


.goreleaser_ui_darwin.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+archives:
+  - builds:
+      - wiretrustee-ui-darwin
+
+brews:
+  -
+    tap:
+      owner: wiretrustee
+      name: homebrew-client-ui
```
mlsmaycon, line 30: should be same repo name as non-ui

PR #303, opened 2022-04-27
User Endpoint

shatoboar:
Exposes endpoint under "/users/" that returns information on users.
Calls idp manager to get information not stored locally (email, name), which in the case of the managed version is auth0.


management/server/account.go
```diff
@@ -214,9 +223,14 @@ func (am *DefaultAccountManager) GetAccountByUserOrAccountId(userId, accountId,
 	return nil, status.Errorf(codes.NotFound, "no valid user or account Id provided")
 }
 
+func isNil(i idp.Manager) bool {
```
mlsmaycon, line 248: Good catch. One thing, the function name is not matching the logic. You are returning true and it isn't nil.

management/server/account.go
```diff
@@ -225,6 +239,53 @@ func (am *DefaultAccountManager) updateIDPMetadata(userId, accountID string) err
...
+	account, err := am.GetAccountById(accountID)
+	if err != nil {
+		return nil, err
+	}
+
+	queriedUsers := make([]*idp.UserData, 0)
+
+	// https://auth0.com/docs/manage-users/user-search/retrieve-users-with-get-users-endpoint#limitations
+	// auth0 limitation of 1000 users via this endpoint
+	for page := 0; page < 20; page++ {
```
mlsmaycon: This logic should belong to the idp manager for auth0

management/server/idp/auth0.go
```diff
@@ -183,6 +186,126 @@ func (c *Auth0Credentials) Authenticate() (JWTToken, error) {
...
+	var list []*UserData
+	for user := range userData {
+		list = append(list, &userData[user])
+	}
+
+	if res.StatusCode != 200 {
+		return nil, fmt.Errorf("unable to get UserData, statusCode %d", res.StatusCode)
+	}
+
+	return list, nil
```
mlsmaycon, line 268: there should be a validation for an empty batch response, either here or at the API

management/server/idp/auth0.go
```diff
@@ -183,6 +186,126 @@ func (c *Auth0Credentials) Authenticate() (JWTToken, error) {
...
+}
+
+// Requests user data from auth0
+// user data: email
+func (am *Auth0Manager) GetUserData(userId string, appMetadata AppMetadata) (*UserData, error) {
```
mlsmaycon: It seems like you can use the same code and endpoint from GetBatchedUserData and only change the query parameter to use user id to retrieve the same result. @braginini what do you think?

mlsmaycon (changes requested):
Besides tests for idp manager and account manager new methods, you can improve the logs as a lot of info logs you've added would be better placed as debug logs.

management/server/account.go
```diff
@@ -225,6 +239,53 @@ func (am *DefaultAccountManager) updateIDPMetadata(userId, accountID string) err
...
+		batch, err := am.idpManager.GetBatchedUserData(accountID, page)
```
mlsmaycon: You should handle the case where there are no IDP manager credentials (self-hosted) where we usually don't call idp manager methods and in this case, `queriedUsers` will always be empty.

braginini: Good point @mlsmaycon 
I also think we need a test for this

management/server/account.go
braginini: @mlsmaycon agree

management/server/account.go
shatoboar: done.

management/server/idp/auth0.go
```diff
@@ -183,6 +186,136 @@ func (c *Auth0Credentials) Authenticate() (JWTToken, error) {
...
+		log.Debugf("Requested batch; %v", batch)
+
+		defer func() {
+			err = res.Body.Close()
+			if err != nil {
+				log.Errorf("error while closing update user app metadata response body: %v", err)
+			}
+		}()
+
+		if len(batch) == 0 {
```
mlsmaycon: this check should be performed after StatusCode as a non 200 status will cause len(batch) == 0

management/server/idp/auth0.go
```diff
@@ -183,6 +186,136 @@ func (c *Auth0Credentials) Authenticate() (JWTToken, error) {
...
+			list = append(list, &batch[user])
+		}
+	}
+
+	return list, nil
+}
+
+// Requests user data from auth0
+// user data: email
+func (am *Auth0Manager) GetUserDataByEmail(userId string, appMetadata AppMetadata) (*UserData, error) {
```
mlsmaycon: name here seems to be misleading as we use the user's id and not email

management/server/idp/auth0.go
shatoboar: Ah thanks, good point

braginini (approved):

PR #305, opened 2022-05-04
Adding peer registration support to JWT

mlsmaycon:
Clients can send JWT as part of the Login message.

* The management will validate the JWT as it does in the API and will register the Peer to the user's account.
* New fields were added to grpc messages in management and client daemon  and its clients were updated
* Peer has one new field, UserID, that will hold the id of the user that registered it
* JWT middleware CheckJWT got splitter and renamed to support validation for non HTTP requests
* Added test for adding new Peer with UserID
* Lots of tests update because of new field


braginini (commented):
Copy() doesn't copy UserId

braginini (changes requested):
Copy() doesn't copy UserId

braginini (approved):

PR #306, opened 2022-05-05
feat(acl): add rules for ACL

gigovich:


management/server/http/handler/rules.go
```diff
@@ -0,0 +1,128 @@
...
+		log.Error(err)
+		http.Redirect(w, r, "/", http.StatusInternalServerError)
+		return
+	}
+
+	writeJSONObject(w, account.Rules)
+}
+
+func (h *Rules) CreateOrUpdateRuleHandler(w http.ResponseWriter, r *http.Request) {
+	account, err := h.getRuleAccount(r)
```
braginini, line 76: I think that this method can be generalized because a similar logic of getting an account from the request is done in Peers and Users

braginini: h.getRuleAccount(r)

management/server/rule.go
```diff
@@ -0,0 +1,97 @@
...
+// Rule of ACL for groups
+type Rule struct {
+	// ID of the rule
+	ID string
+
+	// Name of the rule visible in the UI
+	Name string
+
+	// Source list of groups IDs of peers
+	Source []string
```
braginini, line 25: Do you think it would make sense to have individual peers as well?
I see a rule having groups and peers

management/server/rule.go
```diff
@@ -0,0 +1,97 @@
...
+// TrafficFlowType defines allowed direction of the traffic in the rule
+type TrafficFlowType int
+
+const (
+	// TrafficFlowBidirect allows traffic to both direction
+	TrafficFlowBidirect TrafficFlowType = iota
+)
+
+// Rule of ACL for groups
+type Rule struct {
```
braginini, line 17: Maybe we should call it ACL? 
And all the corresponding go files, handlers rename to ACLS.
Similarly for the http endpoint it might be better to name things as ACL.

What do u think?
@gigovich @mlsmaycon

management/server/http/handler/rules.go
gigovich: my proposal here is:
let's merge this as is, and after that, I can refactor all handlers.

management/server/rule.go
gigovich: It is logical if this use case is not rare, because it adds some mess to UI and implementation.

management/server/rule.go
gigovich: IMHO: Rule is a part of the ACL.

braginini: So as agreed we just rename a Handler to Acls

management/server/rule.go
braginini: so we skip this for now as agreed

management/server/file_store.go
```diff
@@ -163,6 +230,25 @@ func (s *FileStore) SaveAccount(account *Account) error {
 	s.mux.Lock()
 	defer s.mux.Unlock()
 
+	if len(account.Groups) == 0 {
```
braginini: I think it makes sense to move the logic of adding the default group to the AccountManager since it is not a business of the FIleStore

braginini (changes requested):

management/server/account.go
```diff
@@ -47,6 +47,10 @@ type AccountManager interface {
 	GroupAddPeer(accountId, groupID, peerKey string) error
 	GroupDeletePeer(accountId, groupID, peerKey string) error
 	GroupListPeers(accountId, groupID string) ([]*Peer, error)
+	GetRule(accountId, ruleID string) (*Rule, error)
+	SaveRule(accountID string, rule *Rule) error
+	DeleteRule(accountId, ruleID string) error
+	ListRules(accountId string) ([]*Rule, error)
```
braginini, line 58: It might be a good idea to move the implementation of these Rule related methods of DefaultAccountManager to a separate file e.g. rules.go

braginini (changes requested):

management/server/http/handler/acls.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+// ACLs is a handler that returns rules of the account
+type ACLs struct {
+	accountManager server.AccountManager
+	authAudience   string
+	jwtExtractor   jwtclaims.ClaimsExtractor
+}
+
+func NewRules(accountManager server.AccountManager, authAudience string) *ACLs {
```
gigovich: Rename it

management/server/peer.go
```diff
@@ -220,10 +223,28 @@ func (am *DefaultAccountManager) GetNetworkMap(peerKey string) (*NetworkMap, err
 	}
 
 	var res []*Peer
-	for _, peer := range account.Peers {
-		// exclude original peer
-		if peer.Key != peerKey {
-			res = append(res, peer.Copy())
+	rules, err := am.Store.GetPeerSrcRules(account.Id, peerKey)
```
mlsmaycon: fetching only rules where the peer is the source will give you an incomplete list of peers when the peer belongs only to a destination group

management/server/http/handler/rules.go
```diff
@@ -0,0 +1,128 @@
...
+		return
+	}
+
+	var req server.Rule
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	if r.Method == http.MethodPost {
```
mlsmaycon, line 88: We are missing a check for an existing Rule.ID on this handler, so we avoid creating new rules. The same goes for the CreateOrUpdateGroupHandler

mlsmaycon (changes requested):

mlsmaycon (approved):

management/server/http/handler/rules.go
mlsmaycon: nevermind when decoding the request we should expect the rule struct

management/server/http/handler/rules.go
```diff
@@ -0,0 +1,128 @@
...
+// GetAllRulesHandler list for the account
+func (h *Rules) GetAllRulesHandler(w http.ResponseWriter, r *http.Request) {
+	account, err := h.getRuleAccount(r)
+	if err != nil {
+		log.Error(err)
+		http.Redirect(w, r, "/", http.StatusInternalServerError)
+		return
+	}
+
+	writeJSONObject(w, account.Rules)
```
braginini: Any particular reason why we return a map not a list of rules? 
All other endpoints return [peers] or [users].
This one returns a map. 

```
{
  "ca2im93lo1hs0mlug03g": {
    "ID": "ca2im93lo1hs0mlug03g",
    "Name": "Default",
    "Source": [
      "ca2im93lo1hs0mlug030"
    ],
    "Destination": [
      "ca2im93lo1hs0mlug030"
    ],
    "Flow": 0
  }
}

```

I propose returning rules in such a structure:

```
[
   {
      "ID":"ca2im93lo1hs0mlug03g",
      "Name":"Default",
      "Source":[
         "ca2im93lo1hs0mlug030"
      ],
      "Destination":[
         "ca2im93lo1hs0mlug030"
      ],
      "Flow":0
   }
]
```

It will be much easier to handle on the UI side because all the Table components are built to handle an array.
Also this way we are more consistent with our API

**P.S. The same applies to the /api/groups endpoint**

braginini: I also think that returning account.Rules is a bit dangerous because it is our internal structure. 
It might be that we rename fields internally or would like to change how fields are returned.
I'd suggest creating a RuleResponse struct to return (see PeerRersponse for example). Let's be consistent here as well.

**P.S. The same applies to the /api/groups endpoint**

braginini:
```suggestion
		var respBody []*RuleResponse
	for _, rule := range account.Rules {
		respBody = append(respBody, toRuleResponse(rule))
	}

	writeJSONObject(w, respBody)
```

management/server/http/handler/rules.go
```diff
@@ -0,0 +1,128 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+// Rules is a handler that returns rules of the account
+type Rules struct {
+	accountManager server.AccountManager
+	authAudience   string
+	jwtExtractor   jwtclaims.ClaimsExtractor
+}
+
```
braginini, line 49:
```suggestion

type RuleResponse struct {
	ID          string
	Name        string
	Sources      []string
	Destinations []string
	Flow        server.TrafficFlowType
	Enabled bool
}

```

management/server/http/handler/rules.go
```diff
@@ -0,0 +1,128 @@
...
+func (h *Rules) getRuleAccount(r *http.Request) (*server.Account, error) {
+	jwtClaims := h.jwtExtractor.ExtractClaimsFromRequestContext(r, h.authAudience)
+
+	account, err := h.accountManager.GetAccountWithAuthorizationClaims(jwtClaims)
+	if err != nil {
+		return nil, fmt.Errorf("failed getting account of a user %s: %v", jwtClaims.UserId, err)
+	}
+
+	return account, nil
+}
```
braginini, line 175:
```suggestion
}

func toRuleResponse(rule *server.Rule) *RuleResponse {
	return &RuleResponse{
		Name:        rule.Name,
		ID:          rule.ID,
		Destinations: rule.Destination,
		Sources:      rule.Source,
		Flow:        rule.Flow,
		Enabled: true,
	}
}
```

management/server/http/handler/rules.go
braginini: See proposed changes below

braginini:
Another point.

We return Source and Destination groups IDs but no Name. 
In the UI we will need to show the Name of the group in Source and Destination. 
I suggest combining it already on the server-side. Any thoughts? 

This:
```
[
  {
    "ID": "ca2im93lo1hs0mlug03g",
    "Name": "Default",
    "Source": [
      "ca2im93lo1hs0mlug030"
    ],
    "Destination": [
      "ca2im93lo1hs0mlug030"
    ],
    "Flow": 0
  }
]

```

Replace With this 
```
[
  {
    "ID": "ca2im93lo1hs0mlug03g",
    "Name": "Default",
    "Source": [
     { "ca2im93lo1hs0mlug030", "Name": "All" }
    ],
    "Destination": [
      { "ca2im93lo1hs0mlug030", "Name": "All" }
    ],
    "Flow": 0
  }
]

```

braginini (approved):

braginini (changes requested):
It seems like a peer is not removed from the groups when peer is deleted from the account

management Server fails with NPE when:
1. add 2 peers
2. remove 1 peer
3. add 1 peer <-NPE on am.GetNetworkMap in 

```for _, g := range groups {
		for _, pid := range g.Peers {
			peer := account.Peers[pid]
			// exclude original peer
			if peer.Key != peerKey {
				res = append(res, peer.Copy())
			}
		}
	}
```

PR #312, opened 2022-05-12
Improve output of a status command

braginini:
This:
```
Run UP command to log in with SSO (interactive login):

 netbird up 

If you are running a self-hosted version and no SSO provider has been configured in your Management Server,
you can use a setup-key:

 netbird up --management-url <YOUR_MANAGEMENT_URL> --setup-key <YOUR_SETUP_KEY>

More info: https://www.netbird.io/docs/overview/setup-keys


```

Instead of this:
```
run the command "netbird up" to login. If no SSO provider has been set in your management serveryou can use a setup-key, see more at https://www.netbird.io/docs/overview/setup-keys for more info
```


mlsmaycon (approved):
Lgtm

PR #314, opened 2022-05-17
Update daemon server adminURL and managementURL fields

mlsmaycon:
Update daemon server adminURL and managementURL fields and Removed the UP call in the login function

### DIFF
--- a/client/server/server.go
+++ b/client/server/server.go
@@ -90,6 +90,23 @@ func (s *Server) Start() error {
 	return nil
 }
 
+// loginAttempt attempts to login using the provided information. it returns a status in case something fails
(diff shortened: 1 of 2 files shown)


client/server/server.go
```diff
@@ -131,6 +131,12 @@ func (s *Server) Login(_ context.Context, msg *proto.LoginRequest) (*proto.Login
 	s.config = config
 	s.mutex.Unlock()
 
+	if s, _ := state.Status(); s != internal.StatusNeedsLogin && s != internal.StatusLoginFailed {
```
shatoboar: This s seems to be shadowing the server variable, so maybe we can rename it to something like status

shatoboar (changes requested):

braginini (approved):

PR #319, opened 2022-05-21
Fix Peer Deletion & HTTP endpoints

braginini:


management/server/peer.go
```diff
@@ -263,6 +264,10 @@ func (am *DefaultAccountManager) GetNetworkMap(peerKey string) (*NetworkMap, err
 	for _, g := range groups {
 		for _, pid := range g.Peers {
 			peer := account.Peers[pid]
```
mlsmaycon: checking result of getting map is better here.
```Go
peer, ok := account.Peers[pid]
if !ok {
```

mlsmaycon (approved):

PR #333, opened 2022-05-26
fix(acl): update each peer's network when rule,group or peer changed

gigovich:


management/server/group.go
```diff
@@ -115,8 +128,14 @@ func (am *DefaultAccountManager) GroupAddPeer(accountID, groupID, peerKey string
 	if add {
 		group.Peers = append(group.Peers, peerKey)
 	}
+	fmt.Println("+++", group.Peers)
```
mlsmaycon: left-over print?

management/server/group.go
```diff
@@ -134,14 +153,17 @@ func (am *DefaultAccountManager) GroupDeletePeer(accountID, groupID, peerKey str
...
 	}
 
+	account.Network.IncSerial()
 	for i, itemID := range group.Peers {
 		if itemID == peerKey {
 			group.Peers = append(group.Peers[:i], group.Peers[i+1:]...)
-			return am.Store.SaveAccount(account)
+			if err := am.Store.SaveAccount(account); err != nil {
+				return status.Errorf(codes.Internal, "can't save account")
+			}
```
mlsmaycon, line 159: We are missing handling PeerKeyId2SrcRulesId and PeerKeyId2DstRulesId cleanup at this point

management/server/group.go
```diff
@@ -69,7 +77,12 @@ func (am *DefaultAccountManager) DeleteGroup(accountID, groupID string) error {
 
 	delete(account.Groups, groupID)
 
-	return am.Store.SaveAccount(account)
```
mlsmaycon, line 72: We are missing handling PeerKeyId2SrcRulesId and PeerKeyId2DstRulesId cleanup here

management/server/rule.go
```diff
@@ -85,7 +91,12 @@ func (am *DefaultAccountManager) DeleteRule(accountID, ruleID string) error {
 
 	delete(account.Rules, ruleID)
 
-	return am.Store.SaveAccount(account)
+	account.Network.IncSerial()
```
mlsmaycon, line 94: We are missing handling PeerKeyId2SrcRulesId and PeerKeyId2DstRulesId cleanup here too

mlsmaycon (changes requested):

management/server/group.go
gigovich: @mlsmaycon I'm not sure I understand it, because we rebuild these indexes in the `Store.SaveAccount` method.

mlsmaycon: @gigovich on the store we seem to only handle adding rules to PeerKeyId2SrcRulesId and PeerKeyId2DstRulesId maps, but we don't remove a rule from a peer that peer was removed from a group, or when a group is removed from a rule. The same goes for when a peer is removed from the account, there is no delete(PeerKeyId2SrcRulesId, key)

gigovich: Fixed. Reset them before updating in `Store.SaveAccount` https://github.com/netbirdio/netbird/pull/333/files#diff-14bee3eff6b719cef1943d98a2d97bcb4ad24cc1f5429cabde0ec4d30381aa81R243

mlsmaycon: @gigovich this reset is removing the maps of peers from all accounts.

gigovich: https://github.com/netbirdio/netbird/pull/333/commits/fb2cf2c4c945a1a1bf08387c93cdbc2f3bef6ace
@mlsmaycon 

1. Clean indexes when removing peers.
2. Clean only account peers from indexes during the save call.

PR #348, opened 2022-06-02
Handle peer interface config change

braginini:


client/internal/engine.go
```diff
@@ -216,7 +216,33 @@ func (e *Engine) Start() error {
...
+		}
+	}
+
+	// second, terminate the connection and remove them
+	for _, p := range modified {
+		err := e.removePeer(p.GetWgPubKey())
+		if err != nil {
+			return err
+		}
+	}
```
mlsmaycon, line 239: As we discussed, it might make sense to keep the logic to re-add the peer here as well. I would extract the loop code from e.addNewPeers and create a new one for it, maybe e.addNewPeeR to be called by e.addNewPeers and e.modifyPeers

mlsmaycon (approved):

PR #361, opened 2022-06-12
NetBird SSH

braginini:


braginini:
If an agent has been upgraded from any version without SSH support, all others won't receive an update indicating that SSH is now possible to this peer.
Should be fixed

client/internal/config.go
```diff
@@ -145,7 +158,7 @@ func GetConfig(managementURL, adminURL, configPath, preSharedKey string) (*Confi
 	} else {
 		// don't overwrite pre-shared key if we receive asterisks from UI
 		pk := &preSharedKey
-		if preSharedKey == "**********" {
+		if preSharedKey == "**********" || preSharedKey == "" {
```
mlsmaycon: we shouldn't dismiss empty string as sometimes user wants to remove a preshared key from its config

client/internal/engine.go
```diff
@@ -414,6 +420,41 @@ func (e *Engine) handleSync(update *mgmProto.SyncResponse) error {
...
 }
 
+func (e *Engine) updateSSH(sshConf *mgmProto.SSHConfig) error {
+	if sshConf.GetSshEnabled() {
+		// start SSH server if it wasn't running
+		if e.sshServer == nil {
+			//nil sshServer means it has not yet been started
+			var err error
+			e.sshServer, err = e.sshServerFunc(e.config.SSHKey,
+				fmt.Sprintf("%s:%d", e.wgInterface.Address.IP.String(), 2222))
```
mlsmaycon: port should come from config for cases when port 2222 is already in use

client/internal/engine.go
```diff
@@ -414,6 +420,41 @@ func (e *Engine) handleSync(update *mgmProto.SyncResponse) error {
...
+			if err != nil {
+				return err
+			}
+			go func() {
```
mlsmaycon, line 440: In the event of an error (e.g. port in use), we should clean the e.sshServer

client/internal/engine_test.go
```diff
@@ -40,6 +43,125 @@ var (
...
+	key, err := wgtypes.GeneratePrivateKey()
+	if err != nil {
+		t.Fatal(err)
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	engine := NewEngine(ctx, cancel, &signal.MockClient{}, &mgmt.MockClient{}, &EngineConfig{
```
mlsmaycon, line 61: lets avoid using same interface configuration between tests

client/internal/engine.go
```diff
@@ -54,6 +55,8 @@ type EngineConfig struct {
 
 	// UDPMuxSrflxPort default value 0 - the system will pick an available port
 	UDPMuxSrflxPort int
+
+	SSHKey []byte
```
mlsmaycon, line 61: We should distinguish private and public ssh keys, here we store private key but other methods use sshKey for public SSH Key (e.g. Login)

client/ssh/server.go
```diff
@@ -0,0 +1,168 @@
...
+func getShellType() string {
+	shell := os.Getenv("SHELL")
+	if shell == "" {
+		shell = "sh"
+	}
+	return shell
+}
+
+// sessionHandler handles SSH session post auth
+func (srv *DefaultServer) sessionHandler(s ssh.Session) {
```
mlsmaycon: let's avoid using single-letter input variables

client/ssh/server.go
```diff
@@ -0,0 +1,168 @@
...
+			return
+		}
+		err = s.Exit(1)
+		if err != nil {
+			return
+		}
+	}
+}
+
+func (srv *DefaultServer) stdInOut(f *os.File, s ssh.Session) {
```
mlsmaycon: let's avoid using single-letter input variables

client/ssh/util.go
```diff
@@ -0,0 +1,86 @@
...
+const ECDSA KeyType = "ecdsa"
+
+// RSA is key of type rsa
+const RSA KeyType = "rsa"
+
+// RSAKeySize is a size of newly generated RSA key
+const RSAKeySize = 2048
+
+// GeneratePrivateKey creates RSA Private Key of specified byte size
+func GeneratePrivateKey(t KeyType) ([]byte, error) {
```
mlsmaycon: let's avoid using single-letter input variables

client/ssh/window_unix.go
```diff
@@ -0,0 +1,14 @@
...
+
+package ssh
+
+import (
+	"os"
+	"syscall"
+	"unsafe"
+)
+
+func setWinSize(f *os.File, w, h int) {
```
mlsmaycon: let's avoid using single-letter input variables

client/ssh/window_windows.go
```diff
@@ -0,0 +1,9 @@
+package ssh
+
+import (
+	"os"
+)
+
+func setWinSize(f *os.File, w, h int) {
```
mlsmaycon: let's avoid using single-letter input variables

management/server/grpcserver.go
```diff
@@ -290,6 +296,19 @@ func (s *Server) Login(ctx context.Context, req *proto.EncryptedMessage) (*proto
 			return nil, status.Error(codes.Internal, "internal server error")
 		}
 	}
+
```
mlsmaycon, line 299: can't put his comment in line 273, but should we use the peer key from the PeerKeys payload?

management/server/grpcserver.go
```diff
@@ -365,7 +384,8 @@ func toWiretrusteeConfig(config *Config, turnCredentials *TURNCredentials) *prot
 
 func toPeerConfig(peer *Peer) *proto.PeerConfig {
 	return &proto.PeerConfig{
-		Address: fmt.Sprintf("%s/%d", peer.IP.String(), SubnetSize), // take it from the network
+		Address:   fmt.Sprintf("%s/%d", peer.IP.String(), SubnetSize), // take it from the network
+		SshConfig: &proto.SSHConfig{SshEnabled: true},                 //TODO REMOVE THIS HARDCODED VALUE
```
mlsmaycon: remove hardcoded flag

mlsmaycon (changes requested):

mlsmaycon (approved):

PR #364, opened 2022-06-18
Check if new account ID is already being used

mlsmaycon:
Check if the new account ID is already being used

Modify domain only if use is the  account admin


management/server/account.go
```diff
@@ -198,6 +198,11 @@ func BuildManager(
 
 }
 
+func (am *DefaultAccountManager) isNewAccountIDUsed(id string) bool {
```
braginini: I suggest having a method accountExists

management/server/account.go
```diff
@@ -198,6 +192,27 @@ func BuildManager(
 
 }
 
+// newAccount creates a new Account with a generated ID and generated default setup keys.
+// If ID is already in use we try one more time before returning error
```
braginini: I think it might make sense to explain why ID can be in use already
`If ID is already in use (e.g. due to a collision) we try one more time before returning error`

braginini (changes requested):

braginini (approved):

PR #368, opened 2022-06-24
Add client status collection

mlsmaycon:
Add client status recorder package.
We will update the peer's status on every state change.
Local peer, signal, and management states will be 
updated when running or stopping the client.


client/status/status.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// AddPeer adds peer to Daemon status map
+func (d *Status) AddPeer(peerPubKey string) error {
+	d.mux.Lock()
+	defer d.mux.Unlock()
+
+	_, ok := d.peers[peerPubKey]
+	if ok {
+		return errors.New("peer already exist")
```
braginini, line 71: not sure if we should throw an error here. Or maybe at least not throwing an error on the higher level that uses the method? Failure to AddPeer will break the updates coming from the management.

I would rather log a warning here.

client/status/status.go
```diff
@@ -0,0 +1,165 @@
...
+	d.mux.Lock()
+	defer d.mux.Unlock()
+
+	_, ok := d.peers[peerPubKey]
+	if ok {
+		delete(d.peers, peerPubKey)
+		return nil
+	}
+
+	return errors.New("no peer with to remove")
```
braginini, line 88: Same as AddPeer. I'd rather not throw an error and just log a warning.

braginini (changes requested):

client/status/status.go
```diff
@@ -0,0 +1,165 @@
...
+	peerState, ok := d.peers[peerPubKey]
+	if !ok {
+		return PeerState{}, errors.New("peer doesn't exist")
+	}
+
+	return peerState, nil
+}
+
+// UpdatePeerStatus updates peer status
+func (d *Status) UpdatePeerStatus(peerState PeerState) error {
```
braginini: UpdatePeerStatus but the object is called PeerState? Maybe we call it PeerStatus?

client/internal/peer/conn.go
```diff
@@ -343,6 +387,22 @@ func (conn *Conn) cleanup() error {
 
 	conn.status = StatusDisconnected
 
+	peerState, err := conn.statusRecorder.GetPeerStatus(conn.config.Key)
```
braginini: 1. You are modifying and existing state out of the lock of StatusRecorder. I think it makes sense to create a copy of PeerStatus

client/internal/peer/conn.go
```diff
@@ -343,6 +387,22 @@ func (conn *Conn) cleanup() error {
...
+	if err == nil {
+		peerState.ConnStatus = conn.status.String()
+		peerState.ConnStatusUpdate = time.Now()
+		peerState.Direct = false
+		peerState.Relayed = false
+		peerState.RemoteIceCandidateType = ""
+		peerState.LocalIceCandidateType = ""
+		err = conn.statusRecorder.UpdatePeerStatus(peerState)
```
braginini: 2. I would rather not query the status and create a separate method in the statusRecorder to create a status Disconnected and not pollute Conn.
Even there is no need to query status with `peerState, err := conn.statusRecorder.GetPeerStatus(conn.config.Key)`

client/internal/peer/conn.go
```diff
@@ -311,6 +342,19 @@ func (conn *Conn) startProxy(remoteConn net.Conn) error {
...
 	conn.status = StatusConnected
 
+	peerState.ConnStatus = conn.status.String()
+	peerState.ConnStatusUpdate = time.Now()
+	peerState.LocalIceCandidateType = pair.Local.Type().String()
+	peerState.RemoteIceCandidateType = pair.Remote.Type().String()
+	if pair.Local.Type() == ice.CandidateTypeRelay || pair.Remote.Type() == ice.CandidateTypeRelay {
+		peerState.Relayed = true
+	}
+
```
braginini, line 348: I wouldn't modify the status out of the lock of the status recorder. This is error and inconsistency prone. 
I would rather create a copy or a new object to overwrite the value.

client/internal/connect.go
```diff
@@ -68,26 +71,98 @@ func RunClient(ctx context.Context, config *Config) error {
...
+		err = statusRecorder.UpdateSignalStatus(signalState)
+		if err != nil {
+			return err
+		}
+
+		// set disconnected state for management and signal when exiting
+		defer func() {
+			managementState.Connected = false
+
+			err = statusRecorder.UpdateManagementStatus(managementState)
```
braginini: I would create a new state and update without messing up with existing objects

client/internal/connect.go
```diff
@@ -68,26 +71,98 @@ func RunClient(ctx context.Context, config *Config) error {
...
 			}
 			return wrapErr(err)
 		}
 
+		managementState.Connected = true
+
+		err = statusRecorder.UpdateManagementStatus(managementState)
+		if err != nil {
+			return err
+		}
```
braginini: I would rather create a new state and update status within a lock of status recorder

braginini: alternatively create a separate method of statusRecorder that would just update connected state of management, e.g. setManagementConnected(true/false)

braginini (changes requested):

client/internal/connect.go
braginini: alternatively create a separate method of statusRecorder that would just update connected state of management, e.g. setManagementConnected(true/false)

client/status/status.go
```diff
@@ -0,0 +1,165 @@
...
+	d.mux.Lock()
+	defer d.mux.Unlock()
+
+	d.management = managementState
+
+	return nil
+}
+
+// GetStatus gets full status
+func (d *Status) GetStatus() FullStatus {
```
braginini: The status is filled with values in RunClient() go routine. E.g. initially initialized with :
```
func NewStatus() *Status {
	return &Status{
		peers: make(map[string]PeerState),
	}
}
```

Since the status of management, signal, etc is filled in go routine in the run client method (connect), GetStatus() method will be accessible immediately. The user of GetStatus() has to be aware of the possible empty fields or maybe we should initialize status with some dummy values.

client/status/status.go
mlsmaycon: kept the error, but change to log only on the caller

client/status/status.go
mlsmaycon: kept the error, the caller already only logs this error

client/status/status.go
mlsmaycon: update all methods to Update..State

client/internal/peer/conn.go
mlsmaycon: we were creating this copy when returning, but it wasn't safe for future use case where there could be a pointer in its reference.

mlsmaycon: the methods was removed

client/internal/peer/conn.go
mlsmaycon: done

client/internal/connect.go
mlsmaycon: we can't set the connected status separately as it is related to the URL

client/status/status.go
mlsmaycon: we will handle it on the client side

client/internal/connect.go
mlsmaycon: we can't set the connected status separately as it is related to the URL

braginini (approved):

PR #406, opened 2022-07-30
Check and update Agent's Management URL if is legacy

braginini:
Relates to https://github.com/netbirdio/netbird/pull/400

All the existing agents by default connect to port 33073 of the
Management service. This value is also stored in the local config.
All the agents won't switch to the new port 443 
unless explicitly specified in the config.
We want the transition to be smooth for our users, therefore
this PR adds logic to check whether the old port 33073 can be 
changed to 443 and updates the config automatically.

### DIFF
--- a/client/internal/connect.go
+++ b/client/internal/connect.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"github.com/netbirdio/netbird/client/ssh"
 	nbStatus "github.com/netbirdio/netbird/client/status"
+	mgmtcmd "github.com/netbirdio/netbird/management/cmd"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


client/internal/connect.go
```diff
@@ -239,3 +240,57 @@ func connectToManagement(ctx context.Context, managementAddr string, ourPrivateK
...
+		// here we check whether we could switch from the legacy 33073 port to the new 443
+		log.Infof("attempting to switch from the legacy Management URL %s to the new one %s",
+			config.ManagementURL.String(), newURL.String())
+		key, err := wgtypes.ParseKey(config.PrivateKey)
+		if err != nil {
+			log.Infof("couldn't switch to the new Management %s", newURL.String())
+			return config, err
+		}
+
+		_, err = mgm.NewClient(ctx, newURL.Host, key, mgmTlsEnabled)
```
mlsmaycon: we need to run further tests to ensure we are talking to a Management port here as client doesn't validate grpc server

mlsmaycon (approved):

Issue #422, opened 2022-08-15
Self-hosted: Clients are unable to connect to gRPC port :33073. Port clashing?


kid-alan:
Hello there, team!

I highly appreciate the effort you are putting in the project.
I faced an issue while trying to run self-hosted Netbird, so I decided to report and perhaps to help you with solving it.

## An issue
I installed and configured self-hosted Netbird according to https://netbird.io/docs/getting-started/self-hosting
The dashboard works well. I am able to log in and generate auth keys for clients.

**But the clients are failing to connect:**
```
root@node-01:~# netbird up --setup-key <key> --management-url https://netbird.example.com:33073
WARN[2022-08-15T14:44:59Z] retrying Login to the Management service in 1.104660288s due to error rpc error: code = Unknown desc = context deadline exceeded 
WARN[2022-08-15T14:45:10Z] retrying Login to the Management service in 2.160763633s due to error rpc error: code = Unknown desc = context deadline exceeded 
Error: login backoff cycle failed: rpc error: code = Unknown desc = context deadline exceeded
```

Inspecting the logs on the server with `docker-compose logs` — there are no new log entries when attempting to connect to :33073.

**Same behaviour when trying to connect to :443**
Now there are log records from _dashboard_ container, indicating that client is attempting to connect to dashboard, instead of manager.
```
dashboard_1   | 167.71.53.220 - - [15/Aug/2022:14:14:14 +0000] "PRI * HTTP/2.0" 400 150 "-" "-" "-"
dashboard_1   | 167.71.53.220 - - [15/Aug/2022:14:14:15 +0000] "PRI * HTTP/2.0" 400 150 "-" "-" "-"
dashboard_1   | 167.71.53.220 - - [15/Aug/2022:14:14:16 +0000] "PRI * HTTP/2.0" 400 150 "-" "-" "-"
...
```

TCP ports are open and available:
```
root@node-01:~# nc -zv netbird.example.com 33073
Connection to netbird.example.com (167.71.61.23) 33073 port [tcp/*] succeeded!
root@node-01:~# nc -zv netbird.example.com 443
Connection to netbird.example.com (167.71.61.23) 443 port [tcp/*] succeeded!
```

Docker ps output
```
              Name                             Command               State                                   Ports                                 
---------------------------------------------------------------------------------------------------------------------------------------------------
infrastructure_files_coturn_1       docker-entrypoint.sh --log ...   Up                                                                            
[...]

PR #423, opened 2022-08-15
Enable HTTP/2 when loading TLS config from file

braginini:
When creating TLSConfig from provided certificate file, the HTTP/2 support is not enabled.
It works with Certmanager because it adds h2 support.
We can enable it the same way when creating TLSConfig from files:
```
NextProtos: []string{
  "h2", "http/1.1", // enable HTTP/2
  acme.ALPNProto, // enable tls-alpn ACME challenges
},
```  

Thanks to @kid-alan for pointing out the issue in https://github.com/netbirdio/netbird/issues/422


mlsmaycon (approved):
Lets work on updating the io/ioutil usage in a different PR

PR #424, opened 2022-08-16
Add routing support to management service

mlsmaycon:
Management will receive and store routes that are associated with a peer ID

The routes are distributed to peers according to their ACLs


management/server/route.go
```diff
@@ -0,0 +1,362 @@
...
+		Peer:       route.Peer,
+		Metric:     int64(route.Metric),
+		Masquerade: route.Masquerade,
+	}
+}
+
+func (am *DefaultAccountManager) getPeersRoutes(peers []*Peer) []*route.Route {
+	routes := make([]*route.Route, 0)
+	for _, peer := range peers {
+		peerRoutes, err := am.Store.GetPeerRoutes(peer.Key)
```
braginini, line 335: Do we need the GetPeerRoutes method to throw an error if there are no routes? 
I think just returning an empty list should be fine.

management/server/route.go
```diff
@@ -0,0 +1,362 @@
...
+	defer am.mux.Unlock()
+
+	account, err := am.Store.GetAccount(accountID)
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "account not found")
+	}
+
+	routeToUpdate, ok := account.Routes[routeID]
+	if !ok {
+		return nil, status.Errorf(codes.NotFound, "rule %s no longer exists", routeID)
```
braginini: route instead of rule

braginini (changes requested):

management/server/route.go
mlsmaycon: we only log an error here if the am.Store.GetPeerRoutes(peer.Key) returns an error different than NotFound. Which should be mostly system errors, and we want the logs.

management/server/route.go
mlsmaycon: I spotted another one too.

management/server/route.go
mlsmaycon: After discussing, it makes more sense to don't return error store calls when a peer doesn't have error associated with it

braginini (approved):

PR #426, opened 2022-08-17
Support new properties for OIDC auth

braginini:


infrastructure_files/setup.env.example
```diff
@@ -4,13 +4,16 @@
...
-# e.g. dev-24vkclam.us.auth0.com
-NETBIRD_AUTH0_DOMAIN=""
-# e.g. 61u3JMXRO0oOevc7gCkZLCwePQvT4lL0
-NETBIRD_AUTH0_CLIENT_ID=""
-# e.g. https://app.mydomain.com/ or https://app.mydomain.com,
-# Make sure you used the exact same value for Identifier
-# you used when creating your Auth0 API
-NETBIRD_AUTH0_AUDIENCE=""
+# e.g. https://dev-24vkclam.us.auth0.com or http://YOUR-KEYCLOAK-HOST:8080/realms/netbird
+NETBIRD_AUTH_AUTHORITY=""
```
mlsmaycon, line 6: lets keep the example with https for keycloak

infrastructure_files/setup.env.example
```diff
@@ -4,13 +4,16 @@
 
 # Dashboard domain. e.g. app.mydomain.com
 NETBIRD_DOMAIN=""
-# e.g. dev-24vkclam.us.auth0.com
-NETBIRD_AUTH0_DOMAIN=""
```
mlsmaycon, line 8: we should maintain compatibility with users that want to update the script but didn't read the docs. In these cases, they will have an old config file with NETBIRD_AUTH0... variables

infrastructure_files/management.json.tmpl
```diff
@@ -29,9 +29,9 @@
     "Datadir": "",
     "HttpConfig": {
         "Address": "0.0.0.0:$NETBIRD_MGMT_API_PORT",
-        "AuthIssuer": "https://$NETBIRD_AUTH0_DOMAIN/",
-        "AuthAudience": "$NETBIRD_AUTH0_AUDIENCE",
-        "AuthKeysLocation": "https://$NETBIRD_AUTH0_DOMAIN/.well-known/jwks.json",
+        "AuthIssuer": "$NETBIRD_AUTH_AUTHORITY",
```
mlsmaycon, line 32: There is a problem with AuthIssuer when using Auth0 (needs keycloak test). When we enter an AUTH_AUTHORITY without the trailing forward slash (/) our management is unable to validate the token.
```shell
AUTH_AUTHORITY=https://auth0-domain.com
```
One fix would be to keep the keep the trailing forwarding slash:
```shell
"AuthIssuer": "$NETBIRD_AUTH_AUTHORITY/",
```

infrastructure_files/setup.env.example
```diff
@@ -4,13 +4,16 @@
...
+NETBIRD_AUTH_AUTHORITY=""
+# e.g. netbird-client
+NETBIRD_AUTH_CLIENT_ID=""
+# indicates whether to use Auth0 or not: true or false
+NETBIRD_USE_AUTH0=""
+# a list of scopes supported e.g. `openid profile email` for keycloak or `openid profile email api offline_access email_verified` for Auth0
+NETBIRD_AUTH_SUPPORTED_SCOPES=""
+NETBIRD_AUTH_AUDIENCE=""
+# URL of the JWT certificates e.g. https://dev-24vkclam.us.auth0.com/.well-known/jwks.json
+NETBIRD_AUTH_JWT_CERTS=""
```
mlsmaycon, line 15: is the suffix /.well-known/jwks.json default for other OIDC providers? If it is, it is better to build the URL by combining the $NETBIRD_AUTH_AUTHORITY and this suffix.

mlsmaycon (changes requested):

infrastructure_files/setup.env.example
braginini: Nope, I was actually thinking that it might be a better idea to fetch it automatically from the openid-configuration.
But then we need to change management code

infrastructure_files/setup.env.example
braginini: Ok

infrastructure_files/management.json.tmpl
braginini: This property is user set, so it should be the full authority including /

mlsmaycon: ok

mlsmaycon:
we need to fix the docker-compose workflow only. 

The Lint workflow will be fixed in the #424

mlsmaycon (approved):

PR #439, opened 2022-08-23
Update scripts for the self-hosted Oauth 2.0 Device Auth Grant support

braginini:


infrastructure_files/configure.sh
```diff
@@ -63,21 +71,42 @@ export MGMT_VOLUMENAME
...
+    export NETBIRD_AUTH_OIDC_CONFIGURATION_ENDPOINT="https://${NETBIRD_AUTH0_DOMAIN}/.well-known/openid-configuration"
     export NETBIRD_USE_AUTH0="true"
-    export NETBIRD_AUTH_SUPPORTED_SCOPES="openid profile email api offline_access email_verified"
     export NETBIRD_AUTH_AUDIENCE=${NETBIRD_AUTH0_AUDIENCE}
-    export NETBIRD_AUTH_JWT_CERTS="https://${NETBIRD_AUTH0_DOMAIN}/.well-known/jwks.json"
+    export NETBIRD_AUTH_CLIENT_ID=${NETBIRD_AUTH0_CLIENT_ID}
+fi
+
+echo "loading OpenID configuration from ${NETBIRD_AUTH_OIDC_CONFIGURATION_ENDPOINT} to the openid-configuration.json file"
+wget ${NETBIRD_AUTH_OIDC_CONFIGURATION_ENDPOINT} -q -O openid-configuration.json
```
mlsmaycon: let's use curl instead of wget for better support, equivalent command:

```curl ${NETBIRD_AUTH_OIDC_CONFIGURATION_ENDPOINT} -q -o openid-configuration.json```

File: infrastructure_files/configure.sh [hunk 2]
```diff
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+if ! command -v jq &> /dev/null
```
mlsmaycon: lets sync the way we test for software in lines 3 and 14

infrastructure_files/configure.sh
```diff
@@ -63,21 +71,42 @@ export MGMT_VOLUMENAME
...
-    export NETBIRD_AUTH_JWT_CERTS="https://${NETBIRD_AUTH0_DOMAIN}/.well-known/jwks.json"
+    export NETBIRD_AUTH_CLIENT_ID=${NETBIRD_AUTH0_CLIENT_ID}
+fi
+
+echo "loading OpenID configuration from ${NETBIRD_AUTH_OIDC_CONFIGURATION_ENDPOINT} to the openid-configuration.json file"
+wget ${NETBIRD_AUTH_OIDC_CONFIGURATION_ENDPOINT} -q -O openid-configuration.json
+
+export NETBIRD_AUTH_AUTHORITY=$( jq -r  '.issuer' openid-configuration.json )
+export NETBIRD_AUTH_JWT_CERTS=$( jq -r  '.jwks_uri' openid-configuration.json )
+export NETBIRD_AUTH_SUPPORTED_SCOPES=$( jq -r '.scopes_supported | join(" ")' openid-configuration.json )
```
mlsmaycon, line 107: it's a neat way to fetch scopes, but I am concerned about application (client) configuration in the IDP and the scopes supported by them. 

Also, using all the scopes would tell the IDP to add extra information in the claim that is not necessary to the system, like `name given_name family_name nickname picture created_at identities phone address`, but we still missing `api`

mlsmaycon (changes requested):

infrastructure_files/configure.sh
braginini: we can hardcode the values

mlsmaycon (approved):

PR #441, opened 2022-08-24
Add routing peer support

mlsmaycon:
Peer will handle received routes and act as client or routers


client/internal/routemanager/manager.go
```diff
@@ -0,0 +1,183 @@
...
+
+func sendUpdateToClientNetwork(updateChannel chan routesUpdate, updateSerial uint64, routes []*route.Route) {
+	updateChannel <- routesUpdate{
+		updateSerial: updateSerial,
+		routes:       routes,
+	}
+}
+
+func (m *Manager) updateClientNetworks(updateSerial uint64, networks map[string][]*route.Route) {
+	select {
```
braginini: I don't think that we need select here because this function call comes from the UpdateRoutes method which already has the select.

client/internal/routemanager/manager.go
```diff
@@ -0,0 +1,183 @@
...
+	case <-m.ctx.Done():
+		log.Infof("not updating client network because context is done: %v", m.ctx.Err())
+		return
+	default:
+		for id, client := range m.clientNetworks {
+			_, found := networks[id]
+			if !found {
+				log.Debugf("stopping client network watcher, %s", id)
+				go client.stop()
```
braginini: stop() is a context.CancelFunc, no need to run it in a goroutine

client/internal/routemanager/manager.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+}
+
+func (m *Manager) updateClientNetworks(updateSerial uint64, networks map[string][]*route.Route) {
+	select {
+	case <-m.ctx.Done():
+		log.Infof("not updating client network because context is done: %v", m.ctx.Err())
+		return
+	default:
+		for id, client := range m.clientNetworks {
```
braginini: I'd add a comment here to indicate that we are removing routes that do not exist as per the update from the Management service.

client/internal/routemanager/manager.go
```diff
@@ -0,0 +1,183 @@
...
+				log.Debugf("stopping client network watcher, %s", id)
+				go client.stop()
+				delete(m.clientNetworks, id)
+			}
+		}
+
+		for id, routes := range networks {
+			watcher, found := m.clientNetworks[id]
+			if !found {
+				watcher = newClientNetworkWatcher(m.ctx, m.wgInterface, m.statusRecorder, routes[0].Network)
```
braginini: should we call it clientNetwork instead of watcher?

client/internal/routemanager/client.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+	}
+
+	c.routes = updateMap
+	c.updateSerial = update.updateSerial
+}
+
+// stateAndUpdateWatcher is the main point of reacting on client network routing events.
+// All the processing related to the client network should be done here. Thread-safe.
+func (c *clientNetwork) stateAndUpdateWatcher() {
```
braginini: stateAndUpdateWatcher is a typo? 
startAndUpdateWatcher?

client/internal/routemanager/manager.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// Stop stops the manager watchers and clean firewall rules
+func (m *Manager) Stop() {
+	m.stop()
+	m.serverRouter.firewall.CleanRoutingRules()
+}
+
+func sendUpdateToClientNetwork(updateChannel chan routesUpdate, updateSerial uint64, routes []*route.Route) {
+	updateChannel <- routesUpdate{
```
braginini: Might be beneficial to have a clientNetwork.SendUpdate(routesUpdate) function instead.
This prevents exposure/dependency on the internal mechanisms of receiving and sending updates. 
**Optional.**

client/internal/routemanager/client.go
```diff
@@ -0,0 +1,258 @@
...
+			c.network, c.chosenRoute.Peer, err)
+	}
+
+	return nil
+}
+
+func (c *clientNetwork) handleUpdate(update routesUpdate) {
+	if update.updateSerial < c.updateSerial {
+		log.Warnf("received a routes update with smaller serial number, ignoring it")
+		return
```
braginini: If we skip the update, do we need to call `recalculateRouteAndUpdatePeerAndSystem`?
Because in this case, we will recalculate. 
Maybe worth moving up the updateSerial check?

braginini (changes requested):

client/internal/routemanager/client.go
mlsmaycon: true, moved to the peersStateAndUpdateWatcher

client/internal/routemanager/client.go
mlsmaycon: was state, but I renamed it to peersStateAndUpdateWatcher

client/internal/routemanager/manager.go
mlsmaycon: changed to clientNetworkWatcher to avoid type override

client/internal/routemanager/manager.go
mlsmaycon: added sendUpdateToClientNetworkWatcher and moved the goroutine to it

client/internal/routemanager/client.go
```diff
@@ -202,12 +202,13 @@ func (c *clientNetwork) recalculateRouteAndUpdatePeerAndSystem() error {
 	return nil
 }
 
-func (c *clientNetwork) handleUpdate(update routesUpdate) {
-	if update.updateSerial < c.updateSerial {
-		log.Warnf("received a routes update with smaller serial number, ignoring it")
-		return
-	}
+func (c *clientNetwork) sendUpdateToClientNetworkWatcher(update routesUpdate) {
```
braginini, line 224: I'd just call it sendUpdate or update to avoid redundancy

client/internal/routemanager/client.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+	}
+
+	c.routes = updateMap
+	c.updateSerial = update.updateSerial
+}
+
+// peersStateAndUpdateWatcher is the main point of reacting on client network routing events.
+// All the processing related to the client network should be done here. Thread-safe.
+func (c *clientNetwork) peersStateAndUpdateWatcher() {
```
braginini, line 250: this is rather an action than an object, so I'd call it watchStateAndUpdates

braginini (changes requested):

client/internal/routemanager/manager_test.go
```diff
@@ -0,0 +1,370 @@
...
+		name                          string
+		inputInitRoutes               []*route.Route
+		inputRoutes                   []*route.Route
+		inputSerial                   uint64
+		shouldCheckServerRoutes       bool
+		serverRoutesExpected          int
+		clientNetworkWatchersExpected int
+	}{
+		{
+			name:            "Happy Client Path",
```
braginini: I think that we should call the cases with something that explains the flow. Otherwise, it is unclear what Happy or Mixed path means. E.g., 
```
Should create X client routes and remove N whatever routes....
```

client/internal/engine_test.go
```diff
@@ -196,6 +197,7 @@ func TestEngine_UpdateNetworkMap(t *testing.T) {
 		WgPort:       33100,
 	}, nbstatus.NewRecorder())
 	engine.wgInterface, err = iface.NewWGIFace("utun102", "100.64.0.1/24", iface.DefaultMTU)
+	engine.routeManager = routemanager.NewManager(ctx, key.PublicKey().String(), engine.wgInterface, engine.statusRecorder)
```
braginini, line 202: Some tests test route management in isolation from the whole system, but none test it working with the Engine. 
Route management is functionality that is just like the SSH server shouldn't disturb connectivity and updates. 
It is crucial to test because Engine is responsible for triggering route updates when an update comes from the management.

braginini (changes requested):

braginini (approved):

braginini (approved):

PR #443, opened 2022-08-25
Use select for turn credentials and peers update

mlsmaycon:
Also, prevent peer update when SSH is the same

### DIFF
--- a/management/server/updatechannel.go
+++ b/management/server/updatechannel.go
@@ -31,11 +31,15 @@ func (p *PeersUpdateManager) SendUpdate(peer string, update *UpdateMessage) erro
 	p.channelsMux.Lock()
 	defer p.channelsMux.Unlock()
 	if channel, ok := p.peerChannels[peer]; ok {
-		if len(channel) < channelBufferSize {
-			channel <- update
+		select {
+		case channel <- update:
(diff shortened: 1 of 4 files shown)


management/server/updatechannel.go
```diff
@@ -28,8 +31,15 @@ func (p *PeersUpdateManager) SendUpdate(peer string, update *UpdateMessage) erro
...
 	if channel, ok := p.peerChannels[peer]; ok {
-		channel <- update
-		return nil
+		select {
+		case channel <- update:
+			log.Infof("update was sent to channel for peer %s", peer)
+			return nil
+		default:
+			msg := fmt.Errorf("channel for peer %s is %d full", peer, len(channel))
+			log.Info(msg)
```
braginini: I would just log warning here instead of logging error with the info mag

management/server/updatechannel.go
```diff
@@ -28,8 +31,15 @@ func (p *PeersUpdateManager) SendUpdate(peer string, update *UpdateMessage) erro
...
+			return msg
```
braginini: I wouldn't return error either because it can break sending messages to other peers while just one has the issue.

braginini (changes requested):

braginini (approved):

PR #449, opened 2022-09-02
Load WgPort from the configuration file

mlsmaycon:
Added additional common blacklisted interfaces


braginini:
One thing I miss here. How would the other side know about the non-default port?
https://github.com/netbirdio/netbird/blob/553a13588bea8ac9a70c5001379a6ef6ed68fbc2/client/internal/proxy/noproxy.go#L39

client/internal/engine.go
```diff
@@ -388,15 +388,15 @@ func signalCandidate(candidate ice.Candidate, myKey wgtypes.Key, remoteKey wgtyp
 	return nil
 }
 
-func signalAuth(uFrag string, pwd string, myKey wgtypes.Key, remoteKey wgtypes.Key, s signal.Client, isAnswer bool) error {
+func signalAuth(uFrag string, pwd string, myKey wgtypes.Key, myPort int, remoteKey wgtypes.Key, s signal.Client, isAnswer bool) error {
```
braginini: I think that we should be more explicit in calling this field. 
E.g. WireGuard listen port or native port

signal/proto/signalexchange.proto
```diff
@@ -49,4 +49,5 @@ message Body {
   }
   Type type = 1;
   string payload = 2;
+  uint32 port = 3;
```
braginini: I think that we should be more explicit in calling this field.
E.g. WireGuard listen port or native port

braginini (approved):

client/internal/engine.go
```diff
@@ -770,6 +771,12 @@ func (e *Engine) receiveSignalEvents() {
 				return fmt.Errorf("wrongly addressed message %s", msg.Key)
 			}
 
+			if msg.GetBody().GetPort() != 0 {
+				conf := conn.GetConf()
+				conf.ProxyConfig.WgPort = int(msg.GetBody().GetPort())
+				conn.UpdateConf(conf)
+			}
+
```
braginini: I would rather keep this logic inside the conn.
It is not a concern of the Engine to decide.

braginini (dismissed):

mlsmaycon:
LGTM

Issue #458, opened 2022-09-08
feature: automatic user group / wildcard ACL


diegocn:
Hello. This is a feature request for netbird’s team.

As a regular user, I want all devices I add to the network to be peers within “my own group”.

As an admin, I would like to have an ACL that allows all devices added by users would be part of their own group by default, and then, all their devices would be peers.

Idea:
- all users will automatically have their own group/tag created upon first login within the organization
- all their devices will be part of this group/tag by default
- an ACL rule could set this behavior by defining a “special/wildcard” group called “@user_group” (or other better naming) that will be evaluated as each user’s own group automatically

Related to: https://github.com/netbirdio/netbird/issues/457

diegocn:
Hi @nazarewk 

The idea is not to rely on user’s name itself, but on the identity. So, a single identity would have this “self” group, and the peers added using this same identity would be automatically part of this “self” group. 

For example: I first login using my laptop, then login using my smartphone using the same identity for both, so, both peers (devices) would be able to see each other automatically.

Lochnair:
Just want to add that this is the missing feature for me to be able to move to NetBird from Tailscale.

Currently I have in my Tailscale policy:
```json
{
	"src": ["autogroup:member"],
	"dst": ["autogroup:self"],
	"ip":  ["*"],
},
```

This lets me, and other members of my network do as @diegocn describes and reach all of our own devices tied to our identities automatically without adding them to any specific group.

The closest you can get today I guess, is use the API with a script running at regular intervals to automatically create/update groups and policies, should work, but it's not exactly ideal.

masterjuggler:
I also would really appreciate having this feature

> The closest you can get today I guess, is use the API with a script running at regular intervals to automatically create/update groups and policies, should work, but it's not exactly ideal.

@Lochnair Is this something you ended up trying?

Lochnair:
@masterjuggler Afraid not. I've stuck to Tailscale for now, as it does everything I need it to.
This issue can likely be worked around like I suggested, but I also care about #4464, as I do not like the idea of lacking the option to force admin approval for new devices at all.

ZhongWwwHhh:
I think we have the same feature request. 
The member joining from IDP should can connect their own device in default.  

> an ACL rule could set this behavior by defining a “special/wildcard” group called “@user_group” (or other better naming) that will be evaluated as each user’s own group automatically

For security, the system manager might need this policy to allow/disallow this behavior. 

[...]

PR #460, opened 2022-09-09
Add SetupKey auto-groups property 

braginini:
Extend Management API to add support for setup keys auto-tagging feature that allows automatic groups assignment to peers that used the key.
This PR just adds the property to the setup key.

Works as the first feature MVP for this request https://github.com/netbirdio/netbird/issues/458#issue-1366523452


braginini:
Relates to
https://github.com/netbirdio/dashboard/pull/75

management/server/http/setupkeys.go
```diff
@@ -118,54 +89,108 @@ func (h *SetupKeys) HandleKey(w http.ResponseWriter, r *http.Request) {
...
+		log.Errorf("failed getting setup key %s under account %s %v", keyID, account.Id, err)
+		http.Redirect(w, r, "/", http.StatusInternalServerError)
 		return
-	default:
-		http.Error(w, "", http.StatusNotFound)
 	}
-}
 
-func (h *SetupKeys) GetKeys(w http.ResponseWriter, r *http.Request) {
+	writeSuccess(w, key)
```
mlsmaycon, line 110: we should call toResponseBody function before responding.

mlsmaycon, line 110: we should call toResponseBody function before responding.

management/server/http/setupkeys.go
```diff
@@ -118,54 +89,108 @@ func (h *SetupKeys) HandleKey(w http.ResponseWriter, r *http.Request) {
...
+
+	newKey := &server.SetupKey{}
+	newKey.AutoGroups = req.AutoGroups
+	newKey.Revoked = req.Revoked
+	newKey.Name = req.Name
+	newKey.Id = keyID
 
-		err = json.NewEncoder(w).Encode(respBody)
+	newKey, err = h.accountManager.SaveSetupKey(account.Id, newKey)
+	if err != nil {
```
mlsmaycon, line 154: extra if err != nil

management/server/http/setupkeys.go
```diff
@@ -118,54 +89,108 @@ func (h *SetupKeys) HandleKey(w http.ResponseWriter, r *http.Request) {
...
+		http.Redirect(w, r, "/", http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(200)
+	w.Header().Set("Content-Type", "application/json")
+
+	// should be initialized like that, otherwise json marshaller will return null
+	respBody := []*api.SetupKey{}
+	for _, key := range account.SetupKeys {
```
mlsmaycon: we can follow routes new convention and call a list keys function in the account manager.

management/server/http/setupkeys.go
```diff
@@ -190,16 +215,27 @@ func toResponseBody(key *server.SetupKey) *api.SetupKey {
 	} else {
 		state = "valid"
 	}
+
+	// the UpdatedAt field was introduced later, so there might be that some keys have a Zero value (e.g, null in the store file)
+	var updatedAt time.Time
+	if key.UpdatedAt.IsZero() {
```
mlsmaycon: this check is already performed in the account management methods

management/server/setupkey.go
```diff
@@ -83,24 +127,25 @@ func (key *SetupKey) IsOverUsed() bool {
 }
 
 // GenerateSetupKey generates a new setup key
-func GenerateSetupKey(name string, t SetupKeyType, validFor time.Duration) *SetupKey {
+func GenerateSetupKey(name string, t SetupKeyType, validFor time.Duration, autoGroups []string) *SetupKey {
```
mlsmaycon, line 130: we could keep some of these setup key functions and methods private, what do you think?

management/server/http/setupkeys.go
```diff
@@ -190,16 +215,27 @@ func toResponseBody(key *server.SetupKey) *api.SetupKey {
...
+		Key:        key.Key,
+		Name:       key.Name,
+		Expires:    key.ExpiresAt,
+		Type:       string(key.Type),
+		Valid:      key.IsValid(),
+		Revoked:    key.Revoked,
+		UsedTimes:  key.UsedTimes,
+		LastUsed:   key.LastUsed,
+		State:      state,
+		AutoGroups: key.AutoGroups,
```
mlsmaycon, line 225: should we consider returning []api.GroupMinimum?

mlsmaycon (changes requested):

management/server/http/setupkeys.go
braginini: writeSuccess function does toResponseBody

```
func writeSuccess(w http.ResponseWriter, key *server.SetupKey) {
	w.WriteHeader(200)
	w.Header().Set("Content-Type", "application/json")
	err := json.NewEncoder(w).Encode(toResponseBody(key))
	if err != nil {
		http.Error(w, "failed handling request", http.StatusInternalServerError)
		return
	}
}
```

braginini: writeSuccess function does toResponseBody

```
func writeSuccess(w http.ResponseWriter, key *server.SetupKey) {
	w.WriteHeader(200)
	w.Header().Set("Content-Type", "application/json")
	err := json.NewEncoder(w).Encode(toResponseBody(key))
	if err != nil {
		http.Error(w, "failed handling request", http.StatusInternalServerError)
		return
	}
}
```

management/server/http/setupkeys.go
braginini: ok

management/server/http/setupkeys.go
braginini: ok

management/server/http/setupkeys.go
braginini: not really for the list of setup keys, but I will make sure it is only there

management/server/setupkey.go
braginini: Good idea, but it is used somewhere and won't be available anymore.
I think that we should leave for the major refactor of the account manager. There are many functions to remove as well.

mlsmaycon (approved):

PR #479, opened 2022-09-23
Add hostname and userID to the HTTP API peer response

braginini:


management/server/http/api/openapi.yml
```diff
@@ -110,6 +110,12 @@ components:
             ssh_enabled:
               description: Indicates whether SSH server is enabled on this peer
               type: boolean
+            user_id:
```
mlsmaycon, line 102: instead of user_id, we can use the activated_by which accounts for setup key or user

mlsmaycon, line 102: instead of user_id, we can use the activated_by which accounts for setup key or user

braginini: It is tricky on the frontend to then distinguish what was what and do the id mapping. Setup key or a user? 
It is also not very common in the APIs to have such fields that are not a clear reference to a specific entity.
It then has to be a separate field for a key.
Also, I think that the setup key might belong to a user in the future.

mlsmaycon: the idea was to use type to distinguish between types and have a small label there:

`User: user@example.com` or `Setup Key: Office Servers`

But if we will switch setup key to belong to users, then it make sense to have a single field. But until then we need to have a default value indicating that a setup key was used.

braginini: Yeah, but with email/name it will be hard on the frontend to link to other views. Because we can click on that user/key and open the User tab/Setup Key tab. 
So I think that we should keep IDs and then reference on the frontend. It is more flexible.

Additionally, in order to send back the user email we will have to make a call in the management service to fetch this info from user cache, etc. This complicates the API. While frontend already does calls to user list and setup keys list

braginini: Absence of user means the setup key was used. Or I can explicitly add setupkey_id

mlsmaycon: Alright, let's remove the activated_by then

PR #487, opened 2022-09-30
External NAT IP mapping support

tkunicki:
### Support external IP mappings for host interface IPs

The external IP **_must not be behind_** a CGNAT **_and_** port-forwarding for incoming UDP packets from `WgPort` on from the external IP to `WgPort` on the host interface IP **_must be present_**. This can take form of:
- single port-forwarding rule from external IP to host interface IP
- 1:1 DNAT mapping ExternalIP to host interface IP
- NAT DMZ to host interface IP.

#### Configuration
A special value of `stun` will use the STUN server's XORMappedAddress to set the External IP value; this is for installations behind dynamic DHCP assuming the above restrictions are met.

 A single mapping will take the form of `external[/internal]`
- `external` (required):
  - external IP address or `stun` to use STUN to determine the external IP address
- `internal` (optional)
  - the host interface IP address or an interface name
  - if the "internal" argument is present and "stun" is specified for `external`, the STUN request will be generated by a connection originating from the specified host interface IP

#### Examples:
```
 "stun"                 => all interface IPs will be mapped to external IP reported by stun
 "12.34.56.78"          => all interfaces IPs will be mapped to external IP of 12.34.56.78
 "stun/eth0"            => IPv4 assigned to interface eth0 will be mapped to external IP reported by stun
 "stun/10.1.2.3"        => interface IP 10.1.2.3 will be mapped to external IP reported by stun
 "12.34.56.78/eth0"     => IPv4 assigned to interface eth0 will be mapped to external IP of 12.34.56.78
 "12.34.56.78/10.1.2.3" => interface IP 10.1.2.3 will be mapped to external IP of 12.34.56.78
```


client/internal/engine.go
```diff
@@ -857,3 +863,116 @@ func (e *Engine) receiveSignalEvents() {
...
+				// not a properly formattted IP address, maybe it's interface name?
+				internalIP, err = findIPFromInterfaceName(internal)
+				if err != nil {
+					log.Warnf("error finding interface IP for interface '%s', ignoring external mapping '%s': %v", internal, mapping, err)
+					break
+				}
+			}
+		}
+		external = split[0]
+		if external == "stun" {
```
mlsmaycon: This will only be effective if we also set the ice.AgentConfig.NAT1To1IPCandidateType to server Srflx, but at this point, will be better to skip this mode for Nat 1:1  and support only Host Candidates, so keeping the other settings options:


"12.34.56.78"
"12.34.56.78/eth0"
"12.34.56.78/10.0.0.1"

client/internal/engine.go
```diff
@@ -715,6 +720,7 @@ func (e *Engine) connWorker(conn *peer.Conn, peerKey string) {
 		e.syncMsgMux.Lock()
 		conf := conn.GetConf()
 		conf.StunTurn = append(e.STUNs, e.TURNs...)
+		conf.NATExternalIPs = e.parseNATExternalIPMappings()
```
mlsmaycon: not supporting "stun" mode will remove the need to update the config with new NATExternalIPs

client/internal/engine.go
```diff
@@ -857,3 +863,116 @@ func (e *Engine) receiveSignalEvents() {
...
+	}); err != nil {
+		return nil, err
+	}
+	if hErr != nil {
+		return nil, err 
+	}
+	return *hAddr, nil
+}
+
+func findIPFromInterfaceName(ifaceName string) (net.IP, error) {
```
mlsmaycon, line 979: We need to check if the interface name has not been blacklisted as the nat map will be ignored if the ip from the interface is not founded. See here: https://github.com/pion/ice/blob/master/gather.go#L164

client/internal/engine.go
```diff
@@ -857,3 +863,116 @@ func (e *Engine) receiveSignalEvents() {
...
 }
+
+func (e* Engine) parseNATExternalIPMappings() []string {
+	var mappedIPs []string
+	for _, mapping := range e.config.NATExternalIPs {
+		var external, internal string
+		var externalIP, internalIP net.IP
+		var err error
+		split := strings.Split(mapping, "/")
+		if len(split) > 1 {
```
mlsmaycon, line 941: we need to validate the split as it has to be between 1 and 2

client/internal/engine.go
```diff
@@ -857,3 +863,116 @@ func (e *Engine) receiveSignalEvents() {
...
+			externalIP, err = e.findNATExternalIPFromSTUN(internalIP)
+			if err != nil {
+				log.Warnf("error finding external IP via stun, ignoring external IP mapping '%s': %v", mapping, err)
+				break
+			}
+		} else {
+			externalIP = net.ParseIP(external)
+			if externalIP == nil {
+				log.Warnf("invalid external IP, ignoring external IP mapping '%s'", mapping)
+				break
```
mlsmaycon: we should return an empty list to avoid half parsed values and hard to troubleshoot cases

client/internal/engine.go
```diff
@@ -857,3 +863,116 @@ func (e *Engine) receiveSignalEvents() {
...
+	for _, mapping := range e.config.NATExternalIPs {
+		var external, internal string
+		var externalIP, internalIP net.IP
+		var err error
+		split := strings.Split(mapping, "/")
+		if len(split) > 1 {
+			internal = split[1]
+			internalIP = net.ParseIP(internal)
+			if internalIP == nil {
+				// not a properly formattted IP address, maybe it's interface name?
```
mlsmaycon: small typo with formattted

client/internal/engine.go
mlsmaycon: unless there is a case for it when switching networks, but that is a common case for laptops and possible mobile clients under wireless networks. What do you think @braginini?

mlsmaycon (changes requested):
Thanks @tkunicki for the contribution. I left some notes. 

I will ping you on slack regarding tests and the Stun comments.

mlsmaycon:
The Release checks failed because of updated go.sum file. This usually means that go.mod was updated and your local go.sum wasn't pushed

client/internal/engine.go
tkunicki: Can always reintroduce later

client/internal/engine.go
tkunicki: The result is initialized as empty list and only parsable mappings are appended to returned result.  It's possible I don't understand the comment.

mlsmaycon: the comment is more toward making the parseNATExternalIPMappings function return either a valid result list by parsing all the elements from e.config.NATExternalIPs or returns an empty list in case one element fails to parse.

mlsmaycon (approved):
Thanks @tkunicki for this contribution. @braginini I will be adding unit tests here before release

PR #504, opened 2022-10-12
Seticon only when status changes

mlsmaycon:
This prevents a memory leak with the systray lib
when setting the icon every 2 seconds causes a large memory consumption

see https://github.com/getlantern/systray/issues/135


mlsmaycon:
a permanent fix might be merged in https://github.com/getlantern/systray/pull/240

braginini (approved):

PR #517, opened 2022-10-20
Add Open-telemetry support

braginini:
This PR brings open-telemetry metrics to the 
Management service.
The Management service exposes new HTTP endpoint
/metrics on 8081 port by default.
The port can be changed by specifying 
--metrics-port PORT flag when starting the service.


management/server/http/middleware/metrics.go
```diff
@@ -0,0 +1,139 @@
...
+
+		if c, ok := m.httpRequestCounters[metricKey]; ok {
+			c.Add(m.ctx, 1)
+		}
+
+		w := WrapResponseWriter(rw)
+
+		h.ServeHTTP(w, r)
+
+		if !(w.Status() >= 200 || w.Status() < 300) {
```
mlsmaycon: this will log 3XX status, we should log error only on error. 

if w,Status() > 399 should be enough

mlsmaycon (changes requested):

management/server/http/middleware/metrics.go
braginini: true

mlsmaycon (approved):

Issue #534, opened 2022-10-31
Application icon not working in Windows


rgl:
Unfortunately, I mislead myself into thinking `Window.SetIcon` also supported the .ico format (like the systray go module) on Windows, but its not really the case, as its fails with the error:

```
decode image for window icon
2022/10/31 15:04:31   Cause: image: unknown format
2022/10/31 15:04:31   At: C:/Users/rgl/go/pkg/mod/fyne.io/fyne/v2@v2.1.4/internal/driver/glfw/window.go:266
2022/10/31 15:04:31 Fyne error:  Failed to decode image for window icon
2022/10/31 15:04:31   Cause: image: unknown format
2022/10/31 15:04:31   At: C:/Users/rgl/go/pkg/mod/fyne.io/fyne/v2@v2.1.4/internal/driver/glfw/window.go:266
```

So this needs to be fixed.

mlsmaycon:
Thanks for validating that.

PR #535, opened 2022-10-31
correctly set the windows application icon on windows

rgl:
the icon format is not really supported, so this uses a png instead.

this closes https://github.com/netbirdio/netbird/issues/534.


mlsmaycon (approved):
LGTM

PR #543, opened 2022-11-05
Feature/dns protocol

mlsmaycon:
Added DNS update protocol message

Added sync to clients

Update nameserver API with new fields 

Added default NS groups

Added new dns-name flag for the management service append to peer DNS label


management/server/http/api/openapi.yml
```diff
@@ -136,6 +136,9 @@ components:
             ui_version:
               description: Peer's desktop UI version
               type: string
+            dns_label:
+              description: Peer's DNS Label
```
braginini: I think that it should be more clear what this exactly means.

braginini (changes requested):

braginini (approved):

PR #558, opened 2022-11-08
Replace gRPC errors in business logic with internal ones

braginini:


management/server/http/util/util.go
```diff
@@ -0,0 +1,103 @@
...
+
+// WriteError converts an error to an JSON error response.
+// If it is known internal error of type server.Error then it sets the messages from the error, a generic message otherwise
+func WriteError(err error, w http.ResponseWriter) {
+	errStatus, ok := status.FromError(err)
+	httpStatus := http.StatusInternalServerError
+	msg := "internal server error"
+	if ok {
+		switch errStatus.Type() {
+		case status.UserAlreadyExists:
```
mlsmaycon, line 82: I think we should be more clear on Already exists error. Maybe we should return 409 or 422

management/server/http/users.go
```diff
@@ -31,33 +29,34 @@ func NewUserHandler(accountManager server.AccountManager, authAudience string) *
 // UpdateUser is a PUT requests to update User data
 func (h *UserHandler) UpdateUser(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodPut {
-		http.Error(w, "", http.StatusBadRequest)
+		util.WriteError(status.Errorf(status.InvalidArgument, "wrong HTTP method"), w)
```
mlsmaycon: this case should never happen because of mux handler assignment. But in any case, we can return 405 here.

management/server/http/users.go
```diff
@@ -67,40 +66,36 @@ func (h *UserHandler) UpdateUser(w http.ResponseWriter, r *http.Request) {
...
-	writeJSONObject(w, toUserResponse(newUser))
+	util.WriteJSONObject(w, toUserResponse(newUser))
 
 }
 
 // CreateUserHandler creates a User in the system with a status "invited" (effectively this is a user invite).
 func (h *UserHandler) CreateUserHandler(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodPost {
-		http.Error(w, "", http.StatusNotFound)
+		util.WriteError(status.Errorf(status.InvalidArgument, "wrong HTTP method"), w)
```
mlsmaycon: this case should never happen because of mux handler assignment. But in any case, we can return 405 here.

management/server/http/users.go
```diff
@@ -111,37 +106,30 @@ func (h *UserHandler) CreateUserHandler(w http.ResponseWriter, r *http.Request)
...
-	writeJSONObject(w, toUserResponse(newUser))
+	util.WriteJSONObject(w, toUserResponse(newUser))
 }
 
 // GetUsers returns a list of users of the account this user belongs to.
 // It also gathers additional user data (like email and name) from the IDP manager.
 func (h *UserHandler) GetUsers(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodGet {
-		http.Error(w, "", http.StatusBadRequest)
+		util.WriteError(status.Errorf(status.InvalidArgument, "wrong HTTP method"), w)
```
mlsmaycon: this case should never happen because of mux handler assignment. But in any case, we can return 405 here.

mlsmaycon (changes requested):

PR #564, opened 2022-11-15
Fix docker network interface filter

Genteure:
docker network address are assigned on interfaces with name `br-${NETWORK_ID[:12]}`, docker also creates interfaces that start with `veth` but does not assign IPv4 address to them.
I didn't find any docker documentation backing this but this should be easy to verify.


mlsmaycon (approved):
LGTM, thanks @Genteure

PR #584, opened 2022-11-26
Sync peers FQDN

mlsmaycon:
Use stdout and stderr log path only if on Linux and attempt to create the path

Update status system with FQDN fields and status command to display the domain names of remote and local peers

Set some DNS logs to tracing


client/internal/engine.go
```diff
@@ -766,6 +766,10 @@ func (e *Engine) addNewPeer(peerConfig *mgmProto.RemotePeerConfig) error {
 
 		go e.connWorker(conn, peerKey)
 	}
+	err := e.statusRecorder.UpdatePeerFQDN(peerKey, peerConfig.Fqdn)
```
braginini, line 782: What will happen if fqdn is not there (e.g., old server

Also, what will happen if fqdn changes, will it be updated?

braginini (changes requested):

client/internal/engine.go
mlsmaycon: > What will happen if fqdn is not there (e.g., old server

grpc method will return an empty string which is ok

> Also, what will happen if fqdn changes, will it be updated?

Good point. I pushed a fix. Basically, we will always update the peer's FQDN when they won't follow the modify process.

braginini (dismissed):

braginini (approved):

Issue #463, opened 2022-09-10
routes enhancement:  add custom routes to all peers or to a group of peers


jeremyd:
To route traffic from a group of peers toward a specific gateway.  For example, in the use case of exit nodes where traffic crosses to the public internet, each peer that was a client of that exit node would need routes for the external addresses of the wireguard endpoints.

To accomplish this you could assign a custom route consisting of 
```
ip route add <COTURN/NETBIRD MGMT/SIGNAL IP>/32 via <LOCAL NET GATEWAY> dev <INTERFACE>
```
so being able to specify the
* ip range
* gateway
* interface

For a group of peers, or to all peers, this would accomplish the route distribution entirely via netbird!

mlsmaycon:
Thanks for submitting this suggestion @jeremyd 

Regarding the functionality it will need a few more adjustments to avoid forwarding peer connection traffic via the exit node and entering the chicken and egg paradox.

We are looking at a combination of FWMARK and IP RULE for that in Linux, we also want to avoid multiple rules and namespace so we don't loose easy visibility on what we are doing on the peer OS.

jeremyd:
> Regarding the functionality it will need a few more adjustments to avoid forwarding peer connection traffic via the exit node and entering the chicken and egg paradox.
> 
> We are looking at a combination of FWMARK and IP RULE for that in Linux, we also want to avoid multiple rules and namespace so we don't loose easy visibility on what we are doing on the peer OS.

yes, perhaps with FWMARK and IP RULE the routes aren't the best option to use.  Good point.  I'm unsure if anyone would want custom routes in other cases besides exit nodes (or in combination with exit nodes).  I really am looking forward to exit nodes!!  :)  Thank you.

MohammedNoureldin:
Hi @jeremyd,

I am a bit confused.

I am trying to get route all connections of my clients (Windows or Linux peers, for example) to a specific exit node. For this, probably I have to setup a routing peer on that node.

But how should the connections from the clients (peers) supposed to get redirected to that exit node (or accurately to the routing peer on that exit node)?

Should every client (peer) configure his gateway manually to match the IP address of the routing peer?

Or can this gateway adaption happen automatically by Netbird client?

PR #606, opened 2022-12-04
Add network routes distribution groups

mlsmaycon:
## Describe your changes
Updated tests, API, and account manager methods

Sync routes to peers in the distribution groups

Added store upgrade by adding the All group to routes that don't have them

Relates to #463

### Checklist
- [x] Will update documentation in another PR


management/server/file_store.go
```diff
@@ -105,6 +105,19 @@ func restore(file string) (*FileStore, error) {
 		if len(existingLabels) != len(account.Peers) {
 			addPeerLabelsToAccount(account, existingLabels)
 		}
+
+		allGroup, err := account.GetGroupAll()
+		if err != nil {
+			log.Errorf("unable to find the All group, this should happen only when migrate from a version that didn't supported groups. Error: %v", err)
```
braginini: correct grammar : didn't support

management/server/peer.go
```diff
@@ -293,8 +293,9 @@ func (am *DefaultAccountManager) GetNetworkMap(peerPubKey string) (*NetworkMap,
 		return nil, err
 	}
 
-	aclPeers := am.getPeersByACL(account, peerPubKey)
-	routesUpdate := account.GetPeersRoutes(append(aclPeers, account.Peers[peerPubKey]))
+	aclPeers := account.getPeersByACL(peerPubKey)
+	aclRoutes := account.GetPeersRoutes(append(aclPeers, account.Peers[peerPubKey]))
+	routesUpdate := account.filterRoutesByGroups(peerPubKey, aclRoutes)
```
braginini: maybe it makes sense to put the logic of  filterRoutesByGroups to GetPeersRoutes?
do we have any separate logic that would use GetPeersRoutes alone without filterRoutesByGroups?

management/server/account.go
```diff
@@ -156,6 +156,26 @@ func (a *Account) GetPeersRoutes(givenPeers []*Peer) []*route.Route {
...
+			_, found := groupListMap[groupID]
+			if found {
+				filteredRoutes = append(filteredRoutes, r)
+				break
+			}
+		}
+	}
+	return filteredRoutes
+}
+
```
braginini: I would either pass both peer routes and peer groups to the method or fetch them both from the account because it looks a bit odd - one thing is passed as a property (routes), the other one fetched from the account (groups)

braginini: I'd also add a comment describing the result of the function and what it does. It is hard to understand the purpose of this "filter"

management/server/peer.go
mlsmaycon: I've refactored the methods but kept the group filtering in a separated method just in favor of readability

management/server/account.go
mlsmaycon: Noted. I've refactored the method and included your suggestion, plus I've added docs

braginini (approved):

Issue #598, opened 2022-12-02
bug: Server should not create ip routes to networks it is routing to


nazarewk:
I am trying to connect from:
- NixOS machine
- connected to IPv4 only internet
to:
- Bottlerocket ECS peer on AWS
- IPv6-enabled, IPv4-available

It seems like the machine fails after a while and I cannot reconnect until I reboot it.

Originally thought it was on Netbird side, but I'm not so sure anymore.

UPDATE: high availability `ip route` entry for it's own VPC CIDR range is at fault, see https://github.com/netbirdio/netbird/issues/598#issuecomment-1337629243 below

<details>
<summary>Logs</summary>

```
Dec 02 09:32:58 nazarewk-krul systemd[1]: Started A WireGuard-based mesh network that connects your devices into a single private network.
Dec 02 09:32:58 nazarewk-krul netbird[392223]: time="2022-12-02T09:32:58+01:00" level=info msg="starting Netbird service" file="service_controller.go:23"
Dec 02 09:32:58 nazarewk-krul netbird[392223]: time="2022-12-02T09:32:58+01:00" level=info msg="started daemon server: /var/run/netbird/sock" file="service_controller.go:63"
Dec 02 09:32:58 nazarewk-krul netbird[392223]: time="2022-12-02T09:32:58+01:00" level=debug msg="conecting to the Management service api.wiretrustee.com:443" file="connect.go:83"
Dec 02 09:32:59 nazarewk-krul netbird[392223]: time="2022-12-02T09:32:59+01:00" level=debug msg="connected to the Management service api.wiretrustee.com:443" file="connect.go:88"
Dec 02 09:32:59 nazarewk-krul netbird[392223]: time="2022-12-02T09:32:59+01:00" level=debug msg="connected to Signal Service: signal2.wiretrustee.com:443" file="grpc.go:78"
Dec 02 09:32:59 nazarewk-krul netbird[392223]: time="2022-12-02T09:32:59+01:00" level=info msg="using kernel WireGuard" file="iface_linux.go:21"
Dec 02 09:32:59 nazarewk-krul netbird[392223]: time="2022-12-02T09:32:59+01:00" level=debug msg="adding device: wt7" file="iface_linux.go:57"
Dec 02 09:32:59 nazarewk-krul netbird[392223]: time="2022-12-02T09:32:59+01:00" level=debug msg="adding address 100.77.44.6/16 to interface: wt7" file="iface_linux.go:108"
[...]

PR #612, opened 2022-12-06
HA Network Routes: prevent routing directly-accessible networks through VPN interface

nazarewk:
## Describe your changes

It prevents Server from adding rules to route directly accessible/own networks through VPN when the route is in HA mode (has more than 1 entry):

1. the code was checking membership against a single Network Route entry with equality operator on `Peer` attribute:
	1. there is 1 Route per Peer
	2. second (HA) Route entry did not pass the check because it was not the server
	3. rule for routing the traffic through VPN interface is created
	4. host X is routing through host Y
	5. host Y is routing through host X
	6. (i assume) we have a routing loop passing packets back and forth between HA hosts instead of going to the real network
2. The code fixes above issue by checking `NetID` instead of `Peer` values

see the linked issue for more details

fixes: https://github.com/netbirdio/netbird/issues/598

### Checklist
- [x] Is it a bug fix
- [x] Extended the README / documentation, if necessary

### DIFF
--- a/client/internal/routemanager/manager.go
+++ b/client/internal/routemanager/manager.go
@@ -3,13 +3,14 @@ package routemanager
 import (
 	"context"
 	"fmt"
+	"runtime"
+	"sync"
+
 	"github.com/netbirdio/netbird/client/status"
 	"github.com/netbirdio/netbird/client/system"
 	"github.com/netbirdio/netbird/iface"
 	"github.com/netbirdio/netbird/route"
 	log "github.com/sirupsen/logrus"
-	"runtime"
-	"sync"
 )
 
 // Manager is a route manager interface
@@ -147,10 +148,18 @@ func (m *DefaultManager) UpdateRoutes(updateSerial uint64, newRoutes []*route.Ro
 
 		newClientRoutesIDMap := make(map[string][]*route.Route)
 		newServerRoutesMap := make(map[string]*route.Route)
+		ownNetworkIDs := make(map[string]bool)
 
 		for _, newRoute := range newRoutes {
-			// only linux is supported for now
 			if newRoute.Peer == m.pubKey {
+				ownNetworkIDs[getHANetworkID(newRoute)] = true
+			}
+		}
+
+		for _, newRoute := range newRoutes {
+			networkID := getHANetworkID(newRoute)
+			if ownNetworkIDs[networkID] {
+				// only linux is supported for now
 				if runtime.GOOS != "linux" {
 					log.Warnf("received a route to manage, but agent doesn't support router mode on %s OS", runtime.GOOS)
 					continue
@@ -164,8 +173,7 @@ func (m *DefaultManager) UpdateRoutes(updateSerial uint64, newRoutes []*route.Ro
 						system.NetbirdVersion(), newRoute.Network)
 					continue
 				}
-				clientNetworkID := getClientNetworkID(newRoute)
-				newClientRoutesIDMap[clientNetworkID] = append(newClientRoutesIDMap[clientNetworkID], newRoute)
+				newClientRoutesIDMap[networkID] = append(newClientRoutesIDMap[networkID], newRoute)
 			}
 		}
 
--- a/client/internal/routemanager/manager.go
+++ b/client/internal/routemanager/manager.go
@@ -151,21 +151,21 @@ func (m *DefaultManager) UpdateRoutes(updateSerial uint64, newRoutes []*route.Ro
 		ownNetworkIDs := make(map[string]bool)
 
 		for _, newRoute := range newRoutes {
-			if newRoute.Peer == m.pubKey {
-				ownNetworkIDs[getHANetworkID(newRoute)] = true
-			}
-		}
-
-		for _, newRoute := range newRoutes {
 			networkID := getHANetworkID(newRoute)
-			if ownNetworkIDs[networkID] {
+			if newRoute.Peer == m.pubKey {
+				ownNetworkIDs[networkID] = true
 				// only linux is supported for now
 				if runtime.GOOS != "linux" {
 					log.Warnf("received a route to manage, but agent doesn't support router mode on %s OS", runtime.GOOS)
 					continue
 				}
 				newServerRoutesMap[newRoute.ID] = newRoute
-			} else {
+			}
+		}
+
+		for _, newRoute := range newRoutes {
+			networkID := getHANetworkID(newRoute)
+			if !ownNetworkIDs[networkID] {
 				// if prefix is too small, lets assume is a possible default route which is not yet supported
 				// we skip this route management
 				if newRoute.Network.Bits() < 7 {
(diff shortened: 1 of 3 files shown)


client/internal/routemanager/manager.go
```diff
@@ -147,10 +148,18 @@ func (m *DefaultManager) UpdateRoutes(updateSerial uint64, newRoutes []*route.Ro
 
 		newClientRoutesIDMap := make(map[string][]*route.Route)
 		newServerRoutesMap := make(map[string]*route.Route)
+		ownNetIDs := make(map[string]bool)
 
 		for _, newRoute := range newRoutes {
-			// only linux is supported for now
 			if newRoute.Peer == m.pubKey {
+				ownNetIDs[newRoute.NetID] = true
```
mlsmaycon: An HA route is composed of multiple routes that have the same combination of `newRoute.NetID` and the `newRoute. Network`. With that in mind,  The best way to index the map will be with:

`newRoute.NetID + "-" +  newRoute.Network.String()`

You can also use the function `getClientNetworkID`, which generates this HA combination. It makes sense to rename it to `getHANetworkID`


Another point. It will be better to move the OS check and the route registration to this loop, as when creating firewall rules, we use the route ID to comment, and track route and firewall operations. Plus we shouldn't register other peer's routes from the HA group as ours as well.

client/internal/routemanager/manager.go
```diff
@@ -147,10 +148,18 @@ func (m *DefaultManager) UpdateRoutes(updateSerial uint64, newRoutes []*route.Ro
...
+			}
+		}
+
+		for _, newRoute := range newRoutes {
+			// only linux is supported for now
+			if _, ok := ownNetIDs[newRoute.NetID]; ok {
```
mlsmaycon: Moving the server route registration code to the previous loop allows having only code for non-owned network IDs.

Also, as the ownNetIDs values are booleans, you can skip the retrieved check and only check for the map lookup value. e.g.:

```Go
if !ownNetIDs[newRoute.NetID] {
   // client code
   ...
}```

File: client/internal/routemanager/manager_test.go [hunk 3]
```diff
@@ -336,6 +337,55 @@ func TestManagerUpdateRoutes(t *testing.T) {
 			serverRoutesExpected:          0,
 			clientNetworkWatchersExpected: 0,
 		},
+		{
+			name: "HA server should not route it's own networks through VPN",
```
mlsmaycon: I suggest the following name for this test:

`HA server should not register routes from the same HA group`

mlsmaycon (changes requested):
Thank you so much for your PR @nazarewk, it is definitely helping with the case you described. 

I have a few notes that I took into consideration on my request for changes: 

In multi-hop routing cases, you want to route traffic via the VPN interface as the routing peer might be an intermediate router. In that case, will make sense to have different network routes for each hop.

Also, to document, with PR #606 we will be able to prevent similar case by using different distribution groups for routes, but only when different groups are used with routes and peers.

client/internal/routemanager/manager.go
nazarewk: Done the first part, but I am not sure about second part as I'm not familiar with the codebase.

To me it looks like moving the OS check to the first loop would result in those routes being created as `newClientRoutesIDMap` therefore it is not desired?

client/internal/routemanager/manager.go
nazarewk: ~~wouldn't it throw some kind of error when key is missing?~~

nevermind, below works fine:
```go
package main
import "fmt"

func main() {
    ownNetworkIDs := make(map[string]bool)
    ownNetworkIDs["a"] = true
    fmt.Println(ownNetworkIDs["a"])
    fmt.Println(ownNetworkIDs["c"])
    fmt.Println(ownNetworkIDs["b"])
}
```

client/internal/routemanager/manager.go
mlsmaycon: It won't, as the other routes that we will drop should be routed by the other peers in an HA group. Which didn't happen in the previous version of the code, and they ended up registered as client routes which caused the loop behavior you are solving now.

nazarewk: Is this what you meant? [`741bfe4` (#612)](https://github.com/netbirdio/netbird/pull/612/commits/741bfe46945eacbc499a22c40038bf519460faea)

mlsmaycon: Almost, just in the second loop we don't need to register the server routes anymore, so we can just validate if it is not a route the server owns and register as new client route: 

From:
```Go
if ownNetworkIDs[networkID] {
				newServerRoutesMap[newRoute.ID] = newRoute
			} else {
```	
To:
```Go
if !ownNetworkIDs[networkID] {
      // rest of client code
       newClientRoutesIDMap[networkID] = append(newClientRoutesIDMap[networkID], newRoute)
```

nazarewk: So we're effectively deleting `newServerRoutesMap` handling alltogether?

mlsmaycon: No, we are moving it to the first loop as there we will check for owned routes and we can register them already. So checking again in the second loop is not needed

mlsmaycon: Here is an example of what I meant in the previous comments:
```Go
		for _, newRoute := range newRoutes {
			if newRoute.Peer == m.pubKey {
				// only linux is supported for now
				if runtime.GOOS != "linux" {
					log.Warnf("received a route to manage, but agent doesn't support router mode on %s OS", runtime.GOOS)
					continue
				}
				ownNetworkIDs[getHANetworkID(newRoute)] = true
                                newServerRoutesMap[newRoute.ID] = newRoute
			}
		}
```

nazarewk: After introducing this change the test now creates 2 Server routes instead of 3, is it intended?

It is skipping other peer's route alltogether, I guess it is OK

mlsmaycon: Yes, it is; the change is skipping the route with ID `r1` as the routing peer is a remote peer.

client/internal/routemanager/manager.go
```diff
@@ -147,25 +148,32 @@ func (m *DefaultManager) UpdateRoutes(updateSerial uint64, newRoutes []*route.Ro
 
 		newClientRoutesIDMap := make(map[string][]*route.Route)
 		newServerRoutesMap := make(map[string]*route.Route)
+		ownNetworkIDs := make(map[string]bool)
 
 		for _, newRoute := range newRoutes {
-			// only linux is supported for now
+			networkID := getHANetworkID(newRoute)
 			if newRoute.Peer == m.pubKey {
+				// only linux is supported for now
```
mlsmaycon, line 157: We should register the owned network before checking the OS. This is the cause of the failing tests

mlsmaycon (approved):
@nazarewk thank you for the changes and hanging thought all the requests.

PR #616, opened 2022-12-08
Add device flow scope.

aazf:
## Describe your changes

Most of IDP require a base scope, so add the openid as the base scope.

Na

### Checklist
- [x] Is it a bug fix


mlsmaycon:
@aazf thanks for the contribution, we will test with auth0 and keycloak and then approve.

Can you share which idp you had the problem with so we can support users that run into the same issue.

mlsmaycon:
> @mlsmaycon https://github.com/dexidp/dex

thanks @aazf 

I've enabled the workflows, can you update the failing tests?

mlsmaycon:
Thanks @aazf

mlsmaycon:
@aazf the Docker Compose Linux test fails because of some missing secrets.

braginini:
> @mlsmaycon So how to fix it, or everything is fine.

That's okay. All good. We can ignore it now.
The docker compose test requires some secrets that are not available in forks. These are private to the netbird org.

mlsmaycon (approved):

PR #654, opened 2023-01-16
Involve foreground mode switch for up cmd

pappz:
Add new --foreground-mode command line parameter
for 'up' cmd instead of existing log-file workaround

## Describe your changes

During debug the service, I am interested in output of the service. With this option it is easily configurable.
I tried to follow the current code style pattern, nevertheless I recommend to organize the content of the up.go into a new struct and register the upCmd into the cobra via a new function like this ```func registerUpCmd(rootCmd *cobra.Command)```. With it you can avoid global variables and ```init()``` functions.

### Checklist
- [ ] Is it a bug fix

### DIFF
--- a/client/cmd/up.go
+++ b/client/cmd/up.go
@@ -14,8 +14,8 @@ import (
 )
 
 var (
-	daemonModeOff bool
-	upCmd         = &cobra.Command{
+	foregroundMode bool
+	upCmd          = &cobra.Command{
 		Use:   "up",
 		Short: "install, login and start Netbird client",
 		RunE:  upFunc,
@@ -23,7 +23,7 @@ var (
 )
 
 func init() {
-	upCmd.PersistentFlags().BoolVar(&daemonModeOff, "daemon-off", false, "start service in foreground")
+	upCmd.PersistentFlags().BoolVarP(&foregroundMode, "foreground-mode", "F", false, "start service in foreground")
 }
 
 func upFunc(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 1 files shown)


client/cmd/up.go
```diff
@@ -13,121 +13,137 @@ import (
...
+	upCmd         = &cobra.Command{
+		Use:   "up",
+		Short: "install, login and start Netbird client",
+		RunE:  upFunc,
+	}
+)
 
-		cmd.SetOut(cmd.OutOrStdout())
+func init() {
+	upCmd.PersistentFlags().BoolVar(&daemonModeOff, "daemon-off", false, "start service in foreground")
```
mlsmaycon: It is more explicit if we call the flag `foreground-mode` as some users aren't aware of the service being called daemon. 

We can also use BoolVarP and add `F`  as shorthand for the flag.

mlsmaycon (changes requested):
Thanks @pappz for the contribution, this is a long time waiting change. 

I've left a small comment on the flag's name. If you feel is needed, you can also apply the rename to the variable `daemonModeOff`.

client/cmd/up.go
pappz: You are right!
Do you wish ```F``` in upper case?

mlsmaycon: Yes, I think the upper case resembles better with the running mode.

mlsmaycon (approved):
Looks great! Thanks for the contribution.

pappz:
My pleasure

PR #659, opened 2023-01-20
allow specifying --wg-iface and --wg-port

nazarewk:
## Describe your changes

I am working on setting up multiple instances of Netbird on same machine, I need a way to tell the deamon to input different interfaces/ports to those without modifying files by hand.

I got it to work at https://github.com/nazarewk-iac/nix-configs/commit/aeefbf5c

### Checklist
- [ ] Is it a bug fix

### DIFF
--- a/client/cmd/root.go
+++ b/client/cmd/root.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
+	"github.com/netbirdio/netbird/iface"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -49,6 +50,8 @@ var (
 	preSharedKey            string
 	natExternalIPs          []string
 	customDNSAddress        string
+	wgIface                 string
+	wgPort                  int
 	rootCmd                 = &cobra.Command{
 		Use:          "netbird",
 		Short:        "",
@@ -105,19 +108,32 @@ func init() {
 	rootCmd.AddCommand(sshCmd)
 	serviceCmd.AddCommand(runCmd, startCmd, stopCmd, restartCmd) // service control commands are subcommands of service
 	serviceCmd.AddCommand(installCmd, uninstallCmd)              // service installer commands are subcommands of service
-	upCmd.PersistentFlags().StringSliceVar(&natExternalIPs, externalIPMapFlag, nil,
-		`Sets external IPs maps between local addresses and interfaces.`+
-			`You can specify a comma-separated list with a single IP and IP/IP or IP/Interface Name. `+
-			`An empty string "" clears the previous configuration. `+
-			`E.g. --external-ip-map 12.34.56.78/10.0.0.1 or --external-ip-map 12.34.56.200,12.34.56.78/10.0.0.1,12.34.56.80/eth1 `+
-			`or --external-ip-map ""`,
-	)
-	upCmd.PersistentFlags().StringVar(&customDNSAddress, dnsResolverAddress, "",
-		`Sets a custom address for NetBird's local DNS resolver. `+
-			`If set, the agent won't attempt to discover the best ip and port to listen on. `+
-			`An empty string "" clears the previous configuration. `+
-			`E.g. --dns-resolver-address 127.0.0.1:5053 or --dns-resolver-address ""`,
-	)
+
+	for _, cmd := range []*cobra.Command{upCmd, serviceCmd} {
+		cmd.PersistentFlags().StringSliceVar(&natExternalIPs, externalIPMapFlag, nil,
+			`Sets external IPs maps between local addresses and interfaces.`+
+				`You can specify a comma-separated list with a single IP and IP/IP or IP/Interface Name. `+
+				`An empty string "" clears the previous configuration. `+
+				`E.g. --external-ip-map 12.34.56.78/10.0.0.1 or --external-ip-map 12.34.56.200,12.34.56.78/10.0.0.1,12.34.56.80/eth1 `+
+				`or --external-ip-map ""`,
+		)
+		cmd.PersistentFlags().StringVar(&customDNSAddress, dnsResolverAddress, "",
+			`Sets a custom address for NetBird's local DNS resolver. `+
+				`If set, the agent won't attempt to discover the best ip and port to listen on. `+
+				`An empty string "" clears the previous configuration. `+
+				`E.g. --dns-resolver-address 127.0.0.1:5053 or --dns-resolver-address ""`,
+		)
+		// do not set default, because it will override whatever custom value was set
+		cmd.PersistentFlags().StringVar(&wgIface, "wg-iface", "",
+			`WireGuard interface name to use for Netbird client instance. `+
+				fmt.Sprintf("Defaults to %s", iface.WgInterfaceDefault)+
+				`MacOS restricts the tun interface name to utun[0-9]+ pattern. e.g., utun100`)
+		// do not set default, because it will override whatever custom value was set
+		cmd.PersistentFlags().IntVar(&wgPort, "wg-port", 0,
+			"WireGuard port number to use for Netbird client instance. "+
+				fmt.Sprintf("Defaults to %s", iface.WgInterfaceDefault))
+	}
+
 }
 
 // SetupCloseHandler handles SIGTERM signal and exits with success
--- a/client/internal/config.go
+++ b/client/internal/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/exp/slices"
 	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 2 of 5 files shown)


nazarewk:
I guess adding entries to interfaces blacklist would be handy too

nazarewk:
> hello @nazarewk, thanks for the contribution.
> 
> The Pr is still with a Draft status. I have a few comments to add and some requests as well, but I wonder if you want to add more to it.

I have what I personally need in the PR, but I'm up for comments.

client/cmd/root.go
```diff
@@ -94,6 +97,8 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&logFile, "log-file", defaultLogFile, "sets Netbird log path. If console is specified the the log will be output to stdout")
 	rootCmd.PersistentFlags().StringVarP(&setupKey, "setup-key", "k", "", "Setup key obtained from the Management Service Dashboard (used to register peer)")
 	rootCmd.PersistentFlags().StringVarP(&preSharedKey, "preshared-key", "p", "", "Sets Wireguard PreSharedKey property. If set, then only peers that have the same key can communicate.")
+	rootCmd.PersistentFlags().StringVar(&wgIface, "wg-iface", iface.WgInterfaceDefault, "WireGuard interface name to use for Netbird client instance")
```
mlsmaycon: We can add a warning for macOS users:

```
MacOS restricts the tun interface name to utun[0-9]+ pattern. e.g., utun100

File: client/server/server.go [hunk 2]
```diff
@@ -45,13 +46,15 @@ type oauthAuthFlow struct {
 }
 
 // New server instance constructor.
-func New(ctx context.Context, managementURL, adminURL, configPath, logFile string) *Server {
+func New(ctx context.Context, managementURL, adminURL, configPath, logFile, wgInterfaceName string, wgPort int) *Server {
```
mlsmaycon: We are avoiding adding config flags to the daemon service New function because it requires a reinstallation to update them.

Instead, let's also handle the update the internal.ReadConfig function. This way we can use `netbird up --wg-iface wt0 --wg-port 54321` and updating that will require just netbird down and up again.

client/internal/config.go
```diff
@@ -121,7 +124,15 @@ func createNewConfig(input ConfigInput) (*Config, error) {
 		config.AdminURL = newURL
 	}
 
-	config.IFaceBlackList = defaultInterfaceBlacklist
+	if input.WgIface != "" {
```
mlsmaycon, line 185: We need to add similar checks to the ReadConfig function so updates are possible in the up command.

client/cmd/login.go
```diff
@@ -43,6 +44,8 @@ var loginCmd = &cobra.Command{
 				AdminURL:      adminURL,
 				ConfigPath:    configPath,
 				PreSharedKey:  &preSharedKey,
+				WgIface:       wgIface,
```
mlsmaycon: We don't need to add the ports in the login command as it doesn't start the daemon. Keeping on up should be enough and prevents confusion.

client/cmd/root.go
mlsmaycon: let's keep the new flags as part of the up command only. This will prevent confusion too.

mlsmaycon (changes requested):
thanks @nazarewk. I left some comments regarding the input; we are moving things a bit to be configured through the up command to prevent reinstallation of the daemon.

client/cmd/root.go
nazarewk: isn't it also relevant for login?

client/internal/config.go
nazarewk: updated the code, is this what you meant with modifying readConfig?

mlsmaycon: yes, you just missed the port check.

You can add something like this:
```Go
        if input.WgPort != 0 && input.WgPort !=  config.WgPort {
		config.WgPort =  input.WgPort 
		refresh = true
	}
	
	// previous version config upgrade check
	if config.WgPort == 0 {
		config.WgPort = iface.DefaultWgPort
		refresh = true
	}
```

client/internal/config.go
```diff
@@ -203,6 +215,17 @@ func ReadConfig(input ConfigInput) (*Config, error) {
 		config.WgPort = iface.DefaultWgPort
 		refresh = true
 	}
+
+	if input.WgIface == "" {
+		config.WgIface = iface.WgInterfaceDefault
+		refresh = true
+	}
+
+	if !slices.Contains(config.IFaceBlackList, config.WgIface) {
```
mlsmaycon, line 272: You can probably store the previous WgIface and if changed, loop through the config.IFaceBlackList and replace the matching element index with the new interface.

nazarewk:
what is the difference between `netbird up` and `netbird service run/start/stop`? After moving flags to `up` I'm not getting those reflected in [my service file](https://github.com/nazarewk-iac/nix-configs/blob/402b181e095581eb60df5cde87a45abfd4932714/modules/networking/netbird/default.nix#L60-L60).

mlsmaycon:
> what is the difference between `netbird up` and `netbird service run/start/stop`? After moving flags to `up` I'm not getting those reflected in [my service file](https://github.com/nazarewk-iac/nix-configs/blob/402b181e095581eb60df5cde87a45abfd4932714/modules/networking/netbird/default.nix#L60-L60).

The service start and stop only controls the OS system start and stop of a service. The `run` starts netbird as a daemon process, which listens for commands and updates the process execution. 

I saw that you updated the flags to be added with the `service run`, but the problem there is that to update a configuration you need to reinstall the client for most installations. otherwise, every restart will restore any update made with `up`

client/internal/config.go
```diff
@@ -196,20 +208,64 @@ func ReadConfig(input ConfigInput) (*Config, error) {
...
+			}
+			newBlacklist = append(newBlacklist, ifc)
+		}
+		if !replaced {
+			newBlacklist = append(newBlacklist, config.WgIface)
+		}
+		log.Infof("prepending Wireguard Interface Name to interface blacklist %#v, updated to %v (old value %v)", config.WgIface, newBlacklist, config.IFaceBlackList)
+		config.IFaceBlackList = newBlacklist
+		refresh = true
+	}
```
mlsmaycon, line 288:
```suggestion
	if oldIface != config.WgIface {
		for i, ifc := range config.IFaceBlackList {
			if ifc == oldIface {
				ifc = config.WgIface
				config.IFaceBlackList[i] = config.WgIface
				log.Infof("replaced Wireguard Interface Name in the interface blacklist. old name %s, new name", oldIface, config.WgIface)
				refresh = true
				break
			}
		}
	}
```

client/internal/config.go
```diff
@@ -196,20 +208,64 @@ func ReadConfig(input ConfigInput) (*Config, error) {
...
+		log.Infof("new Wireguard Interface Name provided, updated to %#v (old value %#v)", input.WgIface, config.WgIface)
+		config.WgIface = input.WgIface
+		refresh = true
+	}
+
+	if input.WgPort != 0 && input.WgPort != config.WgPort {
+		log.Infof("new Wireguard Port provided, updated to %#v (old value %#v)", input.WgPort, config.WgPort)
+		config.WgPort = input.WgPort
+		refresh = true
+	}
```
mlsmaycon, line 270:
```suggestion
```

client/internal/config.go
```diff
@@ -196,20 +208,64 @@ func ReadConfig(input ConfigInput) (*Config, error) {
...
+		log.Infof("generated config.SSHKey")
 		refresh = true
 	}
 
-	if config.WgPort == 0 {
+	if config.WgPort == 0 && input.WgPort == 0 {
+		log.Infof("using default Wireguard Port: %#v ", iface.DefaultWgPort)
 		config.WgPort = iface.DefaultWgPort
 		refresh = true
 	}
```
mlsmaycon, line 250:
```suggestion
         if input.WgPort != 0 && input.WgPort != config.WgPort {
		log.Infof("new Wireguard Port provided, updated to %d (old value %d)", input.WgPort, config.WgPort)
		config.WgPort = input.WgPort
		refresh = true
	}
	// previous version config upgrade check because WgPort didn't exist
	if config.WgPort == 0 {
		log.Infof("using default Wireguard Port: %d", iface.DefaultWgPort)
		config.WgPort = iface.DefaultWgPort
		refresh = true
	}
```

client/internal/config.go
mlsmaycon: Sorry I wasn't clear on my idea. I've left a code suggestion.

mlsmaycon:
@nazarewk, my apologies, I forgot about the protocol update required to pass the new flags from the client to the daemon. 

You need to update the protocol file here [/client/proto/daemon.proto#L31](https://github.com/netbirdio/netbird/blob/main/client/proto/daemon.proto#L31) and then you can run the [generate.sh](https://github.com/netbirdio/netbird/blob/main/client/proto/generate.sh)

Once that is done, you need to retrieve them in the https://github.com/netbirdio/netbird/blob/main/client/server/server.go#L158 and pass to the configuration calls

With that updating the port will look like this:
```
netbird down
netbird up --wg-port 5000
```

mlsmaycon:
@nazarewk let us know if the requests were too much. I would be happy to join the effort and help you in your branch

nazarewk:
@mlsmaycon I can't continue this right now (maybe next week) and I'm not planning to become Netbird architecture expert so feel free to help and fix the things you mentioned :)

mlsmaycon:
@nazarewk, thanks for the contribution. We have decided to move forward with the flag support in the following PR: https://github.com/netbirdio/netbird/pull/1467

For now, I will close this one, and we can discuss further improvements on a separate ticket/issue.

PR #664, opened 2023-01-27
Use Peer.ID instead of Peer.Key as peer identifier

braginini:
## Describe your changes

Replace Peer.Key as internal identifier with a randomly generated Peer.ID in the Management service.
Every group now references peers by ID instead of a public key.
Every route now references peers by ID instead of a public key.
FileStore does store.json file migration on startup by generating Peer.ID and replacing
all Peer.Key identifier references .

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/management/server/peer.go
+++ b/management/server/peer.go
@@ -101,7 +101,7 @@ func (p *PeerStatus) Copy() *PeerStatus {
 }
 
 // GetPeer looks up peer by its public WireGuard key
-func (am *DefaultAccountManager) GetPeer(peerPubKey string) (*Peer, error) {
+func (am *DefaultAccountManager) GetPeerByKey(peerPubKey string) (*Peer, error) {
 
 	account, err := am.Store.GetAccountByPeerPubKey(peerPubKey)
 	if err != nil {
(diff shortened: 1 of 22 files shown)


management/server/peer.go
```diff
@@ -312,17 +314,22 @@ func (am *DefaultAccountManager) GetPeerByIP(accountID string, peerIP string) (*
...
 
-	account, err := am.Store.GetAccountByPeerPubKey(peerPubKey)
+	account, err := am.Store.GetAccountByPeerID(peerID)
 	if err != nil {
 		return nil, err
 	}
 
-	aclPeers := account.getPeersByACL(peerPubKey)
-	routesUpdate := account.getRoutesToSync(peerPubKey, aclPeers)
+	peer := account.GetPeer(peerID)
```
mlsmaycon, line 324: it might be better to return an error from account.GetPeer, generally speaking, it should never happen, but if it does, let's centralize it in one method?

mlsmaycon: After discussing internally, we are not returning error from the method because of different cases where a not found peer might cause a different kind of error to the APi, e.g. not found is ok for peer deletion, but for route creation no.

Issue #527, opened 2022-10-25
Wrong macOS Version reported in WebUI


vordenken:
In the Netbird WebUI (cloud version) my mac gets reported with OS version 21.6.0 instead of 12.6 - the numberes seemed to be mixed up? 

<img width="369" alt="Bildschirmfoto 2022-10-25 um 08 58 08" src="https://user-images.githubusercontent.com/62395860/197704446-d2190900-ccd5-431d-8b20-16ba13079bc2.png">
<img width="424" alt="Bildschirmfoto 2022-10-25 um 08 57 48" src="https://user-images.githubusercontent.com/62395860/197704455-b043de94-a2f6-46cf-a99a-116c5b8471a2.png">

pascal-fischer:
Just double-checked and PR #666 solved the issue.

PR #666, opened 2023-02-01
Macos version error

tcskiran:
## Describe your changes
Resolved the macOS version error issue https://github.com/netbirdio/netbird/issues/527
image
Now it is showing macOS version instead of the darwin version

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/system/info_darwin.go
+++ b/client/system/info_darwin.go
@@ -18,15 +18,17 @@ func GetInfo(ctx context.Context) *Info {
 	if err != nil {
 		fmt.Println("getInfo:", err)
 	}
-	out, err := exec.Command("sw_vers", "-productVersion").Output()
-	if err != nil {
-		fmt.Println("getInfo, macOS version error", err)
-	}
-	version := strings.TrimSpace(string(out))
 	sysName := string(bytes.Split(utsname.Sysname[:], []byte{0})[0])
 	machine := string(bytes.Split(utsname.Machine[:], []byte{0})[0])
 	release := string(bytes.Split(utsname.Release[:], []byte{0})[0])
-	gio := &Info{Kernel: sysName, OSVersion: version, Core: release, Platform: machine, OS: sysName, GoOS: runtime.GOOS, CPUs: runtime.NumCPU()}
+	out, err := exec.Command("sw_vers", "-productVersion").Output()
+	// If there is an error while getting version, returning darwin version instead
+	if err == nil {
+		version := strings.TrimSpace(string(out))
+		gio := &Info{Kernel: sysName, OSVersion: version, Core: release, Platform: machine, OS: sysName, GoOS: runtime.GOOS, CPUs: runtime.NumCPU()}
+	} else {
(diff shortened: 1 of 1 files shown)


client/system/info_darwin.go
```diff
@@ -16,10 +18,15 @@ func GetInfo(ctx context.Context) *Info {
 	if err != nil {
 		fmt.Println("getInfo:", err)
 	}
+	out, err := exec.Command("sw_vers", "-productVersion").Output()
+	if err != nil {
+		fmt.Println("getInfo, macOS version error", err)
```
mlsmaycon: let's use `release` for `version` in case of an error here.

mlsmaycon (changes requested):
thank you @tcskiran for your contribution!!

I left a small comment for a change. Regarding using `sw_vers` do you know when Apple first introduce it to macOS?

client/system/info_darwin.go
tcskiran: No, I don't know when sw_vers was introduced in macOS

tcskiran: Do you want me to return the darwin version in case we have an error in getting the normal version

mlsmaycon: yes

mlsmaycon: regarding the command introduction, it seems to be there since the early X versions. The link https://www.manpagez.com/man/1/sw_vers/ lists it since 2009.

client/system/info_darwin.go
```diff
@@ -19,7 +21,15 @@ func GetInfo(ctx context.Context) *Info {
...
-	gio := &Info{Kernel: sysName, OSVersion: release, Core: release, Platform: machine, OS: sysName, GoOS: runtime.GOOS, CPUs: runtime.NumCPU()}
+	out, err := exec.Command("sw_vers", "-productVersion").Output()
+	// If there is an error while getting version, returning darwin version instead
+	var gio *Info
+	if err == nil {
+		version := strings.TrimSpace(string(out))
+		gio = &Info{Kernel: sysName, OSVersion: version, Core: release, Platform: machine, OS: sysName, GoOS: runtime.GOOS, CPUs: runtime.NumCPU()}
+	} else {
+		gio = &Info{Kernel: sysName, OSVersion: release, Core: release, Platform: machine, OS: sysName, GoOS: runtime.GOOS, CPUs: runtime.NumCPU()}
+	}
```
mlsmaycon:
```suggestion
	version, err := exec.Command("sw_vers", "-productVersion").Output()
	if err != nil {
	       fmt.Printf("got an error while retrieving macOS version with sw_vers, error: %s. Using darwin version instead.",err)
	       version = []byte(release)
	}
	gio := &Info{Kernel: sysName, OSVersion: strings.TrimSpace(string(out)), Core: release, Platform: machine, OS: sysName, GoOS: runtime.GOOS, CPUs: runtime.NumCPU()}
```

mlsmaycon (approved):
Thanks @tcskiran nice finding!

PR #674, opened 2023-02-07
Refactor Management and Admin URL config

mlsmaycon:
## Describe your changes
We had an issue where the daemon service was overwriting the admin URL every time it was restarted. With the changes, we are making the default admin URL a concern of the config logic.

I've also reduced the number of input parameters for the New function in the daemon logic as it is only called after the service is installed, which kind brake the need for it to receive such inputs. 

Other changes:
- Avoid sending admin or management URLs on service start as it doesn't have an input
- Parse management and admin URL when needed
- Pass empty admin URL on commands to prevent default overwrite

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/config.go
+++ b/client/internal/config.go
@@ -16,27 +16,19 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-// ManagementLegacyPort is the port that was used before by the Management gRPC server.
-// It is used for backward compatibility now.
-// NB: hardcoded from github.com/netbirdio/netbird/management/cmd to avoid import
-const ManagementLegacyPort = 33073
+const (
+	// ManagementLegacyPort is the port that was used before by the Management gRPC server.
+	// It is used for backward compatibility now.
+	// NB: hardcoded from github.com/netbirdio/netbird/management/cmd to avoid import
+	ManagementLegacyPort = 33073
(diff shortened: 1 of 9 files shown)


client/internal/config.go
```diff
@@ -16,28 +16,20 @@ import (
...
-// It is used for backward compatibility now.
-// NB: hardcoded from github.com/netbirdio/netbird/management/cmd to avoid import
-const ManagementLegacyPort = 33073
+const (
+	// ManagementLegacyPort is the port that was used before by the Management gRPC server.
+	// It is used for backward compatibility now.
+	// NB: hardcoded from github.com/netbirdio/netbird/management/cmd to avoid import
+	ManagementLegacyPort = 33073
+	// DefaultManagementURL points to the NetBird's cloud management endpoint
+	DefaultManagementURL = "https://api.wiretrustee.com:443"
```
gigovich, line 25: Why do we still use wiretrustee?

gigovich (approved):
LGTM

client/internal/config.go
mlsmaycon: Mostly because of old clients and because we need to support multiple domains in our cert request logic.

pappz (approved):

Issue #620, opened 2022-12-12
Check if nameserver depends on peer connection


mlsmaycon:
**Describe the problem**
A nameserver configuration that points to a peer or an address behind a routing peer may cause connectivity issues and unwanted behavior.

We should be able to check if the DNS settings depend on a peer connection and only apply it once the connection is live again.

**To Reproduce**
1. Create a nameserver configuration that points to a peer (peer-a)
2. Ensure that it is resolving all connections
3. Share that connection with your peers via distribution groups
4. Disconnect the peer-a

**Expected behavior**
DNS resolution should never fail. In the event we should rollback the DNS resolver and wait for peer connection to be established

**Additional context**
We might need to extend the status recorder system or the dns server

EZPC-Matt:
I had a Debian based VM today where Netbird had put it's own IP in /etc/resolve.conf as the only nameserver but following a VM reboot the WT0 interface had not yet been created and thus the IP didn't exist on the system.

Netbird therefore couldn't resolve the DNS for it's own control plane in order to bring itself back up.

Just upgraded to 0.12.0, not sure what it was before that.

mlsmaycon:
After a chat in slack, the issue happened on a Debian 11 with minimal installation and docker CE. It uses the default systemd init system.

mlsmaycon:
We've improved the unavailability of a DNS server. This should be resolved now

PR #676, opened 2023-02-08
Fix nameserver peer conn check

gigovich:
## Describe your changes
This change adds fallback to the host DNS default resolver when the upstream resolver fails several times.
It also reactivates the upstream resolver in 1 min and tries to use it again.

https://github.com/netbirdio/netbird/issues/620

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/client/internal/dns/host.go
+++ b/client/internal/dns/host.go
@@ -56,6 +56,9 @@ func dnsConfigToHostDNSConfig(dnsConfig nbdns.Config, ip string, port int) hostD
 		serverPort: port,
 	}
 	for _, nsConfig := range dnsConfig.NameServerGroups {
+		if len(nsConfig.NameServers) == 0 {
+			continue
+		}
 		if nsConfig.Primary {
 			config.routeAll = true
 		}
--- a/client/internal/dns/host.go
+++ b/client/internal/dns/host.go
@@ -2,8 +2,9 @@ package dns
 
 import (
 	"fmt"
-	nbdns "github.com/netbirdio/netbird/dns"
 	"strings"
+
+	nbdns "github.com/netbirdio/netbird/dns"
 )
 
 type hostManager interface {
@@ -19,6 +20,7 @@ type hostDNSConfig struct {
 }
 
 type domainConfig struct {
+	disabled  bool
 	domain    string
 	matchOnly bool
 }
--- a/client/internal/dns/mock_test.go
+++ b/client/internal/dns/mock_test.go
@@ -23,12 +23,3 @@ func (rw *mockResponseWriter) Close() error              { return nil }
 func (rw *mockResponseWriter) TsigStatus() error         { return nil }
 func (rw *mockResponseWriter) TsigTimersOnly(bool)       {}
 func (rw *mockResponseWriter) Hijack()                   {}
-
-
-type mockResolver struct {
-  ServeDNSFunc func(w dns.ResponseWriter, r *dns.Msg)
-}
-
-func (rl *mockResolver) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
-  rl.ServeDNSFunc(w, r) 
-}
--- a/client/internal/dns/mock_test.go
+++ b/client/internal/dns/mock_test.go
@@ -1,8 +1,9 @@
 package dns
 
 import (
-	"github.com/miekg/dns"
 	"net"
+
+	"github.com/miekg/dns"
 )
 
 type mockResponseWriter struct {
@@ -23,16 +24,3 @@ func (rw *mockResponseWriter) Close() error              { return nil }
 func (rw *mockResponseWriter) TsigStatus() error         { return nil }
 func (rw *mockResponseWriter) TsigTimersOnly(bool)       {}
 func (rw *mockResponseWriter) Hijack()                   {}
-
-type mockHostManager struct {
-	applyDNSConfigFunc func(config hostDNSConfig) error
-	restoreHostDNSFunc func() error
-}
-
-func (hm *mockHostManager) applyDNSConfig(config hostDNSConfig) error {
-	return hm.applyDNSConfigFunc(config)
-}
-
-func (hm *mockHostManager) restoreHostDNS() error {
-	return hm.restoreHostDNSFunc()
-}
--- a/client/internal/dns/mock_test.go
+++ b/client/internal/dns/mock_test.go
@@ -23,3 +23,16 @@ func (rw *mockResponseWriter) Close() error              { return nil }
 func (rw *mockResponseWriter) TsigStatus() error         { return nil }
 func (rw *mockResponseWriter) TsigTimersOnly(bool)       {}
 func (rw *mockResponseWriter) Hijack()                   {}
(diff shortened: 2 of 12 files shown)


client/internal/dns/mock_test.go
```diff
@@ -23,3 +23,16 @@ func (rw *mockResponseWriter) Close() error              { return nil }
 func (rw *mockResponseWriter) TsigStatus() error         { return nil }
 func (rw *mockResponseWriter) TsigTimersOnly(bool)       {}
 func (rw *mockResponseWriter) Hijack()                   {}
+
+type mockHostManager struct {
```
mlsmaycon: we got the `mockHostConfigurator` in the host.go that might be removed. There is a `newNoopHostMocker` that needs update as well.

client/internal/dns/server.go
```diff
@@ -45,6 +45,7 @@ type DefaultServer struct {
 	runtimePort        int
 	runtimeIP          string
 	previousConfigHash uint64
+	currentConfig      nbdns.Config
```
mlsmaycon: Keeping the whole DNS config might backfire in smaller devices as we might have `CustomZones` with thousands of records. So it might be better to only keep nbdns.Config.NameServerGroups as it will be in most cases, small.

client/internal/dns/mock_test.go
gigovich: Fixed

client/internal/dns/server.go
gigovich: Fixed,
Using [hostDNSConfig](https://github.com/netbirdio/netbird/pull/676/commits/8856e3c2963f773d4fe9db13ba8b724d3c0f01f8) to save state instead of `nbdns.Config`.

client/internal/dns/upstream.go
```diff
@@ -3,56 +3,127 @@ package dns
...
+// will be reset and upstream will be reactivated.
+func (u *upstreamResolver) checkUpstreamFails() {
+	u.mutex.Lock()
+	defer u.mutex.Unlock()
+
+	if u.failsCount.Load() < u.failsTillDeact || u.disabled {
+		return
+	}
+
+	log.WithField("preiod", reactivatePeriod).Warn("upstream resolving is disabled for")
```
mlsmaycon: this line outputs a weird log event:
`time="2023-02-11T12:41:17+01:00" level=warning msg="upstream resolving is disabled for" file="upstream.go:103" preiod=1m0s
`

Might be better to just use Warnf and move the period to the main message

client/internal/dns/server.go
```diff
@@ -382,3 +409,63 @@ func (s *DefaultServer) registerMux(pattern string, handler dns.Handler) {
...
+	nsGroup *nbdns.NameServerGroup,
+	handler dns.Handler,
+) (deactivate func(), reactivate func()) {
+	var removeIndex map[string]int
+	deactivate = func() {
+		s.mux.Lock()
+		defer s.mux.Unlock()
+
+		l := log.WithField("nameservers", nsGroup.NameServers)
+		l.Info("temporary deactivate nameservers group")
```
mlsmaycon: suggest msg:
`temporary deactivate nameservers group due timeout`

client/internal/dns/server.go
```diff
@@ -382,3 +409,63 @@ func (s *DefaultServer) registerMux(pattern string, handler dns.Handler) {
...
+
+		for i, item := range s.currentConfig.domains {
+			if _, found := removeIndex[item.domain]; found {
+				s.currentConfig.domains[i].disabled = true
+				s.deregisterMux(item.domain)
+				removeIndex[item.domain] = i
+			}
+		}
+		if err := s.hostManager.applyDNSConfig(s.currentConfig); err != nil {
+			l.WithError(err).Error("temporary deactivate nameservers group, DNS update apply")
```
mlsmaycon: Suggestion text:
`fail to apply nameserver deactivation on the host`

client/internal/dns/server.go
```diff
@@ -382,3 +409,63 @@ func (s *DefaultServer) registerMux(pattern string, handler dns.Handler) {
...
+		}
+
+		l := log.WithField("nameservers", nsGroup.NameServers)
+		l.Debug("reactivate temporary disabled nameserver group")
+
+		if nsGroup.Primary {
+			s.currentConfig.routeAll = true
+		}
+		if err := s.hostManager.applyDNSConfig(s.currentConfig); err != nil {
+			l.WithError(err).Error("reactivate temporary disabled nameserver group, DNS update apply")
```
mlsmaycon, line 468: Suggestion text:
`fail to apply nameserver reactivation on the host`

client/internal/dns/host.go
```diff
@@ -19,6 +20,7 @@ type hostDNSConfig struct {
 }
 
 type domainConfig struct {
+	disabled  bool
```
mlsmaycon, line 23: We are missing disabled domain config handling on Windows and Darwin

client/internal/dns/upstream.go
```diff
@@ -3,56 +3,127 @@ package dns
...
+	u.deactivate()
```
mlsmaycon: Even if at first look this would not be harmful if executed, I think we should check for the u.ctx.Done() here as u.deactivate will be executed after the new update arrives.

client/internal/dns/server.go
```diff
@@ -219,41 +221,50 @@ func (s *DefaultServer) UpdateDNSServer(serial uint64, update nbdns.Config) erro
 			s.updateSerial = serial
```
mlsmaycon, line 222: I found a bug with the hash configuration, where a nameserver change without replacing the number of IPs caused the hashes to be identical. Can you add `UseStringer:     true,` to `hashstructure.HashOptions` in line 210?

mlsmaycon (changes requested):

client/internal/dns/upstream.go
gigovich: Fixed

client/internal/dns/server.go
gigovich: Fixed

client/internal/dns/server.go
gigovich: Fixed

client/internal/dns/server.go
gigovich: Fixed

client/internal/dns/host.go
gigovich: Added, need help with testing.

client/internal/dns/upstream.go
gigovich: Check added.

client/internal/dns/server.go
gigovich: Fix added

client/internal/dns/host.go
gigovich: Tested in Windows - works as expected.

mlsmaycon: Same for macOS

PR #680, opened 2023-02-09
Mobile prerefactor

pappz:
## Describe your changes

Small code cleaning in the iface package. These changes necessary to get a clean code in case if we involve more platforms.
The OS related functions has been distributed into separate files and it has been mixed with not OS related logic. The goal is to get a clear picture of the layer between Wirequard  and business logic. 

## Main problem
Unfortunately the Android implementation of Wrieguard tool does not fit with the original Wireguard approach.
Until now the service communicated with the wireguard device via wgctrl package on every OS. In Android it is not possible so we have to create some new function what doing almost the same what the wgctrl did. I would like to provide a well layered solution.

## Changes
- Remove unused variables
- Set to private some variables what is not used externally
- We have OS related implementations so remove the 'if runtime.GOOS != "windows"' expression
- Eliminate spagetti style function calls inside WGIface struct across several files.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


iface/iface.go
```diff
@@ -57,39 +49,250 @@ func NewWGIFace(iface string, address string, mtu int) (*WGIface, error) {
...
+	config := wgtypes.Config{
+		PrivateKey:   &key,
+		ReplacePeers: true,
+		FirewallMark: &fwmark,
+		ListenPort:   &port,
+	}
+
+	err = w.configureDevice(config)
+	if err != nil {
+		return fmt.Errorf("received error \"%v\" while configuring interface %s with port %d", err, w.Name, port)
```
gigovich: You could use ` to avoid use \\" and also, it is better to use wrapper [%w](https://go.dev/blog/go1.13-errors) for errors instead of %v

iface/address.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// WGAddress Wireguard parsed address
+type WGAddress struct {
+	IP      net.IP
+	Network *net.IPNet
+}
+
+// newWGAddress parse a string ("1.2.3.4/24") address to WG Address
+func newWGAddress(address string) (WGAddress, error) {
```
gigovich: It looks like it will be better if you save the last name with the `parse` prefix, because usually `new` prefix is used as a constructor name, which not returns errors.

iface/iface.go
```diff
@@ -57,39 +49,250 @@ func NewWGIFace(iface string, address string, mtu int) (*WGIface, error) {
 	return wgIface, nil
 }
 
-// parseAddress parse a string ("1.2.3.4/24") address to WG Address
-func parseAddress(address string) (WGAddress, error) {
-	ip, network, err := net.ParseCIDR(address)
+// GetName returns the interface name
+func (w *WGIface) GetName() string {
+	return w.Name
```
gigovich: There is only one place outside the `iface` package where `w.Name` is used directly and several areas for `w.Address`, so your PR is an excellent place to make these fields private and use getters at that places.

iface/iface.go
```diff
@@ -57,39 +49,250 @@ func NewWGIFace(iface string, address string, mtu int) (*WGIface, error) {
...
+	config := wgtypes.Config{
+		PrivateKey:   &key,
+		ReplacePeers: true,
+		FirewallMark: &fwmark,
+		ListenPort:   &port,
+	}
+
+	err = w.configureDevice(config)
+	if err != nil {
+		return fmt.Errorf(`received error "%v" while configuring interface %s with port %d`, err, w.Name, port)
```
gigovich: Wrap error here also, please.

gigovich (approved):
LGTM, but there is one proposal that can be done in that PR.

braginini (approved):

PR #682, opened 2023-02-10
Add peer login expiration

braginini:
## Describe your changes
This PR adds a peer login expiration logic that requires 
peers created by a user to re-authenticate (re-login) after
a certain threshold of time (24h by default). 

The Account object now has a PeerLoginExpiration
property that indicates the duration after which a peer's
login will expire and a login will be required. Defaults to 24h.

There are two new properties added to the Peer object:
LastLogin that indicates the last time peer successfully used
the Login gRPC endpoint and LoginExpirationEnabled that 
enables/disables peer login expiration.

The login expiration logic applies only to peers that were created
by a user and not those that were added with a setup key.

_P.S. The expiration is disabled always for now. Will be enabled in the future versions.
Next PRs:
HTTP Peer API will have a method to enable/disable expiration.
HTTP Peer API GET Peers will return the indicator that peer's login
has expired.
HTTP API for system settings to globally enable/disable expiration and set expiration duration._

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/grpcserver.go
+++ b/management/server/grpcserver.go
@@ -354,6 +354,11 @@ func (s *GRPCServer) Login(ctx context.Context, req *proto.EncryptedMessage) (*p
 		}
 	}
 
+	expired, left := peer.LoginExpired(24 * time.Hour)
+	if peer.UserID != "" && expired {
+		return nil, status.Errorf(codes.PermissionDenied, "peer login has expired %v ago. Please login once more", left)
+	}
+
 	var sshKey []byte
 	if loginReq.GetPeerKeys() != nil {
 		sshKey = loginReq.GetPeerKeys().GetSshPubKey()
--- a/management/server/grpcserver.go
+++ b/management/server/grpcserver.go
@@ -132,6 +132,11 @@ func (s *GRPCServer) Sync(req *proto.EncryptedMessage, srv proto.ManagementServi
 		return msg
 	}
 
+	expired, left := peer.LoginExpired(24 * time.Hour)
+	if peer.UserID != "" && expired {
+		return status.Errorf(codes.PermissionDenied, "peer login has expired %v ago. Please log in once more", left)
+	}
+
 	syncReq := &proto.SyncRequest{}
 	err = encryption.DecryptMessage(peerKey, s.wgKey, req.Body, syncReq)
 	if err != nil {
@@ -356,7 +361,7 @@ func (s *GRPCServer) Login(ctx context.Context, req *proto.EncryptedMessage) (*p
 
 	expired, left := peer.LoginExpired(24 * time.Hour)
 	if peer.UserID != "" && expired {
-		return nil, status.Errorf(codes.PermissionDenied, "peer login has expired %v ago. Please login once more", left)
+		return nil, status.Errorf(codes.PermissionDenied, "peer login has expired %v ago. Please log in once more", left)
 	}
 
 	var sshKey []byte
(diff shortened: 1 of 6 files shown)


management/server/peer.go
```diff
@@ -58,27 +58,44 @@ type Peer struct {
...
 // Copy copies Peer object
 func (p *Peer) Copy() *Peer {
 	return &Peer{
-		ID:         p.ID,
-		Key:        p.Key,
-		SetupKey:   p.SetupKey,
-		IP:         p.IP,
-		Meta:       p.Meta,
-		Name:       p.Name,
-		Status:     p.Status,
```
pappz, line 74: The Peer.Status is a pointer. Here you copy the address of the p.Status. So if you modify the Peer.Status.connected then it will be applied on the copied Peer.Status.connected also. This is the expectation?

management/server/peer.go
```diff
@@ -100,7 +117,7 @@ func (p *PeerStatus) Copy() *PeerStatus {
 	}
 }
 
-// GetPeer looks up peer by its public WireGuard key
+// GetPeerByKey looks up peer by its public WireGuard key
 func (am *DefaultAccountManager) GetPeerByKey(peerPubKey string) (*Peer, error) {
```
pappz, line 121: For a getter the "Get" word is not necessary in go. But it is not a big deal. Current name is also ok for me.
https://go.dev/doc/effective_go#Getters

management/server/grpcserver.go
```diff
@@ -132,6 +132,11 @@ func (s *GRPCServer) Sync(req *proto.EncryptedMessage, srv proto.ManagementServi
 		return msg
 	}
 
+	expired, left := peer.LoginExpired(24 * time.Hour)
+	if peer.UserID != "" && expired {
```
pappz, line 140: If you are using complex expression I recommend to create for it a separate function. Like hasExpired or something like that. The source code will be easier readable. You do not need to compile in your mind the if expression.
I. e. totally acceptable to write functions just for one line in the scope if the code will be understandable.

management/server/grpcserver.go
```diff
@@ -359,9 +367,23 @@ func (s *GRPCServer) Login(ctx context.Context, req *proto.EncryptedMessage) (*p
 		}
 	}
 
+	// check if peer login has expired
```
pappz, line 374: The feeling to write a inline comment is a good indicator to should write a new function whit a proper name.

management/server/peer.go
braginini: Sounds good. Maybe we could do a refactor in a separate PR and change everywhere?

management/server/peer.go
braginini: That's unrelated to the current changes, was like that for a while. But you are right. Maybe we could do it in a separate PR?

management/server/grpcserver.go
braginini: makes sense, will do

management/server/grpcserver.go
braginini: makes sense

management/server/peer.go
pappz: Separated PR sounds good.

management/server/peer.go
pappz: Not so important. It is just a convention what we can follow in the future if we want.

pappz (approved):

PR #689, opened 2023-02-14
Fallback to console log when can't write log file

braginini:
## Describe your changes

Skip writing log file when permissions don't allow write operations.
Fallback to CONSOLE log output.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/util/log.go
+++ b/util/log.go
@@ -21,7 +21,6 @@ func InitLog(logLevel string, logPath string) error {
 		return err
 	}
 
-	logPath = "/tmp/netbird/netbird/netbird.log"
 	if logPath != "" && logPath != "console" {
 
 		canWrite, _ := canWrite(logPath)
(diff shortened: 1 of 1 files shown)


util/log.go
```diff
@@ -19,24 +21,30 @@ func InitLog(logLevel string, logPath string) error {
 		return err
 	}
 
+	logPath = "/tmp/netbird/netbird/netbird.log"
 	if logPath != "" && logPath != "console" {
```
pappz, line 24: I recommend to move this scope into a new function. And always good idea to avoid the deep if-else statements.
I. e.
```
if logPath == "" {
  return
}
if logPath == "console" {
  return
}

if !canWrite {
 return
}
...

```

pappz: I point to the 25. line. Sry

braginini:
we will close this PR. @mlsmaycon  found a better solution
https://github.com/netbirdio/netbird/pull/692

PR #698, opened 2023-02-20
Proactively expire peers' login per account 

braginini:
## Describe your changes

Goals:
- Enable peer login expiration when adding new peer
- Expire peer's login when the time comes

The account manager triggers peer expiration routine in future if the 
following conditions are true:
- peer expiration is enabled for the account
- there is at least one peer that has expiration enabled and is connected

The time of the next expiration check is based on the nearest peer expiration.
Account manager finds a peer with the oldest last login (auth) timestamp and
calculates the time when it has to run the routine as a sum of the configured
peer login expiration duration and the peer's last login time. 

When triggered, the expiration routine checks whether there are expired peers.
The management server closes the update channel of these peers and updates
network map of other peers to exclude expired peers so that the expired peers
are not able to connect anywhere. 

The account manager can reschedule or cancel peer expiration in the following cases:
- when admin changes account setting (peer expiration enable/disable)
- when admin updates the expiration duration of the account
- when admin updates peer expiration (enable/disable)
- when peer connects (Sync)  
  
P.S. The network map calculation was updated to exclude peers that have login expired.  

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -324,12 +324,13 @@ func (a *Account) GetExpiredPeers() []*Peer {
 // GetNextPeerExpiration returns the minimum duration in which the next peer of the account will expire and true if it was found.
 // If there is no peer that expires this function returns false and a zero time.Duration
 // This function only considers peers that haven't been expired yet and connected.
-func (a *Account) GetNextPeerExpiration() (time.Duration, bool) {
-	nextExpiry := time.Duration(1<<63 - 1) // max duration
+func (a *Account) GetNextPeerExpiration() *time.Duration {
+
 	peersWithExpiry := a.GetPeersWithExpiration()
 	if len(peersWithExpiry) == 0 {
-		return nextExpiry, false
+		return nil
 	}
+	nextExpiry := time.Duration(1<<63 - 1) // max duration
 	for _, peer := range peersWithExpiry {
 		// consider only connected peers because others will require login on connecting to the management server
 		if peer.Status.LoginExpired || !peer.Status.Connected {
@@ -341,7 +342,11 @@ func (a *Account) GetNextPeerExpiration() (time.Duration, bool) {
 		}
 	}
 
-	return nextExpiry, nextExpiry != time.Duration(1<<63-1)
+	if nextExpiry == time.Duration(1<<63-1) {
+		return nil
+	}
+
+	return &nextExpiry
 }
 
 // GetPeersWithExpiration returns a list of peers that have Peer.LoginExpirationEnabled set to true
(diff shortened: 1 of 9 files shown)


management/server/account.go
```diff
@@ -307,6 +308,53 @@ func (a *Account) GetGroup(groupID string) *Group {
 	return a.Groups[groupID]
 }
 
+// GetExpiredPeers returns peers tha have been expired
```
pascal-fischer: Missing a 't' in that

management/server/account.go
```diff
@@ -307,6 +308,53 @@ func (a *Account) GetGroup(groupID string) *Group {
...
+		if expired {
+			peers = append(peers, peer)
+		}
+	}
+
+	return peers
+}
+
+// GetNextPeerExpiration returns the minimum duration in which the next peer of the account will expire and true if it was found.
+// If there is no peer that expires this function returns false and a zero time.Duration
```
pascal-fischer: we return max value not 0 if no peer is found

> If there is no peer that is not yet expired, this function returns false and the max time.Duration value as default

management/server/account.go
```diff
@@ -307,6 +308,53 @@ func (a *Account) GetGroup(groupID string) *Group {
...
+		if peer.Status.LoginExpired || !peer.Status.Connected {
+			continue
+		}
+		_, duration := peer.LoginExpired(a.Settings.PeerLoginExpiration)
+		if duration < nextExpiry {
+			nextExpiry = duration
+		}
+	}
+
+	return nextExpiry, nextExpiry != time.Duration(1<<63-1)
```
pascal-fischer: Maybe go with nil instead of max value for `nextExpiry` if no more expiring peers are available as it is more clean from a coding point of view to not return a value if there is no value available.

management/server/account.go
```diff
@@ -658,6 +711,55 @@ func (am *DefaultAccountManager) UpdateAccountSettings(accountID, userID string,
...
+			peerIDs = append(peerIDs, peer.ID)
+			peer.MarkLoginExpired(true)
+			account.UpdatePeer(peer)
+			err = am.Store.SavePeerStatus(account.Id, peer.ID, *peer.Status)
+			if err != nil {
+				log.Errorf("failed saving peer status while expiring peer %s", peer.ID)
+				// todo return retry?
+				return false, 0
+			}
+		}
```
pascal-fischer: Not sure on this one but I think we do way to much computing overhead here.

`account.GetExpiredPeers()` will return the complete list of expired peers. Even peers that have been expired for months. So in case we we have lots of zombies in the account that are not used anymore but are doing database overrides and peer updates for all of them on any run with the same (unchanged) data.

I would add additional check for status changes at the top of the loop:
```
if peer.Status.LoginExpired {
     continue
}
```

management/server/account.go
```diff
@@ -658,6 +711,55 @@ func (am *DefaultAccountManager) UpdateAccountSettings(accountID, userID string,
...
+
+		log.Debugf("discovered %d peers to expire for account %s", len(peerIDs), account.Id)
```
pascal-fischer, line 745: Log message should also show wrong number of peers. Not newly expired but total expired in account

pascal-fischer (changes requested):

PR #700, opened 2023-02-21
Feat rego default policy

gigovich:
## Describe your changes
Convert rules to Rego policies and allow users to write raw policies to set up connectivity and firewall on the clients.
NB-215
### Checklist
- [ ] Is it a bug fix
- [*] Is a feature enhancement
- [*] It is a refactor
- [*] Created tests that fail without the change (if possible)
- [*] Extended the README / documentation, if necessary


management/server/policy.go
```diff
@@ -0,0 +1,377 @@
...
+
+	// Action of the traffic
+	Action string
+
+	// Port of the traffic
+	Port string
+}
+
+// parseFromRegoResult parses the Rego result to a FirewallRule.
+func (f *FirewallRule) parseFromRegoResult(value interface{}) error {
```
pappz, line 184: With the struct tags and reflect package feature would not be nicer the parse?
```
type FirewallRule struct {
  PeerId string `rego:"ID"`
  ...
}
```

gigovich: Thanks for the note, reflect is not a good choice when you do simple things like this. I build reflect parsers only when I need to parse different types of structs in the same way, in that case, it's ok, but here it will be overkill.

pappz: Thank you your answer. You are right.

management/server/account.go
```diff
@@ -153,6 +157,7 @@ type Account struct {
 	Users                  map[string]*User
 	Groups                 map[string]*Group
 	Rules                  map[string]*Rule
+	Policies               []*Policy
```
mlsmaycon, line 154: For file store, won't it be better to keep a map?

management/server/policy.go
```diff
@@ -0,0 +1,416 @@
...
+}
+
+//go:embed rego/default_policy_module.rego
+var defaultPolicyModule string
+
+//go:embed rego/default_policy.rego
+var defaultPolicy string
+
+// PolicyMeta is the metadata of the policy
+type PolicyMeta struct {
```
mlsmaycon: Meta feels more like a Rule

management/server/policy.go
```diff
@@ -0,0 +1,416 @@
...
+	Description string
+
+	// Disabled status of the policy
+	Disabled bool
+
+	// Query of Rego the policy
+	Query string
+
+	// Meta of the policy
+	Meta *PolicyMeta
```
mlsmaycon: a policy will support a single rule(meta)? usually, policies have multiple sets of rules.

management/server/policy.go
```diff
@@ -0,0 +1,416 @@
...
+	Name string
+
+	// Description of the policy visible in the UI
+	Description string
+
+	// Disabled status of the policy
+	Disabled bool
+
+	// Query of Rego the policy
+	Query string
```
mlsmaycon, line 111: As for the initial release, does it make sense to store the default query until we introduce support for custom queries?

management/server/account.go
gigovich: Golang does not guarantee the order of added items. You can add items to the map several times in the same order, and get them in a different order when you range over it. For policies, it is important to preserve order.

management/server/policy.go

management/server/policy.go

management/server/policy.go
gigovich: This field is updated automatically.

management/server/account.go
```diff
@@ -1262,7 +1227,14 @@ func addAllGroup(account *Account) {
 			Destination: []string{allGroup.ID},
 		}
 		account.Rules = map[string]*Rule{defaultRule.ID: defaultRule}
+
+		defaultPolicy, err := RuleToPolicy(defaultRule)
```
mlsmaycon, line 1266: we can remove the previous block that is adding a Rule to the account object.

management/server/file_store.go
```diff
@@ -113,6 +113,19 @@ func restore(file string) (*FileStore, error) {
...
+		if len(account.Policies) == 0 {
+			account.Policies = make([]*Policy, 0)
+			for _, rule := range account.Rules {
+				policy, err := RuleToPolicy(rule)
+				if err != nil {
+					log.Errorf("unable to migrate rule to policy: %v", err)
+					continue
+				}
+				account.Policies = append(account.Policies, policy)
+			}
```
mlsmaycon, line 126: we should remove the Rules from account once migrated

management/server/policy.go
```diff
@@ -0,0 +1,425 @@
...
+	Action PolicyTrafficActionType `json:"action"`
+
+	// Destinations policy destination groups
+	Destinations []string `json:"destinations"`
+
+	// Sources policy source groups
+	Sources []string `json:"sources"`
+
+	// Port port of the service or range of the ports, and optional protocol (by default TCP)
+	Port string `json:"port"`
```
mlsmaycon: We should skip creating the Port field for now until we have a clear API definition.

management/server/policy.go
```diff
@@ -0,0 +1,425 @@
...
+	Name string
+
+	// Description of the rule visible in the UI
+	Description string
+
+	// Disabled status of rule in the system
+	Disabled bool
+
+	// Action policy accept or drops packets
+	Action PolicyTrafficActionType `json:"action"`
```
mlsmaycon: are annotations really required here?

management/server/policy.go
```diff
@@ -0,0 +1,425 @@
...
+	Destinations []string `json:"destinations"`
+
+	// Sources policy source groups
+	Sources []string `json:"sources"`
+
+	// Port port of the service or range of the ports, and optional protocol (by default TCP)
+	Port string `json:"port"`
+}
+
+// Copy returns a copy of the policy.
```
mlsmaycon: it should be `Copy returns a copy of a policy rule`

management/server/policy.go
```diff
@@ -0,0 +1,425 @@
...
+	if err != nil {
+		log.WithError(err).Error("eval Rego query")
+		return nil, nil
+	}
+
+	if len(evalResult) == 0 || len(evalResult[0].Expressions) == 0 {
+		log.Error("empty Rego query eval result")
+		return nil, nil
+	}
+	expression, ok := evalResult[0].Expressions[0].Value.([]interface{})
```
mlsmaycon: maybe is better to use `expressions`

management/server/policy.go
```diff
@@ -0,0 +1,425 @@
...
+	ID string
+
+	// Name of the rule visible in the UI
+	Name string
+
+	// Description of the rule visible in the UI
+	Description string
+
+	// Disabled status of rule in the system
+	Disabled bool
```
mlsmaycon: We could take the chance and use Enabled instead of Disabled. This will fix the Rule issue that is the only object with Disabled field. 

The reason we had Disabled in rules, was because access control didn't exist in the initial releases of the platform and we had to use Disabled to support a clean migration.

management/server/http/api/openapi.yml
```diff
@@ -393,6 +395,79 @@ components:
...
+        destinations:
+          description: policy destination groups
+          type: array
+          items:
+            $ref: '#/components/schemas/GroupMinimum'
+        action:
+          description: policy accept or drops packets
+          type: string
+          enum: ["accept","drop"]
+        port:
```
mlsmaycon: remove if field is removed from struct

mlsmaycon (changes requested):

management/server/account.go
gigovich: We still have tests where this is important. So let's remove it in the new PR.

mlsmaycon: Let's add a //todo then

gigovich: Fixed

management/server/http/rules_handler.go
```diff
@@ -368,13 +242,13 @@ func (h *RulesHandler) GetRule(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
-		rule, err := h.accountManager.GetRule(account.Id, ruleID, user.Id)
+		policy, err := h.accountManager.GetPolicy(account.Id, ruleID, user.Id)
 		if err != nil {
 			util.WriteError(status.Errorf(status.NotFound, "rule not found"), w)
```
braginini: We should forward the error to the util function itself like that `util.WriteError(err, w)`. 
Plz check other places in this handler.

management/server/rego/default_policy.rego
```diff
@@ -0,0 +1,10 @@
+package netbird
+
+# all rules for peer connectivity and firewall
```
braginini: This comment is saved to the store and duplicated for each policy. Let's remove it

braginini (changes requested):

management/server/policy.go
```diff
@@ -0,0 +1,421 @@
...
+		Name:         pm.Name,
+		Description:  pm.Description,
+		Enabled:      pm.Enabled,
+		Action:       pm.Action,
+		Destinations: pm.Destinations[:],
+		Sources:      pm.Sources[:],
+	}
+}
+
+// ToRule converts the policy rule to a rule.
```
braginini: I think that this should be more clear:
`ToRule converts the PolicyRule to a legacy representation of the Rule (for backwards compatibility)`

braginini (changes requested):

management/server/activity/codes.go
```diff
@@ -19,6 +19,12 @@ const (
 	RuleUpdated
 	// RuleRemoved indicates that a user removed a rule
 	RuleRemoved
+	// PolicyAdded indicates that a user added a new policy
+	PolicyAdded
+	// PolicyUpdated indicates that a user updated a policy
+	PolicyUpdated
+	// PolicyRemoved indicates that a user removed a policy
+	PolicyRemoved
```
braginini, line 27: We need to update `Activity.Message()`, `Activity.StringCode()`, and in the openapi.yml

management/server/http/api/openapi.yml
```diff
@@ -578,8 +649,7 @@ components:
                   "account.create", "account.setting.peer.login.expiration.update", "account.setting.peer.login.expiration.disable", "account.setting.peer.login.expiration.enable",
                   "route.add", "route.delete", "route.update",
                   "nameserver.group.add", "nameserver.group.delete", "nameserver.group.update",
-                  "peer.ssh.disable", "peer.ssh.enable", "peer.rename", "peer.login.expiration.disable", "peer.login.expiration.enable"
-          ]
+                  "peer.ssh.disable", "peer.ssh.enable", "peer.rename", "peer.login.expiration.disable", "peer.login.expiration.enable" ]
```
braginini, line 655: we need to add corresponding policy events here.
And in the dashboard: https://github.com/netbirdio/dashboard/blob/bce75c1ca969def657dbd079dbe49970f7ef1500/src/views/Activity.tsx#L132
https://github.com/netbirdio/dashboard/blob/bce75c1ca969def657dbd079dbe49970f7ef1500/src/views/Activity.tsx#L76

braginini (changes requested):

management/server/http/api/openapi.yml
gigovich: Done https://github.com/netbirdio/dashboard/pull/147

management/server/http/rules_handler.go
gigovich: Fixed

management/server/rego/default_policy.rego
gigovich: Fixed

management/server/policy.go

management/server/activity/codes.go
gigovich: Fixed

management/server/http/api/openapi.yml
```diff
@@ -393,6 +395,75 @@ components:
...
+            $ref: '#/components/schemas/GroupMinimum'
+        action:
+          description: policy accept or drops packets
+          type: string
+          enum: ["accept","drop"]
+      required:
+        - name
+        - sources
+        - destinations
+        - action
```
mlsmaycon, line 431: enabled should be required

management/server/http/api/openapi.yml
```diff
@@ -393,6 +395,75 @@ components:
...
+        rules:
+          description: Policy rule object for policy UI editor
+          type: array
+          items:
+            $ref: '#/components/schemas/PolicyRule'
+      required:
+        - name
+        - description
+        - enabled
+        - query
```
mlsmaycon, line 457: rules should be required

management/server/http/api/openapi.yml
```diff
@@ -574,12 +645,12 @@ components:
                   "setupkey.peer.add", "setupkey.add", "setupkey.update", "setupkey.revoke", "setupkey.overuse",
                   "setupkey.group.delete", "setupkey.group.add",
                   "rule.add", "rule.delete", "rule.update",
+                  "policy.add", "policy.delete", "policy.update",
```
mlsmaycon, line 650: these codes are missing from the activity/codes.go

management/server/policy.go
```diff
@@ -0,0 +1,451 @@
...
+	f.PeerIP = peerIP
+	f.Direction = direction
+	f.Action = action
+	f.Port = port
+
+	return nil
+}
+
+// getRegoQuery returns a initialized Rego object with default rule.
+func (a *Account) getRegoQuery(policies ...*Policy) (rego.PreparedEvalQuery, error) {
```
mlsmaycon: we could remove the input here as `a.Policies` is available, or we convert it to a standalone function.

management/server/http/api/openapi.yml
gigovich: Fixed.

management/server/http/api/openapi.yml
gigovich: Fixed

management/server/http/api/openapi.yml
gigovich: Fixed

management/server/policy.go

mlsmaycon (approved):

braginini (approved):

PR #701, opened 2023-02-22
Allow to create config file next to binary

pappz:
## Describe your changes

Without you set the full path of the config file the Netbird can not 
handle it. We check the path of the given config file's name and 
if it is not contain a folder then skip the step to create a folder 
for it.
You can place the config file next to the binary:
**./netbird  up --config myconfig.json**

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/util/file.go
+++ b/util/file.go
@@ -11,8 +11,7 @@ import (
 // The output JSON is pretty-formatted
 func WriteJson(file string, obj interface{}) error {
 
-	configDir, configFileName := filepath.Split(file)
-	err := os.MkdirAll(configDir, 0750)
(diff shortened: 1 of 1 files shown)


util/file.go
```diff
@@ -100,3 +99,13 @@ func CopyFileContents(src, dst string) (err error) {
 	err = out.Sync()
 	return
 }
+
+func prepareConfigFileDir(file string) (string, string, error) {
```
mlsmaycon, line 103: We can also have a version of the test `"should be written and read successfully"` for this case.

mlsmaycon (approved):

PR #704, opened 2023-02-24
Feature/extend client status cmd to print json or yaml

pascal-fischer:
## Describe your changes
Added cmd flags for the status command in the client to report in json or yaml format.

Refactored existing code to allow reuse for other formats.
Switched to using separate output struct for better sorting and naming of fields in the output string without changing the internally used structs.

### Example outputs:
`netbird status --detail`
[detail.txt](https://github.com/netbirdio/netbird/files/10840499/detail.txt)

`netbird status --json`
[json.txt](https://github.com/netbirdio/netbird/files/10840502/json.txt)

`netbird status --yaml`
[yaml.txt](https://github.com/netbirdio/netbird/files/10840504/yaml.txt)

The flags `--detail`, `--json`, `--yaml` and `--ipv4` are set to mutually exclusive.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


client/cmd/status.go
```diff
@@ -2,25 +2,71 @@ package cmd
...
 	"github.com/netbirdio/netbird/client/system"
 	"github.com/netbirdio/netbird/util"
-	"github.com/spf13/cobra"
-	"google.golang.org/grpc/status"
 )
 
+type peerStateDetailOutput struct {
+	IP                     string    `json:"ip" yaml:"ip"`
+	PubKey                 string    `json:"publicKey" yaml:"publicKey"`
+	FQDN                   string    `json:"fqdn" yaml:"fqdn"`
```
mlsmaycon: we can switch IP and FQDN positions so in yml we will have FQDN as first field in the details array

client/cmd/status.go
```diff
@@ -2,25 +2,71 @@ package cmd
...
+	ConnStatus             string    `json:"connectionStatus" yaml:"connectionStatus"`
+	ConnStatusUpdate       time.Time `json:"connectionStatusUpdate" yaml:"connectionStatusUpdate"`
+	ConnType               string    `json:"connectionType" yaml:"connectionType"`
+	Direct                 bool      `json:"direct" yaml:"direct"`
+	LocalIceCandidateType  string    `json:"localIceCandidateType" yaml:"localIceCandidateType"`
```
mlsmaycon: it might make sense for us to have local and remote as part of a ICE Candidate type object, e.g.:
```yml
iceCandidateType:
  local: 1111
  remote: 4444
  ```

File: client/cmd/status.go [hunk 3]
```diff
@@ -2,25 +2,71 @@ package cmd
...
 	"github.com/netbirdio/netbird/client/proto"
-	nbStatus "github.com/netbirdio/netbird/client/status"
 	"github.com/netbirdio/netbird/client/system"
 	"github.com/netbirdio/netbird/util"
-	"github.com/spf13/cobra"
-	"google.golang.org/grpc/status"
 )
 
+type peerStateDetailOutput struct {
+	IP                     string    `json:"ip" yaml:"ip"`
```
mlsmaycon: we should keep `netbirdIP` label as we might introduce local IPs and other addresses in the future.

client/cmd/status.go
```diff
@@ -2,25 +2,71 @@ package cmd
...
+	PubKey                 string    `json:"publicKey" yaml:"publicKey"`
+	FQDN                   string    `json:"fqdn" yaml:"fqdn"`
+	ConnStatus             string    `json:"connectionStatus" yaml:"connectionStatus"`
```
mlsmaycon: Connection prefix for this and Status update seems unnecessary and generate long keys

client/cmd/status.go
```diff
@@ -2,25 +2,71 @@ package cmd
...
+}
+
+type statusOutputOverview struct {
+	Peers           peersStateOutput      `json:"peers" yaml:"peers"`
+	CliVersion      string                `json:"cliVersion" yaml:"cliVersion"`
+	DaemonVersion   string                `json:"daemonVersion" yaml:"daemonVersion"`
+	DaemonStatus    string                `json:"daemonStatus" yaml:"daemonStatus"`
+	ManagementState managementStateOutput `json:"management" yaml:"management"`
+	SignalState     signalStateOutput     `json:"signal" yaml:"signal"`
+	IP              string                `json:"ip" yaml:"ip"`
```
mlsmaycon: same IP comment from the peer details.

client/cmd/status.go
```diff
@@ -2,25 +2,71 @@ package cmd
...
+type managementStateOutput struct {
+	URL       string `json:"url" yaml:"url"`
+	Connected bool   `json:"connected" yaml:"connected"`
+}
+
+type statusOutputOverview struct {
+	Peers           peersStateOutput      `json:"peers" yaml:"peers"`
+	CliVersion      string                `json:"cliVersion" yaml:"cliVersion"`
+	DaemonVersion   string                `json:"daemonVersion" yaml:"daemonVersion"`
+	DaemonStatus    string                `json:"daemonStatus" yaml:"daemonStatus"`
```
mlsmaycon: It should not be necessary because if the daemon is offline, we won't have the status

client/cmd/status.go
```diff
@@ -2,25 +2,71 @@ package cmd
...
+	ManagementState managementStateOutput `json:"management" yaml:"management"`
+	SignalState     signalStateOutput     `json:"signal" yaml:"signal"`
+	IP              string                `json:"ip" yaml:"ip"`
+	PubKey          string                `json:"publicKey" yaml:"publicKey"`
+	KernelInterface bool                  `json:"usesKernelInterface" yaml:"usesKernelInterface"`
+	FQDN            string                `json:"domain" yaml:"domain"`
```
mlsmaycon: it might make sense for us to use FQDN only, including plain text.

client/cmd/status.go
```diff
@@ -29,65 +75,98 @@ var (
...
+		)
+		return nil
+	}
+
+	if ipv4Flag {
+		cmd.Print(parseInterfaceIP(resp.GetFullStatus().GetLocalPeerState().GetIP()))
+		return nil
+	}
+
+	statusOutputOverview := convertToStatusOutputOverview(resp)
```
mlsmaycon: `statusOutputOverview` overlaps with the the struct statusOutputOverview

client/cmd/status.go
```diff
@@ -29,65 +75,98 @@ var (
...
+
+	statusOutputString := ""
+	if detailFlag {
```
mlsmaycon: maybe a switch case will be better

mlsmaycon (changes requested):
loved the outcome, I left some key naming comments and few adjustments

client/cmd/status.go
pascal-fischer: Ok I will change the order

client/cmd/status.go
pascal-fischer: Sure makes sense

client/cmd/status.go
pascal-fischer: I thought full name looks better, especially for yaml output but I can change it

client/cmd/status.go
pascal-fischer: I was following the existing informations but in case the daemon is offline I should show some different information

client/cmd/status.go
pascal-fischer: I would very much support a switch case (I try to avoid "else If" as much as possible) but I didn't find a proper way to write the statement as I have flags and not a specific variable, so I am open to suggestions

client/cmd/status.go
pascal-fischer: you mean in the json and yaml output name?

client/cmd/status.go
mlsmaycon: something like this:
```Go
switch {
	case detailFlag:
		statusOutputString = parseToFullDetailSummary(statusOutputOverview)
	case jsonFlag:
		statusOutputString, err = parseToJSON(statusOutputOverview)
	case yamlFlag:
		statusOutputString, err = parseToYAML(statusOutputOverview)
	default:
		statusOutputString = parseGeneralSummary(statusOutputOverview, false)
	}
	
	if err != nil {
		return err
	}
```

client/cmd/status.go
mlsmaycon: json, yaml and plain text too.

client/cmd/status.go
mlsmaycon: yeah, it will probably exit on getStatus call

client/cmd/status.go
pascal-fischer: I removed the conn prefix completely and also renamed the status update to `LastStatusUpdate` as it describes a timestamp

client/cmd/status.go
pascal-fischer: I will remove the field

mlsmaycon (approved):

PR #711, opened 2023-03-01
Remove stale peer indices when getting peer by key after removing

braginini:
## Describe your changes

When we delete a peer from an account, we save the account in the file store.
The file store maintains peerID -> accountID and peerKey -> accountID indices.
Those can't be updated when we delete a peer because the store saves the whole account
without a peer already and has no access to the removed peer.
In this PR, we dynamically check if there are stale indices when GetAccountByPeerPubKey 
and GetAccountByPeerID.    

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/file_store_test.go
+++ b/management/server/file_store_test.go
@@ -14,6 +14,44 @@ type accounts struct {
 	Accounts map[string]*Account
 }
 
+func TestStalePeerIndices(t *testing.T) {
+	storeDir := t.TempDir()
+
+	err := util.CopyFileContents("testdata/store.json", filepath.Join(storeDir, "store.json"))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	store, err := NewFileStore(storeDir)
+	if err != nil {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


management/server/file_store_test.go
```diff
@@ -14,6 +14,44 @@ type accounts struct {
...
+	peerKey := "some_peer_key"
+	account.Peers[peerID] = &Peer{
+		ID:  peerID,
+		Key: peerKey,
+	}
+
+	err = store.SaveAccount(account)
+	require.NoError(t, err)
+
+	account.DeletePeer(peerID)
```
gigovich, line 43: Error is not checking.

gigovich (changes requested):
LGTM, need small fix in the tests.

management/server/file_store_test.go
gigovich: Ups, I'm wrong. Everything fine.

gigovich (dismissed):

gigovich (approved):

pappz (approved):

PR #717, opened 2023-03-03
Use the conn state of peer on proper way 

pappz:
## Describe your changes

The ConnStatus is a custom type based on iota like an enum. The 
problem was nowhere used to the benefits of this implementation. All 
ConnStatus instances has been compared with strings. I suppose the 
reason to do it to avoid a circle dependency. In this commit the 
separated status package has been moved to peer package.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


mlsmaycon (approved):
LGTM

Issue #718, opened 2023-03-03
Add support for personal access tokens and service users


pascal-fischer:
**Is your feature request related to a problem? Please describe.**
Being able to create personal access tokens to use for authenticating when sending API requests from 3rd-party applications or when using scripts.

**Describe the solution you'd like**
There should be an option to create personal access tokens for my user to use for authenticating agains the API. For better usability it would be good to have service users that are not bound to a specific person but can be created to hold tokens independently.

MohammedNoureldin:
Hi, how far is it from being ready? May I ask that you also add an initial access token (auto-generated or passed in the CLI) that will expire in for example 1 hour? This will help to automate the whole setup of Netbird.

PR #725, opened 2023-03-06
Adding Personal Access Token generation 

pascal-fischer:
## Describe your changes
Adding backend structs and logic to generate PATs for later usage within the service.

This PR is part of a set of PRs to fully support this feature.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


braginini (changes requested):
I think that we need to extend the Copy() function to copy tokens
https://github.com/netbirdio/netbird/blob/bcac5f7b327c382fce382e49ae82af4f9753198d/management/server/user.go#L94

management/server/user.go
```diff
@@ -95,6 +98,7 @@ func (u *User) Copy() *User {
 		Id:         u.Id,
 		Role:       u.Role,
 		AutoGroups: autoGroups,
+		PATs:       u.PATs,
```
braginini: I think that there should be an array copy. Not just a reference. Otherwise the the old user will keep the reference

braginini (changes requested):

management/server/personal_access_token.go
```diff
@@ -0,0 +1,61 @@
...
+	}, plainToken
+}
+
+func generateNewToken() ([32]byte, string) {
+	secret := randStringRunes(30)
+
+	checksum := crc32.ChecksumIEEE([]byte(secret))
+	encodedChecksum := base62.Encode(checksum)
+	paddedChecksum := fmt.Sprintf("%06s", encodedChecksum)
+	plainToken := "nbp_" + secret + paddedChecksum
```
mlsmaycon: Lets save the prefix in a const

management/server/personal_access_token.go
```diff
@@ -0,0 +1,61 @@
...
+// PersonalAccessToken holds all information about a PAT including a hashed version of it for verification
+type PersonalAccessToken struct {
+	ID             string
+	Description    string
+	HashedToken    [32]byte
+	ExpirationDate time.Time
+	// scope could be added in future
+	CreatedBy string
+	CreatedAt time.Time
+	LastUsed  time.Time
```
mlsmaycon, line 29: we could display partially the beginning of the token in the UI, so it might make sense to store 3 to 4 chars in plain text. maybe that would demand an increase in the char number

management/server/user.go
```diff
@@ -91,10 +94,13 @@ func (u *User) toUserInfo(userData *idp.UserData) (*UserInfo, error) {
 func (u *User) Copy() *User {
 	autoGroups := make([]string, 0)
```
pappz: When you create a new slice try to set a proper capacity. The append is more efficient in this case. I think in this scenario the built in copy function fit better here.

management/server/personal_access_token.go
pascal-fischer: We decided to leave this and reconsider at a later point

braginini (approved):

PR #727, opened 2023-03-06
Exchange proxy mode via signal

mlsmaycon:
## Describe your changes

Before defining if we will use direct or proxy connection we will exchange a message with the other peer if the modes match we keep the decision from the shouldUseProxy function otherwise we skip using direct connection.

Added a feature support message to the signal protocol

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/engine.go
```diff
@@ -353,6 +353,15 @@ func signalCandidate(candidate ice.Candidate, myKey wgtypes.Key, remoteKey wgtyp
 	return nil
 }
 
+func sendSignal(message *sProto.Message, s signal.Client) error {
+	err := s.Send(message)
```
pappz: I think easier if you return with s.Send(message), without check the err inside this scope

client/internal/peer/conn.go
```diff
@@ -399,6 +402,41 @@ func (conn *Conn) startProxy(remoteConn net.Conn, remoteWgPort int) error {
...
+	}
+
+	waitTimeout := time.After(time.Second)
+
+	select {
+	case receivedMSG := <-conn.modeMsg:
+		if receivedMSG.Direct != useProxy {
+			useProxy = true
+		}
+	case <-waitTimeout:
```
pappz: This select cause an extra latency in the connection with older clients. Somehow the client does not publish his version number? If not I do not have better idea than this solution.

mlsmaycon: newer versions are sending their own version in the signal message, I can introduce version validation for this method. See https://github.com/netbirdio/netbird/blob/main/signal/proto/signalexchange.proto#L54

client/internal/peer/conn.go
```diff
@@ -613,3 +656,13 @@ func (conn *Conn) OnRemoteCandidate(candidate ice.Candidate) {
...
 	return conn.config.Key
 }
+
+// OnModeMessage unmarshall the payload message and send it to the mode message channel
+func (conn *Conn) OnModeMessage(message ModeMessage) error {
+	select {
+	case conn.modeMsg <- message:
+		return nil
+	default:
+		return fmt.Errorf("unable to send mode message: channel busy")
```
braginini: This should be `unable to receive mode message: channel busy` or `unable to process mode message: channel busy`

client/internal/peer/conn.go
```diff
@@ -109,6 +119,7 @@ func NewConn(config ConnConfig, statusRecorder *Status) (*Conn, error) {
 		remoteOffersCh: make(chan OfferAnswer),
 		remoteAnswerCh: make(chan OfferAnswer),
 		statusRecorder: statusRecorder,
+		modeMsg:        make(chan ModeMessage),
```
braginini: I suggest use the same naming pattern as remoteAnswerCh and remoteOffersCh

braginini (changes requested):

signal/client/client.go
```diff
@@ -60,3 +72,14 @@ type Credential struct {
 	UFrag string
 	Pwd   string
 }
+
+// ParseFeaturesSupported parses a slice of supported features into FeaturesSupport
+func ParseFeaturesSupported(featuresMessage []uint32) FeaturesSupport {
+	var protoSupport FeaturesSupport
+	for _, feature := range featuresMessage {
+		if feature == DirectCheck {
+			protoSupport.DirectCheck = true
```
pappz, line 81: I think here the function should return

braginini (approved):

PR #732, opened 2023-03-09
Fix connstate indication

pappz:
## Describe your changes

Fix the status indication in the client service. The status of the 
management server and the signal server was incorrect if the network 
connection was broken. Basically the status update was not used by 
the management and signal library.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/server/server.go
+++ b/client/server/server.go
@@ -97,6 +97,8 @@ func (s *Server) Start() error {
 
 	if s.statusRecorder == nil {
 		s.statusRecorder = peer.NewRecorder(config.ManagementURL.String())
+	} else {
+		s.statusRecorder.UpdateManagementAddress(config.ManagementURL.String())
 	}
 
 	go func() {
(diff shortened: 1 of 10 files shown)


client/server/server.go
```diff
@@ -386,7 +386,7 @@ func (s *Server) Up(callerCtx context.Context, _ *proto.UpRequest) (*proto.UpRes
 	}
 
 	if s.statusRecorder == nil {
-		s.statusRecorder = peer.NewRecorder()
+		s.statusRecorder = peer.NewRecorder(s.config.ManagementURL.String())
```
mlsmaycon, line 391: To avoid management URL caching, we need to clean the `s.statusRecorder ` on `server.Down` calls. So when the user wants to update the management URL with the `netbird up -m https://newManager` the URL will reflect the change.

mlsmaycon (changes requested):

braginini (approved):

mlsmaycon (approved):

PR #739, opened 2023-03-13
Add version info command to signal server

pappz:
## Describe your changes

Add version command to signal and management servers.

The version information will be filled during build time.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/signal/cmd/root.go
+++ b/signal/cmd/root.go
@@ -7,6 +7,8 @@ import (
 	"runtime"
 
 	"github.com/spf13/cobra"
+
+	"github.com/netbirdio/netbird/version"
 )
 
 const (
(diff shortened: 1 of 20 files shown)


signal/cmd/root.go
```diff
@@ -15,14 +15,18 @@ const (
 )
 
 var (
+	// will be replaced with the release version when using goreleaser
+	version = "development"
```
mlsmaycon: we moved away from the global var to keep version in a single package `system`, have a look at https://github.com/netbirdio/netbird/blob/main/client/cmd/version.go, main benefits of this approach is that we can use the system.NetbirdVersion() call in other places and keep consistency

mlsmaycon (changes requested):

signal/cmd/root.go
pappz: That package is inside in the client tree. Should I move out the version logic to outer, to the root?

mlsmaycon (approved):

PR #745, opened 2023-03-16
PAT persistence

pascal-fischer:
## Describe your changes
Added support to store PATs with a user. Also allow lookup of respected user and account by provided PAT.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/account.go
```diff
@@ -1112,6 +1119,47 @@ func (am *DefaultAccountManager) redeemInvite(account *Account, userID string) e
 	return nil
 }
 
+// GetAccountFromPAT returns Account and User associated with a personal access token
+func (am *DefaultAccountManager) GetAccountFromPAT(token string) (*Account, *User, error) {
+	if len(token) != PATLength {
+		return nil, nil, fmt.Errorf("token invalid")
+	}
+
+	prefix := token[:len(PATPrefix)]
```
pappz, line 1128: If a user send an invalid token (shorter then len(PATPrefix) then this line can cause exception.

pappz (changes requested):

management/server/account.go
pascal-fischer: but the line above already checks if the token has the proper length, afterwards I just check the parts of the token (that are either valid or not but data should be there)

pappz: True!

management/server/account.go
```diff
@@ -1112,6 +1119,47 @@ func (am *DefaultAccountManager) redeemInvite(account *Account, userID string) e
...
+	if prefix != PATPrefix {
+		return nil, nil, fmt.Errorf("token invalid")
+	}
+	secret := token[len(PATPrefix) : len(PATPrefix)+PATSecretLength]
+	encodedChecksum := token[len(PATPrefix)+PATSecretLength : len(PATPrefix)+PATSecretLength+PATChecksumLength]
+
+	verificationChecksum, err := base62.Decode(encodedChecksum)
+	if err != nil {
+		return nil, nil, fmt.Errorf("token invalid")
```
gigovich: Here and below, could you please use this form of error handling:
`fmt.Errorf("token invalid: %w", err)`

pappz (dismissed):

management/server/account_test.go
```diff
@@ -458,6 +459,65 @@ func TestDefaultAccountManager_GetAccountFromToken(t *testing.T) {
...
+	}
+	err := store.SaveAccount(account)
+	if err != nil {
+		t.Fatalf("Error when saving account: %s", err)
+	}
+
+	am := DefaultAccountManager{
+		Store:                   store,
+		cacheMux:                sync.Mutex{},
+		cacheLoading:            nil,
```
gigovich: Why do you explicitly set nil?

management/server/account_test.go
```diff
@@ -458,6 +459,65 @@ func TestDefaultAccountManager_GetAccountFromToken(t *testing.T) {
...
+		peersUpdateManager:      nil,
+		idpManager:              nil,
+		cacheManager:            nil,
+		ctx:                     nil,
+		eventStore:              nil,
+		singleAccountMode:       false,
```
gigovich: Or false?

management/server/file_store.go
```diff
@@ -312,6 +344,42 @@ func (s *FileStore) GetAccountBySetupKey(setupKey string) (*Account, error) {
 	return account.Copy(), nil
 }
 
+// GetTokenIDByHashedToken returns the id of a personal access token by its hashed secret
+func (s *FileStore) GetTokenIDByHashedToken(token string) (string, error) {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+
+	tokenID, tokenIDFound := s.HashedPAT2TokenID[token]
```
gigovich: It's not mandatory or even a recommendation, but usually, as check variable names mostly used `ok`.

management/server/account.go
pascal-fischer: my initial thought here was that I don't want to show the user why his token is invalid for security reasons. But I guess as it is open source anyway I can show it (or hide it on a higher level). Will change it

management/server/account_test.go
pascal-fischer: No particular reason. IDE created object with all fields. I will remove the unused ones

management/server/file_store.go
pascal-fischer: Thanks, will do that from now on.

management/server/user.go
```diff
@@ -189,6 +193,61 @@ func (am *DefaultAccountManager) CreateUser(accountID, userID string, invite *Us
...
+	}
+
+	user := account.Users[userID]
+	if user == nil {
+		return status.Errorf(status.NotFound, "user not found")
+	}
+
+	user.PATs[pat.ID] = pat
+
+	err = am.Store.SaveAccount(account)
```
gigovich: `return am.Store.SaveAccount(account)`

gigovich (approved):
LGTM, but it is good if some minor style fixes will be added.

management/server/user.go
pascal-fischer: I kept that in two lines to improve readability because return type of SaveAccount wouldn't be intuitive to me (would have to check func definition) but I can do single line

gigovich (approved):

PR #769, opened 2023-03-28
Fix broken auto-generated Rego rule

gigovich:
## Describe your changes
Default Rego policy generated from the rules in some cases is broken.
This change fixes the Rego template for rules to generate policies.

Also, file store load constantly regenerates policy objects from rules.
It allows updating/fixing of the default Rego template during releases.

Maybe related to https://github.com/netbirdio/netbird/discussions/762

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


management/server/policy.go
```diff
@@ -218,39 +221,31 @@ func (f *FirewallRule) parseFromRegoResult(value interface{}) error {
...
-			continue
-		}
-		queries = append(queries, rego.Module(fmt.Sprintf("netbird-%d", i), p.Query))
-	}
-	return rego.New(queries...).PrepareForEval(context.TODO())
-}
-
-// getPeersByPolicy returns all peers that given peer has access to.
-func (a *Account) getPeersByPolicy(peerID string) ([]*Peer, []*FirewallRule) {
+func (a *Account) getRegoQuery(peerID string, qeuryNumber int, query string) ([]*Peer, []*FirewallRule) {
```
mlsmaycon: We should rename this method to something that reflects the returning values. 

Also, there is a typo with queryNumber

management/server/file_store.go
```diff
@@ -121,15 +121,25 @@ func restore(file string) (*FileStore, error) {
 			store.PrivateDomain2AccountID[account.Domain] = accountID
 		}
 
-		// if no policies are defined, that means we need to migrate Rules to policies
-		if len(account.Policies) == 0 {
+		// always regenerate policy objects which reflects Rules
```
mlsmaycon: Lets add a Todo here to evaluate policy templating as is the main cause of this regeneration.

management/server/policy.go

management/server/file_store.go
gigovich: Fixed

mlsmaycon (approved):

PR #771, opened 2023-03-28
Add grpc keep alive for management service

pappz:
## Describe your changes

Application level keep alive PoC for gRPC.

The concept is that mgmt server send the keep alive messages 
periodically.
Because the current setup does not support the both way communication 
the server inject keep alive (empty) messages in the Sync stream. So 
the keep alive procedure will start after the Sync call happens.

The application level keep alive will be applied only on that case when the 
server is behind reveres proxy. The server check it with the proper 
HTTP headers  to ensure to necessary this feature or not.

Because in the client has a bug what can cause exception I involved 
the version information into the gRPC function calls and the server 
send keep alive messages only if it is supported by the client.

Example nginx config:
```
server {
  listen *:80 http2;

  server_name signal.localhost;

  location / {
    grpc_set_header x-netbird-peer '$remote_addr:$remote_port';
    grpc_pass grpc://0.0.0.0:82;
  }
}

```

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


ykorzikowski:
Is there an ETA when this feature will be ready? I am thinking to disable my rproxy for netbird until this feature is done. I see this issue is occuring more often since latest release.

pappz:
@ykorzikowski could you send me the version information of your proxy?

xiaolei0125:
@pappz   I use Cloudflare proxy my netbird management and signal service,  But netbird client got disconnect and reconnect every 100 second.  

 this caused by cloudflare 524 timeoout:  https://developers.cloudflare.com/support/troubleshooting/cloudflare-errors/troubleshooting-cloudflare-5xx-errors/#error-524-a-timeout-occurred.  

>  Error 524 indicates that Cloudflare successfully connected to the origin web server, but the origin did not provide an HTTP response before the default 100 second connection timed out.

Can you continue add grpc keep alive for management and signal service ?  Thanks!

pktiuk:
This feature seems to be a good idea.

iamnotbobby:
@pktiuk 

I apologize for the year-later reply, but I'd love to see an implementation of keep-alive in gRPC. I'm running it behind Cloudflare and it can be a bit annoying having it fill up my logs.

PR #775, opened 2023-03-30
Add single line installer

bcmmbaga:
## Describe your changes
This is a shell script that installs the Netbird applications (CLI and UI) on a Linux or macOS system.

The shell script is designed to work with various Linux operating systems, including but not limited to Ubuntu, Fedora, Debian, CentOS, Raspberry Pi OS, Arch Linux, Alpine Linux, etc. The supported architectures include amd64, arm64, and 386. It can also be used on Mac OS, both on Intel and Apple Silicon architectures.

### Features
- Detects for officially supported package managers (apt, dnf, yum, brew) and installs via the respective package manager.
- Installs appropriate binaries based on the system's architecture and operating system type if package manager is not supported.
- Skip netbird-ui installation for Linux systems that lack a desktop environment or for systems that have an unsupported architecture.

For testing
`curl -fsSL https://raw.githubusercontent.com/bcmmbaga/netbird/main/release_files/install.sh | sh`

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/release_files/install.sh
+++ b/release_files/install.sh
@@ -62,13 +62,13 @@ add_apt_repo() {
 }
 
 add_rpm_repo() {
-cat <<-EOF | sudo tee /etc/yum.repos.d/wiretrustee.repo
-[Wiretrustee]
-name=Wiretrustee
-baseurl=https://pkgs.wiretrustee.com/yum/
+cat <<-EOF | sudo tee /etc/yum.repos.d/netbird.repo
+[Netbird]
+name=Netbird
+baseurl=https://pkgs.netbird.io/yum/
 enabled=1
 gpgcheck=0
-gpgkey=https://pkgs.wiretrustee.com/yum/repodata/repomd.xml.key
+gpgkey=https://pkgs.netbird.io/yum/repodata/repomd.xml.key
 repo_gpgcheck=1
 EOF
 } 
@@ -211,7 +211,7 @@ install_netbird() {
     dnf)
         add_rpm_repo
         sudo dnf -y install dnf-plugin-config-manager
-        sudo dnf config-manager --add-repo /etc/yum.repos.d/wiretrustee.repo
+        sudo dnf config-manager --add-repo /etc/yum.repos.d/netbird.repo
         sudo dnf -y install netbird
 
         if ! $SKIP_UI_APP; then 
--- a/release_files/install.sh
+++ b/release_files/install.sh
@@ -48,6 +48,19 @@ download_release_binary() {
     fi
 }
 
+add_apt_repo() {
+    sudo apt-get update
+    sudo apt-get install ca-certificates gnupg -y
+        
+    curl -sSL https://pkgs.wiretrustee.com/debian/public.key \
+    |  gpg --dearmor --output /usr/share/keyrings/wiretrustee-archive-keyring.gpg
+
+    APT_REPO="deb [signed-by=/usr/share/keyrings/wiretrustee-archive-keyring.gpg] https://pkgs.wiretrustee.com/debian stable main"
+    echo "$APT_REPO" | sudo tee /etc/apt/sources.list.d/wiretrustee.list
+
+    sudo apt-get update
+}
+
 add_rpm_repo() {
 cat <<-EOF | sudo tee /etc/yum.repos.d/wiretrustee.repo
 [Wiretrustee]
@@ -181,16 +194,7 @@ install_netbird() {
     # only the CLI will be installed
     case "$PACKAGE_MANAGER" in
     apt)
-        sudo apt-get update
-        sudo apt-get install ca-certificates gnupg -y
-        
-        curl -sSL https://pkgs.wiretrustee.com/debian/public.key \
-        |  gpg --dearmor --output /usr/share/keyrings/wiretrustee-archive-keyring.gpg
-
-        APT_REPO="deb [signed-by=/usr/share/keyrings/wiretrustee-archive-keyring.gpg] https://pkgs.wiretrustee.com/debian stable main"
-        echo "$APT_REPO" | sudo tee /etc/apt/sources.list.d/wiretrustee.list
-
-        sudo apt-get update
+        add_apt_repo
         sudo apt-get install netbird -y
         
         if ! $SKIP_UI_APP; then 
--- a/release_files/install.sh
+++ b/release_files/install.sh
@@ -100,6 +100,11 @@ install_netbird() {
     # Checks if SKIP_UI_APP env is set
     if [ -z "$SKIP_UI_APP" ]; then
         SKIP_UI_APP=false
+    else
+        if $SKIP_UI_APP; then 
+            echo "SKIP_UI_APP has been set to true in the environment"
+            echo "Netbird UI installation will be omitted based on your preference"
+        fi
     fi
 
     # Identify OS name and default package manager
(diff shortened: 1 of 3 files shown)


braginini:
Great job, @bcmmbaga! Could you please drop a simple description of the main features/functionality of the script? Maybe also mentioing the OS it supports, etc.

Danke!

release_files/install.sh
```diff
@@ -0,0 +1,251 @@
...
+    # download and copy binaries to INSTALL_DIR
+    download_release_binary "$CLI_APP"
+    if ! $SKIP_UI_APP; then 
+        download_release_binary "$UI_APP"
+    fi  
+}
+
+install_netbird() {
+    # Checks if SKIP_UI_APP env is set
+    if [ -z "$SKIP_UI_APP" ]; then
```
mlsmaycon, line 141: I think that besides checking if the SKIP_UI_APP was provided, you can check if the environment that the script is being executed have a desktop environment installed, see example: https://unix.stackexchange.com/a/645761

release_files/install.sh
```diff
@@ -0,0 +1,251 @@
...
+        mv "$UNZIPPED_FOLDER"/Info.plist "${INSTALL_DIR}/"
+        mv "$UNZIPPED_FOLDER"/_CodeSignature/ "${INSTALL_DIR}/"
+    else
+        tar -xzvf "$BINARY_NAME"
+        sudo mv "${1%_"${BINARY_BASE_NAME}"}" "$INSTALL_DIR"
+    fi
+}
+
+add_rpm_repo() {
+cat <<-EOF | sudo tee /etc/yum.repos.d/wiretrustee.repo
```
mlsmaycon: we can rename the repo file to `netbird.repo` and `pkgs.wiretrustee.com` domain to `pkgs.netbird.io`

release_files/install.sh
```diff
@@ -0,0 +1,251 @@
...
+            fi
+		;;
+	esac
+    fi
+
+    # Run the installation, if a desktop environment is not detected
+    # only the CLI will be installed
+    case "$PACKAGE_MANAGER" in
+    apt)
+        sudo apt-get update
```
mlsmaycon: we could follow `add_rpm_repo` and wrap most of the commands here to a separate function

release_files/install.sh
```diff
@@ -0,0 +1,251 @@
...
+    download_release_binary "$CLI_APP"
+    if ! $SKIP_UI_APP; then 
+        download_release_binary "$UI_APP"
+    fi  
+}
+
+install_netbird() {
+    # Checks if SKIP_UI_APP env is set
+    if [ -z "$SKIP_UI_APP" ]; then
+        SKIP_UI_APP=false
```
mlsmaycon, line 142: Let's print a info that the UI installation is being skipped or not

release_files/install.sh
```diff
@@ -0,0 +1,251 @@
...
+            echo "sudo nixos-rebuild switch"
+			exit 0
+        fi   
+
+        install_native_binaries
+    ;;
+    esac
+
+    # Start client daemon service if only CLI is installed
+    if ! $SKIP_UI_APP; then
```
mlsmaycon: we can keep these commands execution in all cases, but I would change the if do the following:

```shell
   netbird status > /dev/null 2>&1
    if [ $? -ne 0 ]; then
```

release_files/install.sh
```diff
@@ -0,0 +1,251 @@
...
+                SKIP_UI_APP=true
+            fi
+
+            # Allow netbird UI installation for linux running desktop enviroment
+            if [ -z "$XDG_CURRENT_DESKTOP" ];then
+                    SKIP_UI_APP=true
+            fi
+
+            # Check the availability of a compatible package manager
+            if [ -x "$(command -v apt)" ]; then
```
mlsmaycon, line 172: lets print the package manager to the user

release_files/install.sh
```diff
@@ -0,0 +1,251 @@
...
+    if ! $SKIP_UI_APP; then
+        sudo netbird service install 2>/dev/null
+        sudo netbird service start 2>/dev/null
+    fi
+
+    echo "Installation has been finished. To connect, you need to run NetBird by executing the following command:"
+    echo ""
+    echo "sudo netbird up"
+}
+
```
mlsmaycon, line 286: We should check if at least the cli is installed before proceeding. We should check if the process is running too, to prevent failures. 

```shell
if [ -x "$(command -v yum)" ]; then
   netbird status > /dev/null 2>&1
    if [ $? -ne 0 ]; then
        echo "netbird service is running, please stop it before proceeding"
    fi    
    echo "netbird seems to be installed already, please remove it before proceeding"
    exit 1
fi    
```

release_files/install.sh
```diff
@@ -0,0 +1,251 @@
...
+    DOWNLOAD_URL="${BASE_URL}/${VERSION}/${BINARY_NAME}"
+
+    echo "Installing $1 from $DOWNLOAD_URL"
+    cd /tmp && curl -LO "$DOWNLOAD_URL" 
+    
+    if [ "$OS_TYPE" = "darwin" ] && [ "$1" = "$UI_APP" ]; then
+        # Unzip the app
+        unzip -q -o "$BINARY_NAME"
+
+        INSTALL_DIR="/Applications/NetBird UI.app/Contents"
```
mlsmaycon: Because of signature reasons, after decompressing you can follow homebrew approach and move the whole dir to the `/Applications/Netbird\ UI.app` e.g.:
```shell
mv /tmp/netbird_ui_darwin_arm64 /Applications/Netbird\ UI.app
```

release_files/install.sh
```diff
@@ -0,0 +1,251 @@
...
+        sudo dnf config-manager --add-repo /etc/yum.repos.d/wiretrustee.repo
+        sudo dnf -y install netbird
+
+        if ! $SKIP_UI_APP; then 
+            sudo dnf -y install netbird-ui
+        fi
+    ;;
+    pacman)
+        sudo pacman -Syy
+        sudo pacman -Sy --needed --noconfirm git base-devel go
```
mlsmaycon: I got this error on arch:

```shell
warning: Public keyring not found; have you run 'pacman-key --init'?
downloading required keys...
error: keyring is not writable
error: keyring is not writable
error: keyring is not writable
error: keyring is not writable
error: keyring is not writable
error: keyring is not writable
error: keyring is not writable
error: keyring is not writable
error: keyring is not writable
error: required key missing from keyring
error: failed to commit transaction (unexpected error)
Errors occurred, no packages were upgraded.
```

mlsmaycon (changes requested):
@bcmmbaga thanks for the PR. I've left some notes that might be helpful to avoid some issues with the script. The new domain is new and we didn't had a change to update our docs, but they are working.

release_files/install.sh
bcmmbaga: This is already implemented for linux OS. [L115](https://github.com/bcmmbaga/netbird/blob/37f1733769abe32d940b0978bdc6fb7b68d7920e/release_files/install.sh#L115)

release_files/install.sh
bcmmbaga: I have tested it on Manjaro and encountered no problems. It is possible that the issue you are experiencing is due to the Pacman keyring being absent.

PR #779, opened 2023-04-03
Bind implementation

braginini:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/peer/stdnet.go
```diff
@@ -3,9 +3,9 @@
 package peer
 
 import (
-	"github.com/pion/transport/v2/stdnet"
+	"github.com/netbirdio/netbird/client/internal/stdnet"
 )
 
 func (conn *Conn) newStdNet() (*stdnet.Net, error) {
-	return stdnet.NewNet()
+	return stdnet.NewNet(nil, conn.config.InterfaceBlackList)
```
pappz: In this construction make no sense for two different stdnet.go and stdnet_android.go Both of them doing the same thing at the end. 
I recommend you to create two constructor in the stdnet pkg. On of them with blacklist and one of them with discoverer and with blacklist. In this case in the UpdateInterfaces function you can avoid the if/else statement because you can create two separated function.

pappz (changes requested):

iface/tun_windows.go
```diff
@@ -32,6 +36,40 @@ func (c *tunDevice) Create() error {
...
+	}
+
+	// We need to create a wireguard-go device and listen to configuration requests
+	tunDevice := device.NewDevice(tunIface, c.iceBind, device.NewLogger(device.LogLevelSilent, "[wiretrustee] "))
+	err = tunDevice.Up()
+	if err != nil {
+		return tunIface, err
+	}
+
+	uapi, err := c.getUAPI(c.name)
```
pappz: I do not see where do you close the uapi resource

iface/tun_windows.go
```diff
@@ -32,6 +36,40 @@ func (c *tunDevice) Create() error {
...
+	}
+
+	go func() {
+		for {
+			uapiConn, uapiErr := uapi.Accept()
+			if uapiErr != nil {
+				log.Traceln("uapi Accept failed with error: ", uapiErr)
+				continue
+			}
+			go tunDevice.IpcHandle(uapiConn)
```
pappz: Where do you close the uapiConn?

pappz (changes requested):

iface/tun_windows.go
braginini: Just added this to the Close() method

iface/tun_windows.go
```diff
@@ -2,33 +2,74 @@ package iface
...
+
+	tunIface, err := tun.CreateTUN(c.name, c.mtu)
+	if err != nil {
+		return nil, err
+	}
+	// We need to create a wireguard-go device and listen to configuration requests
+	tunDevice := device.NewDevice(tunIface, c.iceBind, device.NewLogger(device.LogLevelSilent, "[netbird] "))
+	err = tunDevice.Up()
+	if err != nil {
+		return tunIface, err
```
pappz: In this case the tunDevice never will be closed.

iface/tun_windows.go
```diff
@@ -2,33 +2,74 @@ package iface
...
+	}
+
+	uapi, err := c.getUAPI(c.name)
+	if err != nil {
+		return nil, err
```
pappz, line 65: In this case the tunDevice never will be closed.

iface/tun_windows.go
```diff
@@ -2,33 +2,74 @@ package iface
...
+	}
+	c.uapi = uapi
+
+	go func() {
```
pappz, line 68: I think this thread never exit. It will loop forever. This is true for tun_unix.go also

client/internal/engine.go
```diff
@@ -166,68 +166,77 @@ func (e *Engine) Stop() error {
...
-	}
-
-	networkName := "udp"
-	if e.config.DisableIPv6Discovery {
-		networkName = "udp4"
-	}
-
 	transportNet, err := e.newStdNet()
 	if err != nil {
 		log.Warnf("failed to create pion's stdnet: %s", err)
```
pappz: I think it should be log.ErrorF because if it happens than it is a real problem what we want to solve (by the way the client can goes well).

pappz (changes requested):

client/internal/peer/conn_test.go
```diff
@@ -1,6 +1,7 @@
 package peer
 
 import (
+	"github.com/netbirdio/netbird/client/internal/stdnet"
```
pappz: Please move this import to end of the import's list

client/internal/proxy/direct.go
```diff
@@ -0,0 +1,57 @@
+package proxy
+
+import (
+	log "github.com/sirupsen/logrus"
```
pappz, line 4: Please order the imports

pappz (changes requested):

client/internal/proxy/direct.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// DirectNoProxy is used when there is no need for a proxy between ICE and WireGuard.
+// This is possible in either of these cases:
+// - peers are in the same local network
+// - one of the peers has a public static IP (host)
+// DirectNoProxy will just update remote peer with a remote host and fixed WireGuard port (r.g. 51820).
+// In order DirectNoProxy to work, WireGuard port has to be fixed for the time being.
+type DirectNoProxy struct {
+	config Config
```
pappz, line 15: I see on several place in this repo to use a huge "config" struct instead of pass arguments via constructor. This is a really bad pattern because for the developer hard to understand the scope of the current checked code and logic of the system. In the current situation the DirectNoProxy does not use all parameters of Config. So why we store it in memory? If we check deeper the proxy.Config we can realize  that we completely eliminate it because almost the all information what carry is initialized on during the process's startup time and it will never changing but we duplicate these information for every peer.

So I think this is a great time to do a quick and simple refactor.
What I recommend:
- Delete the proxy.Config and pass all necessary arguments to the constructors.
- The PreSharedKey is come from cfg file and it has never changing in runtime. So in this part of the code make no sens to take care about it. Update the iface.wGConfigurer.updatePeer function and remove this argument.
- WgListenAddr probably the same situation. We pass the local wg port within peer.ConnConfig and proxyConfig to the peer.Conn and both config store the same information use only one of them or create a new function in the iface.WGIface for get back this info.
- in iface.wGConfigurer the keepalive time is using the default time always. You can remove this parameter from the function.

pappz (changes requested):

client/internal/proxy/direct.go
braginini: We initially used a config struct to avoid constantly changing the signature of methods. In the case of a proxy, this might be unnecessary because there are just a few parameters. So it makes sense to do the refactor you suggested. But I'd instead do it in a separate PR. DOes this make sense?

pappz: I understand your point and it is true and good. In case the proxyConfig I think all information is available peer.ConnConfig also.
In a separated pull request the workflow is nicer but it cause extra work for you. I think the feature/bind is already a huge change so does not matter more one extra change. I suppose if you merge the bind into the main branch you never want to revert it any more.

iface/bind/bind.go
```diff
@@ -0,0 +1,208 @@
...
+	sharedConn net.PacketConn
+	udpMux     *UniversalUDPMuxDefault
+
+	// below are fields initialized on creation
+	transportNet transport.Net
+	mu           sync.Mutex
+}
+
+// NewICEBind create a new instance of ICEBind with a given transportNet and an interfaceFilter function.
+// The interfaceFilter function is used to exclude interfaces from hole punching (the IPs of that interfaces won't be used as connection candidates)
```
pappz: The comment is deprecated I think. The interfaceFilter is not applied here.

iface/bind/bind.go
```diff
@@ -0,0 +1,208 @@
+package bind
+
+import (
+	"errors"
```
pappz, line 4: Please sort the imports

pappz (changes requested):

iface/bind/bind.go
```diff
@@ -0,0 +1,208 @@
...
+	return func(buff []byte) (int, conn.Endpoint, error) {
+		n, endpoint, err := c.ReadFrom(buff)
+		if err != nil {
+			return 0, nil, err
+		}
+		e, err := netip.ParseAddrPort(endpoint.String())
+		if err != nil {
+			return 0, nil, err
+		}
+		if !stun.IsMessage(buff[:20]) {
```
pappz: This step is risky. If you do not check the length of the array then it can cause exception.

pappz (changes requested):

iface/bind/bind.go
```diff
@@ -0,0 +1,208 @@
...
+	)
+	if err != nil {
+		return nil, 0, err
+	}
+	return conn, uaddr.Port, nil
+}
+
+func parseSTUNMessage(raw []byte) (*stun.Message, error) {
+	msg := &stun.Message{
+		Raw: append([]byte{}, raw...),
```
pappz: Maybe does not need to append

pappz (changes requested):

iface/bind/bind.go
```diff
@@ -0,0 +1,208 @@
...
+	if err1 != nil {
+		return err1
+	}
+
+	return err2
+}
+
+// SetMark sets the mark for each packet sent through this Bind.
+// This mark is passed to the kernel as the socket option SO_MARK.
+func (b *ICEBind) SetMark(mark uint32) error {
```
pappz, line 163: Is it ok to doing nothing this function?

pappz (approved):

Issue #623, opened 2022-12-12
Option to provide device name during registration


pnowy:
**Is your feature request related to a problem? Please describe.**

When single user would like to multiple devices would be nice to have option to provide device name by CLI parameter (like `--name`) as could be hard to guess by NetBird operator who registered the device.

**Describe the solution you'd like**

Provide the device name manually to CLI or somehow automatically get the metadata info who registered the device

**Describe alternatives you've considered**

Have an option to configure JWT claim from IDP and add some metadata which user registered the device (but probably this will be solved by activity events showed by @braginini)

cclements:
This would also be useful for me.  An --dev-name argument following --setup-key would be helpful, especially when using the same setup key for multiple endpoints.

lfarkas:
it'd be very useful eg in case of home assistant or any other way when netbirn run inside a container. in this case it's very annoying that the hostname will be a docker container id eg: 049e78cb8a5d

PR #789, opened 2023-04-06
feat: add support for custom device hostname

Thunderbottom:
## Describe your changes

Adds support for `--hostname` (or `-n`) flag to be passed during `netbird up`, allowing users to set up a custom hostname for devices. Particularly helpful in cases where the device automatically logs into netbird on spawn, or in other cases for orchestration.

fixes https://github.com/netbirdio/netbird/issues/623

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


lfarkas:
i really like to add this feature

mlsmaycon (commented):
@Thunderbottom thanks for this contribution and many others. I like where you are going here, but it might make sense for use to take a different route. Recently, we added a context check for the android system.GetInfo check, where we read the hostname from a context key, perhaps for a smaller PR, we can do the same. 

See the links below for reference:

https://github.com/netbirdio/netbird/blob/main/client/system/info_android.go#L34
https://github.com/netbirdio/netbird/blob/main/client/android/client.go#L74

As this would update the System.Info, you probably won't need to change management code.

Thunderbottom:
Yeah, that makes much more sense. I've made my changes and pushed to this PR to add support for all systems. Let me know if it's fine by you.

client/cmd/up.go
```diff
@@ -55,6 +56,10 @@ func upFunc(cmd *cobra.Command, args []string) error {
 
 	ctx := internal.CtxInitState(cmd.Context())
 
+	if hostName != "" {
```
mlsmaycon, line 59: we need to add the context update to the login command as well.

client/cmd/up.go
```diff
@@ -55,6 +56,10 @@ func upFunc(cmd *cobra.Command, args []string) error {
...
+		ctx = context.WithValue(ctx, system.DeviceNameCtxKey, hostName)
```
mlsmaycon, line 61: we decided to add the `//nolint` for this lint check

client/system/info_test.go
```diff
@@ -24,3 +24,11 @@ func Test_UIVersion(t *testing.T) {
 	got := GetInfo(ctx)
 	assert.Equal(t, want, got.UIVersion)
 }
+
+func Test_CustomHostname(t *testing.T) {
+	ctx := context.WithValue(context.Background(), DeviceNameCtxKey, "custom-host")
```
mlsmaycon, line 30: please add `//nolint` here too

mlsmaycon (changes requested):

mlsmaycon (approved):
thanks for the contribution @Thunderbottom

PR #791, opened 2023-04-06
Add service user logic and endpoints to management server

pascal-fischer:
## Describe your changes
- extend user object with flag for service users
- update user get endpoint to set query filter for service users (or not)
- split service user creation / inviting from idp
- allow user deletion (only for service users)

Whats missing:
- permission validation of operations in AccountManager (will be discussed separatly, for now accesscontrol middleware will block the requests)
- activity events (require update for all user related events to set isServiceUser flag to distinguish)

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/user.go
+++ b/management/server/user.go
@@ -126,7 +126,7 @@ func NewUser(id string, role UserRole, isServiceUser bool, serviceUserName strin
 	}
 }
 
-// NewRegularUser creates a new user with role UserRoleAdmin
+// NewRegularUser creates a new user with role UserRoleUser
 func NewRegularUser(id string) *User {
 	return NewUser(id, UserRoleUser, false, "", []string{})
 }
@@ -288,13 +288,6 @@ func (am *DefaultAccountManager) DeleteUser(accountID, executingUserID string, t
 	if err != nil {
 		return err
 	}
-
-	// refresh only needs to happen if we support delete of regular users
-
-	// _, err = am.refreshCache(account.Id)
-	// if err != nil {
-	// 	return err
-	// }
 
 	return nil
 }


management/server/user.go
```diff
@@ -101,34 +106,88 @@ func (u *User) Copy() *User {
...
-		AutoGroups: []string{},
+		Id:              id,
+		Role:            role,
+		IsServiceUser:   isServiceUser,
+		ServiceUserName: serviceUserName,
+		AutoGroups:      autoGroups,
 	}
 }
 
 // NewRegularUser creates a new user with role UserRoleAdmin
```
pappz: Invalid comment

pappz (changes requested):

management/server/user.go
```diff
@@ -193,8 +252,55 @@ func (am *DefaultAccountManager) CreateUser(accountID, userID string, invite *Us
...
+	delete(account.Users, targetUserID)
+
+	err = am.Store.SaveAccount(account)
+	if err != nil {
+		return err
+	}
+
+	// refresh only needs to happen if we support delete of regular users
+
+	// _, err = am.refreshCache(account.Id)
```
pappz: Please remove unused codes

management/server/user.go
pappz, line 1: - The UserStatusDisabled is unused. Please remove
- In the inviteNewUser the user's status will be UserStatusInvited or not?

pappz (changes requested):

management/server/user.go
pascal-fischer: Regarding the UserStatusDisabled. I would keep it because it comes from the API protocol and there we use all 3 constants. Just in code we don't need it right now but I would rather keep it. 

Regarding the UserStatusInvited, you are right, it is set in user.go L84

management/server/user.go
pascal-fischer: done

management/server/user.go
pascal-fischer: done

pappz (approved):

PR #799, opened 2023-04-09
Feat upgrade management api policy

gigovich:
## Describe your changes
Fixes for management policies API to handle UI changes

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


management/server/http/api/openapi.yml
```diff
@@ -450,44 +450,85 @@ components:
...
-          description: policy destination groups
-          type: array
-          items:
-            $ref: '#/components/schemas/GroupMinimum'
         action:
-          description: policy accept or drops packets
+          description: Policy rule accept or drops packets
           type: string
           enum: ["accept","drop"]
+        bidirect:
```
mlsmaycon: it will be better to use bidirectional, also the description:
```yaml
        bidirectional:
          description: Define if the rule is applicable in both directions, sources, and destinations.
          type: boolean
```

management/server/http/policies_handler.go
```diff
@@ -100,26 +105,49 @@ func (h *Policies) UpdatePolicy(w http.ResponseWriter, r *http.Request) {
 	if req.Rules != nil {
```
mlsmaycon: we should allow empty rules in request and return an error

management/server/http/policies_handler.go
```diff
@@ -149,6 +177,7 @@ func (h *Policies) CreatePolicy(w http.ResponseWriter, r *http.Request) {
 	var req api.PostApiPoliciesJSONRequestBody
 	err = json.NewDecoder(r.Body).Decode(&req)
 	if err != nil {
+		log.Errorf("failed to decode request body: %v", err)
```
mlsmaycon: left over?

management/server/http/policies_handler.go
```diff
@@ -170,23 +199,45 @@ func (h *Policies) CreatePolicy(w http.ResponseWriter, r *http.Request) {
 		for _, r := range req.Rules {
```
mlsmaycon: same comment as update policy

management/server/http/policies_handler.go
```diff
@@ -170,23 +199,45 @@ func (h *Policies) CreatePolicy(w http.ResponseWriter, r *http.Request) {
...
 				pr.Action = server.PolicyTrafficActionAccept
-			case api.PolicyRuleActionDrop:
+			case api.PolicyRuleUpdateActionDrop:
 				pr.Action = server.PolicyTrafficActionDrop
 			default:
 				util.WriteError(status.Errorf(status.InvalidArgument, "unknown action type"), w)
 				return
 			}
+
+			switch r.Protocol {
```
mlsmaycon: these 2 switch statements could be in functions as the checks are the same as in update policy

management/server/http/policies_handler.go
```diff
@@ -261,12 +312,15 @@ func (h *Policies) GetPolicy(w http.ResponseWriter, r *http.Request) {
 func toPolicyResponse(account *server.Account, policy *server.Policy) *api.Policy {
 	cache := make(map[string]api.GroupMinimum)
 	ap := &api.Policy{
-		Id:          policy.ID,
 		Name:        policy.Name,
 		Description: policy.Description,
 		Enabled:     policy.Enabled,
 		Query:       policy.Query,
 	}
+	if policy.ID != "" {
```
mlsmaycon: policy id should never be empty, right?

management/server/policy.go
```diff
@@ -66,6 +90,15 @@ type PolicyRule struct {
 
 	// Sources policy source groups
 	Sources []string
+
+	// Bidirect allows connection in both directions
+	Bidirect bool
```
mlsmaycon: same as API spec, lets use Birectional

management/server/policy.go
```diff
@@ -78,6 +111,9 @@ func (pm *PolicyRule) Copy() *PolicyRule {
 		Action:       pm.Action,
 		Destinations: pm.Destinations[:],
 		Sources:      pm.Sources[:],
+		Bidirect:     pm.Bidirect,
+		Protocol:     pm.Protocol,
+		Ports:        pm.Ports,
```
mlsmaycon: missing `[:]`

management/server/peer.go
```diff
@@ -423,6 +423,9 @@ func (am *DefaultAccountManager) GetNetworkMap(peerID string) (*NetworkMap, erro
 	if err != nil {
 		return nil, err
 	}
+	if len(account.Policies) > 0 {
```
mlsmaycon: left over?

mlsmaycon (changes requested):

management/server/http/api/openapi.yml
gigovich: Fixed

management/server/http/policies_handler.go
gigovich: Fixed

management/server/http/policies_handler.go
gigovich: Fixed

management/server/http/policies_handler.go
gigovich: Fixed

management/server/http/policies_handler.go
gigovich: I did "extract" refactoring, so I moved the common code to `savePolicy` helper method, so there is no more duplicate code.

management/server/http/policies_handler.go
gigovich: Fixed

management/server/policy.go

management/server/policy.go

management/server/peer.go
gigovich: Fixed

management/server/http/policies_handler.go
mlsmaycon: Policy ID should never be empty, so we don't need to check it.

management/server/http/policies_handler.go
```diff
@@ -261,22 +250,32 @@ func (h *Policies) GetPolicy(w http.ResponseWriter, r *http.Request) {
...
+		id := policy.ID
+		ap.Id = &id
+	}
 	if len(policy.Rules) == 0 {
```
mlsmaycon: policy.Rules should never be empty either, in this case we should panic or return nothing with a error 500

mlsmaycon (changes requested):

management/server/http/policies_handler.go
gigovich: Fixed

management/server/http/policies_handler.go
gigovich: Fixed

mlsmaycon (approved):

PR #801, opened 2023-04-11
Add route management for Android interface

pappz:
## Describe your changes

Support client route management feature on Android

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/routemanager/manager.go
```diff
@@ -1,9 +1,130 @@
...
+	clientNetworks map[string]*clientNetwork
+	serverRouter   *serverRouter
+	statusRecorder *peer.Status
+	wgInterface    *iface.WGIface
+	pubKey         string
+}
+
+// NewManager returns a new route manager
+func NewManager(ctx context.Context, pubKey string, wgInterface *iface.WGIface, statusRecorder *peer.Status) *DefaultManager {
+	mCTX, cancel := context.WithCancel(ctx)
```
pascal-fischer, line 36: I would use a more descriptive struct name than just Manager. Why not use RoutesManager like the package is called?

client/internal/routemanager/server_nonandroid.go
```diff
@@ -0,0 +1,121 @@
...
+		if err != nil {
+			return fmt.Errorf("couldn't initialize firewall containers, got err: %v", err)
+		}
+	}
+
+	for routeID := range m.routes {
+		update, found := routesMap[routeID]
+		if !found || !update.IsEqual(m.routes[routeID]) {
+			serverRoutesToRemove = append(serverRoutesToRemove, routeID)
+			continue
```
pascal-fischer: continue should have no effect

client/internal/routemanager/server_nonandroid.go
```diff
@@ -0,0 +1,121 @@
...
+
+	for _, routeID := range serverRoutesToRemove {
+		oldRoute := m.routes[routeID]
+		err := m.removeFromServerNetwork(oldRoute)
+		if err != nil {
+			log.Errorf("unable to remove route id: %s, network %s, from server, got: %v",
+				oldRoute.ID, oldRoute.Network, err)
+		}
+		delete(m.routes, routeID)
+	}
```
pascal-fischer, line 58: Why do we first collect all toDeleteRoutes and then in a second run delete them? why not both in one loop? as we do for adding, here we do one round

pappz: This code come from the original implementation. The scope of this pull request is out of the code cleaning.

client/internal/routemanager/manager.go
pappz: This code come from the original implementation. The scope of this pull request is out of the code cleaning.

PR #804, opened 2023-04-12
Feat firewall controller nftables

gigovich:
## Describe your changes
Add nftables firewall manager for client

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/firewall/nftables/manager_linux.go
```diff
@@ -0,0 +1,347 @@
...
+			cType,
+		)
+		if err != nil {
+			return nil, nil, err
+		}
+		m.filterChainIPv4 = chain
+		return m.tableIPv4, m.filterChainIPv4, nil
+	}
+	if m.filterChainIPv6 != nil {
+		return m.tableIPv4, m.filterChainIPv6, nil
```
mlsmaycon: should be tableIPv6

client/firewall/nftables/manager_linux_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	rule, err := manager.AddFiltering(
+		ip,
+		fw.ProtocolTCP,
+		&fw.Port{Values: []int{53}},
+		fw.DirectionSrc,
+		fw.ActionDrop,
+		"",
+	)
+	require.NoError(t, err, "failed to add rule")
```
mlsmaycon, line 36: Let's check if the rules were added to the system. See example: https://github.com/netbirdio/netbird/blob/main/client/internal/routemanager/nftables_linux_test.go#L162

client/firewall/nftables/manager_linux_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	err = manager.DeleteRule(rule)
+	require.NoError(t, err, "failed to delete rule")
```
mlsmaycon, line 82: same as above

client/internal/engine.go
```diff
@@ -1109,6 +1116,9 @@ func (e *Engine) protoRuleToFirewallRule(r *mgmProto.FirewallRule) firewall.Rule
 		protocol = firewall.ProtocolUDP
 	case "icmp":
 		protocol = firewall.ProtocolICMP
+	default:
+		log.Errorf("invalid protocol, skipping firewall rule: %q", r.Protocol)
+		return nil
```
mlsmaycon, line 1136: We should not apply partial update from management, so its better to attempt to convert all firewall rules form the update and if any of them are invalid abort the call to applyFirewallRules

client/internal/engine_test.go
```diff
@@ -946,6 +946,12 @@ func TestEngine_firewallManager(t *testing.T) {
 	if runtime.GOOS != "linux" {
 		t.Skipf("firewall manager not supported in the: %s", runtime.GOOS)
 		return
+	} else {
+		_, err := exec.LookPath("iptables")
+		if err != nil {
+			t.Skipf("iptables not found: %v", err)
```
mlsmaycon: nftables should cover this case, no?

mlsmaycon (changes requested):
it seems like we are missing the management.proto file as we got an update for management/proto/management.pb.go

client/firewall/nftables/manager_linux.go
gigovich: Fixed

client/firewall/nftables/manager_linux_test.go
gigovich: Fixed

client/internal/engine_test.go
gigovich: Fixed

client/firewall/nftables/manager_linux_test.go
gigovich: Fixed

client/internal/engine.go
gigovich: It is better to validate the rule in other places. That error in this method is critical and should not happen at all and mostly added by me for debugging reasons, because work is not finished yet in the logic layer.

mlsmaycon: that's the point, if there is an error, the whole ACL update must not proceed and no changes should be applied

gigovich: Fixed in the engine, if the rule applies failed in the update batch, the update canceled already applied rules from this batch removed.

mlsmaycon: Ok, with the ACL engine we should have this decoupled. Let merge and then work on it there.

mlsmaycon (approved):

PR #805, opened 2023-04-12
Feat linux firewall support

gigovich:
## Describe your changes
Update the client's engine to apply firewall rules received from the manager (results of ACL policy).

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


client/firewall/iptables/manager_linux.go
```diff
@@ -38,66 +55,88 @@ func Create() (*Manager, error) {
 
 	ipv6Client, err := iptables.NewWithProtocol(iptables.ProtocolIPv6)
 	if err != nil {
-		return nil, fmt.Errorf("ip6tables is not installed in the system or not supported")
+		log.Errorf("ip6tables is not installed in the system or not supported")
```
mlsmaycon: lets print the error among the message.

client/firewall/iptables/manager_linux.go
```diff
@@ -38,66 +55,88 @@ func Create() (*Manager, error) {
...
+
+	var portValue string
+	if port != nil && port.Values != nil {
+		// TODO: we support only one port per rule in current implementation of ACLs
+		portValue = strconv.Itoa(port.Values[0])
 	}
-	if port == nil || port.Values == nil || (port.IsRange && len(port.Values) != 2) {
-		return nil, fmt.Errorf("invalid port definition")
+
+	ruleID := uuid.New().String()
```
mlsmaycon, line 98: if we want to be able to troubleshoot issues, like users sending applied firewall rules, would be better to have an policy ID together with this generated ID

client/internal/engine.go
```diff
@@ -1039,6 +1054,92 @@ func (e *Engine) close() {
...
 
+	if e.firewallManager != nil {
+		if err := e.firewallManager.Reset(); err != nil {
+			log.Warnf("failed resetting firewall: %v", err)
+		}
+	}
+}
+
+// applyFirewallRules to the local firewall manager processed by ACL policy.
+func (e *Engine) applyFirewallRules(rules []*mgmProto.FirewallRule) {
```
mlsmaycon: as discussed, lets create an ACL controller layer that will be handling the updates and passing to firewall managers

client/firewall/iptables/manager_linux.go
mlsmaycon: like: `uuid-policyID-or-name`

management/proto/management.proto
```diff
@@ -297,4 +300,14 @@ message NameServer {
   string IP = 1;
   int64  NSType = 2;
   int64  Port = 3;
-}
\ No newline at end of file
+}
+
+// FirewallRule represents a firewall rule
+message FirewallRule {
```
mlsmaycon, line 306: lets share the rule id or name too?

mlsmaycon (commented):
I've added some comments to check after PRs 804 and 799 are merged

client/firewall/iptables/manager_linux.go
gigovich: Fixed

client/internal/engine.go
gigovich: Refactored, extracted as ACL module

client/internal/acl/manager.go
```diff
@@ -0,0 +1,145 @@
+package acl
+
+import (
+	"net"
+	"strconv"
+	"sync"
+
+	"github.com/netbirdio/netbird/client/firewall"
+	mgmProto "github.com/netbirdio/netbird/management/proto"
+	log "github.com/sirupsen/logrus"
```
pappz: Please move logrus import into separated block from Netbirdio relatead imports.

pappz (dismissed):

client/internal/acl/manager.go
gigovich: Fixed

braginini (approved):

Issue #806, opened 2023-04-13
Add support for rotating signing keys on expiry


Thunderbottom:
**Describe the problem**
Currently, signing keys are loaded on management startup, and are never refreshed on key expiry. IdPs often rotate these keys at a set interval for added security.

**To Reproduce**

1. Set up an IdP that rotates signing key at set intervals, eg. [dexIdP](https://dexidp.io)
2. Integrate the IdP with Netbird and start the Netbird management server, this should load the signing key from the IdP on startup.
3. Wait for the signing key to expire, in the case of dexIdP, the default expiry is set to `6h`.
4. Try to login to the management service and see:
![image](https://user-images.githubusercontent.com/11243138/231723454-9e3fa3d3-ae2f-4a91-87b0-c47174b55c49.png)

**Additional management service logs:**
```
infrastructure_files-management-1  | 2023-04-06T12:55:11Z DEBG management/server/jwtclaims/jwtValidator.go:135: error parsing token: unable to find appropriate key
infrastructure_files-management-1  | 2023-04-06T12:55:11Z DEBG management/server/http/middleware/auth_middleware.go:60: Error when validating JWT claims: Error parsing token: unable to find appropriate key
infrastructure_files-management-1  | 2023-04-06T12:55:11Z ERRO management/server/telemetry/http_api_metrics.go:150: HTTP response 4198917813: GET /api/users status 401
infrastructure_files-management-1  | 2023-04-06T12:55:11Z DEBG management/server/jwtclaims/jwtValidator.go:135: error parsing token: unable to find appropriate key
infrastructure_files-management-1  | 2023-04-06T12:55:11Z DEBG management/server/http/middleware/auth_middleware.go:60: Error when validating JWT claims: Error parsing token: unable to find appropriate key
infrastructure_files-management-1  | 2023-04-06T12:55:11Z ERRO management/server/telemetry/http_api_metrics.go:150: HTTP response 1761936853: GET /api/users status 401
infrastructure_files-management-1  | 2023-04-06T12:55:11Z DEBG management/server/jwtclaims/jwtValidator.go:135: error parsing token: unable to find appropriate key
infrastructure_files-management-1  | 2023-04-06T12:55:11Z DEBG management/server/http/middleware/auth_middleware.go:60: Error when validating JWT claims: Error parsing token: unable to find appropriate key
[...]

PR #808, opened 2023-04-14
Add support for refreshing signing keys on expiry

bcmmbaga:
## Describe your changes
Introduces a new flag `--idp-sign-key-refresh-enabled` that enables the Cache-Control header to be checked when fetching signing keys to determine the rotation period for signing keys. This feature will automatically refresh the signing key upon its expiry.

* Add support for rotating signing keys on expiry  ([#806][i806])

[i806]: https://github.com/netbirdio/netbird/issues/806
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/cmd/management.go
+++ b/management/cmd/management.go
@@ -80,6 +80,7 @@ var (
 			if err != nil {
 				return fmt.Errorf("failed reading provided config file: %s: %v", mgmtConfig, err)
 			}
+			config.HttpConfig.KeyRotationEnabled = UseKeyCacheHeaders
 
 			tlsEnabled := false
 			if mgmtLetsencryptDomain != "" || (config.HttpConfig.CertFile != "" && config.HttpConfig.CertKey != "") {
@@ -186,6 +187,7 @@ var (
 				config.HttpConfig.AuthIssuer,
 				config.GetAuthAudiences(),
 				config.HttpConfig.AuthKeysLocation,
+				config.HttpConfig.KeyRotationEnabled,
 			)
 			if err != nil {
 				return fmt.Errorf("failed creating JWT validator: %v", err)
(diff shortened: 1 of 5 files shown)


Thunderbottom:
Thanks for the quick response!

Although, I believe this should be the default behavior and not a flag. Most, if not all, IdPs rotate the signing keys after certain time intervals so it'll make sense to enable it by default.

bcmmbaga:
> Although, I believe this should be the default behavior and not a flag. Most, if not all, IdPs rotate the signing keys after certain time intervals so it'll make sense to enable it by default.

some IDP's may not rotate their signing keys automatically, which means that enabling this feature by default could result in additional calls to the endpoint every time a user requests access, potentially leading to rate limiting calls.

Thunderbottom:
I believe a better way to do this would be to check if the cache-control header exists with the expiry defined the first time netbird starts, and then use that to check locally every time someone tries to access. There's no need for another flag for doing that.

pascal-fischer:
I was reviewing the code and it looks good so far but if I am not mistaken we fetch the keys once on startup. When they are expired every following token validation will send a get request to get the new keys. Is there a way for us to store the refreshed keys? so we only request new keys once and store them until next expiration?

I also think it would be a good idea to set the key rotation as the default behaviour, but only if we keep the request count low.

bcmmbaga:
> I was reviewing the code and it looks good so far but if I am not mistaken we fetch the keys once on startup. When they are expired every following token validation will send a get request to get the new keys.

Currently, the keys are only retrieved once during startup and are not retrieved again after they have expired for Identity Providers that have automatic key rotation.

>  Is there a way for us to store the refreshed keys? so we only request new keys once and store them until next expiration?

Identity Providers that have a  auto key rotation in x interval  will also require the retrieval of a new key after it has expired

pascal-fischer (dismissed):

pascal-fischer (approved):

pascal-fischer (dismissed):

pascal-fischer (approved):

PR #818, opened 2023-04-19
fix: check if dns listener can bind before switching

Thunderbottom:
## Describe your changes

Currently, there's no check if DNS listener with custom address can bind to the specified `IP:Port`. In cases where the DNS listener fails to bind, the system's `resolvconf` is still changed even though the DNS listener is not running. This causes DNS to spectacularly fail on the client system until the user runs `netbird down`, replacing the netbird `resolvconf` with the original file.

This PR tries to fix the issue by not letting netbird change `resolvconf` if the DNS listener fails to bind in any case.

### Checklist
- [X] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/dns/server_nonandroid.go
+++ b/client/internal/dns/server_nonandroid.go
@@ -97,15 +97,20 @@ func NewDefaultServer(ctx context.Context, wgInterface *iface.WGIface, customAdd
 }
 
 // Start runs the listener in a go routine
-func (s *DefaultServer) Start() {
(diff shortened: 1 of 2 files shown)


mlsmaycon:
The PR #902 will solve that in part. 

Initially, we will support a fake address with userspace agents (windows, macOS, and older Linux) then we will add the same support to kernel agents. 

Closing this PR for the time being. Thanks @Thunderbottom  for the contribution.

PR #820, opened 2023-04-19
Drop Rego use manual code to calculate network map

gigovich:
## Describe your changes
Remove Rego engine to calculate network map and peer firewall rules for peer.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


management/server/policy.go
```diff
@@ -243,175 +170,102 @@ type FirewallRule struct {
...
 				continue
 			}
-			peers = append(peers, peer)
-			peersSeen[peer.ID] = struct{}{}
-		}
-		for _, rule := range r {
-			if _, ok := ruleSeen[rule.id]; ok {
-				continue
+
+			sourcePeers, peerInSources := getPeers(rule.Destinations)
```
mlsmaycon: should be rule.Sources

management/server/policy.go
```diff
@@ -243,175 +170,102 @@ type FirewallRule struct {
...
-
-	evalResult, err := stmt.Eval(
-		context.TODO(),
-		rego.EvalInput(input),
-	)
-	if err != nil {
-		log.WithError(err).Error("eval Rego query")
-		return nil, nil
-	}
+// getPeersByPolicy returns all peers that given peer has access to
```
mlsmaycon: as we are getting peers and firewall rules, we can probably rename the method to `getPeerConnectionResources` and modify the comments. 

Another thing, the returning definition, `(peers []*Peer, rules []*FirewallRule)`, would be better: `(connPeers []*Peer, fwRules []*FirewallRule)`

mlsmaycon (changes requested):

management/server/policy.go
gigovich: Fixed. I also added an extra test case because the old test did not cover direction connects.

management/server/policy.go
gigovich: I renamed it, simplified this function, and extracted the closure function as a generator and another as a regular helper function. Add comments to all of them.

mlsmaycon (approved):

PR #825, opened 2023-04-24
Use local time zone for display last update changes

mlsmaycon:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/cmd/status_test.go
+++ b/client/cmd/status_test.go
@@ -11,16 +11,13 @@ import (
 	"github.com/netbirdio/netbird/version"
 )
 
-var _ = initTZ()
-
-func initTZ() bool {
+func init() {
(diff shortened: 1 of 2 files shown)


client/cmd/status_test.go
```diff
@@ -11,6 +11,18 @@ import (
 	"github.com/netbirdio/netbird/version"
 )
 
+var _ = initTZ()
+
+func initTZ() bool {
```
pappz: It is not clear for me why do not you put this logic into a init() function instead of initTZ() and why return it back with unused bool.

mlsmaycon: I've updated to use the init function

pappz (approved):

PR #826, opened 2023-04-26
Share kernel Wireguard port with raw socket

mlsmaycon:
## Describe your changes
This is initial changes to support shared port between stun (ICE agent) and
the kernel Wireguard

Implement single port mode for execution with kernel Wireguard interfaces using a raw socket listener.

BPF filters ensure that only stun packets hit the raw socket

Removed a lot of the proxy logic and direct mode exchange

now we are doing an extra hole punch to the remote wireguard port for best-effort cases and support to old client's direct mode.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


client/stunlistener/listener_linux.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	ipv4Instructions, ipv6Instructions, err := getBPFInstructions(uint32(port))
+	if err != nil {
+		return nil, fmt.Errorf("getBPFInstructions failed with: %s", err)
+	}
+
+	err = socket4.SetBPF(ipv4Instructions)
+	if err != nil {
+		return nil, fmt.Errorf("socket4.SetBPF failed with: %s", err)
```
pappz: Here we should close the socket4 and socket6 also

client/stunlistener/listener_linux.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+	if socket6 != nil {
+		err = socket6.SetBPF(ipv6Instructions)
+		if err != nil {
+			return nil, fmt.Errorf("socket6.SetBPF failed with: %s", err)
```
pappz: Here we should close the socket4 and socket6 also

pappz (changes requested):

client/internal/peer/conn.go
```diff
@@ -356,14 +358,39 @@ func (conn *Conn) startProxy(remoteConn net.Conn, remoteWgPort int) error {
...
+		return err
+	}
+	addr.Port = remoteWgPort
+
+	mux, ok := conn.config.UDPMuxSrflx.(*bind.UniversalUDPMuxDefault)
+	if !ok {
+		return fmt.Errorf("invalid udp mux conversion")
+	}
+	_, err = mux.GetSharedConn().WriteTo([]byte{1}, addr)
+	if err != nil {
```
pappz, line 388: Unnecessary error check

pappz (changes requested):

client/stunlistener/listener_linux.go
```diff
@@ -0,0 +1,366 @@
...
+
+// read start a read loop for a specific receiver and sends the packet to the packetDemux channel
+func (s *StunListener) read(receiver receiver) {
+	for {
+		buf := make([]byte, 1500)
+		n, addr, err := receiver(s.ctx, buf, 0)
+		select {
+		case <-s.ctx.Done():
+			return
+		case s.packetDemux <- rcvdPacket{n, addr, buf[:], err}:
```
pappz: Make no sense to create a new buf slice with the same content. Just pass the buf variable.

client/stunlistener/listener_linux.go
```diff
@@ -0,0 +1,366 @@
...
+	if s.conn6 != nil {
+		errGrp.Go(s.conn6.Close)
+	}
+	return errGrp.Wait()
+}
+
+// read start a read loop for a specific receiver and sends the packet to the packetDemux channel
+func (s *StunListener) read(receiver receiver) {
+	for {
+		buf := make([]byte, 1500)
```
pappz: Is this loop can cause performance issue? If yes we should use the sync.Pool instead of create new slice all the time.

pappz (changes requested):

client/stunlistener/listener_linux.go
```diff
@@ -0,0 +1,366 @@
...
+	var payload gopacket.Payload
+	var parser *gopacket.DecodingLayerParser
+	var ip net.IP
+
+	if sa, isIPv4 := pkt.addr.(*unix.SockaddrInet4); isIPv4 {
+		ip = sa.Addr[:]
+		parser = gopacket.NewDecodingLayerParser(layers.LayerTypeIPv4, &ip4layer, &udp, &payload)
+	} else if sa, isIPv6 := pkt.addr.(*unix.SockaddrInet6); isIPv6 {
+		ip = sa.Addr[:]
+		parser = gopacket.NewDecodingLayerParser(layers.LayerTypeUDP, &udp, &payload)
```
pappz: The base is UDP. I think it should be LayerTypeIPv6

pappz (changes requested):

client/stunlistener/listener_linux.go
mlsmaycon: We receive the UDP layer for IPv6 address. It is similar in the bpf filter

braginini (dismissed):

sharedsock/sock_linux.go
```diff
@@ -0,0 +1,327 @@
...
+	} else if sa, isIPv6 := pkt.addr.(*unix.SockaddrInet6); isIPv6 {
+		ip = sa.Addr[:]
+		parser = gopacket.NewDecodingLayerParser(layers.LayerTypeUDP, &udp, &payload)
+	} else {
+		return -1, nil, fmt.Errorf("received invalid address family")
+	}
+
+	decodedLayers := make([]gopacket.LayerType, 0, 3)
+
+	err = parser.DecodeLayers(pkt.buf[:], &decodedLayers)
```
pappz, line 250: At this point the buf is may half empty. Maybe we should handle the N variable to reduca the len of the slice.

pappz (approved):

PR #836, opened 2023-05-05
Userspace packet filtering

gigovich:
## Describe your changes
Add TUN device wrapper and userspace packet filter to use it as firewall manager for ACL controller.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


pappz:
I see in the iface/tun_wrapper_unix.go you embed the **tun.Device** and because of it you wrap all exported functions. I think you can remove the all exported function from the source code except the Read, Write of course. So you can get a shorter code.
I recommend this changes:
```
// TunWrapper to override Read or Write of packets
type TunWrapper struct {
	tun.Device
	filter PacketFilter
}

// newTunInjection constructor function
func newTunInjection(device tun.Device) *TunWrapper {
	return &TunWrapper{
		Device: device,
	}
}
```

gigovich:
https://github.com/netbirdio/netbird/pull/836#issuecomment-1537957936
@pappz thanks, rewrote it. It was done without embedding for debugging reasons.

pappz (approved):

pappz (approved):

pappz (approved):

iface/device_wrapper.go
```diff
@@ -0,0 +1,68 @@
...
+	DropInput(packetData []byte) bool
+
+	// DropOutput traffic filter
+	DropOutput(packetData []byte) bool
+}
+
+// DeviceWrapper to override Read or Write of packets
+type DeviceWrapper struct {
+	tun.Device
+	filter PacketFilter
```
pappz, line 25: This variable get value from outside of this struct. It is not looks thread safe. Is it ok that the filter variable has not been protected during the **func Read**?

pappz (changes requested):

iface/device_wrapper.go
gigovich: Fixed. Added mutex and setter.

pappz (approved):

PR #846, opened 2023-05-09
Block user through HTTP API

braginini:
## Describe your changes

The new functionality allows to block a user in the Management service .
Blocked users loose access to the Dashboard, aren't able to modify network map,
and all of their connected devices disconnect and set to "login expired" state.

Technically all above achieved with the updated PUT /api/users endpoint,
that was extended with is_blocked field.    

image

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/http/users_handler_test.go
+++ b/management/server/http/users_handler_test.go
@@ -32,16 +32,19 @@ var usersTestAccount = &server.Account{
 			Id:            existingUserID,
 			Role:          "admin",
 			IsServiceUser: false,
+			AutoGroups:    []string{"group_1"},
 		},
 		regularUserID: {
 			Id:            regularUserID,
 			Role:          "user",
 			IsServiceUser: false,
+			AutoGroups:    []string{"group_1"},
 		},
 		serviceUserID: {
 			Id:            serviceUserID,
 			Role:          "user",
 			IsServiceUser: true,
+			AutoGroups:    []string{"group_1"},
 		},
 	},
 }
@@ -71,7 +74,7 @@ func initUsersTestData() *UsersHandler {
 				}
 				return key, nil
 			},
-			DeleteUserFunc: func(accountID string, executingUserID string, targetUserID string) error {
+			DeleteUserFunc: func(accountID string, initiatorUserID string, targetUserID string) error {
 				if targetUserID == notFoundUserID {
 					return status.Errorf(status.NotFound, "user with ID %s does not exists", targetUserID)
 				}
@@ -211,6 +214,18 @@ func TestUpdateUser(t *testing.T) {
 			expectedGroups:     []string{"group_2", "group_3"},
 			requestBody:        bytes.NewBufferString("{\"role\":\"admin\",\"auto_groups\":[\"group_3\", \"group_2\"],\"is_service_user\":false, \"is_blocked\": false}"),
 		},
+		{
+			name:               "Should_Fail_Because_AutoGroups_Is_Absent",
+			requestType:        http.MethodPut,
+			requestPath:        "/api/users/" + regularUserID,
+			expectedStatusCode: http.StatusBadRequest,
+			expectedUserID:     regularUserID,
+			expectedBlocked:    false,
+			expectedRole:       "admin",
+			expectedStatus:     "blocked",
+			expectedGroups:     []string{"group_2", "group_3"},
+			requestBody:        bytes.NewBufferString("{\"role\":\"admin\",\"is_service_user\":false, \"is_blocked\": false}"),
+		},
 	}
 
 	userHandler := initUsersTestData()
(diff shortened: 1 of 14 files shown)


management/server/http/users_handler_test.go
```diff
@@ -145,6 +161,108 @@ func TestGetUsers(t *testing.T) {
...
+				t.Fatalf("I don't know what I expected; %v", err)
+			}
+
+			respBody := &api.User{}
+			err = json.Unmarshal(content, &respBody)
+			if err != nil {
+				t.Fatalf("response content is not in correct json format; %v", err)
+			}
+
+			assert.Contains(t, tc.expectedUserID, respBody.Id)
```
pascal-fischer: what's the idea behind contains if we have equals in the next line?

management/server/http/users_handler_test.go
```diff
@@ -145,6 +161,108 @@ func TestGetUsers(t *testing.T) {
...
+		{
+			name:               "Update_Block_User",
+			requestType:        http.MethodPut,
+			requestPath:        "/api/users/" + regularUserID,
+			expectedStatusCode: http.StatusOK,
+			expectedUserID:     regularUserID,
+			expectedBlocked:    true,
+			expectedRole:       "user",
+			expectedStatus:     "blocked",
+			expectedGroups:     []string{"group_1"},
```
pascal-fischer, line 190: each of those test cases is also testing "add groups" maybe we should limit change to the object per test case to the scope of the test case (makes more clear what failed). We could update the initial data if we want to avoid the empty value (same we could add for isServiceUser or isBlocked)

management/server/peer.go
```diff
@@ -676,6 +686,19 @@ func (am *DefaultAccountManager) LoginPeer(login PeerLogin) (*Peer, *NetworkMap,
...
 }
 
+func checkIfPeerOwnerIsBlocked(peer *Peer, account *Account) error {
+	if peer.AddedWithSSOLogin() {
+		user, err := account.FindUser(peer.UserID)
+		if err != nil {
+			return status.Errorf(status.PermissionDenied, "user doesn't exist or was blocked")
+		}
+		if user.IsBlocked() {
+			return status.Errorf(status.PermissionDenied, "user doesn't exist or was blocked")
```
pascal-fischer: I would split the error messages (even tho user does not exists should be rather hard in this case as we have any option to delete and the user had to have existed at one point to add the peer)

management/server/user.go
```diff
@@ -457,9 +487,9 @@ func (am *DefaultAccountManager) GetAllPATs(accountID string, executingUserID st
 	return pats, nil
 }
 
-// SaveUser saves updates a given user. If the user doesn't exit it will throw status.NotFound error.
-// Only User.AutoGroups field is allowed to be updated for now.
-func (am *DefaultAccountManager) SaveUser(accountID, userID string, update *User) (*UserInfo, error) {
+// SaveUser saves updates to the given user. If the user doesn't exit it will throw status.NotFound error.
+// Only User.AutoGroups, User.Role, and User.Blocked fields are allowed to be updated for now.
+func (am *DefaultAccountManager) SaveUser(accountID, initiatorUserID string, update *User) (*UserInfo, error) {
```
pascal-fischer, line 487: I like the name initiatorUserID, we should adapt the other functions as well

management/server/user.go
```diff
@@ -472,56 +502,106 @@ func (am *DefaultAccountManager) SaveUser(accountID, userID string, update *User
...
+		return nil, status.Errorf(status.PermissionDenied, "admins can't block or unblock themselves")
+	}
+
+	if initiatorUser.IsAdmin() && initiatorUserID == update.Id && update.Role != UserRoleAdmin {
+		return nil, status.Errorf(status.PermissionDenied, "admins can't change their role")
 	}
 
 	// only auto groups, revoked status, and name can be updated for now
 	newUser := oldUser.Copy()
 	newUser.AutoGroups = update.AutoGroups
```
pascal-fischer: Autogroups update needs to be removed here, is updated after the check is all groups exist further down

management/server/user.go
```diff
@@ -51,15 +51,27 @@ type User struct {
...
 }
 
-// IsAdmin returns true if user is an admin, false otherwise
+// IsBlocked returns true if the user is blocked, false otherwise
+func (u *User) IsBlocked() bool {
+	return u.Blocked
+}
+
+// Block marks user as blocked
+func (u *User) Block() {
```
pascal-fischer: the Block() function is only used once, in tests. Everywhere else we access the object directly. If we have a Block operation we should also have unblock.

management/server/user.go
```diff
@@ -472,56 +502,106 @@ func (am *DefaultAccountManager) SaveUser(accountID, userID string, update *User
...
+				newGroupID, update.Id)
+		}
+	}
+	newUser.AutoGroups = update.AutoGroups
 
 	account.Users[newUser.Id] = newUser
 
+	if oldUser.Blocked != update.Blocked {
+		// expire peers that belong to the user who's getting blocked
+		if update.Blocked {
```
pascal-fischer: that could be one chained if `!oldUser.isBlocked() && update.isBlocked()`

management/server/user.go
```diff
@@ -472,56 +502,106 @@ func (am *DefaultAccountManager) SaveUser(accountID, userID string, update *User
...
-		addedGroups := difference(newUser.AutoGroups, oldUser.AutoGroups)
-		for _, g := range removedGroups {
-			group := account.GetGroup(g)
-			if group != nil {
-				am.storeEvent(userID, oldUser.Id, accountID, activity.GroupRemovedFromUser,
-					map[string]any{"group": group.Name, "group_id": group.ID, "is_service_user": newUser.IsServiceUser, "user_name": newUser.ServiceUserName})
-			} else {
-				log.Errorf("group %s not found while saving user activity event of account %s", g, account.Id)
-			}
+		if update.AutoGroups != nil {
```
pascal-fischer, line 573: I think Autogroups should never be nil?? If it can be are we sure nil can not mean empty list of groups? because then the remove part might be needed

management/server/user.go
```diff
@@ -472,56 +502,106 @@ func (am *DefaultAccountManager) SaveUser(accountID, userID string, update *User
...
-					map[string]any{"group": group.Name, "group_id": group.ID, "is_service_user": newUser.IsServiceUser, "user_name": newUser.ServiceUserName})
-			} else {
-				log.Errorf("group %s not found while saving user activity event of account %s", g, account.Id)
+			for _, g := range addedGroups {
+				group := account.GetGroup(g)
+				if group != nil {
+					am.storeEvent(initiatorUserID, oldUser.Id, accountID, activity.GroupAddedToUser,
+						map[string]any{"group": group.Name, "group_id": group.ID, "is_service_user": newUser.IsServiceUser, "user_name": newUser.ServiceUserName})
+				} else {
+					log.Errorf("group %s not found while saving user activity event of account %s", g, account.Id)
```
pascal-fischer: This code will never get reached as we do the check if groups exist already further up. It feels like we do this dance with groups twice, maybe we can join this. If we update the groups before we save the newUser Object we do not need the check in the top of the function

management/server/user_test.go
```diff
@@ -550,3 +541,115 @@ func TestUser_GetUsersFromAccount_ForUser(t *testing.T) {
...
+		t.Fatal(err)
+		return
+	}
+
+	regularUserID := "regularUser"
+
+	tt := []struct {
+		name           string
+		adminUserID    string
+		regularUserID  string
```
pascal-fischer: `adminUserID` and `regularUserId` have no influence on the test so can be removed form test case and hardcoded via constants down below

pascal-fischer (changes requested):

management/server/user.go
braginini: makes sense

management/server/user.go
braginini: makes sense

management/server/user_test.go
braginini: makes sense

pascal-fischer (approved):

PR #860, opened 2023-05-15
Configure CLI flags properly

pascal-fischer:
## Describe your changes
The flags in the cli do not match the functionality of the different commands. We have flags set as global flags even tho they are disregarded completly e.g. 
```
service run -h
runs Netbird as service

Usage:
  netbird service run [flags]

Flags:
  -h, --help   help for run

Global Flags:
      --admin-url string        Admin Panel URL [http|https]://[host]:[port] (default "https://app.netbird.io:443")
  -c, --config string           Netbird config file location (default "/etc/netbird/config.json")
      --daemon-addr string      Daemon service address to serve CLI requests [unix|tcp]://[path|host:port] (default "unix:///var/run/netbird.sock")
  -n, --hostname string         Sets a custom hostname for the device
      --log-file string         sets Netbird log path. If console is specified the the log will be output to stdout (default "/var/log/netbird/client.log")
  -l, --log-level string        sets Netbird log level (default "info")
  -m, --management-url string   Management Service URL [http|https]://[host]:[port] (default "https://api.wiretrustee.com:443")
      --preshared-key string    Sets Wireguard PreSharedKey property. If set, then only peers that have the same key can communicate.
  -k, --setup-key string        Setup key obtained from the Management Service Dashboard (used to register peer)
```

netbird service run only allows logfile and loglevel so proper response would be:
```
service run -h
runs Netbird as service

Usage:
  netbird service run [flags]

Flags:
  -h, --help               help for run
      --log-file string    sets Netbird log path. If console is specified the the log will be output to stdout (default "/var/log/netbird/client.log")
  -l, --log-level string   sets Netbird log level (default "info")
```

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/cmd/down.go
+++ b/client/cmd/down.go
@@ -2,14 +2,20 @@ package cmd
 
 import (
 	"context"
-	"github.com/netbirdio/netbird/util"
 	"time"
+
+	"github.com/netbirdio/netbird/util"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"github.com/netbirdio/netbird/client/proto"
 )
+
+func init() {
+	downCmd.PersistentFlags().StringVarP(&logLevel, "log-level", "l", "info", "sets Netbird log level")
+	downCmd.PersistentFlags().StringVar(&daemonAddr, "daemon-addr", defaultDaemonAddr, "Daemon service address to serve CLI requests [unix|tcp]://[path|host:port]")
+}
 
 var downCmd = &cobra.Command{
 	Use:   "down",
--- a/client/cmd/login.go
+++ b/client/cmd/login.go
@@ -19,6 +19,18 @@ import (
 	"github.com/netbirdio/netbird/client/system"
 )
 
+func init() {
+	loginCmd.PersistentFlags().StringVarP(&configPath, "config", "c", defaultConfigPath, "Netbird config file location")
+	loginCmd.PersistentFlags().StringVar(&adminURL, "admin-url", "", fmt.Sprintf("Admin Panel URL [http|https]://[host]:[port] (default \"%s\")", internal.DefaultAdminURL))
+	loginCmd.PersistentFlags().StringVarP(&managementURL, "management-url", "m", "", fmt.Sprintf("Management Service URL [http|https]://[host]:[port] (default \"%s\")", internal.DefaultManagementURL))
+	loginCmd.PersistentFlags().StringVarP(&logLevel, "log-level", "l", "info", "sets Netbird log level")
+	loginCmd.PersistentFlags().StringVar(&logFile, "log-file", defaultLogFile, "sets Netbird log path. If console is specified the the log will be output to stdout")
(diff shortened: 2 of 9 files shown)


mlsmaycon:
Thanks @pascal-fischer the update will be really great and bring less confusion for the CLI command. I've summarized the flags and which commands they should belong to. In the `Purpose` column, I've added the goals of each flag according to the command it is running with. Can you check if there is any missing peace?

| Flag | Commands | Purpose |
|--------|--------|--------|
| config  | `ssh`,`up`,`login`, `install` and `run` | - `up` and `login` in foreground mode use the flag to read peer information to send to management <br/>- In `ssh` and `run` commands, the flag has the same behavior as the `up` command, but run is always executed the action in foreground<br/>- `install` Sets the `run` command with the flag when configuring  the host's init system |
| log-level | All except `version` | - In foreground mode it defines all agent execution log-level<br/>- In CLI mode (daemon client) it defines the cli's log-level<br/>- The `install` command adds this to the configure `run` command| 
| log-file | All except `version` | - In foreground mode it defines all agent execution log-level<br/>- In CLI mode (daemon client) it defines the cli's log-level<br/>- The `install` command adds this to the configure `run` command<br/>- The `login` command uses this flag with `console` value to run in foreground|
| management-url | `up`,`login`, `install` and `run` | - `up` and `login` in foreground mode use the flag to read peer information to send to management <br/>- `run` same as `up` behavior, but run is always executed the action in foreground<br/>- `install` Sets the `run` command with the flag when configuring  the host's init system |
| admin-url | `up` and `login` | Configure the Admin URL used by the UI client. Useful in RDP or headless configuration.|
| foreground-mode | `up` | Runs the CLI in foreground (needs to be replicated to the login command)|
| daemon-addr |`install` and `run` | - `run` uses this flag to define the listening address of the daemon process<br/>- `install` **Should** set the `run` command with the flag when configuring  the host's init system |
| hostname |`up` and `login` | Sends the desired hostname metadata to the management system |
| preshared-key |`up` and `login` | Sends the desired pre-shared key to the daemon process to be stored in the config file. It requires running `netbird down` before to take effect|
| setup-key | `up` and `login` | Sends the desired hostname metadata to the management system |
| external-ip-map |`up` and `login` | Sends the desired external ip map to the daemon process to be stored in the config file. It requires running `netbird down` before to take effect|
| dns-resolver-address | `up` and `login` | Sends the desired dns resolver listener address to the daemon process to be stored in the config file. It requires running `netbird down` before to take effect|
| port |`ssh` | port used for ssh connection |
| detail |`status` | enables extra detailed information from the status command |
| json | `status` | enables extra detailed information in json format from the status command |
| yaml | `status` | enables extra detailed information in yaml format from the status command |
| ipv4 | `status` | filter output to display only the node's ip |
| filter-by-ips | `status` | filter the status output |
| filter-by-status | `status` |  filter the status output |

pascal-fischer:
For this PR I tried to set the flag as I see usages in code. That does not 100% match with what you wrote. Maybe we need to doublecheck if the code is missing something or if I missed a usage somewhere.

braginini (approved):

PR #868, opened 2023-05-17
Feature/android route notification

pappz:
## Describe your changes

Add new feature to notify the user when new client route has arrived.
Refactor the initial route handling. I move every route logic into the route 
manager package.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/routemanager/notifier.go
```diff
@@ -0,0 +1,86 @@
...
+	n.routeListener = listener
+}
+
+func (n *notifier) setInitialClientRoutes(clientRoutes []*route.Route) {
+	ids := make([]string, 0)
+	for _, r := range clientRoutes {
+		ids = append(ids, r.ID)
+	}
+	sort.Strings(ids)
+	n.initialRouteIDs = ids
```
mlsmaycon: we should use unique network ranges instead of IDs as they are the ones that go with interface creation and will make reconnection required

mlsmaycon (changes requested):

mlsmaycon (approved):

PR #902, opened 2023-05-25
Feat fake dns address

gigovich:
## Describe your changes
Add DNS packet catch hook.
Rename SetFiltering to SetFilter.
Add DNS response writer.
Handle userspace network in fake DNS server.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/client/internal/dns/response_writer.go
+++ b/client/internal/dns/response_writer.go
@@ -7,7 +7,6 @@ import (
 	"github.com/google/gopacket"
 	"github.com/google/gopacket/layers"
 	"github.com/miekg/dns"
-	"golang.zx2c4.com/wireguard/device"
 
 	"github.com/netbirdio/netbird/iface"
 )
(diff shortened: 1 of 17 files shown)


client/internal/dns/response_writer.go
```diff
@@ -0,0 +1,102 @@
...
+		ipv6.SrcIP, ipv6.DstIP = ipv6.DstIP, ipv6.SrcIP
+		ip = ipv6
+	}
+
+	if err := udp.SetNetworkLayerForChecksum(ip.(gopacket.NetworkLayer)); err != nil {
+		return 0, fmt.Errorf("failed to set network layer for checksum: %v", err)
+	}
+
+	// Serialize the packet
+	buffer := gopacket.NewSerializeBuffer()
```
pappz, line 67: If this function is thread safe then do not need create new buffer all the time. It is unnecessary memory allocation.

iface/device_wrapper.go
pappz, line 1: I have an improvement idea. In the Write function I do not think we need to create an additional slice for filter out the deleted []bytes. If think enough to set an empty []byte slice of the original **bufs [][]byte** on the desired position. So with it we can avoid huge amount of memcopy  I suggest it because somewhere else we already apply this trick. Maybe around the STUN pkg filter.

client/internal/dns/response_writer.go
gigovich: We create a response write object for each DSN request (async requests), so we can't use single byte buffer without blocking. We can use a buffers pool but it is not necessary yet.

iface/device_wrapper.go
gigovich: We can do this improvement but not in this PR because it is not related to this feature and need extra testing.

pappz (approved):

mlsmaycon (approved):
LGTM

PR #906, opened 2023-05-26
Add retry to sending signal message

mlsmaycon:
## Describe your changes

Increased the default send timeout from 2 to 5

Added a max of 4 retries  with an increased timeout after the second attempt

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/signal/client/grpc.go
+++ b/signal/client/grpc.go
@@ -24,6 +24,8 @@ import (
 	"github.com/netbirdio/netbird/signal/proto"
 )
 
+const defaultSendTimeout = 5 * time.Second
+
 // ConnStateNotifier is a wrapper interface of the status recorder
 type ConnStateNotifier interface {
 	MarkSignalDisconnected()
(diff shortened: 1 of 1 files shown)


signal/client/grpc.go
```diff
@@ -322,9 +324,23 @@ func (c *GrpcClient) Send(msg *proto.Message) error {
...
+			attemptTimeout = time.Duration(attempt) * defaultSendTimeout * time.Second
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), attemptTimeout)
+
+		_, err = c.realClient.Send(ctx, encryptedMessage)
+
+		cancel()
+
+		if err == nil {
+			break
```
braginini: I would just return here

braginini (dismissed):

pascal-fischer (approved):

signal/client/grpc.go
```diff
@@ -322,14 +324,29 @@ func (c *GrpcClient) Send(msg *proto.Message) error {
...
-		return err
+	attemptTimeout := defaultSendTimeout
+
+	for attempt := 0; attempt < 4; attempt++ {
+		select {
+		case <-c.ctx.Done():
+			return fmt.Errorf("not running message attempt. client context is done: %s", c.ctx.Err())
+		default:
+		}
+		if attempt > 1 {
```
pappz, line 330: if you start the iteration from 1 instead of 0 then this if statement no need to.

mlsmaycon: we decided to have the following attempts timeout:
1st and 2nd with the default value 5s
3rd with 10s
4th with 15s

PR #915, opened 2023-05-30
Extend protocol and firewall manager to handle old management

gigovich:
## Describe your changes
With old management, the new clients always will receive an empty list of firewall rules, and by default, we block all traffic. To avoid this, let’s modify the protocol and add `FirewallRulesIsEmpty` field (same as `RemotePeersIsEmpty`). When this field is not set and rules are empty, we work with the old management version, and we need to add special rules to input and output chains to allow all traffic.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -635,7 +635,7 @@ func (e *Engine) updateNetworkMap(networkMap *mgmProto.NetworkMap) error {
 		// we have old version of management without rules handling, we should allow all traffic
 		allowByDefault := len(networkMap.FirewallRules) == 0 && !networkMap.FirewallRulesIsEmpty
 		if allowByDefault {
-			log.Warn("received a firewall update from an older NetBird management version, allowing all traffic from the NetBird network")
+			log.Warn("this peer is connected to a NetBird Management service with an older version. Allowing all traffic from connected peers")
 		}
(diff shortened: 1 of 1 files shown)


client/internal/engine.go
```diff
@@ -631,7 +631,10 @@ func (e *Engine) updateNetworkMap(networkMap *mgmProto.NetworkMap) error {
 	}
 
 	if e.acl != nil {
-		e.acl.ApplyFiltering(networkMap.FirewallRules)
+		// if we got empty rules list but management not set networkMap.FirewallRulesIsEmpty flag
+		// we have old version of management without rules handling, we should allow all traffic
+		allowByDefault := len(networkMap.FirewallRules) == 0 && !networkMap.FirewallRulesIsEmpty
```
mlsmaycon, line 636: Let's add the following warning log:
"received a firewall update from an older NetBird management version, allowing all traffic from the NetBird network."

mlsmaycon (changes requested):

client/internal/engine.go
gigovich: Done

mlsmaycon (approved):

mlsmaycon (approved):

PR #941, opened 2023-06-07
Decouple docs generation

pascal-fischer:
## Describe your changes
Adding seperate github actions workflow to decouple docs update from general release flow to only run if openapi doc is updated on release.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -1,0 +1,24 @@
+name: update docs
+
+on:
+  push:
+    tags:
+      - 'v*'
+    branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,24 @@
+name: update docs
+
+on:
+  push:
+    tags:
+      - 'v*'
+    branches:
```
mlsmaycon: we don't want it to run on pushes to the main branch

PR #967, opened 2023-06-19
Feature/permanent dns

pappz:
## Describe your changes

Because on Android system hard to update DNS settings the system will set our DNS service as permanent DNS server on mobile system.
This PR modify the original logic. Now the DNS service can functional in two different way
1. the DNS service will start listen when get update from mgm server
2. the DNS service start listening immediately when it instantiated

In the second case the upstream DNS server will be the DNS servers from the host machine.
The relevant Android code: https://github.com/netbirdio/react-native-app/pull/24

### Android
The Android code track the network changes and update the DNS configuration when it has been changed on the host machine. For example when switch from 4G to WIFI.

### Test
Because this change refactor the whole logic necessary to test all combination of the supported systems
- Mac, Win, Linux, Android
- On Linux with WG bind interface and with kernel module support
- On Linux with systemd, resolve.conf

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/android/dns_list.go
+++ b/client/android/dns_list.go
@@ -1,0 +1,21 @@
+package android
+
+// DNSList is a wrapper of []string
+type DNSList struct {
+	items []string
+}
+
+// Add new DNS address to the collection
+func (array *DNSList) Add(s string) {
+	array.items = append(array.items, s)
+}
+
+// Get return an element of the collection
+func (array *DNSList) Get(i int) string {
+	return array.items[i]
+}
+
+// Size return with the size of the collection
+func (array *DNSList) Size() int {
+	return len(array.items)
+}
--- a/client/android/dns_list.go
+++ b/client/android/dns_list.go
@@ -1,5 +1,7 @@
 package android
 
+import "fmt"
+
 // DNSList is a wrapper of []string
 type DNSList struct {
 	items []string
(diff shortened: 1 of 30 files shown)


client/internal/dns/server_test.go
```diff
@@ -478,7 +522,8 @@ func TestDNSServerStartStop(t *testing.T) {
 func TestDNSServerUpstreamDeactivateCallback(t *testing.T) {
 	hostManager := &mockHostConfigurator{}
 	server := DefaultServer{
-		dnsMux: dns.DefaultServeMux,
+		//dnsMux: dns.DefaultServeMux,
```
pascal-fischer: some comment left

pappz: Hmm, interesting. I can not remember well why I did it. I will double check this test.

pascal-fischer (approved):

client/android/dns_list.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+// Add new DNS address to the collection
+func (array *DNSList) Add(s string) {
+	array.items = append(array.items, s)
+}
+
+// Get return an element of the collection
+func (array *DNSList) Get(i int) string {
+	return array.items[i]
```
gigovich: We can get panic here, better to check the range of i.

gigovich (approved):
LGTM

client/android/dns_list.go
pappz: Thank you! I pushed the fix and I also updated the Android code

mlsmaycon (approved):

mlsmaycon (approved):

PR #970, opened 2023-06-19
ACL firewall manager fix/improvement

gigovich:
## Describe your changes
Fix the issue with rule squashing, it contained an issue when calculating the total amount of IPs in the Peer map (doesn't include offline peers). That is why squashing did not work.
Also, this commit changes the rules apply behavior. Instead policy:
1. Apply all rules from the network map
2. Remove all previously applied rules.
We do:
1. Apply only new rules
2. Remove outdated rules

Why the first variant was implemented: when you have a drop policy it is important in which order your rules are and you need a totally clean previous state to apply the new. But in the release, we didn't include a drop policy so we can do this improvement.

### Checklist
- [x] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/client/firewall/port.go
+++ b/client/firewall/port.go
@@ -1,5 +1,9 @@
 package firewall
 
+import (
+	"strconv"
+)
+
 // Protocol is the protocol of the port
 type Protocol string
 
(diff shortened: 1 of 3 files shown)


mlsmaycon (approved):

mlsmaycon:
Testes on systems with iptables and nftables showed a great increase in performance.

PR #972, opened 2023-06-20
Don't create setup keys on new account

braginini:
## Describe your changes

Since we enabled one-time setup key display,
there is no need to generate keys by default.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -1373,33 +1373,27 @@ func addAllGroup(account *Account) error {
 }
 
 // newAccountWithId creates a new Account with a default SetupKey (doesn't store in a Store) and provided id
-func newAccountWithId(accountId, userId, domain string) *Account {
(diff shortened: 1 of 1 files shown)


management/server/account.go
```diff
@@ -1373,33 +1373,27 @@ func addAllGroup(account *Account) error {
...
 		DisabledManagementGroups: make([]string, 0),
 	}
-	log.Debugf("created new account %s with setup key %s", accountId, defaultKey.Key)
+	log.Debugf("created new account %s", accountID)
 
 	acc := &Account{
-		Id:               accountId,
-		SetupKeys:        setupKeys,
+		Id:               accountID,
+		SetupKeys:        map[string]*SetupKey{},
```
pascal-fischer: would create the map together with the rest up in L1383

pascal-fischer (approved):

PR #978, opened 2023-06-21
Fix disabled DNS resolver fail

gigovich:
## Describe your changes
When DNS server runs in a fake resolver, there is an issue with incorrect handling disable state because we still receive updates from the management. Add extra checking and fix panic. 
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/dns/server.go
+++ b/client/internal/dns/server.go
@@ -105,6 +105,10 @@ func NewDefaultServer(ctx context.Context, wgInterface *iface.WGIface, customAdd
 		defaultServer.enabled = hasValidDnsServer(initialDnsCfg)
 	}
 
+	if wgInterface.IsUserspaceBind() {
+		defaultServer.evelRuntimeAddressForUserspace()
+	}
+
 	return defaultServer, nil
 }
 
@@ -117,7 +121,9 @@ func (s *DefaultServer) Initialize() (err error) {
 		return nil
 	}
 
-	s.evalRuntimeAddress()
+	if !s.wgInterface.IsUserspaceBind() {
+		s.evalRuntimeAddress()
+	}
 	s.hostManager, err = newHostManager(s.wgInterface)
 	return
 }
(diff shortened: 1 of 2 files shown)


client/internal/dns/server_test.go
```diff
@@ -276,6 +279,96 @@ func TestUpdateDNSServer(t *testing.T) {
 	}
 }
 
+func TestDNSFakeResolverHandleUpdates(t *testing.T) {
+	ov := os.Getenv("NB_WG_KERNEL_DISABLED")
```
pappz, line 282: I am not sure this solution is good pattern. What will be happen if the system run test in parallel? 
On an another branch I solved it with a moc interface. https://github.com/netbirdio/netbird/pull/967/commits/d222c3b3ba84483f0b95e310ed6e21bf6df42f47

pappz (changes requested):

client/internal/dns/server_test.go
gigovich: Sure, it's better to use mocks here, but let's do it after we merge your branch.

client/internal/dns/server.go
```diff
@@ -127,12 +127,13 @@ func (s *DefaultServer) listen() {
...
+		s.setListenerStatus(true)
 		go func() {
-			s.setListenerStatus(true)
-			defer s.setListenerStatus(false)
-
 			hookID := s.filterDNSTraffic()
 			s.fakeResolverWG.Wait()
+
+			s.mux.Lock()
+			defer s.mux.Unlock()
```
pappz: I feel it is not fully thread safe.
With this lock you protect the **listenerIsRunning** variable but you modify that variable from a different thread.
So  you call the s.fakeResolverWG.Done() in 280 line and right after modify the **listenerIsRunning** variable but it does not means that in this thread the s.mux.Lock() will be invoked before you modify.

pappz (changes requested):

client/internal/dns/server.go
gigovich: No, this lock is used to protect RemovePacketHook, but anyway, I'm working to rewrite this part totally and drop this waitgroup at all

pappz: It good to know. On my another task I used it in wrong way I think.

pappz: In this case I recommend you to use channel withe a lock instead of wg.  With it you can avoid the weakness (negative write) of wg. The code style will be ugly but by the way :)

gigovich: When I get a negative number, it means we have something wrong with calls, and it helps to catch bugs.
Anyway - the workgroup was removed :)

pappz: You definitely found the greatest solution! I like it :)

client/internal/dns/server.go
```diff
@@ -201,17 +198,24 @@ func (s *DefaultServer) Stop() {
...
 func (s *DefaultServer) stopListener() error {
+	if s.wgInterface != nil && s.wgInterface.IsUserspaceBind() && s.listenerIsRunning {
+		// udpFilterHookID here empty only in the unit tests
+		if filter := s.wgInterface.GetFilter(); filter != nil && s.udpFilterHookID != "" {
+			if err := filter.RemovePacketHook(s.udpFilterHookID); err != nil {
+				log.Errorf("unable to remove DNS packet hook: %s", err)
+			}
+		}
+		s.udpFilterHookID = ""
+		s.listenerIsRunning = false
```
pappz, line 220: Here we can return. Do not need to check again the s.listener.IsRunning

pappz (changes requested):

client/internal/dns/server.go
gigovich: Fixed

pappz (approved):

PR #989, opened 2023-06-27
Add API Endpoint for Resending User Invitations in Auth0

bcmmbaga:
## Describe your changes

Create a new API endpoint `POST /api/users/{userId}/invite` to  resend user invites to users that didn’t activate their accounts before the expiration time

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/http/api/openapi.yml
+++ b/management/server/http/api/openapi.yml
@@ -169,12 +169,12 @@ components:
     UserInviteRequest:
       type: object
       properties:
-        email:
-          description: User's Email to send invite to
+        user_id:
+          description: Users ID send invite to
           type: string
-          example: demo@netbird.io
+          example: google-oauth2|277474792786460067937
       required:
-        - email
+        - user_id
     PeerMinimum:
       type: object
       properties:
@@ -1287,7 +1287,7 @@ paths:
             type: string
           description: The unique identifier of a user
       requestBody:
-        description: PersonalAccessToken create parameters
+        description: User invite information
         content:
           application/json:
             schema:
--- a/management/server/http/api/openapi.yml
+++ b/management/server/http/api/openapi.yml
@@ -166,15 +166,6 @@ components:
         - role
         - auto_groups
         - is_service_user
-    UserInviteRequest:
-      type: object
-      properties:
-        user_id:
-          description: Users ID to send invite to
-          type: string
-          example: google-oauth2|277474792786460067937
-      required:
-        - user_id
     PeerMinimum:
       type: object
       properties:
@@ -1286,12 +1277,6 @@ paths:
           schema:
             type: string
           description: The unique identifier of a user
-      requestBody:
-        description: User invite information
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/UserInviteRequest'
       responses:
         '200':
           description: Invite status code
--- a/management/server/http/api/openapi.yml
+++ b/management/server/http/api/openapi.yml
@@ -170,7 +170,7 @@ components:
       type: object
       properties:
         user_id:
-          description: Users ID send invite to
+          description: Users ID to send invite to
           type: string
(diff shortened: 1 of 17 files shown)


management/server/http/api/openapi.yml
```diff
@@ -1262,6 +1271,39 @@ paths:
...
+        - TokenAuth: [ ]
+      parameters:
+        - in: path
+          name: userId
+          required: true
+          schema:
+            type: string
+          description: The unique identifier of a user
+      requestBody:
+        description: PersonalAccessToken create parameters
```
pascal-fischer: description needs to be updated

management/server/http/handler.go
```diff
@@ -113,6 +113,7 @@ func (apiHandler *apiHandler) addUsersEndpoint() {
 	apiHandler.Router.HandleFunc("/users/{userId}", userHandler.UpdateUser).Methods("PUT", "OPTIONS")
 	apiHandler.Router.HandleFunc("/users/{userId}", userHandler.DeleteUser).Methods("DELETE", "OPTIONS")
 	apiHandler.Router.HandleFunc("/users", userHandler.CreateUser).Methods("POST", "OPTIONS")
+	apiHandler.Router.HandleFunc("/users/{userId}/invite", userHandler.InviteUser).Methods("POST", "OPTIONS")
```
pascal-fischer, line 116: Why do we need userID in the path? as far as I can see it is not used so /users/invite should be fine

management/server/http/users_handler_test.go
```diff
@@ -340,6 +351,69 @@ func TestCreateUser(t *testing.T) {
...
+	}{
+		{
+			name:           "Invite User",
+			requestType:    http.MethodPost,
+			requestPath:    "/api/users/" + existingUserID + "/invite",
+			expectedStatus: http.StatusOK,
+			requestBody:    bytes.NewBuffer(userToInviteReqPayload),
+		},
+		{
+			name:           "Invite User with Existing User",
```
pascal-fischer: The test case names should represent/describe whats tested within this case so would be found user email and not found user email

pascal-fischer (changes requested):

pascal-fischer:
General question for my understanding: The idea is to re-send an invite to an already invited person if the invite is already expired. As we don't  have a trigger, this is supposed to be a manual operation? probably re-send invite button in UI that the admin has to actively click right?

management/server/http/handler.go
pascal-fischer: We just had a talk and it might make more sense to drop email and go with userID directly?

bcmmbaga: > Why do we need userID in the path? as far as I can see it is not used so /users/invite should be fine

The userID is used in setting initiator ID when storing the event

management/server/http/users_handler.go
```diff
@@ -208,6 +208,42 @@ func (h *UsersHandler) GetAllUsers(w http.ResponseWriter, r *http.Request) {
...
+		util.WriteErrorResponse("couldn't parse JSON request", http.StatusBadRequest, w)
+		return
+	}
+
+	if req.UserId == "" {
+		util.WriteErrorResponse("request body is missing the required field user_id", http.StatusBadRequest, w)
+		return
+	}
+
+	err = h.accountManager.InviteUser(account.Id, user.Id, req.UserId)
```
pascal-fischer: We do not need the request body at all. We have the userID as a path variable in /users/{userId}/invite. You can access path variables like this 
```
vars := mux.Vars(r)
targetUserID := vars["userId"]
if len(targetUserID) == 0 {
	util.WriteError(status.Errorf(status.InvalidArgument, "invalid user ID"), w)
	return
}
```

or instead of Len do map access check

pascal-fischer (approved):

PR #991, opened 2023-06-28
Improve userspace filter performance

pascal-fischer:
## Describe your changes
Switched rule set from array to map (as lookup table grouped by IP) to reduce number of rules checked for each packet when filtering) this way increasing throughput.

To test:
- throughput when thousands of rules active
- startup time when thousands of rules active
- memory consumption between old and new impl.
- functionality

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/client/firewall/uspfilter/uspfilter.go
+++ b/client/firewall/uspfilter/uspfilter.go
@@ -23,8 +23,8 @@ type IFaceMapper interface {
 
 // Manager userspace firewall manager
 type Manager struct {
-	outgoingRules []Rule
-	incomingRules []Rule
+	outgoingRules map[string][]Rule
+	incomingRules map[string][]Rule
 	rulesIndex    map[string]int
 	wgNetwork     *net.IPNet
 	decoders      sync.Pool
@@ -62,6 +62,8 @@ func Create(iface IFaceMapper) (*Manager, error) {
 				return d
 			},
 		},
+		outgoingRules: make(map[string][]Rule),
+		incomingRules: make(map[string][]Rule),
 	}
 
 	if err := iface.SetFilter(m); err != nil {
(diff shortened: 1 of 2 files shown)


client/firewall/uspfilter/uspfilter.go
```diff
@@ -21,11 +21,13 @@ type IFaceMapper interface {
 	SetFilter(iface.PacketFilter) error
 }
 
+// RuleSet is a set of rules grouped by a string key
+type RuleSet map[string]Rule
```
gigovich, line 25: Unfortunately we can't use map, it should be list, because in go no key order saved: https://github.com/golang/go/issues/54500
This means, if we will support DROP policy in the future, order of processing rules will be important.
So need @mlsmaycon approve of this change

client/firewall/uspfilter/uspfilter.go
```diff
@@ -224,33 +224,33 @@ func (m *Manager) dropFilter(packetData []byte, rules []Rule, isIncomingPacket b
...
+	case layers.LayerTypeIPv6:
+		if isIncomingPacket {
+			ip = d.ip6.SrcIP
+		} else {
+			ip = d.ip6.DstIP
+		}
+	}
+
+	if len(rules["0.0.0.0"]) > 0 || len(rules["::"]) > 0 {
+		return false
```
gigovich: I'm not sure that it is correct. Because we decide to not drop packet just based on the IP address and ignore protocol and ports.

gigovich (changes requested):

gigovich:
@pascal-fischer it also fails during testing.

Script to reproduce on any account with several peers:

1. Disable default rule
2. Add rule with TCP port
3. Run agent agent in the docker container with userspace.
4. In same container exec bash and try to do any request to internet - it fails.
5. Do same test with binary built from master - it works.

client/firewall/uspfilter/uspfilter.go
pascal-fischer: Yeah on that part I was not 100% sure

client/firewall/uspfilter/uspfilter.go
pascal-fischer: So the order of rules is important for us? Would it make sense to extend the object with a priority field? I could also revert the map of ids and only leave it for IPs to keep the performance improvement so I have map of arrays. For the IPs the order should not matter as they re not related

gigovich (approved):
LGTM

About, rules order, for now we decided to ignore until we allow to use DROP verdict.

PR #1007, opened 2023-07-10
Remove unused netbird UI dependencies

bcmmbaga:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/.goreleaser_ui.yaml
+++ b/.goreleaser_ui.yaml
@@ -74,6 +74,8 @@ nfpms:
       - src: client/ui/disconnected.png
         dst: /usr/share/pixmaps/netbird.png
     dependencies:
+      - libappindicator-gtk3
+      - libappindicator
       - netbird
 
 uploads:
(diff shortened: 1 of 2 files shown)


.goreleaser_ui.yaml
```diff
@@ -11,6 +11,8 @@ builds:
       - amd64
     ldflags:
       - -s -w -X github.com/netbirdio/netbird/version.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.CommitDate}} -X main.builtBy=goreleaser
+    tags:
+      - legacy_appindicator
```
mlsmaycon, line 15: Building the app with this tag will have any effect on the Debian packages?

bcmmbaga: So far, there have been no observed effects. I did tests on 2 deb distros:
- Ubuntu:  works without installing any dependecies
- fedora: works but I had to install [appindicator](https://extensions.gnome.org/extension/615/appindicator-support/) to work properly

mlsmaycon: ok, then let's remove the app indicator dependency from both, RPM and DEB packages and update our docs, and dashboard, ok?

bcmmbaga: for RPM-based distro, the `libappindicator-gtk3` and `libappindicator` packages are still required. Should we include step to install these in docs?

mlsmaycon: yes, lets include the additional package as part of the doc and not a requirement, because the installation might fail if there is one dependency missing in a particular OS repository.

mlsmaycon (approved):

PR #1036, opened 2023-07-28
Keep confiured nameservers as fallback

pascal-fischer:
## Describe your changes
Sometimes if the client is stopped forcefully the old DNS configuration was not restored. This was casuing nameresolution to fail because the netbird resolver was down. At the same time the netbird resolver could not start because it could not connect to management due to failing name resolution.

In this PR the netbird resolver was chnaged to keep the previous configured nameserver as fallback in case the netbird resolver is not resopnding.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


gigovich (approved):
LGTM

PR #1042, opened 2023-08-02
Update setup key auto_groups description

mlsmaycon:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/http/api/openapi.yml
+++ b/management/server/http/api/openapi.yml
@@ -327,7 +327,7 @@ components:
           type: string
           example: valid
         auto_groups:
-          description: List of group IPs to auto-assign to peers registered with this key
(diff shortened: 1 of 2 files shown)


management/server/http/api/openapi.yml
```diff
@@ -375,7 +375,7 @@ components:
           type: boolean
           example: false
         auto_groups:
-          description: Setup key groups to auto-assign to peers registered with this key
+          description: List of group IPs to auto-assign to peers registered with this key
```
bcmmbaga: Considering clarity, would it be better to rename the` group IPs` to `group names`?

mlsmaycon: it should be group IDs, I will fix that

bcmmbaga (approved):

PR #1057, opened 2023-08-07
Fix/ebpf free

pappz:
## Describe your changes

Fix eBPF free call and add debug logs.
Because of a typo we never detached the loaded eBPF program. 
It can cause problems if we start to reuse the proxy factory.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/wgproxy/proxy_userspace.go
+++ b/client/internal/wgproxy/proxy_userspace.go
@@ -20,6 +20,7 @@ type WGUserSpaceProxy struct {
 
 // NewWGUserSpaceProxy instantiate a user space WireGuard proxy
 func NewWGUserSpaceProxy(wgPort int) *WGUserSpaceProxy {
+	log.Debugf("instantiate new userspace proxy")
 	p := &WGUserSpaceProxy{
 		localWGListenPort: wgPort,
 	}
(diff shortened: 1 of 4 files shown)


client/internal/wgproxy/proxy_userspace.go
```diff
@@ -20,6 +20,7 @@ type WGUserSpaceProxy struct {
 
 // NewWGUserSpaceProxy instantiate a user space WireGuard proxy
 func NewWGUserSpaceProxy(wgPort int) *WGUserSpaceProxy {
+	log.Debugf("instantiate new userspace proxy")
```
mlsmaycon, line 23: we should close the conn when calling Free for userspace proxy too?

pappz: You are right but not exactly. Behind the eBPF proxy and the user space proxy has different life circle logic. In case of eBPF we instantiate it in a very early state but of the user space proxy does not have a global state. We instantiate a new proxy when we create a peer conn. And we free it up when we close the peer conn.
So basically for the Free() function make no sense in case of user space proxy and CloseConn() function make no sense in case of eBPF proxy. I left comments in the code at the relevant places.
```
// Free doing nothing because this implementation of proxy does not have global state
```

mlsmaycon: Got it, thanks for clarifying.

mlsmaycon (approved):

PR #1030, opened 2023-07-25
Implement SQLite Store using gorm and lookup table

surik:
## Describe your changes

This pull request is currently in draft mode, and my aim is to experimentally implement a Store interface using SQLite and Gorm. For this implementation, I have chosen to follow a document-oriented approach instead of utilizing the full power of relational databases. The primary objective of this pull request is to showcase my approach and draw attention to any potential weak points. Additionally, I want to test the existing Store Interface with this implementation and identify areas where improvements can be made.

Goals:

1. Test the Store interface using SQLite and Gorm to gauge its effectiveness and identify strengths and weaknesses.
1. Evaluate whether certain methods, such as GetAllAccount, can be removed to avoid loading the entire database into memory, considering it is only used in two places.
1. Address concurrent modification concerns by exploring the use of locks, specifically focusing on Global and Account locks. I want to assess whether using transactions can help alleviate potential concurrency issues.
Benchmark Results:

```
$ go test  ./management/server -run Benchmark_ -bench=.  -benchmem
goos: darwin
goarch: amd64
pkg: github.com/netbirdio/netbird/management/server
cpu: Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
BenchmarkTest_StoreWrite/FileStore_Write_100-8               163          13118187 ns/op         5638586 B/op       8789 allocs/op
BenchmarkTest_StoreWrite/SqliteStore_Write_100-8             199           5097194 ns/op           47591 B/op        773 allocs/op
BenchmarkTest_StoreWrite/FileStore_Write_500-8                50          23935032 ns/op        12289436 B/op      15859 allocs/op
BenchmarkTest_StoreWrite/SqliteStore_Write_500-8             100          11405159 ns/op           47724 B/op        777 allocs/op
BenchmarkTest_StoreWrite/FileStore_Write_1000-8               27          41880172 ns/op        25661102 B/op      30031 allocs/op
BenchmarkTest_StoreWrite/SqliteStore_Write_1000-8            100          17335683 ns/op           47777 B/op        777 allocs/op
BenchmarkTest_StoreWrite/FileStore_Write_2000-8               13          89936823 ns/op        45821132 B/op      58825 allocs/op
BenchmarkTest_StoreWrite/SqliteStore_Write_2000-8             55          30195228 ns/op           47928 B/op        782 allocs/op
BenchmarkTest_StoreRead/FileStore_Read_100-8              437596              2821 ns/op            2856 B/op         26 allocs/op
[...]

PR #1065, opened 2023-08-08
Implement SQLite Store using gorm and relational approach

surik:
## Describe your changes

This approach involves loading associated tables into slices and then converting them into maps. This technique is introducing additional fields prefixed with 'G' (shows Gorm relation), allowing for granular resource management without many changes to existing structures. While this approach introduces complexities, it serves the goal of achieving a functional implementation while minimizing disruptions to the current architecture.

One significant advantage is the elimination of the complexity associated with the lookup table used in the document-oriented approach. This simplification not only enhances query performance but also contributes to improved write speed, especially as the number of accounts grows.

Another key advantage introduced by the new relational approach is the ability to operate on separated data structures, such as accounts, users, and peers, independently. This level of separation empowers more granular and focused updates, ensuring that modifications to one resource do not require loading the entire account. But this may require additional changes in the Store interface and the Account Manager.

#### Benchmark Results:

```
$ go test  ./management/server -run Benchmark_ -bench=.  -benchmem
goos: darwin
goarch: amd64
pkg: github.com/netbirdio/netbird/management/server
cpu: Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
BenchmarkTest_StoreWrite/FileStore_Write_100-8         	     146	  14532698 ns/op	 6130954 B/op	    8549 allocs/op
BenchmarkTest_StoreWrite/SqliteStore_Write_100-8       	     278	   4623969 ns/op	  210247 B/op	    2714 allocs/op
BenchmarkTest_StoreWrite/FileStore_Write_500-8         	      44	  25448032 ns/op	13313935 B/op	   15779 allocs/op
BenchmarkTest_StoreWrite/SqliteStore_Write_500-8       	     242	   5269740 ns/op	  210193 B/op	    2714 allocs/op
BenchmarkTest_StoreWrite/FileStore_Write_1000-8        	      25	  45887216 ns/op	26595502 B/op	   30004 allocs/op
BenchmarkTest_StoreWrite/SqliteStore_Write_1000-8      	     205	   5795530 ns/op	  210193 B/op	    2715 allocs/op
BenchmarkTest_StoreWrite/FileStore_Write_2000-8        	      12	 100217523 ns/op	50800290 B/op	   58819 allocs/op
BenchmarkTest_StoreWrite/SqliteStore_Write_2000-8      	     178	   6024594 ns/op	  210188 B/op	    2715 allocs/op
BenchmarkTest_StoreRead/FileStore_Read_100-8           	  434073	      2602 ns/op	    3192 B/op	      27 allocs/op
BenchmarkTest_StoreRead/SqliteStore_Read_100-8         	    1483	    694025 ns/op	  118992 B/op	    2113 allocs/op
BenchmarkTest_StoreRead/FileStore_Read_500-8           	  412197	      2588 ns/op	    3192 B/op	      27 allocs/op
BenchmarkTest_StoreRead/SqliteStore_Read_500-8         	    1731	    676320 ns/op	  118894 B/op	    2113 allocs/op
BenchmarkTest_StoreRead/FileStore_Read_1000-8          	  439562	      2681 ns/op	    3192 B/op	      27 allocs/op
BenchmarkTest_StoreRead/SqliteStore_Read_1000-8        	    1508	    708517 ns/op	  118790 B/op	    2113 allocs/op
```

We can see that the write speed of the relational approach is better compared with the result shown in #1030. 
Unfortunately, the reading speed is worse but this might be due to the amount of copying I have to do when loading an account. This can be improved.

#### ERD diagram:

image

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/sqlite_store.go
+++ b/management/server/sqlite_store.go
@@ -287,19 +287,21 @@ func (s *SqliteStore) GetAccount(accountID string) (*Account, error) {
 	var account Account
 
 	result := s.db.Model(&account).
-		Preload("SetupKeysG").
-		Preload("PeersG").
-		Preload("UsersG").
-		Preload("UsersG.PATsG").
-		Preload("GroupsG").
-		Preload("RulesG").
-		Preload("Policies").
-		Preload("Policies.Rules").
-		Preload("RoutesG").
-		Preload("NameServerGroupsG").
+		Preload("UsersG.PATsG"). // have to be specifies as this is nester reference
+		Preload(clause.Associations).
 		First(&account, "aid = ?", accountID)
 	if result.Error != nil {
 		return nil, status.Errorf(status.NotFound, "account not found")
+	}
+
+	// we have to manually preload policy rules as it seems that gorm preloading doesn't do it for us
+	for i, policy := range account.Policies {
+		var rules []*PolicyRule
+		err := s.db.Model(&PolicyRule{}).Find(&rules, "policy_id = ?", policy.ID).Error
+		if err != nil {
+			return nil, status.Errorf(status.NotFound, "account not found")
+		}
+		account.Policies[i].Rules = rules
 	}
 
 	account.SetupKeys = make(map[string]*SetupKey, len(account.SetupKeysG))
--- a/management/server/sqlite_store.go
+++ b/management/server/sqlite_store.go
@@ -99,8 +99,7 @@ func (s *SqliteStore) AcquireAccountLock(accountID string) (unlock func()) {
 }
 
 func (s *SqliteStore) SaveAccount(account *Account) error {
-	for id, key := range account.SetupKeys {
-		key.Id = id
+	for _, key := range account.SetupKeys {
 		account.SetupKeysG = append(account.SetupKeysG, *key)
 	}
 
@@ -139,7 +138,17 @@ func (s *SqliteStore) SaveAccount(account *Account) error {
 	}
 
 	err := s.db.Transaction(func(tx *gorm.DB) error {
-		result := tx.Select(clause.Associations).Delete(account)
+		result := tx.Select(clause.Associations).Delete(account.Policies, "account_id = ?", account.Id)
+		if result.Error != nil {
+			return result.Error
+		}
+
+		result = tx.Select(clause.Associations).Delete(account.UsersG, "account_id = ?", account.Id)
+		if result.Error != nil {
+			return result.Error
+		}
+
+		result = tx.Select(clause.Associations).Delete(account)
 		if result.Error != nil {
 			return result.Error
 		}
@@ -210,7 +219,7 @@ func (s *SqliteStore) GetAccountByPrivateDomain(domain string) (*Account, error)
 
 func (s *SqliteStore) GetAccountBySetupKey(setupKey string) (*Account, error) {
 	var key SetupKey
-	result := s.db.Select("account_id").First(&key, "key = ?", setupKey)
+	result := s.db.Select("account_id").First(&key, "key = ?", strings.ToUpper(setupKey))
 	if result.Error != nil {
 		return nil, status.Errorf(status.NotFound, "account not found: index lookup failed")
 	}
@@ -265,7 +274,10 @@ func (s *SqliteStore) GetAllAccounts() (all []*Account) {
 	}
 
 	for _, account := range accounts {
-		all = append(all, account.Copy()) // TODO: copy and delete gorm models
+		//all = append(all, account.Copy()) // TODO: copy and delete gorm models
+		if acc, err := s.GetAccount(account.Id); err == nil {
+			all = append(all, acc)
+		}
 	}
 
 	return all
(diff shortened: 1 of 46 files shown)


surik:
I managed to make all tests pass with SQLite implementation. Additionally, the testing matrix is extended to cover both file and SQLite store scenarios. The only exception is Windows as I keep getting:

```
testing.go:1206: TempDir RemoveAll cleanup: remove C:\Windows\TEMP\TestSqlite_GetTokenIDByHashedToken2617861885\001\store.db: The process cannot access the file because it is being used by another process.
```

This needs further investigation about the cleanup procedure. For now, testing for Windows only happens with the file store.

surik:
I managed to make all tests pass with SQLite implementation. The testing matrix is extended to cover both file and SQLite store scenarios. The only exception is Windows as I keep getting:

```
testing.go:1206: TempDir RemoveAll cleanup: remove C:\Windows\TEMP\TestSqlite_GetTokenIDByHashedToken2617861885\001\store.db: The process cannot access the file because it is being used by another process.
```

This seems related to the cleanup procedure and needs further investigation. For now, SQLite store tests are disabled for Windows.

I also added a sub-command that can help with the migration of json storage to the SQLite store:

```
./management migration --help
Migrate JSON file store to SQLite store. Please make a backup of the store json file before running this command.

This command reads the content of {datadir}/store.json and migrates it to {datadir}/store.db that can be used by SQLite store driver.

Usage:
  netbird-mgmt migration [--datadir directory] [--log-file console] [flags]
```

By default, management uses JSON file store as before. To use SQLite there is an option in the config file `StoreKind` which can be set either to `Sqlite` or `JsonFile`.

surik:
I managed to fix Sqlite store tests on Windows by not using `cache=shared`. Also, made a few manual tests locally and could not find any issues. Saying that this one is ready for review.

management/server/sqlite_store.go
```diff
@@ -0,0 +1,423 @@
...
+		&installation{},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &SqliteStore{db: db, storeFile: file, installationPK: 1}, nil
+}
+
+// NewSqliteStoreFromFileStor restores a store from FileStore and stores SQLite DB in the file located in datadir
```
pappz: Typo mistake in the "NewSqliteStoreFromFileStor".

pappz (changes requested):

management/server/sqlite_store.go
```diff
@@ -0,0 +1,423 @@
...
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &SqliteStore{db: db, storeFile: file, installationPK: 1}, nil
+}
+
+// NewSqliteStoreFromFileStore restores a store from FileStore and stores SQLite DB in the file located in datadir
+func NewSqliteStoreFromFileStore(filestore *FileStore, dataDir string, metrics telemetry.AppMetrics) (*SqliteStore, error) {
```
pappz, line 73: Looks like that the migration of the "installation id" is missing.

pappz (changes requested):

pappz (changes requested):
Because the global lock doing nothing then [this](https://github.com/netbirdio/netbird/pull/1065/files#diff-998d2ca8f48ecbf2d1a259f3851048dacc8f8ae1dfc13c0e69a34e74eaae817aR734) line may cause problem. Here we write out the account but without account lock.

surik:
The global lock was implemented for SQLite. I managed to eliminate the usage of one global lock-in [here](https://github.com/netbirdio/netbird/pull/1178). We may consider addressing the rest later.

Additionally, I implemented a rollback command in case someone wants to migrate back from SQLite to the JSON store.

Also, I made a few correctness checks by migrating a test `store.json` to `store.db` and then back to `store.json`. This may look like the following:
```
# you need store.json in CWD
$ go build -o mgmt ./management
$ ./mgmt sqlite-migration up --log-file console --datadir .
$ mv store.json store.before.json
$ ./mgmt sqlite-migration down --log-file console --datadir .
```

Then you have both `store.before.json` and `store.json`. Just run diff on them or use any other tool to compare two JSONs. There may be some differences mostly related to the way how JSON is serialized. You may see things like this:

```
12c12
<                     "AccountID": "",
---
>                     "AccountID": "b000jehgmce4vh1360cc",
...
312c312
<                     "Peers": null
---
>                     "Peers": []
...
3332,3334c3334,3338
<             "Routes": null,
<             "NameServerGroups": null,
<             "DNSSettings": null,
---
>             "Routes": {},
>             "NameServerGroups": {},
>             "DNSSettings": {
>                 "DisabledManagementGroups": null
>             },
```

Which mostly are equivalents in the current file store implementation.

Management should behave the same way with `store.json` after `up` and `down` migration.

management/server/sqlite_store.go
```diff
@@ -0,0 +1,458 @@
...
+
+func (s *SqliteStore) GetAllAccounts() (all []*Account) {
+	var accounts []Account
+	result := s.db.Find(&accounts)
+	if result.Error != nil {
+		return all
+	}
+
+	for _, account := range accounts {
+		//all = append(all, account.Copy()) // TODO: copy and delete gorm models
```
pappz: Is this comment still valid?

surik: Good caught. Removed.

pappz (dismissed):

mlsmaycon (approved):

PR #1069, opened 2023-08-09
Improve account copying

surik:
With this fix, all nested slices and pointers will be copied by value. 
Also, this fixes tests to compare the original and copy account by their values by marshaling them to JSON strings.

Before that, they were copying the pointers that also passed the simple `=` compassion (as the addresses match).

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


braginini (dismissed):
LGTM
Thanks for the PR!

braginini (approved):

PR #1075, opened 2023-08-11
Update GitHub Actions and Enhance golangci-lint

surik:
## Describe your changes

This PR showcases the implementation of additional linter rules. I've updated the `golangci-lint` GitHub Actions to the latest available version. This update makes sure that the tool works the same way locally - assuming being updated regularly - and with the GitHub Actions.

I've also taken care of keeping all the GitHub Actions up to date, which helps our code stay current. But there's one part, `goreleaser` that's a bit tricky to test on our computers. So, it's important to take a close look at that.

To make it easier to understand what I've done, I've made separate changes for each thing that the new linters found. This should help the people reviewing the changes see what's going on more clearly. Some of the changes might not be obvious at first glance.

### Things to consider for the future

- CI runs on Ubuntu so the static analysis only happens for Linux. Consider running it for the rest: Darwin, Windows

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/routemanager/client.go
+++ b/client/internal/routemanager/client.go
@@ -155,8 +155,9 @@ func (c *clientNetwork) startPeersStatusChangeWatcher() {
 
 func (c *clientNetwork) removeRouteFromWireguardPeer(peerKey string) error {
 	state, err := c.statusRecorder.GetPeer(peerKey)
+	// TODO: should error and non Connected statuses be split?
 	if err != nil || state.ConnStatus != peer.StatusConnected {
-		return nil
+		return nil //nolint:nilerr
 	}
 
 	err = c.wgInterface.RemoveAllowedIP(peerKey, c.network.String())
--- a/client/internal/routemanager/client.go
+++ b/client/internal/routemanager/client.go
@@ -155,9 +155,11 @@ func (c *clientNetwork) startPeersStatusChangeWatcher() {
 
 func (c *clientNetwork) removeRouteFromWireguardPeer(peerKey string) error {
 	state, err := c.statusRecorder.GetPeer(peerKey)
-	// TODO: should error and non Connected statuses be split?
-	if err != nil || state.ConnStatus != peer.StatusConnected {
-		return nil //nolint:nilerr
+	if err != nil {
+		return err
+	}
+	if state.ConnStatus != peer.StatusConnected {
+		return nil
 	}
 
 	err = c.wgInterface.RemoveAllowedIP(peerKey, c.network.String())
--- a/management/server/idp/authentik.go
+++ b/management/server/idp/authentik.go
@@ -3,10 +3,6 @@ package idp
 import (
 	"context"
 	"fmt"
-	"github.com/golang-jwt/jwt"
-	"github.com/netbirdio/netbird/management/server/telemetry"
-	log "github.com/sirupsen/logrus"
-	"goauthentik.io/api/v3"
 	"io"
 	"net/http"
 	"net/url"
@@ -14,6 +10,11 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/golang-jwt/jwt"
+	"github.com/netbirdio/netbird/management/server/telemetry"
+	log "github.com/sirupsen/logrus"
+	"goauthentik.io/api/v3"
 )
 
 // AuthentikManager authentik manager client instance.
@@ -236,6 +237,7 @@ func (am *AuthentikManager) UpdateUserAppMetadata(userID string, appMetadata App
 	if err != nil {
 		return err
 	}
+	defer resp.Body.Close()
 
 	if am.appMetrics != nil {
 		am.appMetrics.IDPMetrics().CountUpdateUserAppMetadata()
@@ -267,6 +269,7 @@ func (am *AuthentikManager) GetUserDataByID(userID string, appMetadata AppMetada
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
 
 	if am.appMetrics != nil {
 		am.appMetrics.IDPMetrics().CountGetUserDataByID()
@@ -294,6 +297,7 @@ func (am *AuthentikManager) GetAccount(accountID string) ([]*UserData, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
 
 	if am.appMetrics != nil {
 		am.appMetrics.IDPMetrics().CountGetAccount()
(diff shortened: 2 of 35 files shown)


surik:
Hi @braginini @mlsmaycon what do you think about this one?

client/internal/routemanager/client.go
```diff
@@ -155,8 +155,9 @@ func (c *clientNetwork) startPeersStatusChangeWatcher() {
 
 func (c *clientNetwork) removeRouteFromWireguardPeer(peerKey string) error {
 	state, err := c.statusRecorder.GetPeer(peerKey)
+	// TODO: should error and non Connected statuses be split?
```
bcmmbaga: Is there a specific use case or design consideration behind this `TODO` comment?

management/server/idp/authentik.go
```diff
@@ -236,6 +237,7 @@ func (am *AuthentikManager) UpdateUserAppMetadata(userID string, appMetadata App
 	if err != nil {
 		return err
 	}
+	defer resp.Body.Close()
```
bcmmbaga, line 240: minor the response is io.NopCloser, means it's a no-op when the caller does Close() on the body. It won't harm to call Close(), but it won't actually do anything because of the NopClose.

management/server/route.go
```diff
@@ -104,7 +105,9 @@ func (am *DefaultAccountManager) checkPrefixPeerExists(accountID, peerID string,
 
 	routesWithPrefix := account.GetRoutesByPrefix(prefix)
 
-	if err != nil {
+	// TODO: This place is not clear. We've already checked the error above but there is additional logic.
```
bcmmbaga: It's important to have a clear understanding of what this part is doing before proceeding merging this. @mlsmaycon can you help here?

bcmmbaga (changes requested):

management/server/route.go
mlsmaycon: this second err check can be removed, as it is a left over from a refactor, the loop provides the necessary check already.

client/internal/routemanager/client.go
surik: Hi @bcmmbaga, the `nilerr` linter pointed to this so I wanted to get attention as I may not be fully aware of the logic here. The `removeRouteFromWireguardPeer` returns `nil` when  `err` is not nil. Should we split `if err != nil || state.ConnStatus != peer.StatusConnected` to two cases?

management/server/idp/authentik.go
bcmmbaga: @surik Can you also remove the close call in these lines? [line#272](https://github.com/netbirdio/netbird/pull/1075/commits/f89804c53f3bab5981a7a598fc15817c77158d66#diff-cc2813e3131a9e44547b4296fb2c8ce749fa988b8010cd8aa9b9c7c7c0276fc8R272), [line#300](https://github.com/netbirdio/netbird/pull/1075/commits/f89804c53f3bab5981a7a598fc15817c77158d66#diff-cc2813e3131a9e44547b4296fb2c8ce749fa988b8010cd8aa9b9c7c7c0276fc8R300), [line#337](https://github.com/netbirdio/netbird/pull/1075/commits/f89804c53f3bab5981a7a598fc15817c77158d66#diff-cc2813e3131a9e44547b4296fb2c8ce749fa988b8010cd8aa9b9c7c7c0276fc8R337), [line#397](https://github.com/netbirdio/netbird/pull/1075/commits/f89804c53f3bab5981a7a598fc15817c77158d66#diff-cc2813e3131a9e44547b4296fb2c8ce749fa988b8010cd8aa9b9c7c7c0276fc8R397), [line#425](https://github.com/netbirdio/netbird/pull/1075/commits/f89804c53f3bab5981a7a598fc15817c77158d66#diff-cc2813e3131a9e44547b4296fb2c8ce749fa988b8010cd8aa9b9c7c7c0276fc8R425), [line#479](https://github.com/netbirdio/netbird/pull/1075/commits/f89804c53f3bab5981a7a598fc15817c77158d66#diff-cc2813e3131a9e44547b4296fb2c8ce749fa988b8010cd8aa9b9c7c7c0276fc8R479) and [line#496](https://github.com/netbirdio/netbird/pull/1075/commits/f89804c53f3bab5981a7a598fc15817c77158d66#diff-cc2813e3131a9e44547b4296fb2c8ce749fa988b8010cd8aa9b9c7c7c0276fc8R479)

client/internal/routemanager/client.go
bcmmbaga: @mlsmaycon, could you assist with this please?

management/server/idp/authentik.go
surik: @bcmmbaga Yes, sure. Do you prefer to completely disable `bodyclose` linter or only make an exception for these lines?

client/internal/routemanager/client.go
mlsmaycon: we can rewrite it to call `c.wgInterface.RemoveAllowedIP(peerKey, c.network.String())` only when the error is nil and the peer is connected

management/server/idp/authentik.go
bcmmbaga: We should not disable the check. Only remove the body close on authentik.go, as it is a no-op

surik: OK, I see. I'm afraid the linter may complain then. Let me try it out.

surik: Yes, it complains. See https://github.com/netbirdio/netbird/actions/runs/6073134678/job/16474466067

```
management/server/idp/authentik.go:236:10: response body must be closed (bodyclose)
                Execute()
                       ^
management/server/idp/authentik.go:267:87: response body must be closed (bodyclose)
        user, resp, err := am.apiClient.CoreApi.CoreUsersRetrieve(ctx, int32(userPk)).Execute()
                                                                                             ^
management/server/idp/authentik.go:294:98: response body must be closed (bodyclose)
        userList, resp, err := am.apiClient.CoreApi.CoreUsersList(ctx).Attributes(accountFilter).Execute()
                                                                                                        ^
management/server/idp/authentik.go:330:72: response body must be closed (bodyclose)
        userList, resp, err := am.apiClient.CoreApi.CoreUsersList(ctx).Execute()
                                                                              ^
management/server/idp/authentik.go:389:101: response body must be closed (bodyclose)
        user, resp, err := am.apiClient.CoreApi.CoreUsersCreate(ctx).UserRequest(createUserRequest).Execute()
```

I believe `bodycheck` just wants consistency for all. I'm not familiar with `goauthentik` package. Does no-op close obvious here without reading the code of the `Execute` method? If not and if this doesn't hurt to call it maybe it makes sense to keep closing them. Let me know what you think.

bcmmbaga: Okay, let's include them

client/internal/routemanager/client.go
surik: I've addressed it in https://github.com/netbirdio/netbird/pull/1075/commits/e4bf3bcadf3dcacdb6ff70b78588cc6b9096b751

bcmmbaga (approved):

PR #1095, opened 2023-08-18
Fix bug with updating Keycloak user attributes

Alpha5k:
## Describe your changes

When using Keycloak as an IdP, NetBird will attempt to write the attributes `wt_pending_invite` and `wt_account_id` to the user's profile the first time they log in. This is done through the Keycloak Admin REST API via a PUT request to `/admin/{realm}/users/{id}`. However, an undocumented function of this API endpoint is it overwrites *all* attributes on the account, including the `email`, `firstName`, and `lastName` fields. If these fields are not specified in the request it overwrites them to blank values.

This can be replicated by creating an account in Keycloak, then logging into a NetBird instance for the first time. You will see that after login the name and email fields are now blank along with any custom attributes in the profile.

Since the Keycloak API doesn't have a PATCH endpoint, we can fix this by getting the user's profile first then appending the attributes to it.

### Checklist
- [X] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/idp/keycloak.go
+++ b/management/server/idp/keycloak.go
@@ -78,6 +78,8 @@ type keycloakProfile struct {
 	CreatedTimestamp int64                  `json:"createdTimestamp"`
 	Username         string                 `json:"username"`
 	Email            string                 `json:"email"`
+	FirstName        string                 `json:"firstName"`
+	LastName         string                 `json:"lastName"`
 	Attributes       keycloakUserAttributes `json:"attributes"`
 }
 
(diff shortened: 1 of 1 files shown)


braginini:
Great stuff @Alpha5k! Thank you for the contribution. We will test this today/tomorrow.

mlsmaycon:
Hello @Alpha5k, thanks for the contribution. We have worked on removing the attribute updates for all IDPs. 

See https://github.com/netbirdio/netbird/pull/1147

With that, you can give only read permissions to the keycloak integration.

Issue #743, opened 2023-03-15
MTU configuration


n0k0m3:
**Is your feature request related to a problem? Please describe.**
I want to be able to change MTU (similar to vanilla Wireguard or Netmaker), either through local configuration or management UI

**Describe the solution you'd like**
Expose MTU configuration to /etc/netbird/config.json or management UI, and config will either be loaded with `-c` or update when `up` is called 

**Describe alternatives you've considered**
None, as NetBird doesn't expose MTU configuration to client

**Additional context**
None

ThHirsch:
I am looking for a solution/fix as well.
We were not able to get Netbird working on secure connections (https, ssh, ...) after (initial) successfull handshake with the servicec to the target client, if the source client is in a networt where the router is connected to a DOCIS3.1 cable modem (it seems it adds/needs additional encapsulation then).
Especially if the connection is then going to send the data in bigger packets, which then get fragmented because of the MTU and breaking the secure connection checks then.  (at typical MTU Problem?!).
Fiddling with the MTU on other levels of the connection did not help - so it seems Netbirds client MTU must be configurable. BTW - I am on MAC OSX 14.2.1
So please consider to include this in the published client, as I am not a delvelooer and I am not able to compile own client with the fixes/patches that seem to be provided by others.

maisen20:
@ThHirsch have you found a workaround to fix this problem? It seems that i have the same problem as you, a network of mine is also connected to the internet via a DOCIS3.1 cable modem.

I think this is why all pings between the networks work without problems (from A to B and B to A), but I can only access all WebUIs and SSH servers in one way. So from network B to A everything works fine, but from network A to B I can only ping all clients in the remote network and nothing more (no WebUI access and no SSH connections of/to clients in Network B).

ThHirsch:
@maisen20 :  No, we ended up in configuring the (lower) MTU on the (WLAN-)Network-Card in the OS (MacOSX in my case).
It seems, if it is configured there, Wireguard/Netbird will 'magically' honor this then (or at least adapt to it).
Hopt this helps for your situation.

alexcupertme:
Is there any updates on this problem? We also have troubles with inability to change MTU while connecting hosts with SSH

ThHirsch:
As tis is still bugging me in our environments, I am keen to see this pcked up and solved.
Especially, as  OpenDUT seemed to have contributed some proposals and even code for a solution?
Is there any progres?
Is it on any roadmap?
[...]

PR #1101, opened 2023-08-22
Custom MTU value for wireguard interface

reimarstier:
The default maximum transmission unit (MTU) for a ethernet network interface is 1500 bytes. By default the wireguard interface is created with a lower MTU value, e.g. 1420 bytes. In case of netbird it is created with a MTU value of 1280 bytes because there is some overhead for tunnel communication accounted for. 
To run some custom applications it might be necessary to have the full 1500 bytes of an ethernet frame. For such use cases I would like to configure a higher MTU than the default value. This will have a negative performance impact to the network communication but it is the lesser of two evil if the other option is that the application won't work. 

This PR adds the field `WgIfaceMtu` to the netbird configuration `config.json` to pass a custom MTU value for the wireguard interface. The value may be passed as a command line parameter to the login command as follows:
```
netbird login --mtu 1500 --management-url https://netbird.local:443 --setup-key xxx
netbird up

# to check the MTU value
root@peer1:/etc/netbird# ip a show dev wt0
5: wt0: <POINTOPOINT,NOARP,UP,LOWER_UP> mtu 1500 [...]
```

## Describe your changes
This allows to pass another [MTU](https://en.wikipedia.org/wiki/Maximum_transmission_unit) value for the wireguard interface. 

See issue #743 

### Checklist
- [x] Extended the README / documentation, if necessary


braginini:
hey @reimarstier 
Thank you for the contribution! 

I assume that this implementation will require a modification of the config file after installation followed by a restart. Some manual work but should be ok with smaller networks.

reimarstier:
> hey @reimarstier Thank you for the contribution!

Gladly ;-)

> I assume that this implementation will require a modification of the config file after installation followed by a restart. Some manual work but should be ok for smaller networks.

Exactly. The workflow to change the MTU would require some manual steps but that would be okay for me at least.
```
netbird login --management-url <TBD> --setup-key <TBD>
# Then edit /etc/netbird/config.json
# grep -i mtu config.json 
    "WgIfaceMtu": 1500,
netbird up
root@peer1:/etc/netbird# ip a show dev wt0
5: wt0: <POINTOPOINT,NOARP,UP,LOWER_UP> mtu 1500 [...]
```


> A more scalable approach would be [CLI](https://github.com/netbirdio/netbird/blob/e586eca16cf4aa95b18afd4585e042cc693a5aa9/client/cmd/root.go#L90) and the local [daemon](https://github.com/netbirdio/netbird/blob/e586eca16cf4aa95b18afd4585e042cc693a5aa9/client/proto/daemon.proto#L31) extension. 

I agree this would be more scalable. But I would rather keep this PR small and create another PR to add this functionality.
Currently there is another PR open that addresses a very similar issue: #659
As of now this PR would be a nice small addition without breaking anything or causing any conflicts.

> Would you be able to do it?

Yes, I think so. But I'd need to get to know a few more things about netbirds CLI logic first.
What else needs to be done to add this?

reimarstier:
As of now with the following test workflow it is working:
```
cd client
rm -rf dist/
GOOS=linux GOARCH=amd64 ~/go/bin/goreleaser build --single-target --snapshot
sudo systemctl stop netbird
sudo cp ./dist/netbird_linux_amd64_v1/netbird /usr/bin/

sudo netbird service uninstall
sudo netbird service install

sudo systemctl start netbird
sudo systemctl status netbird
netbird login --mtu 1538 --management-url https://netbird.local:443 --setup-key xxx
netbird up
netbird status
ip l show dev wt0
```
What is working:
- modifying the config.json and restarting the netbird service
- using the cli as before without the mtu value
- using the cli to set the custom mtu value with the login subcommand
- updating mtu with login command
- omitting the mtu value on present custom configuration will leave the mtu value unchanged

What is not yet working:
- using mtu parameter as argument to the up command
- using the mtu value in the client ui

mlsmaycon:
Hello @reimarstier, We need to put this PR on hold for the next release as we will check a few aspects of the change to ensure it doesn't impact some of the buffers in other parts of the code. 

We should get back to you soon.

reimarstier:
Rebased, updated and tested for v0.23.9.

mlsmaycon:
Hello @reimarstier sorry for the delay to get back on this topic. 

To prevent issues with lower buffers we also need to update the following files with a configurable MTU value:

```
- in eBPF Proxy the buffer sizes are hardcoded
https://github.com/netbirdio/netbird/blob/6d4240a5ae2c1bc16a469dd407a8a6a858c833f0/client/internal/wgproxy/proxy_ebpf.go#L130C1-L130C1
https://github.com/netbirdio/netbird/blob/6d4240a5ae2c1bc16a469dd407a8a6a858c833f0/client/internal/wgproxy/proxy_ebpf.go#L150

- in userpsace ICE proxy the buffer is hardcoded
https://github.com/netbirdio/netbird/blob/6d4240a5ae2c1bc16a469dd407a8a6a858c833f0/client/internal/wgproxy/proxy_userspace.go#L66
https://github.com/netbirdio/netbird/blob/6d4240a5ae2c1bc16a469dd407a8a6a858c833f0/client/internal/wgproxy/proxy_userspace.go#L89

- in our BIND implementation the buffer is harcdoded
https://github.com/netbirdio/netbird/blob/6d4240a5ae2c1bc16a469dd407a8a6a858c833f0/iface/bind/udp_mux_universal.go#L76

- in sharedSock implementation the buffer is hardcoded
https://github.com/netbirdio/netbird/blob/6d4240a5ae2c1bc16a469dd407a8a6a858c833f0/sharedsock/sock_linux.go#L211
```

We are also looking into having this setting coming from the management system, where an administrator sets a value for a group of peers which they know the expected MTU. e.g. cloud hosts in AWS. What do you think about this other approach?

Thanks again for your contribution.

reimarstier:
Hey @mlsmaycon ,

Sorry for not coming back to this earlier. 

I think you are making a good point. I'm looking into configuring the buffers for all components. This ist definitlely important to me.

How do you think this setting for a group may be communicated? I'm a little bit lost there, this touches so many more places of the netbird components. Can you give me some directions?

ThHirsch:
Ist setting/managing MTU available in current builds? 
I am asking, because this issue/ticket is 'closed' but can't see in this comment thread, if it's now 'done'

KingSteve032:
> Ist setting/managing MTU available in current builds? I am asking, because this issue/ticket is 'closed' but can't see in this comment thread, if it's now 'done'

It doesn’t look like it has yet. Would be a nice feature to have. I have network controlled sdr radios that need a mtu higher than 1280.

ThHirsch:
@KingSteve032   it seems this was 'closed', because this issue is tracked in https://github.com/netbirdio/netbird/issues/743 now - which is still 'open' . So I advice to monitor #743 then.

PR #1106, opened 2023-08-24
Add auto-update feature in netbird script for binary installation

bcmmbaga:
## Describe your changes

This pull request addresses the need to enhance the installer script by introducing a new parameter `--update` to trigger updates. The goal is to streamline the update process for binary installations and provide a better experience for users.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/release_files/install.sh
+++ b/release_files/install.sh
@@ -28,7 +28,7 @@ download_release_binary() {
     BASE_URL="https://github.com/${OWNER}/${REPO}/releases/download"
     BINARY_BASE_NAME="${VERSION#v}_${OS_TYPE}_${ARCH}.tar.gz"
 
-    # for Darwin, download the signed Netbird-UI
+    # for Darwin, download the signed NetBird-UI
     if [ "$OS_TYPE" = "darwin" ] && [ "$1" = "$UI_APP" ]; then
         BINARY_BASE_NAME="${VERSION#v}_${OS_TYPE}_${ARCH}_signed.zip"
     fi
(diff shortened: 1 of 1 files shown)


release_files/install.sh
```diff
@@ -303,4 +303,43 @@ install_netbird() {
...
+
+      sudo netbird service stop
+      sudo netbird service uninstall
+      install_native_binaries
+
+      sudo netbird service install
+      sudo netbird service start
+    fi
+  else
+     echo "Netbird installation was done using a package manager. please use package to update"
```
braginini:
```suggestion
     echo "Netbird installation was done using a package manager. Please use your system's package manager to update"
```

braginini (approved):

Aqua1ung:
So, for _binary installations_, how does one _update_ NetBird? Just run `install.sh` again?

bcmmbaga:
> So, for _binary installations_, how does one _update_ NetBird? Will running `curl -fsSL https://pkgs.netbird.io/install.sh | sh` again do it?

You can update by running `curl -fsSL https://pkgs.netbird.io/install.sh | sh -s -- --update`

Aqua1ung:
> You can update by running `curl -fsSL https://pkgs.netbird.io/install.sh | sh -s -- --update`

I am guessing this an intelligent update (meaning no update if none needed), right?

bcmmbaga:
> Thanks. I am guessing this an _intelligent_ update (meaning no update if none needed), right?

Correct. In case you are running the latest version, there will be no updates.

PR #1113, opened 2023-08-30
Handle user delete

gigovich:
## Describe your changes
Implement user deletion across all IDP-ss. Expires all user peers 
when the user is deleted. Users are permanently removed from a local 
store, but in IDP, we remove Netbird attributes for the user 
untilUserDeleteFromIDPEnabled setting is not enabled for the account.

To test, an admin user should remove any additional users.

Until the UI incorporates this feature, use a curl **DELETE** request 
targeting the **/users/<USER_ID>** management endpoint. Note that this 
request only removes user attributes and doesn't trigger a delete 
from the IDP.

To enable user removal from the IdP, set **UserDeleteFromIDPEnabled**
to true in account settings. Until we have a UI for this, make this 
change directly in the store file.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/user.go
```diff
@@ -330,12 +330,40 @@ func (am *DefaultAccountManager) DeleteUser(accountID, initiatorUserID string, t
 		return status.Errorf(status.PermissionDenied, "only admins can delete service users")
```
surik: We probably should change the error message as now we can delete not only service users

management/server/user_test.go
```diff
@@ -424,25 +424,6 @@ func TestUser_DeleteUser_ServiceUser(t *testing.T) {
 	assert.Nil(t, store.Accounts[mockAccountID].Users[mockServiceUserID])
 }
 
-func TestUser_DeleteUser_regularUser(t *testing.T) {
```
surik: Is deletion of regular users anyhow different apart from activity events? Maybe consider reworking it to a positive scenario and keeping it?

management/server/user.go
```diff
@@ -330,12 +330,40 @@ func (am *DefaultAccountManager) DeleteUser(accountID, initiatorUserID string, t
...
+		meta = map[string]any{"name": targetUser.ServiceUserName}
+		eventAction = activity.ServiceUserDeleted
+	} else {
+		meta = map[string]any{"email": targetUserEmail}
+		eventAction = activity.UserDeleted
+
+	}
+	am.storeEvent(initiatorUserID, targetUserID, accountID, eventAction, meta)
+
+	if !isNil(am.idpManager) {
```
surik, line 361: Small thing but we can move this check inside `deleteUserFromIDP` as it is closed to the place we `idpManager` is being used.

management/server/account.go
```diff
@@ -738,18 +741,19 @@ func (a *Account) UserGroupsRemoveFromPeers(userID string, groups ...string) {
 
 // BuildManager creates a new DefaultAccountManager with a provided Store
 func BuildManager(store Store, peersUpdateManager *PeersUpdateManager, idpManager idp.Manager,
-	singleAccountModeDomain string, dnsDomain string, eventStore activity.Store,
+	singleAccountModeDomain string, dnsDomain string, eventStore activity.Store, userDeleteFromIDPEnabled bool,
```
surik, line 744: Looks like in the test we always have `userDeleteFromIDPEnabled = false` there may be benefits of also testing the path when it is enabled.

pappz: Yes you are right but what can I proof with it? We can do test to verify to call the external idpManager.DeleteUser() or not. But in this case the idpManager could be a moc interface so at the end we will verify that the moc implementation works fine or not.
But I am not a test expert. If you think it have any benefits in unit test layer then I will write tests for it.

management/server/user.go
pappz: Good catch

management/server/user.go
pappz: I think that would be more complicated. If put this nil check into **deleteUserFromIDP** then I need create a new **error** (like, idp manager is not instantiated) and it must to handle in this line. On other case if I do not handle this error just ignore it then the **delteUserFromIDP** will doing nothing in some cases and it maybe not so clear.
What do you think?

surik: No problem, lets keep it as it is then

management/server/account.go
bcmmbaga: Enabling the `userDeleteFromIDPEnabled=true` testing option will exclusively assess the mock scenario and won't involve actual deletion through the IDP API. We can address this by conducting manual testing for each IDP with the flag enabled.

surik: Yes, more likely it will be just checking if the correct method is being called and certain error paths. But you might be right there are not many benefits in it currently.

bcmmbaga (approved):
Looks good to me! Please note that we need to update the Auth0 management API permission to include `delete:users` for users who intend to remove users from the IDP after deletion.

surik (approved):

PR #1118, opened 2023-09-01
Integrate Rosenpass

pascal-fischer:
## Describe your changes
This PR aims to integrate [Rosenpass](https://rosenpass.eu) with NetBird. It adds a manager for Rosenpass that starts a Rosenpass server and handles the managed peers. It uses the [cunicu/go-rosenpass](https://github.com/cunicu/go-rosenpass) implementation. Rosenpass will then negotiate a pre shared-key every 2 minutes and apply it to the wireguard connection.

The Feature can be enabled by setting [a flag] during the `netbird up` command [currently it uses a env variable `NB_RP_ENABLED`]. 

If two peers are both support and have the Rosenpass feature enabled they will create a post-quantum secure connection. If one of the peers or both don't have this feature enabled or are running an older version that does not have this feature yet, the NetBird client will fallback to plain Wireguard connection without pre-shared keys for those connections (keeping Rosenpass negotiation for the rest).

Additionally this PR includes an update of all Github Actions workflows to use go version 1.21.0 as this is a requirement for the integration. 

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


pascal-fischer:
Whats missing in this PR before it can be ready is solving the 2-minute-problem and replacing the env variable with a flag to enable rosenpass feature. Furthermore the error handling could be improved e.g. if we add the peer to the rosenpass server but adding the peer to the WgHandler fails we should either try to fix it or remove the peer from the handler as well.

pascal-fischer:
closed in favor of https://github.com/netbirdio/netbird/pull/1153

PR #1133, opened 2023-09-08
Extend activity store with deleted emails

pappz:
## Describe your changes
Store the email address of the deleted user in activity db.
The information will be encrypted with AES.

Refactor the events API endpoint. In the response the arguments 
will be extended with the email of the initiator and if relevant then 
the email of the target.

- Add new store encryption key config parameter to the management config.
The key must to be 32 byte in base64 format.
- The user delete handler will store the deleted email address in the activity store
- The get all events handler map the emails based on the user id from the activity store.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/management/cmd/management.go
+++ b/management/cmd/management.go
@@ -23,6 +23,7 @@ import (
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/h2c"
 
+	"github.com/netbirdio/netbird/management/server/activity"
 	"github.com/netbirdio/netbird/management/server/activity/sqlite"
 	httpapi "github.com/netbirdio/netbird/management/server/http"
 	"github.com/netbirdio/netbird/management/server/jwtclaims"
@@ -146,10 +147,20 @@ var (
 			if disableSingleAccMode {
 				mgmtSingleAccModeDomain = ""
 			}
-			eventStore, err := sqlite.NewSQLiteStore(config.Datadir)
+			eventStore, key, err := initEventStore(config.Datadir, config.DataStoreEncryptionKey)
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to initialize database: %s", err)
 			}
+
+			if key != "" {
+				log.Debugf("update config with activity store key")
+				config.DataStoreEncryptionKey = key
+				err := updateMgmtConfig(mgmtConfig, config)
+				if err != nil {
+					return fmt.Errorf("failed to write out store encryption key: %s", err)
+				}
+			}
+
 			accountManager, err := server.BuildManager(store, peersUpdateManager, idpManager, mgmtSingleAccModeDomain,
 				dnsDomain, eventStore, userDeleteFromIDPEnabled)
 			if err != nil {
@@ -286,6 +297,20 @@ var (
 		},
 	}
 )
+
+func initEventStore(dataDir string, key string) (activity.Store, string, error) {
+	var err error
+	if key == "" {
+		log.Debugf("generate new activity store encryption key")
+		key, err = sqlite.GenerateKey()
+		if err != nil {
+			return nil, "", err
+		}
+	}
+	store, err := sqlite.NewSQLiteStore(dataDir, key)
+	return store, key, err
+
+}
 
 func notifyStop(msg string) {
 	select {
@@ -438,6 +463,10 @@ func loadMgmtConfig(mgmtConfigPath string) (*server.Config, error) {
 	}
 
 	return loadedConfig, err
+}
+
+func updateMgmtConfig(path string, config *server.Config) error {
+	return util.WriteJson(path, config)
 }
 
 // OIDCConfigResponse used for parsing OIDC config response
--- a/management/server/activity/event.go
+++ b/management/server/activity/event.go
@@ -15,7 +15,7 @@ type Event struct {
 	// InitiatorID is the ID of an object that initiated the event (e.g., a user)
 	InitiatorID string
 	// InitiatorEmail is the email address of an object that initiated the event. This will be set on deleted users only
-	InitiatorEmail *string
+	InitiatorEmail string
 	// TargetID is the ID of an object that was effected by the event (e.g., a peer)
 	TargetID string
 	// AccountID is the ID of an account where the event happened
--- a/management/server/activity/event.go
+++ b/management/server/activity/event.go
@@ -14,10 +14,13 @@ type Event struct {
 	ID uint64
 	// InitiatorID is the ID of an object that initiated the event (e.g., a user)
 	InitiatorID string
+	// InitiatorEmail is the email address of an object that initiated the event. This will be set on deleted users only
+	InitiatorEmail *string
 	// TargetID is the ID of an object that was effected by the event (e.g., a peer)
 	TargetID string
 	// AccountID is the ID of an account where the event happened
 	AccountID string
+
 	// Meta of the event, e.g. deleted peer information like name, IP, etc
 	Meta map[string]any
 }
@@ -31,12 +34,13 @@ func (e *Event) Copy() *Event {
 	}
 
 	return &Event{
-		Timestamp:   e.Timestamp,
-		Activity:    e.Activity,
-		ID:          e.ID,
-		InitiatorID: e.InitiatorID,
-		TargetID:    e.TargetID,
-		AccountID:   e.AccountID,
-		Meta:        meta,
+		Timestamp:      e.Timestamp,
+		Activity:       e.Activity,
+		ID:             e.ID,
+		InitiatorID:    e.InitiatorID,
+		InitiatorEmail: e.InitiatorEmail,
+		TargetID:       e.TargetID,
+		AccountID:      e.AccountID,
+		Meta:           meta,
 	}
 }
--- a/management/server/activity/sqlite/crypt.go
+++ b/management/server/activity/sqlite/crypt.go
@@ -18,8 +18,11 @@ type EmailEncrypt struct {
 func GenerateKey() (string, error) {
 	key := make([]byte, 32)
 	_, err := rand.Read(key)
+	if err != nil {
+		return "", err
+	}
 	readableKey := base64.StdEncoding.EncodeToString(key)
-	return readableKey, err
+	return readableKey, nil
 }
 
 func NewEmailEncrypt(key string) (*EmailEncrypt, error) {
--- a/management/server/activity/sqlite/crypt.go
+++ b/management/server/activity/sqlite/crypt.go
@@ -1,0 +1,78 @@
+package sqlite
+
+import (
+	"bytes"
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+)
+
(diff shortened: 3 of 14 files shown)


management/server/http/api/openapi.yml
```diff
@@ -913,6 +913,10 @@ components:
           description: The ID of the initiator of the event. E.g., an ID of a user that triggered the event.
           type: string
           example: google-oauth2|123456789012345678901
+        initiator_email:
+          description: The e-mail address of the initiator of the event. E.g., an e-mail of a user that triggered the event.
+          type: string
```
surik, line 918: Would it be beneficial to add `format: email` here?

management/cmd/management.go
```diff
@@ -440,6 +465,10 @@ func loadMgmtConfig(mgmtConfigPath string) (*server.Config, error) {
 	return loadedConfig, err
 }
 
+func updateMgmtConfig(path string, config *server.Config) error {
```
surik, line 468: Looks like this function is called only once. I wonder if we need this wrapper or maybe instead better to call `util.WriteJson(path, config)` directly?

management/server/http/events_handler.go
```diff
@@ -45,22 +45,50 @@ func (h *EventsHandler) GetAllEvents(w http.ResponseWriter, r *http.Request) {
 		util.WriteError(err, w)
 		return
 	}
-	events := make([]*api.Event, 0)
+	events := make([]*api.Event, 0, len(accountEvents))
```
surik: We can avoid using append as we know the length of the slice. What do you think about something like this?

```go
events := make([]*api.Event, len(accountEvents))
for i, e := range accountEvents {
	events[i] = toEventResponse(e)
}
```

management/server/http/events_handler.go
```diff
@@ -45,22 +45,50 @@ func (h *EventsHandler) GetAllEvents(w http.ResponseWriter, r *http.Request) {
...
+
+	emails := make(map[string]string)
+	for _, ui := range userInfos {
+		emails[ui.ID] = ui.Email
+	}
+
+	// fill event with email of initiator
+	for _, event := range events {
+		if event.InitiatorEmail == "" {
+			event.InitiatorEmail = emails[event.InitiatorId]
```
surik: Can we have a situation when `emails` doesn't have a key `event.InitiatorId`? If so maybe writing a log message or returning an error can help.

management/server/activity/event.go
```diff
@@ -31,12 +34,13 @@ func (e *Event) Copy() *Event {
...
-		ID:          e.ID,
-		InitiatorID: e.InitiatorID,
-		TargetID:    e.TargetID,
-		AccountID:   e.AccountID,
-		Meta:        meta,
+		Timestamp:      e.Timestamp,
+		Activity:       e.Activity,
+		ID:             e.ID,
+		InitiatorID:    e.InitiatorID,
+		InitiatorEmail: e.InitiatorEmail,
```
surik, line 41: This one copies the address, not the value. I wonder if this is what we want. Would it be simpler to have `InitiatorEmail string`?

management/server/activity/sqlite/crypt.go
```diff
@@ -0,0 +1,78 @@
...
+
+type EmailEncrypt struct {
+	block cipher.Block
+}
+
+func GenerateKey() (string, error) {
+	key := make([]byte, 32)
+	_, err := rand.Read(key)
+	readableKey := base64.StdEncoding.EncodeToString(key)
+	return readableKey, err
```
surik: Probably should not return `readableKey` if `err != nil`

management/server/activity/sqlite/sqlite.go
```diff
@@ -24,19 +26,28 @@ const (
 		"meta TEXT," +
 		" target_id TEXT);"
 
-	selectStatement = "SELECT id, activity, timestamp, initiator_id, target_id, account_id, meta" +
-		" FROM events WHERE account_id = ? ORDER BY timestamp %s LIMIT ? OFFSET ?;"
+	creatTableAccountEmailQuery = `CREATE TABLE IF NOT EXISTS deleted_users (id TEXT NOT NULL, email TEXT NOT NULL);`
+
+	selectStatement = `SELECT events.id, activity, timestamp, initiator_id, i.email as "initiator_email", target_id, t.email as "target_email", account_id, meta
```
surik, line 31: More likely outside of this PR but I believe having indices here may benefit performance.

management/server/activity/sqlite/sqlite.go
```diff
@@ -140,6 +188,22 @@ func (store *Store) Save(event *activity.Event) (*activity.Event, error) {
 	return eventCopy, nil
 }
 
+func (store *Store) SaveWithDeletedUserEmail(event *activity.Event, email string) (*activity.Event, error) {
+	email = store.emailEncrypt.Encrypt(email)
+
+	stmt, err := store.db.Prepare(insertDeleteUserStatement)
```
surik: I suggest not preparing a statement each time we call this method. Let's do it once in `NewSQLiteStore`. Honestly, I would expect golangci-lint to complain about it. Could it be that the PR needs a rebase on the main branch?

surik (changes requested):
Looks good from my first view. I added few question/comments.

management/server/http/api/openapi.yml
pappz: Good idea, I will extend with it.

management/cmd/management.go
pappz: I think the one line functions sometimes good idea. A function name can explain better what we do in the code. I write source code not just for the compiler I write code for another developer also.

management/server/http/events_handler.go
pappz: Good point, but in the case of the append function there is really minimal overhead. It does not allocate new memory space, it just updates the attributes of the slice.

management/server/http/events_handler.go
pappz: Well, if everything goes well it should not happen. It could happen only that case if the database is corrupted. I do not think so we should handle this error on this layer but I will consider.

management/server/activity/event.go
pappz: In the db we do not store for every events email address. In this case a pointer what could be nil express better the db structure. But definitely you are right it would be simpler a simple string. I will check what can I do.

management/server/activity/sqlite/crypt.go
pappz: I will fix it

management/server/activity/sqlite/sqlite.go
pappz: Good point. If you feel motivation to create a PR it would be a good improvements.

management/server/activity/sqlite/sqlite.go
pappz: Could you explain more what do you think exactly? Maybe I do not understand something but looks like that we do not use this function in a batch event or something like that.

management/server/http/api/openapi.yml
pappz: Unfortunately I have to say I can not use your suggestion because it depend on multiple library update and it can cause unpredictable issues. It is too much risk at this point.

surik: Oh, I see. No need to do this now then. Thank you for verifying!

management/server/activity/sqlite/sqlite.go
surik: My concern here is that every time `SaveWithDeletedUserEmail` is being called we recreate a statement by  `store.db.Prepare`. I can't explain it any better than in this document: https://go.dev/doc/database/prepared-statements.
```
When you expect to execute the same SQL repeatedly, you can use an sql.Stmt to prepare the SQL statement in advance, then execute it as needed.
```

See how the rest of the statements are prepared in `NewSQLiteStore`: https://github.com/netbirdio/netbird/blob/bb791d59f3cf55437d2e7749f9df2759ae019b80/management/server/activity/sqlite/sqlite.go#L57

management/server/activity/sqlite/sqlite.go
```diff
@@ -74,15 +109,36 @@ func processResult(result *sql.Rows) ([]*activity.Event, error) {
...
 			TargetID:    target,
 			AccountID:   account,
 			Meta:        meta,
-		})
+		}
+
+		if initiatorEmail != nil {
+			email, err := store.emailEncrypt.Decrypt(*initiatorEmail)
+			if err != nil {
+				log.Errorf("failed to decrypt data: %s", *initiatorEmail)
```
surik: I'm not very sure if loging `initiatorEmail` is helpful as it may be cryptic. Maybe ID and AccountID are better?

management/server/http/api/openapi.yml
```diff
@@ -929,6 +933,7 @@ components:
         - activity
         - activity_code
         - initiator_id
+        - initiator_email
```
surik, line 936: I guess we also keep track of target_email. Do we need to return it in the response?

management/server/activity/sqlite/sqlite.go
pappz: If the decryption fail then the encrypted data is the most interesting information. The AccountID and ID does not carry any relevant information but if it necessary we can figure it out based on the encrypted string from the DB.

management/server/http/api/openapi.yml
pappz: Not exactly. We response the target email in the "meta" map because different type of actions point to different type of resource. For example if I delete a DNS setting or a route setting then make no sense to send target_email, because we do not talk about a user.

management/server/activity/sqlite/sqlite.go
surik: OK, I see your point. Also, I have another thought. Could we have a situation in which the encryption is broken in a way that the email information will be exposed? Database and logger backend may have different access tiers. In my understanding, the combination of ID and AccountID is unique to find the broken record and minimize the exposure.

pappz: In this case will see log line like this:
```
2023-09-12T14:03:22+02:00 ERRO management/server/activity/sqlite/sqlite.go:115: failed to decrypt data: O5n/22BXKs14OkF5Mx/o3uO9H/Ub9Ms6bF/YNzRLWoo=
2023-09-12T14:03:22+02:00 WARN management/server/http/events_handler.go:81: failed to resolve email for initiator: auth0|64fb0efda99d862a48616d96
```
Will appear the real e-mail address in the logs in case if the db contains the email in unencrypted format. But I understand you. If you wish I will modify this log message. Easy modification.

surik: I would do it but maybe I am too paranoid. @mlsmaycon what do you think?

management/server/http/api/openapi.yml
surik: Thank you for the clarification. I think it is fine then.

management/server/activity/sqlite/sqlite.go
mlsmaycon: @pappz, the user ID will be best for this output. in the end we can map it back to the failed hash using it

surik (approved):

PR #1145, opened 2023-09-13
fix NETBIRD_SIGNAL_PORT not working with custom port

Fantu:
Use NETBIRD_SIGNAL_PORT variable instead of the static port for signal container in the docker-compose template to make setting of custom signal port working

## Describe your changes
trying to use custom signal port I saw it was not working and signal container still use the default one.
after a fast look to commit that implement it and comparing with other variable for custom port working I found the missing in docker template that still have the static port to default.
This patch add the use of variable instead of the static port.
From a test seems now working correctly with custom port.

### Checklist
- [X] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/infrastructure_files/docker-compose.yml.tmpl
+++ b/infrastructure_files/docker-compose.yml.tmpl
@@ -36,7 +36,7 @@ services:
     volumes:
       - $SIGNAL_VOLUMENAME:/var/lib/netbird
     ports:
-      - 10000:80
+      - $NETBIRD_SIGNAL_PORT:80
   #      # port and command for Let's Encrypt validation
   #      - 443:443
   #    command: ["--letsencrypt-domain", "$NETBIRD_LETSENCRYPT_DOMAIN", "--log-file", "console"]


mlsmaycon (changes requested):
Hello @Fantu Thanks for the contribution. 

To avoid further cases where we miss this configuration, can you please add a test for this port update?

You can add a unique value in the https://github.com/netbirdio/netbird/blob/main/infrastructure_files/tests/setup.env

And then do a grep check similar to the ones in the workflow file: https://github.com/netbirdio/netbird/blob/main/.github/workflows/test-infrastructure-files.yml#L85

Fantu:
@mlsmaycon thanks for reply
I don't have enough experience on these parts, I had never used netbird (only fast test on debug issue related to signal) and like other things my employer told me to prepare the netbird server for his use and/or that of others and then later told me to fix it given the functioning problem of the signal part and seeing which was urgent for him I looked into it and found/fixed the problem (was easy and fast without study the parts in this case) and after testing that it worked I made the pull request to help the project

I didn't see other custom ports in tests/setup.env so I'm not able to guess how to do the correct implementation without know how it works

mlsmaycon (approved):
@Fantu no worries, thanks again for the contribution. 

We will add the tests in a new PR.

Issue #1126, opened 2023-09-04
Netbird and Keycloak connect to LDAP don't work


emferrandi:
**Describe the problem**
I'm facing an issue using Keycloak and Active Directory with Netbird.
I connected Keycloak with Windows Active Directory and I manage to connect to Netbird.
And on the Peers page, I see this error : " Request failed with status code 401. Please refresh the page if the issue continues. invalid jwt"
When I connect to Netbird using an internal user (not imported from AD), it works perfectly.

**To Reproduce**
1. Install Netbird and Keycloak in self-hosted mode
2. Configure Keycloak to connect to LDAP
3. Try to connect to user imported from LDAP

**Expected behavior**
See Peer information from the URL 

**Screenshots**
![image](https://github.com/netbirdio/netbird/assets/92516978/14c4b1ec-89f7-4196-8f00-32d67f7094cd)

**Additional context**
Difference between user : 
Classic user in Keycloak :
```
{
    "id":"3f3ce19d-52cf-4579-b5b7-d499d6006e32",
    "createdTimestamp":1690813563692,
    "username":"user",
    "enabled":true,
    "totp":false,
    "emailVerified":false,
    "firstName":"user",
    "lastName":"user",
    "email":"user@mail.com",
    "attributes":{
        "wt_pending_invite":["false"],
        "wt_account_id":["cj3shulxxxx738xxxxg"]
    },
    "disableableCredentialTypes":[],
    "requiredActions":[],
    "notBefore":0,
    "access":{
        "manageGroupMembership":true,
        "view":true,
        "mapRoles":true,
        "impersonate":false,
        "manage":true
    }
}
```
And user imported from AD :
```
{
    "id":"1cxxx9d7-f0xx-4fxx-bbxx-18749xxxxba7b",
    "createdTimestamp":1693477962721,
    "username":"UserAD",
    "enabled":true,
    "totp":false,
    "emailVerified":false,
    "firstName":"Théo",
    "lastName":"Casas",
    "email":"UserAD@mail.com",
    "federationLink":"738xx269-d0xx-49xx-xx09-4223xxxxa336",
    "attributes":{
        "LDAP_ENTRY_DN":["CN=UserAD,OU=Utilisateurs,OU=xxx,DC=xxx,DC=xxx"],
        "createTimestamp":["20190516124504.0Z"],
        "modifyTimestamp":["20230831104958.0Z"],
        "LDAP_ID":["xxx"]
    },
    "disableableCredentialTypes":[],
[...]

Issue #1115, opened 2023-08-30
Google Workspace IdP not setting email addresses properly


marcbudofsky:
**Describe the problem**
With a clean install of v0.22.7 and leveraging Google Workspace as the IdP, users are only showing ID's with no email attribute set. Based on #959, `NETBIRD_MGMT_IDP="none"` has been defined in setup.env to fix the Invalid JWT issue. I'm assuming the lack of email addresses is a side effect.

**To Reproduce**

1. Pull latest https://github.com/netbirdio/netbird.git
2. Copy infrastructure_files/setup.env.template to infrastructure_files/setup.env (overwrite existing file from old version)
3. Fill in correct values as identified in the Google Workspace setup guide: https://docs.netbird.io/selfhosted/identity-providers#google-workspace
4. Re-run ./configure.sh
5. Run docker compose pull && docker compose down && docker compose up -d

**Expected behavior**
Email addresses are set in Netbird when a user entry is created.

**Screenshots**
![users](https://github.com/netbirdio/netbird/assets/104649/6fb0e4c1-f7ef-4523-b2f0-b077f3835c78)
![user-details](https://github.com/netbirdio/netbird/assets/104649/92d097c4-b3c5-4cfe-8fed-d3e4fad0ee08)

**Additional context**
Setting `NETBIRD_MGMT_IDP="google"` caused an Invalid JWT error and rendered Netbird unusable.

mosi0815:
I just found the problem that makes the Google IDP unusable. On the first login with a new user, Netbird tries to set the wt_account_id attribute in the users app_metadata custom schema. 
This fails because of an invalid metadata payload in the update query. It contains an additional field that Netbird didn't add to the custom schema beforehand.
I'm neither a GO expert nor very familiar with the Netbird codebase but I got it working by changing the AppMetadata struct in idp.go like this:
```go
type AppMetadata struct {
        // WTAccountID is a NetBird (previously Wiretrustee) account id to update in the IDP
        // maps to wt_account_id when json.marshal
        WTAccountID     string `json:"wt_account_id,omitempty"`
        WTPendingInvite *bool  `json:"wt_pending_invite,omitempty"`
        WTInvitedBy     string `json:"wt_invited_by_email,omitempty"`
}
```

Google needs a couple of seconds to update the metadata. So the invalid JWT error still appears after the first login but after a refresh of the page a couple of seconds later everything works as expected.

mosi0815:
Update:
[...]

PR #1147, opened 2023-09-14
Optimize Cache and IDP Management

bcmmbaga:
## Describe your changes

This pull request modifies the IdP and cache manager(s) to prevent the sending of app metadata to the upstream IDP on self-hosted instances. As a result, the IdP will now load all users from the IdP without filtering based on accountID.

* Netbird and Keycloak connect to LDAP don't work - [#1126](https://github.com/netbirdio/netbird/issues/1126)
* Google Workspace IdP not setting email addresses properly - [#1115](https://github.com/netbirdio/netbird/issues/1115)
* Fix bug with updating Keycloak user attributes - [#1095](https://github.com/netbirdio/netbird/pull/1095)

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


management/server/account.go
```diff
@@ -928,6 +928,27 @@ func (am *DefaultAccountManager) warmupIDPCache() error {
...
+	// in cases like this, we expect it to return all users in an "unset" field.
+	// We iterate over the users in the "unset" field, look up their AccountID in our store, and
+	// update their AppMetadata with the AccountID.
+	if unsetData, ok := userData["unset"]; ok {
+		for _, user := range unsetData {
+			accountID, err := am.Store.GetAccountByUser(user.ID)
+			if err == nil {
+				data := userData[accountID.Id]
+				if data == nil {
+					data = make([]*idp.UserData, 0)
```
pappz: Better if you set the capacity to 1.

pappz (changes requested):

management/server/account.go
```diff
@@ -928,6 +928,27 @@ func (am *DefaultAccountManager) warmupIDPCache() error {
...
+			accountID, err := am.Store.GetAccountByUser(user.ID)
+			if err == nil {
+				data := userData[accountID.Id]
+				if data == nil {
+					data = make([]*idp.UserData, 1)
+				}
+
+				user.AppMetadata.WTAccountID = accountID.Id
+
+				userData[accountID.Id] = append(data, user)
```
pappz, line 946: It provide incorrect slice because of the slice size is already 1. If you append more element then the size will be 2.
You can do two thinks. Make the slice with 0 size and with 1 capacity and use the append or point directly to the proper elements of the data slice instead of append (data[0] = user)

pappz (changes requested):

pappz (approved):

mlsmaycon (approved):

PR #1155, opened 2023-09-19
Add direct write file to avoid moving docker mounted files

### LINKED ISSUE #1154: Request failed with status code 502
**Describe the problem**
Installed Netbird by using getting-started-with-zitadel.sh on a OVH VPS but after accessing console I'm gettin 502 error 
obraz

netbird-management-1 container keep crashing with following error message
`2023-09-19T19:46:44Z INFO management/server/telemetry/app_metrics.go:161: enabled application metrics and exposing on http://0.0.0.0:8081
Error: failed to write out store encryption key: rename /etc/netbird/.2772331671management.json /etc/netbird/management.json: device or resource busy
`

**To Reproduce**

Follow Quickstart guide

**Expected behavior**
Working console

mlsmaycon:
fixes #1154
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/util/file.go
+++ b/util/file.go
@@ -91,9 +91,9 @@ func DirectWriteJson(file string, obj interface{}) error {
 }
 
 func openOrCreateFile(file string) (*os.File, error) {
-	_, err := os.Stat(file)
+	s, err := os.Stat(file)
 	if err == nil {
-		return os.Open(file)
+		return os.OpenFile(file, os.O_WRONLY, s.Mode())
 	}
 
 	if !os.IsNotExist(err) {
--- a/util/file.go
+++ b/util/file.go
@@ -104,6 +104,7 @@ func openOrCreateFile(file string) (*os.File, error) {
 	if err2 != nil {
 		return nil, err2
 	}
+	//no:lint
 	err2 = targetFile.Chmod(0640)
 	if err != nil {
 		_ = targetFile.Close()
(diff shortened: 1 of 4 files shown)


util/file.go
```diff
@@ -54,6 +56,63 @@ func WriteJson(file string, obj interface{}) error {
...
+	}
+
+	targetFile, err := os.Create(file)
+	if err != nil {
+		return nil, err
+	}
+	//no:lint
+	err = targetFile.Chmod(0640)
+	if err != nil {
+		_ = targetFile.Close()
```
pappz, line 115: We should close the file in all cases independently from the err of the chmod

pappz (changes requested):

util/file.go
braginini: It is closed in the DirectWriteJson function

pappz: I think no. Because we schedule the close with defer in line 72. and in case of chmod err we never reach the 72. line.

pappz: Ahh, you are right

pappz (changes requested):

pappz:
Something is not correct. It provide invalid json file.
This is the end of my file:
```
            "Scope": "openid",
            "UseIDToken": false,
            "RedirectURLs": null
        }
    },
    "PKCEAuthorizationFlow": null
}ow": null
}
```

pappz (approved):

PR #1156, opened 2023-09-20
Use sync.Map in updatechannel

surik:
## Describe your changes

The [sync.Map](https://pkg.go.dev/sync#Map) is part of Golang and looks a better fit for the case of concurrently accessing the map of the update channels. 

```
The Map type is optimized for two common use cases: 
(1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or 
(2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. 
In these two cases, the use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.
```

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


surik:
I split out the metric to its own PR #1158

surik:
I'm closing it as it may need better consideration to understand the impact.

PR #1161, opened 2023-09-23
feat: start daemon on demand (fixes #444 partially)

### LINKED ISSUE #444: Add option to don't automatically connect on reboot
Today the client connects automatically every time the service is started and the peer is authenticated

We should add an option to prevent this behavior if the user doesn't want to keep a constant connection

oskardotglobal:
## Describe your changes
This is still a WIP, but it is a proposal to fix the auto-start problem by allowing users to opt-out out of automatically starting netbird on boot.
I do this by trying to start the daemon whenever you try to connect. This at least allows users to manually disable auto-start if required.

[#444](https://github.com/netbirdio/netbird/issues/444)

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/client/cmd/root.go
+++ b/client/cmd/root.go
@@ -110,8 +110,8 @@ func init() {
 	rootCmd.AddCommand(loginCmd)
 	rootCmd.AddCommand(versionCmd)
 	rootCmd.AddCommand(sshCmd)
-	serviceCmd.AddCommand(runCmd, startCmd, stopCmd, restartCmd) // service control commands are subcommands of service
-	serviceCmd.AddCommand(installCmd, uninstallCmd)              // service installer commands are subcommands of service
+	serviceCmd.AddCommand(runCmd, startCmd, stopCmd, restartCmd, enableCmd, disableCmd) // service control commands are subcommands of service
+	serviceCmd.AddCommand(installCmd, uninstallCmd)                                     // service installer commands are subcommands of service
 	upCmd.PersistentFlags().StringSliceVar(&natExternalIPs, externalIPMapFlag, nil,
 		`Sets external IPs maps between local addresses and interfaces.`+
 			`You can specify a comma-separated list with a single IP and IP/IP or IP/Interface Name. `+
--- a/client/cmd/service_controller.go
+++ b/client/cmd/service_controller.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"github.com/netbirdio/netbird/client/internal"
 	"net"
 	"os"
 	"strings"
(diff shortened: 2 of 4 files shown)


oskardotglobal:
By the way, I've taken a look at the service package and it seems like there's currently no way to use that to enable / disable the service

oskardotglobal:
To be honest, I kind of forgot this PR existed; I did completely revamp how I did this without ever committing it.

With these latest changes, the client will just not connect when the daemon is started unless this is changed via the config. However, it still defaults to connecting by default so this PR is backwards compatible. This functionality can be toggled with the `netbird service disable` and `netbird service enable` commands.

mlsmaycon:
Hello @oskardotglobal thanks for the awesome contribution. I have some remarks as we see this feature being released for new and existing users:

Ideally, all clients should autostart from the beginning, the majority of users are fine with that and offer an easy way to always be connected.

With that in mind, we should update the configuration option introduced in the PR because the default value for bool in Go is `false`, so let's rename the configuration from `AutoStart` to `DisableAutoConnect`. Then, when loading from an existing configuration, `DisableAutoStart` will be `false`, keeping the current behavior of connecting on boot.

Other note, the `netbird service disable` might drive people into the wrong conclusion and assume that this disabled the system service all along. It is better for us to work with: `netbird service install --disable-auto-connect`. We can also add this flag to the `netbird up` command.

Let us know what do you think.

oskardotglobal:
Sure, I'll make the proposed naming changes.
Maybe the flag should be global?

oskardotglobal:
I've made the changes and moved the flag to the `up` command as I thought it'd be better since it's that way for the rosenpass flag

client/cmd/up.go
```diff
@@ -65,6 +65,26 @@ func upFunc(cmd *cobra.Command, args []string) error {
 		ctx = context.WithValue(ctx, system.DeviceNameCtxKey, hostName)
 	}
 
+	if cmd.Flag(disableAutoConnectFlag).Changed {
```
mlsmaycon: We need to move the flag check to the `runInForegroundMode` and `runInDaemonMode` functions.

mlsmaycon (changes requested):

mlsmaycon:
Thanks, @oskardotglobal; I've added one more note about where to add the set the flag.

mlsmaycon:
Thanks @oskardotglobal we will review it today and if tests goes ok it will be part of the 0.26 release

client/cmd/up.go
```diff
@@ -135,6 +147,25 @@ func runInDaemonMode(ctx context.Context, cmd *cobra.Command) error {
 		return err
 	}
 
+  if cmd.Flag(disableAutoConnectFlag).Changed {
```
mlsmaycon: @oskardotglobal for the daemon mode, you need to move the flag check to after the loginRequest initiallization:

https://github.com/netbirdio/netbird/pull/1161/files#diff-0db67eedcdec93665cc9dea08a99e38d7ed16f94db2e98f842af921340815542R206

And it should be similar to the rosenpassEnabled flag. 

This will require an update of the grpc protocol file with a new message field. See https://github.com/netbirdio/netbird/blob/main/client/proto/daemon.proto#L59

mlsmaycon (changes requested):
Validated the flag and functionality in the configuration file changes and it is working. 
We need to update it so that the change goes as part of the CLI request to the daemon process. If you are not sure about this change, let us know, we can help you out.

client/cmd/up.go
oskardotglobal: I updated the protocol but I can't seem to find what to do with the value

oskardotglobal: Where do I actually update the config after receiving the field inside the message?

oskardotglobal: Never mind, figured it out

mlsmaycon (approved):

mlsmaycon:
Thanks, @oskardotglobal, for the awesome contribution!

oskardotglobal:
To be honest, this has been bothering me ever since I started using netbird ;), but I'm still happy to help

PR #1190, opened 2023-10-02
Prevent return extra userData

mlsmaycon:
## Describe your changes

If there is a difference between local and cached data, we trigger a cache refresh; as we remove users from the local store and potentially from the remote IDP, we need to switch the source of truth to the local store to prevent unwanted endless cache for cases where the removal from the IDP fails or for cases where the userDeleteFromIDPEnabled got enabled after the first user deletion.

This also aligns with the change in the warmup method in PR #1147

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -637,8 +637,8 @@ func (a *Account) Copy() *Account {
 	}
 
 	routes := map[string]*route.Route{}
-	for id, route := range a.Routes {
-		routes[id] = route.Copy()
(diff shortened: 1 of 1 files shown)


braginini:
It is unclear what was the issue

mlsmaycon:
> It is unclear what was the issue , could you please describe in the ticket?

updated, have a look

braginini (approved):

PR #1193, opened 2023-10-04
Improve updateAccountPeers by bypassing AM and using account directly

surik:
### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


mlsmaycon (changes requested):
Removing the error check removes the need for this method to return an error altogether.

mlsmaycon (approved):

PR #1210, opened 2023-10-10
Add Pagination for IdP Users Fetch

bcmmbaga:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/idp/okta.go
+++ b/management/server/idp/okta.go
@@ -196,7 +196,7 @@ func (om *OktaManager) GetAllAccounts() (map[string][]*UserData, error) {
 	return indexedUsers, nil
 }
 
-// getAllUsers returns all users in an Okta account filtered by customer ID.
+// getAllUsers returns all users in an Okta account.
 func (om *OktaManager) getAllUsers() ([]*UserData, error) {
 	qp := query.NewQueryParams(query.WithLimit(200))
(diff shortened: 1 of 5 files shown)


management/server/idp/okta.go
```diff
@@ -169,57 +170,75 @@ func (om *OktaManager) GetAccount(accountID string) ([]*UserData, error) {
...
+				om.appMetrics.IDPMetrics().CountRequestStatusError()
+			}
+			return nil, fmt.Errorf("unable to get all accounts, statusCode %d", resp.StatusCode)
+		}
+
+		userList = append(userList, paginatedUsers...)
+	}
+
+	users := make([]*UserData, 0)
+	for _, user := range userList {
```
surik, line 232: We probably can minimize allocations:
```go
users := make([]*UserData, len(userList))
for i, user := range userList {
  userData, err := parseOktaUser(user)
  if err != nil {
      return nil, err
  }
  users[i] = userData
}
```

bcmmbaga: I allocated it as users capacity by creating a slice with an initial length of zero and a capacity equal to the length of the userList

`users := make([]*UserData, 0, len(userList))`

surik (approved):

PR #1211, opened 2023-10-11
Fix nil pointer exception in group delete

pappz:
## Describe your changes

Fix group delete panic

In case if in the db the DNSSettings is null then can cause panic in delete group function 
because this field is pointer and it was not checked. Because of in the future implementation 
this variable will be filled in any case then make no sense to keep the pointer type.

Fix DNSSettings copy function

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -179,7 +179,7 @@ type Account struct {
 	Policies               []*Policy
 	Routes                 map[string]*route.Route
 	NameServerGroups       map[string]*nbdns.NameServerGroup
-	DNSSettings            *DNSSettings
+	DNSSettings            DNSSettings
 	// Settings is a dictionary of Account settings
 	Settings *Settings
 }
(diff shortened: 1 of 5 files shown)


management/server/account.go
```diff
@@ -512,13 +512,11 @@ func (a *Account) getUserGroups(userID string) ([]string, error) {
...
-	if a.DNSSettings != nil {
-		for _, groupID := range a.DNSSettings.DisabledManagementGroups {
-			_, found := peerGroups[groupID]
-			if found {
-				enabled = false
-				break
-			}
+	for _, groupID := range a.DNSSettings.DisabledManagementGroups {
+		_, found := peerGroups[groupID]
+		if found {
```
surik, line 517: could be this `if _, found := peerGroups[groupID]; found {`

surik (approved):

management/server/account.go
pappz: What is the difference?

surik: just cosmetic based on my understanding, feel free to ignore

pappz: ok :)
For my eyes fit better this but I try to flexible in the future.

PR #1218, opened 2023-10-13
Fix update script's failure to update netbird-ui in binary installation

bcmmbaga:
## Describe your changes

Resolve the problem with the update script that prevents netbird-ui from updating during binary installation.

Introduce the variable UPDATE_NETBIRD. Now we can upgrade the binary installation with:
```
export UPDATE_NETBIRD=true
curl -fsSL https://pkgs.netbird.io/install.sh | sh
```

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


braginini:
NetBird UI doesn't stop when upgrade happens. Therefore, the currently running version is not updated and the user needs to restart it. Is this the expected behaviour? Should we send a stop signal?

bcmmbaga:
> NetBird UI doesn't stop when upgrade happens. Therefore, the currently running version is not updated and the user needs to restart it. Is this the expected behaviour? Should we send a stop signal?

Yes, for now, this is the expected behavior. The script is designed to only stop the NetBird service during upgrades. However, if it's necessary to close the UI as well, we can send a signal to do so

release_files/install.sh
```diff
@@ -284,14 +284,21 @@ is_bin_package_manager() {
   fi
 }
 
+stop_running_netbird_ui() {
+  if pgrep -x "netbird-ui" >/dev/null; then
```
mlsmaycon: is pgrep on most devices?

release_files/install.sh
```diff
@@ -284,14 +284,21 @@ is_bin_package_manager() {
...
+    echo "NetBird UI is running. Stopping it..."
+    pkill -x "netbird-ui"
```
mlsmaycon: same for pkill, let's use `ps ` and `grep` combination?

mlsmaycon (approved):

PR #1230, opened 2023-10-18
Make possible set IdpSignKeyRefreshEnabled from setup.env

Fantu:
## Describe your changes
IdpSignKeyRefreshEnabled is default to false but with some idps on token expire of logged users netbird always give error and return usable only on server restart so I think is useful make easier/faster set it on server configuration

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


mlsmaycon (approved):
Thanks @Fantu for your contribution

PR #1232, opened 2023-10-19
Feature/update check

pappz:
## Describe your changes

Periodically fetch the latest available version, and the UI will show a new menu for the download link. It checks both the daemon version and the UI version.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


version/update.go
```diff
@@ -0,0 +1,184 @@
...
+
+	content, err := io.ReadAll(resp.Body)
+	if err != nil {
+		log.Errorf("failed to read content: %s", err)
+		return false
+	}
+
+	lastAvailable, err := goversion.NewVersion(string(content))
+	if err != nil {
+		log.Errorf("faield to parse the version string: %s", err)
```
pascal-fischer: typo

version/update.go
```diff
@@ -0,0 +1,184 @@
...
+var (
+	versionURL = "https://pkgs.netbird.io/releases/latest/version"
+)
+
+// Update fetch the version info periodically and notify the onUpdateListener in case the UI version or the
+// daemon version are deprecated
+type Update struct {
+	uiVersion     *goversion.Version
+	daemonVersion *goversion.Version
+	lastAvailable *goversion.Version
```
pascal-fischer: I was confused in the beginning. Maybe we can rename the var to latestAvailable

version/update.go
```diff
@@ -0,0 +1,184 @@
...
+// SetDaemonVersion update the currently running daemon version. If new version is available it will trigger
+// the onUpdateListener
+func (u *Update) SetDaemonVersion(newVersion string) bool {
+	daemonVersion, err := goversion.NewVersion(newVersion)
+	if err != nil {
+		daemonVersion, _ = goversion.NewVersion("0.0.0")
+	}
+
+	u.versionsLock.Lock()
+	if u.daemonVersion != nil && u.daemonVersion.Equal(daemonVersion) {
```
pascal-fischer, line 74: I am not sure if it should be equal or grater than as we are setting it to 0.0.0.0 in case of error. Then override the current value with 0 and next time the current lates would be recognised as new

pappz: I do not understand well your case.
So for example if the daemon version is default "development" string, we set the version to "0.0.0". In the update check will indicate that new version is available. I think this is the expected behavior.
I use equal instead of grater then because we would like to handle the downgrade events also.

pascal-fischer (approved):

PR #1237, opened 2023-10-21
Support pinning version during binary install

glemsom:
## Support for specific version installation
For installations using the binary release method (using the official installer script), it would be nice to be able to define a specific version to install.

A user/developer can choose to define the `NETBIRD_RELEASE` variable during installation, to pin a specific version during installation. 
If `NETBIRD_RELEASE` is not defined, we default to the current behavior of `latest`

#### Example usage (once merged)
```
export SKIP_UI_APP=true; export NETBIRD_RELEASE=v0.23.7; curl https://pkgs.netbird.io/install.sh | sh
```

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/release_files/install.sh
+++ b/release_files/install.sh
@@ -44,7 +44,7 @@ get_release() {
 }
 
 download_release_binary() {
-    VERSION=$(get_release $NETBIRD_RELEASE)
+    VERSION="$(get_release $NETBIRD_RELEASE)"
     BASE_URL="https://github.com/${OWNER}/${REPO}/releases/download"
     BINARY_BASE_NAME="${VERSION#v}_${OS_TYPE}_${ARCH}.tar.gz"
 
--- a/release_files/install.sh
+++ b/release_files/install.sh
@@ -23,19 +23,28 @@ if command -v sudo > /dev/null && [ "$(id -u)" -ne 0 ]; then
     SUDO="sudo"
 fi
 
-get_latest_release() {
+if [ -z ${NETBIRD_RELEASE+x} ]; then 
+    NETBIRD_RELEASE=latest
+fi
+
+get_release() {
+    local RELEASE=$1
+    if [ "$RELEASE" == "latest" ]; then
+        local TAG="latest"
+    else
+        local TAG="tags/${RELEASE}"
+    fi
     if [ -n "$GITHUB_TOKEN" ]; then
-          curl -H  "Authorization: token ${GITHUB_TOKEN}" -s "https://api.github.com/repos/${OWNER}/${REPO}/releases/latest" \
+          curl -H  "Authorization: token ${GITHUB_TOKEN}" -s "https://api.github.com/repos/${OWNER}/${REPO}/releases/${TAG}" \
               | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'
     else
-          curl -s "https://api.github.com/repos/${OWNER}/${REPO}/releases/latest" \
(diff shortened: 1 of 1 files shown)


mlsmaycon:
Hello @glemsom, thanks for your contribution. 

Please check the annotations from Codacy regarding double quotes and we should be good to go.

release_files/install.sh
```diff
@@ -23,19 +23,28 @@ if command -v sudo > /dev/null && [ "$(id -u)" -ne 0 ]; then
...
-          curl -s "https://api.github.com/repos/${OWNER}/${REPO}/releases/latest" \
+          curl -s "https://api.github.com/repos/${OWNER}/${REPO}/releases/${TAG}" \
               | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'
     fi
-
 }
 
 download_release_binary() {
-    VERSION=$(get_latest_release)
+    VERSION="$(get_release $NETBIRD_RELEASE)"
```
mlsmaycon:
```suggestion
    VERSION=$(get_release "$NETBIRD_RELEASE")
```

mlsmaycon: @glemsom can you apply the suggestion above?

glemsom: @mlsmaycon  Sure :)
Fixed in e7f70fc88b489f38df4d4f652fe3c2a9de659341

release_files/install.sh
```diff
@@ -23,19 +23,28 @@ if command -v sudo > /dev/null && [ "$(id -u)" -ne 0 ]; then
     SUDO="sudo"
 fi
 
-get_latest_release() {
+if [ -z ${NETBIRD_RELEASE+x} ]; then
```
mlsmaycon: This seems to be failing. Can you try the following?

```suggestion
if [ "x-$NETBIRD_RELEASE" = "x-" ]; then 
```

release_files/install.sh
```diff
@@ -23,19 +23,28 @@ if command -v sudo > /dev/null && [ "$(id -u)" -ne 0 ]; then
...
 fi
 
-get_latest_release() {
+if [ -z ${NETBIRD_RELEASE+x} ]; then 
+    NETBIRD_RELEASE=latest
+fi
+
+get_release() {
+    local RELEASE=$1
+    if [ "$RELEASE" == "latest" ]; then
```
mlsmaycon:
```suggestion
    if [ "$RELEASE" = "latest" ]; then
```

glemsom: Sorry for the late reply.
You're right. I only tested my change with bash.
I read through the workflows, and see we test using ubuntu:latest, which default to dash as shell - which in return does not support the "double equal-sign" that bash does.
Sorry about that !

_(I've now imported your validation actions to my fork)_  :smile: 

We should for sure keep the validation action as-is, as it mimic the current recommended one-line installation behavior on the website. _(I should just have tested my change better!)_

This is now corrected to be POSIX compatible. (efc8164bae8325a6cb2937350fde2e268df56dd7)
I imported the validation workflows to my fork - seems to go through now :)

braginini:
@glemsom 
I'm not coming here with a review of the PR but just wanted to personally say Thank You for the contribution :) The whole NetBird team appreciates it!

mlsmaycon (approved):
Thank you @glemsom, that's a great improvement.

glemsom:
You are welcome. And thanks for an awesome product  :)

PR #1244, opened 2023-10-23
Feature/add iOS support

pascal-fischer:
## Describe your changes
This PR includes the NetBirdSDK that can be converted into an objective-c framework using gomobile and all the necessary changes to integrate iOS with our existing go code.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/engine.go
```diff
@@ -205,8 +205,7 @@ func (e *Engine) Start() error {
 			go e.mobileDep.DnsReadyListener.OnReady()
 		}
 	} else if e.dnsServer == nil {
-		// todo fix custom address
-		e.dnsServer, err = dns.NewDefaultServer(e.ctx, e.wgInterface, e.config.CustomDNSAddress)
+		e.dnsServer, err = dns.NewDefaultServer(e.ctx, e.wgInterface, e.config.CustomDNSAddress, e.mobileDep.InterfaceName, wgAddr)
```
pappz: I do not think so we should pass the e.mobileDep.InterfaceName and wgAddr. The e.wgInterface knows these informations. But regarding the interface name in the engine config and mobileDep will be different name what is incorrect. So remove the interfaceName from the MobileDependency and overwrite the config.WgIface (engine config) in the connect.go:Run function

client/internal/dns/server.go
```diff
@@ -109,14 +118,16 @@ func newDefaultServer(ctx context.Context, wgInterface WGIface, dnsService servi
...
+		interfaceName: interfaceName,
+		wgAddr:        wgAddr,
 	}
 
 	return defaultServer
 }
 
 // Initialize instantiate host manager and the dns service
-func (s *DefaultServer) Initialize() (err error) {
+func (s *DefaultServer) Initialize(manager IosDnsManager) (err error) {
```
pappz: In general to pass a nil to an exported function is bad practice. I do not know when can I set nil. And looks like that it depend from runtime (GOOS) variable when it will be used.  Hard to follow the logic when you try to write unit test.
So instead of a new function parameter just create a new constructor for the DefaultServer. We already have 2 because of the Android and feel free to add an iOS related.

iface/tun_ios.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (t *tunDevice) Close() (err error) {
+	if t.device != nil {
+		t.device.Close()
+	}
+
+	return
+}
+
+func (t *tunDevice) routesToString(routes []string) string {
```
pappz: Looks unused function

pappz (changes requested):

client/internal/dns/server.go
```diff
@@ -19,9 +20,14 @@ type ReadyListener interface {
 	OnReady()
 }
 
+// IosDnsManager is a dns manager interface for iosß
```
pappz: Typo: iosß

client/internal/dns/server.go
```diff
@@ -131,7 +142,11 @@ func (s *DefaultServer) Initialize() (err error) {
 		}
 	}
 
-	s.hostManager, err = newHostManager(s.wgInterface)
+	if runtime.GOOS == "ios" {
```
pappz: I am not sure in that make sense for this condition. You can call the newHostManager with s.wgInterface, manager arguments. The proper implementations of the host managers will handle the nil things.

client/ios/NetBirdSDK/client.go
```diff
@@ -0,0 +1,223 @@
+package NetBirdSDK
```
pappz, line 1: Good package names are short and clear. They are lower case, with no under_scores or mixedCaps. They are often simple nouns. https://go.dev/blog/package-names

client/internal/engine.go
pascal-fischer: I had some trouble with getting data from the wgInterface that it was empty. I will try to overwrite it

client/internal/dns/server.go
pascal-fischer: Yeah I didn't like that either. I will have a look at Android and see if I can do similar for iOS

client/internal/dns/server.go
pascal-fischer: Agree

client/ios/NetBirdSDK/client.go
pascal-fischer: I know but for this package I changed the name to that format as this is how the gomobile then names the framework on Swift and I followed the Swift convention as it seemed to be the lesser problem. The iOS package in go will never be referenced but the Swift package will be referenced throughout the whole app.

client/internal/dns/upstream_test.go
```diff
@@ -49,15 +49,15 @@ func TestUpstreamResolver_ServeDNS(t *testing.T) {
 			timeout:             upstreamTimeout,
 			responseShouldBeNil: true,
 		},
-		//{
```
pappz, line 52: If these scenarios are not valid anymore then please remove from the code. The git will remember for it.

client/system/info_ios.go
```diff
@@ -0,0 +1,45 @@
...
+
+// GetInfo retrieves and parses the system information
+func GetInfo(ctx context.Context) *Info {
+
+	// Convert fixed-size byte arrays to Go strings
+	sysName := extractOsName(ctx, "sysName")
+	swVersion := extractOsVersion(ctx, "swVersion")
+
+	gio := &Info{Kernel: sysName, OSVersion: swVersion, Core: swVersion, Platform: "unknown", OS: sysName, GoOS: runtime.GOOS, CPUs: runtime.NumCPU()}
+	// systemHostname, _ := os.Hostname()
```
pappz: Please remove the unused line

pascal-fischer: Done

client/internal/dns/upstream_test.go
pascal-fischer: I didn't even notice, goland just fixed the syntax but I agree. I will remove it

iface/tun_ios.go
pascal-fischer: Removed

client/internal/dns/mockServer.go
pappz, line 1: Could you rename the mockServer.go to snake_case

client/internal/engine.go
```diff
@@ -207,25 +208,34 @@ func (e *Engine) Start() error {
...
 			go e.mobileDep.DnsReadyListener.OnReady()
 		}
-	} else if e.dnsServer == nil {
-		// todo fix custom address
-		e.dnsServer, err = dns.NewDefaultServer(e.ctx, e.wgInterface, e.config.CustomDNSAddress)
-		if err != nil {
-			e.close()
-			return err
+	case "ios":
+		if e.dnsServer == nil {
```
pappz, line 212: Do you see the reason to check the e.dnsServer == nil? We do not check any other variables like this (routeManager, firewall, wginterface, etc...). But the problem is that the Engine code do not handle multiple invocation of the exported Start() function.

iface/wg_configurer_ios.go
```diff
@@ -0,0 +1,165 @@
...
+	"net"
+	"strings"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
+)
+
+var (
+	errFuncNotImplemented = errors.New("function not implemented")
```
pappz: unused variable

iface/wg_configurer_ios.go
pappz, line 1: This file looks fully compatible with the wg_configurer_android.go. Could you merge it?

iface/ipc_parser_ios.go
pappz, line 1: Looks identical with ipc_perser_android.go

client/internal/dns/upstream.go
```diff
@@ -35,20 +36,50 @@ type upstreamResolver struct {
 	mutex            sync.Mutex
 	reactivatePeriod time.Duration
 	upstreamTimeout  time.Duration
+	lIP              net.IP
+	lNet             *net.IPNet
```
pappz: Looks like unused variable

client/internal/dns/upstream.go
```diff
@@ -35,20 +36,50 @@ type upstreamResolver struct {
...
+	lName            string
```
pappz: Looks like unused variable

client/internal/dns/upstream.go
```diff
@@ -35,20 +36,50 @@ type upstreamResolver struct {
...
 
 	deactivate func()
 	reactivate func()
 }
 
-func newUpstreamResolver(parentCTX context.Context) *upstreamResolver {
+func getInterfaceIndex(interfaceName string) (int, error) {
+	iface, err := net.InterfaceByName(interfaceName)
+	if err != nil {
+		log.Errorf("unable to get interface by name error: %s", err)
```
pappz: Duplicated error message. We will print this err on line 69

client/internal/dns/upstream.go
```diff
@@ -35,20 +36,50 @@ type upstreamResolver struct {
...
+
+func newUpstreamResolver(parentCTX context.Context, interfaceName string, wgAddr string) *upstreamResolver {
 	ctx, cancel := context.WithCancel(parentCTX)
+
+	// Specify the local IP address you want to bind to
+	localIP, localNet, err := net.ParseCIDR(wgAddr) // Should be our interface IP
+	if err != nil {
+		log.Errorf("error while parsing CIDR: %s", err)
+	}
+	index, err := getInterfaceIndex(interfaceName)
```
pappz: If detect the error here why do we allow to run forward the code? Without these variables is the iOS app could works well?

client/internal/dns/upstream.go
pascal-fischer: removed

client/internal/dns/mockServer.go
pascal-fischer: done

pappz (dismissed):

client/internal/dns/upstream_ios.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	ios.upstreamClient = ios
+
+	return ios, nil
+}
+
+func (u *upstreamResolverIOS) exchange(upstream string, r *dns.Msg) (rm *dns.Msg, t time.Duration, err error) {
+	client := &dns.Client{}
+	upstreamHost, _, err := net.SplitHostPort(upstream)
+	if err != nil {
```
pappz, line 46: Incorrect error handling. If err != nil then we should not use upstreamHost.

pascal-fischer: The logic is only checking what client to use. If we should use private or default client for the exchange. If the upstream can not be split properly the default client will be used and the exchange will fail anyway. But I can add an explicit return to force a failure if we can't detect what client to use

pappz (approved):

PR #1268, opened 2023-10-30
Extends management  user and group structure

bcmmbaga:
## Describe your changes

Extends the user and group structure in management to include additional fields `Issued` and `IntegrationReference`.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/group.go
```diff
@@ -34,6 +34,8 @@ type Group struct {
 
 	// Peers list of the group
 	Peers []string `gorm:"serializer:json"`
+
+	IntegrationReference IntegrationReference `gorm:"serializer:json"`
```
surik: I suggest we use [gorm embedded struct](https://gorm.io/docs/models.html#Embedded-Struct) here. Then we have the ability to search by the field inside. 

```go
IntegrationReference IntegrationReference `gorm:"embedded;embeddedPrefix:integration_ref_"
```

We need to change the `ID` field of `IntegrationReference` to something else otherwise we may have similar issues as [here](https://github.com/go-gorm/gorm/issues/6517)

management/server/user.go
```diff
@@ -59,6 +72,11 @@ type User struct {
 	Blocked bool
 	// LastLogin is the last time the user logged in to IdP
 	LastLogin time.Time
+
+	// Issued of the user
+	Issued string
```
surik: Let's think about setting a default value for this newly added field. For `sqlite` we can do it via gorm by adding this tag `gorm:"default:api"` to the field. For `filestore` this more likely should happen when we read the file on startup.

Also as we add a new field should we reflect it in [the OpenAPI spec](https://github.com/netbirdio/netbird/blob/main/management/server/http/api/openapi.yml)?

Also, it may make sense to extend `users_handler_test.go` to check that the field is properly handled.

management/server/http/users_handler.go
```diff
@@ -77,6 +83,7 @@ func (h *UsersHandler) UpdateUser(w http.ResponseWriter, r *http.Request) {
 		Role:       userRole,
 		AutoGroups: req.AutoGroups,
 		Blocked:    req.IsBlocked,
+		Issued:     existingUser.Issued,
```
surik: We want to keep `IntegrationReference` from existing user. The same is true for the group handler.

surik (changes requested):

management/server/user.go
```diff
@@ -270,9 +296,11 @@ func (am *DefaultAccountManager) inviteNewUser(accountID, userID string, invite
...
 	role := StrRoleToUserRole(invite.Role)
 	newUser := &User{
-		Id:         idpUser.ID,
-		Role:       role,
-		AutoGroups: invite.AutoGroups,
+		Id:                   idpUser.ID,
+		Role:                 role,
+		AutoGroups:           invite.AutoGroups,
+		Issued:               invite.Issued,
+		IntegrationReference: invite.IntegrationReference,
```
pappz, line 301: In case of CreateUser **user.go:236** API endpoint the UserInfo does not contain the IntegrationReference but here you add it to the User structure. Is it correct?

management/server/user.go
bcmmbaga: Yes, issues is added to the user api response

management/server/user.go
bcmmbaga: Yes, it is correct because the CreateUser method will also be called by integration

surik (approved):

PR #1270, opened 2023-10-30
Parse original resolv.conf

pappz:
Handle original search domains in resolv.conf type implementations.

- parse the original resolv.conf file
- merge the search domains
- ignore the domain keyword
- append any other config lines (sortstlist, options)
- fix read origin resolv.conf from bkp in resolvconf implementation
- fix line length validation
- fix number of search domains validation

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/internal/dns/file_linux.go
```diff
@@ -138,15 +112,108 @@ func (f *fileConfigurator) restore() error {
...
+		if len(listOfDomains) >= fileMaxNumberOfSearchDomains {
+			// lets log all skipped domains
+			log.Infof("already appended %d domains to search list. Skipping append of %s domain", fileMaxNumberOfSearchDomains, dConf.domain)
+			continue
+		}
+		if charCount > fileMaxLineCharsLimit {
+			// lets log all skipped domains
+			log.Infof("search list line is larger than %d characters. Skipping append of %s domain", fileMaxLineCharsLimit, dConf.domain)
+			continue
+		}
```
pascal-fischer: The current implementation does not account for already existing searchdomains in our file so the count only counts of what we add. @mlsmaycon and I where thinking if the limits make sense or if we can drop them. I tested on my test machine with exceeding the limits and it still worked file.

pascal-fischer (approved):

pascal-fischer (approved):

PR #1278, opened 2023-11-01
fix some typo spotted with codespell

Fantu:
I skipped all results related to code that must not be changed, there is only change in management/server/http/accounts_handler_test.go which I have some doubts about

### Checklist
- [X] Is it a bug fix
- [ ] It is a refactor


Fantu:
"Test Code Windows" workflow failed on "Download wintun" for internal server error, same error trying on manually download so I suppose shold be started again when wintun download return working

mlsmaycon:
This is awesome @Fantu could you please add a github action workflow for codespell? 

https://github.com/codespell-project/actions-codespell

It can be added in another job as part of the workflow https://github.com/netbirdio/netbird/blob/main/.github/workflows/golangci-lint.yml

mlsmaycon:
> "Test Code Windows" workflow failed on "Download wintun" for internal server error, same error trying on manually download so I suppose shold be started again when wintun download return working

It was a temporary server-side issue. After restarting the flow it is running again

Fantu:
@mlsmaycon can you check the changes in management/server/http/accounts_handler_test.go which I have some doubts about?
about codespell there was many results skipped after check in source code, probably I can add some exclude to make working for actual scan but there may be new results in the future which however are part of the code (and which also should not be modified) which will cause the test to fail
I don't know if possible do a workflow con fail "tolerated"
or do you want to make sure that the PRs that make it fail for results to be excluded also update the codespell workflow?

mlsmaycon:
> @mlsmaycon can you check the changes in management/server/http/accounts_handler_test.go which I have some doubts about? about codespell there was many results skipped after check in source code, probably I can add some exclude to make working for actual scan but there may be new results in the future which however are part of the code (and which also should not be modified) which will cause the test to fail I don't know if possible do a workflow con fail "tolerated" or do you want to make sure that the PRs that make it fail for results to be excluded also update the codespell workflow?

I will check the management checks. 

Regarding codespell workflow, in that case is better to do in another PR and we can check what is breaking and why.

mlsmaycon (approved):

mlsmaycon:
Looks good, thanks @Fantu

mlsmaycon:
@Fantu I've created the following PR with the job: https://github.com/netbirdio/netbird/pull/1281

PR #1281, opened 2023-11-01
Add codespell job

mlsmaycon:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -1,2 +1,2 @@
-ERRO
-clientA
+ERROR
+client
--- a/.codespellignore
+++ b/.codespellignore
@@ -1,2 +1,0 @@
-erro
-clienta
--- a/.codespellignore
+++ b/.codespellignore
@@ -1,2 +1,2 @@
-ERROR
-client
+erro
+clienta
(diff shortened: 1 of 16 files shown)


Fantu:
@mlsmaycon I saw
```
with:
  only_warn: 1
```
that make job not failing but only report problem found, I think should be added so will not result in fail for result that should be added in exclusion

Fantu:
@mlsmaycon the continue error I suppose can be for:
`Words are case sensitive based on **how they are written in the dictionary file**.`
so for:
`clientA ==> client`
if I understand good should be excluded "client" but if correct this seems to me bad thing that don't permit exclusion check of result instead (or in alternative)

mlsmaycon:
> @mlsmaycon the continue error I suppose can be for:
> 
> > Words are case sensitive based on **how they are written in the dictionary file**.
> 
> so for: `clientA ==> client` if I understand good should be excluded "client" but if correct this seems to me bad thing that don't permit exclusion check of result instead (or in alternative)

trying that, thanks for the tip

Fantu:
still fails, so seems there is a bug that should be reported
in the meantime I suggest you put
`only_warn: 1`

Fantu:
try erro and clienta, all lower case

surik (approved):

PR #1311, opened 2023-11-15
Add non-deletable service user

bcmmbaga:
## Describe your changes

Introduce a non-deletable service user, which can only be created by invoking the AccountManager's `CreateUser` method rather than through API calls. Additionally, ensure the exclusion of these users from the list of retrieved account service users in API calls.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/http/users_handler.go
+++ b/management/server/http/users_handler.go
@@ -207,7 +207,7 @@ func (h *UsersHandler) GetAllUsers(w http.ResponseWriter, r *http.Request) {
 			util.WriteError(status.Errorf(status.InvalidArgument, "invalid service_user query parameter"), w)
 			return
 		}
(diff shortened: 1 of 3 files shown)


management/server/http/users_handler.go
```diff
@@ -198,7 +198,9 @@ func (h *UsersHandler) GetAllUsers(w http.ResponseWriter, r *http.Request) {
 	users := make([]*api.User, 0)
 	for _, r := range data {
```
surik, line 199: I wonder if we can only have one check like this:

```go
if r.NonDeletable {
	continue
}
```

And do not check for `!r.NonDeletable` below at all

surik (approved):

Issue #1316, opened 2023-11-17
Add “official” systemd units?


hg:
Hello,

netbird has been [maintained](https://aur.archlinux.org/packages/netbird) in the Arch Linux User repository since 2021, and recently Arch developers [have expressed](https://aur.archlinux.org/packages/netbird#comment-944094) interest in packaging it in the official repository.

Would you be willing to merge hand-rolled systemd `.service` files instead of (or in addition to) `netbird service install`?

- the client config is a templated unit which allows the user to create & start multiple client instances by using `systemctl enable --now netbird@foo netbird@bar netbird@baz` (the part after the `@` matches the name of the config file in `/etc/netbird`)
- all units provide partial sandboxing from the rest of the system (#378, although netbird pretty much requires root or `CAP_NET_ADMIN` these days, they were much stricter in the past). The flags have been there for years and are pretty well tested by now.
- they can be used by other Linux distributions as the project gets more adoption

The files are [here](https://github.com/hg/aur/tree/master/netbird). I can prepare a PR if you're interested, or you can just copy them from there.

If you'd rather stay with `netbird service install`, Arch (and others in the future) will probably ship their own units.

Thanks.

mlsmaycon:
Hello @hg this is awesome. We are interested in the PR. 

We are organizing the packaging files in the [release_files](https://github.com/netbirdio/netbird/tree/main/release_files) path, but we are open to suggestions to make it clear. 

Thanks for reaching out and for maintaining the package in the Arch Linux User repository.

lfarkas:
i also vote for a separate service file not the generated one. it'd be much easier on all distro to use the same official service file!

Issue #1079, opened 2023-08-12
systemd unit file improvements


sebastianschauenburg:
**Describe the problem**
After a reboot, netbird fails to start using the systemd unit file ```netbird.service```. Example of an error which is visible:

    aug 11 12:48:14 hostname systemd[7480]: netbird.service: Failed to set up standard output: No such file or directory

I'm using a Ubuntu 23.04 system with netbird 0.22.3 (installed via apt).

**To Reproduce**
1. Set up /var/log as RAM log with (for example) [log2ram](https://github.com/azlux/log2ram).
2. Reboot the system
3. View the error logs

**Expected behavior**
No errors and a working netbird interface

**Additional context**

```/var/log/netbird``` is created at a certain point, but was not synced to the persistent /var/log (if available). So effectively, after a reboot, the ```/var/log/netbird``` directory does not exist. This in turn means it cannot be used for stdout and stderr.
I tried creating it with an ```ExecStartPre``` using ```mkdir``` inside the unit file, but attaching the stdout and stderr apparently happens before that point and as such, errors with:

    aug 11 12:49:36 hostname systemd[8997]: netbird.service: Failed at step STDOUT spawning /usr/bin/mkdir: No such file or directory

**Proposed solution**
__Add__ this to the unit file, which ensures ```client.log``` can be created / updated:

    ExecStartPre=/usr/bin/mkdir /var/log/netbird

__Remove__ the following lines, since that output should be captured inside of systemd (journalctl) anyway:

    StandardOutput=file:/var/log/netbird/netbird.out
    StandardError=file:/var/log/netbird/netbird.err

nazarewk:
> For future reference: the canonical way to fix this is to add:
> 
> ```ini
> CacheDirectory=netbird
> ConfigurationDirectory=netbird
> LogDirectory=netbird
> RuntimeDirectory=netbird
> StateDirectory=netbird
> ```
> 
> under the `[Service]` section.
> 
> systemd will then make sure to create the appropriate directories and give your service full access to them _before_ starting the daemon:
> 
>     * `/var/cache/netbird`
> 
>     * `/etc/netbird`
> 
>     * `/var/log/netbird`
> 
>     * `/var/run/netbird`
> 
>     * `/var/lib/netbird`
> 
> 
> It's not immediately obvious how to add this to `netbird service install` since netbird relies on a third-party (not really) package for its service management, and the `Option` field I see there does not "accept" these keys.

Note there is no `LogDirectory` in systemd

[...]

PR #1318, opened 2023-11-17
systemd units for the client, management & signal services

hg:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

---

Continuing from #1316.

### install & upgrade

I did not change `install.sh` yet as this PR uses a templated systemd unit for the client, and thus `install.sh` will need a bit of work to make it safely migrate old clients. I'd like to know first if you want this at all.

### logging

Logs get written to the console where they get picked up by journald, since this is where you would typically expect to see them on a systemd-based Linux distribution. Users get automatic compression and log rotation for free.

It's trivial to make each client use a separate log file if you'd rather keep using text logs, though.

### sanbox

Sandbox flags are quite conservative: they mostly deny access to home directories and things netbird shouldn't have to worry about (like changing mount points).

It isn't difficult to make all daemons *appear* to work with a non-root user and somewhat limited privileges (with `CAP_NET_ADMIN`), but I am not sure that it doesn't introduce some subtle breakage somewhere. It would probably make a bad default for now.

### customization

`env` should be copied into `/etc/default/netbird{,-management,-signal}` and is the same for all three. It's there to allow the user to pass additional configuration flags without overriding the unit.

### directories

These directives:

```ini
RuntimeDirectory=netbird
StateDirectory=netbird
LogDirectory=netbird
ConfigurationDirectory=netbird
```

create `/etc/netbird` and `/var/{log,lib,run}/netbird` (if necessary), make sure the daemon can access them, and only then start the daemon. This will prevent problems such as #1079 for those users who prefer and configure text logs. If someday netbird is made to run under a restricted user, you won't have to `chown -R` them in upgrade scripts since systemd will do it for you.

What do you think?

### DIFF
--- a/release_files/systemd/env
+++ b/release_files/systemd/env
@@ -1,0 +1,3 @@
+# Extra flags you might want to pass to the daemon
+FLAGS=""
+
--- a/release_files/systemd/netbird-management.service
+++ b/release_files/systemd/netbird-management.service
@@ -1,0 +1,41 @@
+[Unit]
+Description=Netbird Management
+Documentation=https://netbird.io/docs
+After=network-online.target syslog.target
+Wants=network-online.target
+
+[Service]
+Type=simple
+EnvironmentFile=-/etc/default/netbird-management
+ExecStart=/usr/bin/netbird-mgmt management $FLAGS
+Restart=on-failure
+RestartSec=5
+TimeoutStopSec=10
+CacheDirectory=netbird
+ConfigurationDirectory=netbird
+LogDirectory=netbird
+RuntimeDirectory=netbird
+StateDirectory=netbird
+
+# sandboxing
+LockPersonality=yes
+MemoryDenyWriteExecute=yes
+NoNewPrivileges=yes
+PrivateMounts=yes
+PrivateTmp=yes
+ProtectClock=yes
+ProtectControlGroups=yes
+ProtectHome=yes
+ProtectHostname=yes
+ProtectKernelLogs=yes
+ProtectKernelModules=yes
+ProtectKernelTunables=yes
+ProtectSystem=yes
+RemoveIPC=yes
+RestrictNamespaces=yes
+RestrictRealtime=yes
+RestrictSUIDSGID=yes
+
+[Install]
+WantedBy=multi-user.target
+
--- a/release_files/systemd/netbird-signal.service
+++ b/release_files/systemd/netbird-signal.service
@@ -1,0 +1,41 @@
+[Unit]
+Description=Netbird Signal
+Documentation=https://netbird.io/docs
+After=network-online.target syslog.target
+Wants=network-online.target
+
+[Service]
+Type=simple
+EnvironmentFile=-/etc/default/netbird-signal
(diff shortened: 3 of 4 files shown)


mlsmaycon:
Hello @hg thanks for the contribution and description. 

Here are the answers for your questions:
> ### install & upgrade
> I did not change `install.sh` yet as this PR uses a templated systemd unit for the client, and thus `install.sh` will need a bit of work to make it safely migrate old clients. I'd like to know first if you want this at all.

By default, the install.sh is using the Arch User repository. We don't have experience with how we can make a transition from an AUR installation to the extras repo. I would prefer to update the install.sh once the package is published. 

Regarding the binary install mode, we could keep it as is because it is, by definition, a custom kind of installation.

> ### logging
> Logs get written to the console where they get picked up by journald, since this is where you would typically expect to see them on a systemd-based Linux distribution. Users get automatic compression and log rotation for free.
> 
> It's trivial to make each client use a separate log file if you'd rather keep using text logs, though.

The only problem I can see with this is that only Arch (for now) would be using journald for logs; overall, community members would have to keep that in mind when helping others to troubleshoot issues.  On the other hand, the current log system used in Netbird has compression and rotation capabilities for application logs but doesn't cover stdout or stderr logs. 

Perhaps making the `--log-file console` configurable with the env file would be a good option.

> ### sanbox
> Sandbox flags are quite conservative: they mostly deny access to home directories and things netbird shouldn't have to worry about (like changing mount points).

> It isn't difficult to make all daemons _appear_ to work with a non-root user and somewhat limited privileges (with `CAP_NET_ADMIN`), but I am not sure that it doesn't introduce some subtle breakage somewhere. It would probably make a bad default for now.

We are looking into adding some session and certificate capabilities and other checks that might need access to a user's home directory, but it won't be an issue for us to check if the software gets affected by the current settings in the templates.

Another point, since 0.20 we added support to eBPF programs and Raw sockets; for newer kernels on Ubuntu, we could use `CAP_NET_RAW` and `CAP_BPF`. But we saw that it didn't work in other systems, so in our docs, we decided to add SYS_ADMIN and SYS_RESOURCE to cover more systems. 

> ### customization
> `env` should be copied into `/etc/default/netbird{,-management,-signal}` and is the same for all three. It's there to allow the user to pass additional configuration flags without overriding the unit.

This is great, thanks.

> ### directories
> These directives:
> 
> ```ini
> RuntimeDirectory=netbird
> StateDirectory=netbird
> LogDirectory=netbird
> ConfigurationDirectory=netbird
> ```

> create `/etc/netbird` and `/var/{log,lib,run}/netbird` (if necessary), make sure the daemon can access them, and only then start the daemon. This will prevent problems such as #1079 for those users who prefer and configure text logs. If someday netbird is made to run under a restricted user, you won't have to `chown -R` them in upgrade scripts since systemd will do it for you.

Excellent, thanks for the tip.

Let me know what do you think about the answers.

mlsmaycon (approved):
Thanks @hg for the contribution and notes. I think this is a good start to decouple the packaging and the binary.

PR #1331, opened 2023-11-27
Add account deletion endpoint

mlsmaycon:
## Describe your changes
Adding support to account owners to delete an account

This will remove all users from local, and if `--user-delete-from-idp` is set it will remove from the remote IDP

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -64,6 +64,7 @@ type AccountManager interface {
 	GetAccountByUserOrAccountID(userID, accountID, domain string) (*Account, error)
 	GetAccountFromToken(claims jwtclaims.AuthorizationClaims) (*Account, *User, error)
 	GetAccountFromPAT(pat string) (*Account, *User, *PersonalAccessToken, error)
+	DeleteAccount(accountID, userID string) error
 	MarkPATUsed(tokenID string) error
 	GetUser(claims jwtclaims.AuthorizationClaims) (*User, error)
 	ListUsers(accountID string) ([]*User, error)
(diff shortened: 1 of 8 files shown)


management/server/sqlite_store.go
```diff
@@ -202,6 +203,41 @@ func (s *SqliteStore) SaveAccount(account *Account) error {
...
+		if result.Error != nil {
+			return result.Error
+		}
+
+		result = tx.Select(clause.Associations).Delete(account)
+		if result.Error != nil {
+			return result.Error
+		}
+
+		result = tx.Session(&gorm.Session{FullSaveAssociations: true})
```
surik: This is not needed as the deletion happened on line 220

Also, we can have a helper function with the set of Delete calls as above. Then the same function can be called from SaveAccount. Having this in one place makes the future modification easier.

management/server/account.go
```diff
@@ -1004,6 +1005,57 @@ func (am *DefaultAccountManager) warmupIDPCache() error {
...
+	for _, otherUser := range account.Users {
+		if otherUser.IsServiceUser {
+			continue
+		}
+
+		if otherUser.Id == userID {
+			continue
+		}
+
+		deleteUserErr := am.deleteRegularUser(account, userID, otherUser.Id)
```
surik, line 1038: Can we get a situation when going through `account.Users` a few users got deleted but then on attempt to delete another one got an error. Will the follow-up call of `DeleteAccount` handle this properly?

mlsmaycon: Yes, this should be the case. The operation will halt, and the user can try to delete it again. This time, they will have fewer users to delete.

Another point is that we can't rollback the account state as we might have invited users with their unique IDs.

management/server/sqlite_store.go
mlsmaycon: true, I've removed the extra call.

surik (approved):

PR #1340, opened 2023-11-29
add owner role support

mlsmaycon:
## Describe your changes
This PR adds support to Owner roles. 

The owner role has a similar access level as the admin, but it has the power to delete the account.

Besides that, the role has the following constraints:
- The role can only be transferred. So, only a user with the owner role can transfer the owner role to a new user
- It can't be assigned to users being invited
- It can't be assigned to service users

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/group_test.go
+++ b/management/server/group_test.go
@@ -57,7 +57,7 @@ func TestDefaultAccountManager_DeleteGroup(t *testing.T) {
 		{
 			"integration",
 			"grp-for-integration",
-			"only admins service user can delete integration group",
+			"only service users with admin power can delete integration group",
 		},
 	}
 


management/server/user.go
```diff
@@ -99,7 +102,7 @@ func (u *User) LastDashboardLoginChanged(LastLogin time.Time) bool {
 
 // IsAdmin returns true if the user is an admin, false otherwise
```
surik: Maybe doc string has to be updated to mention owner role too

management/server/user.go
```diff
@@ -207,10 +215,14 @@ func (am *DefaultAccountManager) createServiceUser(accountID string, initiatorUs
 	if executingUser == nil {
 		return nil, status.Errorf(status.NotFound, "user not found")
 	}
-	if executingUser.Role != UserRoleAdmin {
+	if !executingUser.IsAdmin() {
 		return nil, status.Errorf(status.PermissionDenied, "only admins can create service users")
```
surik: I don't know if we consider the owner as a subset of the admin role. But maybe the error log needs to be adjusted,

management/server/user.go
```diff
@@ -416,7 +431,7 @@ func (am *DefaultAccountManager) DeleteUser(accountID, initiatorUserID string, t
 	if executingUser == nil {
 		return status.Errorf(status.NotFound, "user not found")
 	}
-	if executingUser.Role != UserRoleAdmin {
+	if !executingUser.IsAdmin() {
 		return status.Errorf(status.PermissionDenied, "only admins can delete users")
```
surik: Maybe the error needs adjustment as IsAdmin() return true for owner

PR #1350, opened 2023-12-01
Fix IPv6 format for DNS address received from android

mlsmaycon:
## Describe your changes

this adds the address in the expected format in Go [ipv6]:port

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/dns/server.go
+++ b/client/internal/dns/server.go
@@ -421,14 +421,7 @@ func (s *DefaultServer) updateLocalResolver(update map[string]nbdns.SimpleRecord
 }
 
 func getNSHostPort(ns nbdns.NameServer) string {
-	log.Debugf("getNSHostPort: %s : %d", ns.IP.String(), ns.Port)
-	if ns.IP.Is4() {
-		return fmt.Sprintf("%s:%d", ns.IP.String(), ns.Port)
-	}
-
-	n := fmt.Sprintf("[%s]:%d", ns.IP.String(), ns.Port)
-	log.Debugf("formated ns: %s", n)
-	return n
+	return fmt.Sprintf("%s:%d", ns.IP.String(), ns.Port)
 }
 
 // upstreamCallbacks returns two functions, the first one is used to deactivate


client/internal/dns/server.go
```diff
@@ -421,7 +421,14 @@ func (s *DefaultServer) updateLocalResolver(update map[string]nbdns.SimpleRecord
...
 
 func getNSHostPort(ns nbdns.NameServer) string {
-	return fmt.Sprintf("%s:%d", ns.IP.String(), ns.Port)
+	log.Debugf("getNSHostPort: %s : %d", ns.IP.String(), ns.Port)
+	if ns.IP.Is4() {
+		return fmt.Sprintf("%s:%d", ns.IP.String(), ns.Port)
+	}
+
+	n := fmt.Sprintf("[%s]:%d", ns.IP.String(), ns.Port)
+	log.Debugf("formated ns: %s", n)
```
surik: codespell complain here

pappz: I think these two debug log is too much. We will print out "formatted ns:..." in case of ipv6 only.

client/internal/dns/server.go
```diff
@@ -488,7 +495,18 @@ func (s *DefaultServer) addHostRootZone() {
...
-		handler.upstreamServers[n] = fmt.Sprintf("%s:53", ua)
+		a, err := netip.ParseAddr(ua)
+		if err != nil {
+			log.Errorf("invalid upstream IP address: %s, error: %s", ua, err)
+			continue
+		}
+
+		ipString := ua
+		if !a.Is4() {
+			ipString = fmt.Sprintf("[%s]", ua)
```
pappz, line 499: the port number is not required here?

pappz (changes requested):

client/internal/dns/server.go
mlsmaycon: it is added below

surik (approved):

pappz (approved):

WolfgangDpunkt:
> this adds the address in the expected format in Go [ipv6]:port

I think this would also have to be adapted for the Linux client.
Every ipv6 DNS address that I enter via the Management Gui results in this error message on a debian server client:

`dial udp: address 2001:4860:4860::8888:53: too many colons in address`

PR #1387, opened 2023-12-14
Support status filter by names

mlsmaycon:
## Describe your changes

users can filter status based on peer fully qualified names.

e.g., netbird status -d --filter-by-names peer-a.netbird.cloud,peer-b.netbird.cloud

Enable detailed information when using just filter flags.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


pascal-fischer (approved):
One addition that would be nice would be to filter by hostname, like a match domain that instead of using full fqdn only pass the hostname to that flag

pascal-fischer (approved):

PR #1479, opened 2024-01-17
Add OS Version posture checks 

surik:
## Describe your changes

This PR extends the existing `/api/posture-checks` endpoint with additional OS version checks.

```json
{
	"name": "Default",
	"description": "This checks if the peer is running on Android >=9, iOS >= 17, any macOS, or Linux kernel > 6.0 but not Windows",
	"checks": {
		"os_version_check": {
			"android": {
				"min_version": "9"
			},
			"darwin": {
				"min_version": "0"
			},
			"ios": {
				"min_version": "14.2"
			},
			"linux": {
				"min_kernel_version": "6.0"
			}
		}
	}
}
```

* Version checks for Android, Darwin(macOS), and iOS are based on OS version. `min_version` has to be provided.
* Version checks for Linux and Windows are based on the kernel version. `min_kernel_version` has to be provided.
* To allow any version of a particular OS set `min_version` or `min_kernel_version`.
* To disallow a particular OS simply skip the definition completely as in the example above for Windows.

After this PR is merged both `os_version_check` and `nb_version_check` can be provided in the same check:

```json
{
	"name": "Default",
	"description": "This checks if the peer is running on Android >=9, iOS >= 17, any macOS, or Linux kernel > 6.0 but not Windows",
	"checks": {
		"nb_version_check": {
			"min_version": "0.25.0"
		},
		"os_version_check": {
			"linux": {
				"min_kernel_version": "6.0"
			}
		}
	}
}
```

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/posture/os_version.go
```diff
@@ -0,0 +1,96 @@
...
+	Windows *MinKernelVersionCheck
+}
+
+var _ Check = (*OSVersionCheck)(nil)
+
+func (c *OSVersionCheck) Check(peer nbpeer.Peer) (bool, error) {
+	peerGoOS := peer.Meta.GoOS
+	switch peerGoOS {
+	case "android":
+		return checkMinVersion(peerGoOS, peer.Meta.Core, c.Android)
```
surik: Move from `peer.Meta.Core` to `peer.Meta.OsVersion` after https://github.com/netbirdio/netbird/pull/1495

management/server/http/posture_checks_handler.go
```diff
@@ -199,27 +208,44 @@ func validatePostureChecksUpdate(req api.PostureCheckUpdate) error {
...
 	}
 
+	if osVersionCheck := req.Checks.OsVersionCheck; osVersionCheck != nil &&
+		osVersionCheck.Android != nil && osVersionCheck.Android.MinVersion == "" &&
+		osVersionCheck.Darwin != nil && osVersionCheck.Darwin.MinVersion == "" &&
+		osVersionCheck.Ios != nil && osVersionCheck.Ios.MinVersion == "" &&
+		osVersionCheck.Linux != nil && osVersionCheck.Linux.MinKernelVersion == "" &&
+		osVersionCheck.Windows != nil && osVersionCheck.Windows.MinKernelVersion == "" {
+		return status.Errorf(status.InvalidArgument, "minimum version for at least one OS in the OS version check shouldn't be empty")
+	}
```
bcmmbaga: Consider triggering an error if a single OS version check is set, but its corresponding requirement is not set. Currently, an error is only triggered when all OS versions are set and empty.

```suggestion
	if osVersionCheck := req.Checks.OsVersionCheck; osVersionCheck != nil &&
		osVersionCheck.Android != nil && osVersionCheck.Android.MinVersion == "" ||
		osVersionCheck.Darwin != nil && osVersionCheck.Darwin.MinVersion == "" ||
		osVersionCheck.Ios != nil && osVersionCheck.Ios.MinVersion == "" ||
		osVersionCheck.Linux != nil && osVersionCheck.Linux.MinKernelVersion == "" ||
		osVersionCheck.Windows != nil && osVersionCheck.Windows.MinKernelVersion == "" {
		return status.Errorf(status.InvalidArgument, "minimum version for at least one OS in the OS version check shouldn't be empty")
	}
```

surik: Good caught! I've changed the conditions but split the conditions for easy readability. Also, added tests to cover the cases for this function.

bcmmbaga (approved):

PR #1481, opened 2024-01-18
Add getGroupByName method

pascal-fischer:
This PR extends the account manager to be able to look up a group by its name.

### DIFF
--- a/management/server/group.go
+++ b/management/server/group.go
@@ -73,7 +73,7 @@ func (am *DefaultAccountManager) GetGroup(accountID, groupID string) (*Group, er
 	return nil, status.Errorf(status.NotFound, "group with ID %s not found", groupID)
 }
 
-// GetGroupByName
+// GetGroupByName filters all groups in an account by name and returns the one with the most peers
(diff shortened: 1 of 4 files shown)


util/common.go
```diff
@@ -22,3 +22,13 @@ func FileExists(path string) bool {
 	_, err := os.Stat(path)
 	return err == nil
 }
+
+// Contains checks if a slice of strings contains a specific string.
+func Contains(slice []string, str string) bool {
```
surik: golang 1.21 has a similar function in the standard library: https://pkg.go.dev/slices#Contains

management/server/group.go
```diff
@@ -73,6 +73,39 @@ func (am *DefaultAccountManager) GetGroup(accountID, groupID string) (*Group, er
 	return nil, status.Errorf(status.NotFound, "group with ID %s not found", groupID)
 }
 
+// GetGroupByName
```
surik: Maybe mention here that this returns a single group with most peers.

util/common.go
pascal-fischer: ah good catch, then I don't need this. Will remove, thanks

surik (approved):

PR #1493, opened 2024-01-23
Support disabled resolved stub server mode

pappz:
## Describe your changes
In the case of disabled stub listeren the list of name servers is unordered. The solution is to configure the resolv.conf file directly instead of dbus API.
Because third-party services also can manipulate the DNS settings the agent watch the resolv.conf file and keep it up to date.

- apply file type DNS manager if in the name server list does not exist the 127.0.0.53 address
- watching the resolv.conf file with inotify service and overwrite all the time if the configuration has changed and it invalid
- fix resolv.conf generation algorithm

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/dns/file_linux.go
+++ b/client/internal/dns/file_linux.go
@@ -4,10 +4,13 @@ package dns
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 
+	"github.com/illarion/gonotify"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -23,11 +26,17 @@ const (
 )
 
 type fileConfigurator struct {
-	originalPerms os.FileMode
+	originalPerms   os.FileMode
+	inotify         *gonotify.Inotify
+	inotifyWg       sync.WaitGroup
+	inotifyLock     sync.Mutex
+	nbSearchDomains []string
+	nbNameserverIP  string
 }
 
 func newFileConfigurator() (hostManager, error) {
-	return &fileConfigurator{}, nil
+	fc := &fileConfigurator{}
+	return fc, nil
 }
 
 func (f *fileConfigurator) supportCustomPort() bool {
@@ -58,22 +67,39 @@ func (f *fileConfigurator) applyDNSConfig(config HostDNSConfig) error {
 		}
 	}
 
-	searchDomainList := searchDomains(config)
+	f.inotifyLock.Lock()
+	f.nbSearchDomains = searchDomains(config)
+	f.nbNameserverIP = config.ServerIP
+	defer f.inotifyLock.Unlock()
 
-	originalSearchDomains, nameServers, others, err := parseResolvConf(fileDefaultResolvConfBackupLocation)
+	resolvConf, err := parseBackupResolvConf()
 	if err != nil {
 		log.Error(err)
 	}
 
-	searchDomainList = mergeSearchDomains(searchDomainList, originalSearchDomains)
+	f.stopWatchFileChanges()
+
+	err = f.updateConfig(resolvConf)
+	if err != nil {
+		return err
+	}
+	f.watchFileChanges()
+	return nil
+}
+
+func (f *fileConfigurator) updateConfig(cfg *resolvConf) error {
+	f.inotifyLock.Lock()
+	defer f.inotifyLock.Unlock()
+	searchDomainList := mergeSearchDomains(f.nbSearchDomains, cfg.searchDomains)
+	nameServers := f.generateNsList(cfg)
 
 	buf := prepareResolvConfContent(
 		searchDomainList,
-		append([]string{config.ServerIP}, nameServers...),
-		others)
+		nameServers,
+		cfg.others)
 
 	log.Debugf("creating managed file %s", defaultResolvConfPath)
-	err = os.WriteFile(defaultResolvConfPath, buf.Bytes(), f.originalPerms)
+	err := os.WriteFile(defaultResolvConfPath, buf.Bytes(), f.originalPerms)
 	if err != nil {
 		restoreErr := f.restore()
 		if restoreErr != nil {
@@ -87,6 +113,7 @@ func (f *fileConfigurator) applyDNSConfig(config HostDNSConfig) error {
 }
 
 func (f *fileConfigurator) restoreHostDNS() error {
+	f.stopWatchFileChanges()
 	return f.restore()
 }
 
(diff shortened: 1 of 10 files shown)


client/internal/dns/file_parser_linux.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func parseBackupResolvConf() (*resolvConf, error) {
+	return parseResolvConfFile(fileDefaultResolvConfBackupLocation)
+}
+
+func parseResolvConfFile(resolvConfFile string) (*resolvConf, error) {
+	file, err := os.Open(resolvConfFile)
+	if err != nil {
+		return nil, err
```
lixmal: Do you mind enriching this with some context for easier debugging?

 ```suggestion
		return nil, fmt.Errorf("open: %w", err)
```

File: client/internal/dns/file_parser_linux.go [hunk 2]
```diff
@@ -0,0 +1,98 @@
...
+	}
+	defer file.Close()
```
lixmal: Can we log the error here? Could give us some hint if there's something wonky going on with the filesystem
```suggestion
	defer func() {
		if err := file.Close(); err != nil {
			log.Errorf("failed closing %s: %s", resolvConfFile, err)
		}
	}()
```

client/internal/dns/file_parser_linux.go
```diff
@@ -0,0 +1,98 @@
...
+
+	cur, err := os.ReadFile(resolvConfFile)
+	if err != nil {
+		return nil, err
```
lixmal:
```suggestion
		return nil, fmt.Errorf("read: %w", err)
```

client/internal/dns/file_parser_linux.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	if len(cur) == 0 {
+		return nil, fmt.Errorf("file is empty")
+	}
+
+	rconf := &resolvConf{
+		nameServers: make([]string, 0),
```
lixmal: We don't need this, append takes care of that.
```suggestion
```

client/internal/dns/file_parser_linux.go
```diff
@@ -0,0 +1,98 @@
...
+		others:      make([]string, 0),
```
lixmal: Same here
```suggestion
```

client/internal/dns/file_parser_linux_test.go
```diff
@@ -0,0 +1,96 @@
...
+		expectedOther  []string
+	}{
+		{
+			input: `domain chello.hu
+search chello.hu
+nameserver 192.168.0.1
+`,
+			expectedSearch: []string{"chello.hu"},
+			expectedNS:     []string{"192.168.0.1"},
+			expectedOther:  []string{},
```
lixmal, line 25: Can we test this part as well: `expectedOther`?

client/internal/dns/host_linux.go
```diff
@@ -103,3 +103,20 @@ func getOSDNSManagerType() (osManagerType, error) {
 	}
 	return fileManager, nil
 }
+
+// checkStub checks if the stub resolver is disabled in systemd-resolved. If it is disabled, we fallback to file manager.
```
lixmal:
```suggestion
// checkStub checks if the stub resolver is disabled in systemd-resolved. If it is disabled, we fall back to file manager.
```

client/internal/dns/file_linux.go
```diff
@@ -59,22 +62,35 @@ func (f *fileConfigurator) applyDNSConfig(config HostDNSConfig) error {
...
 	if err != nil {
 		log.Error(err)
 	}
 
-	searchDomainList = mergeSearchDomains(searchDomainList, originalSearchDomains)
+	f.repair.stopWatchFileChanges()
+
+	err = f.updateConfig(nbSearchDomains, nbNameserverIP, resolvConf)
+	if err != nil {
+		return err
```
lixmal, line 77: Can we enrich this with some context?
```suggestion
		return fmt.Errorf("parse backup resolv.conf: %w", err)
```

client/internal/dns/file_repair_linux.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (f *repair) watchFileChanges(nbSearchDomains []string, nbNameserverIP string) {
+	if f.inotify != nil {
+		return
+	}
+
+	log.Infof("start to watch resolv.conf")
+	inotify, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.Errorf("failed to start inotify watcher for resolv.conf: %s", err)
```
lixmal, line 50: Can we return this error instead? Then we can log it upstream as non-critical.

client/internal/dns/file_repair_linux.go
```diff
@@ -0,0 +1,151 @@
...
+		if event.Has(et) {
+			ok = true
+			break
+		}
+	}
+	if !ok {
+		return false
+	}
+
+	operatioFileSymlink := fmt.Sprintf("%s~", f.operationFile)
```
lixmal:
```suggestion
	operationFileSymlink := fmt.Sprintf("%s~", f.operationFile)
```

client/internal/dns/file_repair_linux.go
```diff
@@ -0,0 +1,151 @@
...
+	if event.Name == f.operationFile || event.Name == operatioFileSymlink {
```
lixmal:
```suggestion
	if event.Name == f.operationFile || event.Name == operationFileSymlink {
```

client/internal/dns/file_repair_linux_test.go
```diff
@@ -0,0 +1,130 @@
...
+				cancel()
+				return nil
+			}
+
+			r := newRepair(operationFile, updateFn)
+			r.watchFileChanges([]string{"netbird.cloud"}, "10.0.0.1")
+
+			err = os.WriteFile(operationFile, []byte(tt.touchedConfContent), 0755)
+			if err != nil {
+				t.Fatalf("failed to wrtie out resolv.conf: %s", err)
```
lixmal:
```suggestion
				t.Fatalf("failed to write out resolv.conf: %s", err)
```

client/internal/dns/file_repair_linux_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run("test", func(t *testing.T) {
+			t.Parallel()
+			workDir := t.TempDir()
+			operationFile := workDir + "/resolv.conf"
+			err := os.WriteFile(operationFile, []byte(tt.resolvConfContent), 0755)
+			if err != nil {
+				t.Fatalf("failed to wrtie out resolv.conf: %s", err)
```
lixmal:
```suggestion
				t.Fatalf("failed to write out resolv.conf: %s", err)
```

client/internal/dns/file_repair_linux_test.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+
+			<-ctx.Done()
+
+			r.stopWatchFileChanges()
+
+			if changed != tt.wantChange {
+				t.Errorf("unexpected result: want: %v, get: %v", tt.wantChange, changed)
```
lixmal:
```suggestion
				t.Errorf("unexpected result: want: %v, got: %v", tt.wantChange, changed)
```

client/internal/dns/file_repair_linux.go
```diff
@@ -0,0 +1,151 @@
...
+				log.Warnf("failed to parse resolv conf: %s", err)
+				continue
+			}
+
+			log.Debugf("check resolv.conf parameters: %s", rConf)
+			if !isNbParamsMissing(nbSearchDomains, nbNameserverIP, rConf) {
+				log.Tracef("resolv.conf still correct, skip the update")
+				continue
+			}
+			log.Info("broken params in resolv.conf, repair it...")
```
lixmal:
```suggestion
			log.Info("broken params in resolv.conf, repairing it...")
```

lixmal (dismissed):

client/internal/dns/file_parser_linux.go
pappz: It is true but it is not guaranteed we will append any element. It could be empty nameServers, others, and search domains. In this case, the nameServers will be nil.  What do you think?

lixmal: I think a nil slice is as good as an empty one, but either way it should be consistent. So `searchDomains` should be initialized as well

pappz: Good point!

lixmal (approved):

client/internal/dns/file_linux.go
pappz: You are right but in the updateConfig(...) we return with a long error message. Does it make sense to add more info here?

client/internal/dns/file_repair_linux.go
pappz: In general yes but I think in this case is not a good idea to provide error return variable. If it return with err the developer could think that the watchFile is not started well so does not need to stop it. But if we check deeper this function can happen issue around the last line. The best way is to implement a "self close" procedure in case of error but hard to implement it in thread safe way.

lixmal (approved):

PR #1494, opened 2024-01-23
Restore dns on unclean shutdown

lixmal:
## Describe your changes

1. When the Netbird agent terminates ungracefully (kill, power outage, panic, etc.), the system DNS settings are not restored.

    This can result in broken DNS resolution (e.g. a DNS server provided by a peer), in which case the agent won't be able to connect to the management server.
    
    To fix this issue we create an indicator file after adjusting the system DNS settings.
    This file is deleted once the settings are restored.
    
    If the agent terminates ungracefully it won't be able to delete this file. Therefore when it is started again it will detect the existence of this file and restore the previous DNS state.
    
    
    - On Linux
    
      `/var/lib/netbird/resolv.conf`, contains the old `/etc/resolv.conf`
      `/var/lib/netbird/manager`, contains the name of the previous host manager and last netbird IP address
      
      - manager `file`
      
          DNS settings are restored by copying the `/var/lib/netbird/resolv.conf` back to `/etc/resolv.conf`

           Test:
          - Uninstall `systemd-resolved`, `openresolv`/`resolvconf`
          - `echo nameserver 9.9.9.9 | sudo tee /etc/resolv.conf`
          - Watch `/var/log/netbird/client.log` for restore message and `/etc/resolv.conf`
          - Start agent
          - `killall -9 netbird`
          - Start agent

     
       - manager `resolvconf`
      
           DNS settings are restored by running `restoreHostDNS()` (runs `resolvconf -d <interface>`)

           Test:
            - Uninstall `systemd-resolved`
            - Install `openresolv` package
            - `echo "nameserver 9.9.9.9" | sudo resolvconf -x -a <physical int>`
            - Watch `/var/log/netbird/client.log` for restore message and `/etc/resolv.conf`
            - Start agent
            - `killall -9 netbird`
            - Start agent

       - manager `systemd`
      
           DNS settings are restored by running `restoreHostDNS()` (runs dbus methods`)
        
          Test:
            - Watch `/var/log/netbird/client.log` for restore message and `resolvectl status`
            - Start agent
            - `killall -9 netbird`
            - Start agent

        - manager `networkManager`
      
           DNS settings are restored by running `restoreHostDNS()` (runs dbus methods`)
        
          Test:
            - Uninstall `systemd-resolved`
            - Edit `/etc/NetworkManager/NetworkManager.conf`, add `dns=dnsmasq` in the main section
            - `sudo nmcli general reload` or `sudo nmcli networking off; sudo nmcli networking on` 
            - Watch `/var/log/netbird/client.log` for restore message and `nmcli device show | grep -i dns`  
            - Start agent
            - `killall -9 netbird`
            - Start agent

    - On Windows
    
      `%PROGRAMDATA%\Netbird\unclean_shutdown_dns.txt`, contains the interface `guid` that is required to set up the host manager.
      
      DNS settings are restored by running `restoreHostDNS()` (removes registry keys)

      Test:
        - Watch `%PROGRAMDATA%\Netbird\client.log` for restore message and `SYSTEM\CurrentControlSet\Services\Dnscache\Parameters\DnsPolicyConfig\NetBird-Match` registry
        - Start agent
        - Start PowerShell as admin
        - `Get-Process -Name netbird | Stop-Process -Force`
        - (Re)start agent
    
    - On macOS
    
      `/var/lib/netbird/unclean_shutdown_dns`, empty
      
      DNS settings are restored by running `restoreHostDNS()` (runs `scutil remove`).

      Test:
        - Watch `/var/log/netbird/client.log` for restore message and `scutil --dns` for DNS settings
        - Start agent
        - `sudo killall -9 netbird`
        - Start agent

    - mobile should not be affected
        

2. Moves the `restore DNS` logic before the `remove interface` logic so we don't get errors for missing interfaces 

3. Improves error handling and error messages for debugging

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/dns/host_darwin.go
+++ b/client/internal/dns/host_darwin.go
@@ -5,14 +5,10 @@ package dns
 import (
 	"bufio"
 	"bytes"
-	"errors"
 	"fmt"
 	"io"
-	"io/fs"
 	"net/netip"
-	"os"
 	"os/exec"
-	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -32,8 +28,6 @@ const (
 	scutilPath                          = "/usr/sbin/scutil"
 	searchSuffix                        = "Search"
 	matchSuffix                         = "Match"
-
-	fileUncleanShutdownFileLocation = "/var/lib/netbird/unclean_shutdown_dns"
 )
 
 type systemConfigurator struct {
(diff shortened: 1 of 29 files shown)


client/internal/dns/host_linux.go
```diff
@@ -120,3 +163,85 @@ func checkStub() bool {
...
+
+	log.Warnf("restoring unclean shutdown dns settings via previously detected manager: %s", osManagerTypeStr)
+
+	// determine os manager type, so we can invoke the respective restore action
+	osManagerType, err := newOsManagerType(osManagerTypeStr)
+	if err != nil {
+		return fmt.Errorf("detect previous host manager: %w", err)
+	}
+
+	// the only real thing we need is the interface name
```
pappz: If it is true I recommend to eliminate the WGIface dependency from the osManagers. The code will be more cleaner, shorter and will be easier to write proper unit tests.

client/internal/dns/host_darwin.go
```diff
@@ -266,7 +298,51 @@ func runSystemConfigCommand(command string) ([]byte, error) {
 	cmd.Stdin = strings.NewReader(command)
 	out, err := cmd.Output()
 	if err != nil {
-		return nil, fmt.Errorf("got error while running system configuration command: \"%s\", error: %s", command, err)
+		return nil, fmt.Errorf("running system configuration command: \"%s\", error: %w", command, err)
 	}
 	return out, nil
 }
+
+func CheckUncleanShutdown(_ string) error {
```
pappz: I do not see sense of _.

lixmal: Removed

client/internal/dns/host_linux.go
lixmal: Good point. Somehow I was under the impression that we need the whole interface struct, but that is only the case for windows.

Fixed in 2644718189e3486b95c5ffd53ac012453835432b

pappz (approved):

Issue #683, opened 2023-02-11
limit Client Server trust 


Darin755:
**Is your feature request related to a problem? Please describe.**
I'm a little concerned with the level of trust that netbird puts onto the server and the network operator. I know its convenient to allow some control of the clients though the dashboard but I think this could be easily misused to harm the clients that aren't servers

**Describe the solution you'd like**
I would be nice to have a "don't trust the server" option when installing netbird that would block connections to the local client at the client level. This would protect labtops and desktops that usually don't have as good as security compared to servers.

**Other considerations**
I am also a little worried about the power entrusted to the server. Creating a single server that connects everything could allow a corrprate network to be compromised though one single machine. I know that netbird can't protect against third party services running on clients but I think its a good idea to make the clients assume the server has already been compromised.

mlsmaycon:
Thanks for your request @Darin755, can you share a bit more of how you see the "don't trust the server" working?

We are working on better ACLs at the moment which will control inbound access to clients and we will add flags to disable SSH and DNS at the client level.

Darin755:
- Its great that you're adding ACL's but they wouldn't fix the issue. My fear is that the netbird server will become a high value target for attacker that want to harm a individual or organization. Currently netbird puts all of the trust into one server and when that server gets compromised it could be used to compomise every device in the network. What I would like to see is a more distributed approach to security that allows for the clients to detect if something's wrong before its to late.

 - I am also worried about the server having ssh access to the clients with the new ssh feature. I think it would be really bad if someone was able to access ssh on all of a companies devices.

 - I recently been using i2p for remote access. Its design is overly complicated for what netbird is trying to do but I think netbird could learn something from i2p. One of the biggest benifits of i2p is its zero trust architecture. I would like to see netbird add the ability to have multiple servers that talk to each other over a a secure channel. Each server should be setup in a way that doesn't give it power over the other servers. If one server is compromised it should be rejected from the network and traffic should be rerouted.

Sorry this is so long. I am mostly just typing out loud.

mlsmaycon:
Thanks, @Darin755 for sharing more about your concerns, which are definitely helpful for us to build a better system. 

We are always considering and researching options to improve security in the system while keeping usability simple and more updates towards that will come in the following months.

Regarding the trust with the management server, I forgot to mention in the previous message that we have the option to use Wireguard's pre-shared keys with the flag `--preshared-key` to control tunnel establishment.

CrimsonFez:
[...]

Issue #509, opened 2022-10-17
Client side control of SSH.


dragon2611:
Can we please make SSH support something that can be disabled from the client side?

I quite like the way tailscale handles it where if you want to use their SSH implementation you have to connect the client with an argument passed so it tells the control plane SSH is available before you can switch on tailscale SSH.

Zorlin:
This is hugely important, otherwise all my Tailscale boxes basically have root on each other :/

tconnard:
This is a big issue for me.
I don't want the cloud hosted (either managed or on a vm) management for the overlay network to be able to bypass access control for ssh.
I already have my own ssh management in place

grzybniak:
hello, any news here?

tjarbo:
Any progress on this topic? Maybe @braginini (as you posted on #852)?
As highlighted in https://github.com/netbirdio/netbird/issues/683#issuecomment-1806477634, this SSH feature (among others) is really a no-go as the netbird server/network operator is currently too powerful. ACLs etc. are not a mitigation for this threat.

jonathanspw:
+1

codyro:
+1

PR #1508, opened 2024-01-30
Disable SSH server by default on client side and add the flag --allow-server-ssh to enable it

charnesp:
## Describe your changes
I add in the configuration of the Netbird client the option to allow (or not) the launch of an SSH server on the peer.

The client-side control of the SSH server option on the client side of Netbird create a root access to all peers by Netbird administrators, thus an important security issue. 

This changes the default behavior for **new peers**, by requiring the agent to be executed with allow-server-ssh set to true in order for the management configuration to take effect. See example commands below:

To enable SSH
```shell
netbird down
netbird up --allow-server-ssh
```

To disable SSH
```shell
netbird down
netbird up --allow-server-ssh=false
```

> Existing peers will have SSH enabled in the same way as the previous version. You can disable it with the commands above.

https://github.com/netbirdio/netbird/issues/509 and https://github.com/netbirdio/netbird/issues/683

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


charnesp:
Hi @mlsmaycon I see you tried to relaunch the test pipeline. By looking at the logs, it appears it fail because it tried to test the SSH server connectivity. As in this pull request the SSH server is disabled by default (to ensure a zero-trust behavior of the agent) it will always fail. 
Thus 2 options : 
* Enabling the SSH server by default (with the potential zero-trust issue if the agent is not correctly configured)
* Modyfing the test pipeline to handle this configuration.

What do you recommend ?

mlsmaycon:
> Hi @mlsmaycon I see you tried to relaunch the test pipeline. By looking at the logs, it appears it fail because it tried to test the SSH server connectivity. As in this pull request the SSH server is disabled by default (to ensure a zero-trust behavior of the agent) it will always fail. Thus 2 options :
> 
> * Enabling the SSH server by default (with the potential zero-trust issue if the agent is not correctly configured)
> * Modyfing the test pipeline to handle this configuration.
> 
> What do you recommend ?

Hello @charnesp, first of all, thanks for submitting the PR; 

I have two points to add about the PR and the tests:

1. about the PR: We need to consider the case when existing users are using the client and require SSH, so for retro-compatibility, we should use a `*bool` type for `Config.ServerSSHAllowed` and when there is an existing configuration file with `Config.ServerSSHAllowed  == nil` we set the value to `true`. 
2. for the tests: We should modify them to validate both cases.

What do you think? do you need some help there?

charnesp:
Hi @mlsmaycon , it seems indeed logical for existing users. 

I pushed a modification of the code following your advise using pointers and tested it. It should cover your case.

Thank you

client/internal/config.go
```diff
@@ -88,6 +90,16 @@ func ReadConfig(configPath string) (*Config, error) {
...
 		}
+
+		// Should be redondant with createEngineConfig declaration on connect.go, but easier to read
+		if config.ServerSSHAllowed == nil {
+			config.ServerSSHAllowed = util.True()
+		}
+
+		// DEBUG
+		log.Infof("ServerSSHAllowed value")
+		log.Info(config.ServerSSHAllowed)
```
bcmmbaga: we should remove these debug statements

charnesp: Indeed. I did that on this commit : https://github.com/netbirdio/netbird/pull/1508/commits/4c8c9766bce1d7a90d485c698dd07d5c324fc4a3

bcmmbaga: That commit did remove from `client/internal/connect.go` but there are still debug statements in  `client/internal/config.go`

charnesp: My bad... It is corrected

bcmmbaga:
we need to fix these tests [TestEngine_SSH](https://github.com/netbirdio/netbird/actions/runs/7899388378/job/21566572455?pr=1508#step:7:48)

client/internal/config.go
```diff
@@ -280,6 +293,12 @@ func update(input ConfigInput) (*Config, error) {
 		refresh = true
 	}
 
+	if input.ServerSSHAllowed != nil {
+		log.Infof("SSH allowed flag set")
+		config.ServerSSHAllowed = input.ServerSSHAllowed
+		refresh = true
+	} 
+
```
bcmmbaga, line 311: If there is already existing configuration, we need to update the `ServerSSHAllowed` field to `true` if it is currently `nil`, in order to avoid breaking the older clients. 

```
if config.ServerSSHAllowed == nil {
		config.ServerSSHAllowed = util.True()
		refresh = true
}
```

client/internal/config.go
```diff
@@ -88,6 +90,12 @@ func ReadConfig(configPath string) (*Config, error) {
 		if _, err := util.ReadJson(configPath, config); err != nil {
 			return nil, err
 		}
+
+		// Should be redondant with createEngineConfig declaration on connect.go, but easier to read
+		if config.ServerSSHAllowed == nil {
+			config.ServerSSHAllowed = util.True()
+		}
```
bcmmbaga: We need to move this from here to `update`. why? This will only update the memory-loaded configuration, but the client configuration file will not be updated. See this [comment](https://github.com/netbirdio/netbird/pull/1508#discussion_r1492748875).

client/internal/config.go
```diff
@@ -280,6 +293,12 @@ func update(input ConfigInput) (*Config, error) {
 		refresh = true
 	}
 
+	if input.ServerSSHAllowed != nil {
+		log.Infof("SSH allowed flag set")
```
bcmmbaga, line 297: You can also remove this

bcmmbaga: You can also remove this log message.

client/internal/config.go
charnesp: Understood, done in the last commit

charnesp:
@bcmmbaga I guess the last pull should be ok concerning the TestEngine_SSH.

bcmmbaga (approved):

charnesp:
Hi @bcmmbaga do you need my help for the branch conflicts?

bcmmbaga:
> Hi @bcmmbaga do you need my help for the branch conflicts?

Yes, please resolve them.

bcmmbaga (approved):

mlsmaycon:
Thanks, @charnesp, for the awesome contribution!

charnesp:
@mlsmaycon my pleasure. Thank you for your amazing work and solution!

dillfrescott:
Thank you so much for doing this!!!

PR #1510, opened 2024-01-30
Extract peer real IP from Load Balancer when possible

surik:
Otherwise, get it from the GRPC peer address.

This PR extends management config with a new `TrustedHTTPProxies` option that accepts a list of network prefixes ex: `"192.168.1.1/32", "10.0.0.0/16"` that is used for evaluation of client real ip according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#selecting_an_ip_address

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/grpcserver.go
+++ b/management/server/grpcserver.go
@@ -3,7 +3,6 @@ package server
 import (
 	"context"
 	"fmt"
-	"net"
 	"net/netip"
 	"strings"
 	"time"
@@ -136,9 +135,9 @@ func extractIps(ctx context.Context) (string, string) {
 		}
 
 		if grcpPeerExtracted {
-			ip, err := net.ResolveTCPAddr("", grpcPeer.Addr.String())
+			addrPort, err := netip.ParseAddrPort(grpcPeer.Addr.String())
 			if err == nil {
-				return ip.IP.String(), peerAddr
+				return addrPort.Addr().String(), peerAddr
 			}
 			log.Errorf("when resolve ip address %s", err)
 		}
--- a/management/server/grpcserver.go
+++ b/management/server/grpcserver.go
@@ -3,6 +3,8 @@ package server
 import (
 	"context"
 	"fmt"
+	"net"
+	"net/netip"
 	"strings"
 	"time"
 
@@ -11,6 +13,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	gRPCPeer "google.golang.org/grpc/peer"
 	"google.golang.org/grpc/status"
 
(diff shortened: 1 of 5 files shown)


management/server/grpcserver.go
```diff
@@ -109,17 +112,52 @@ func (s *GRPCServer) GetServerKey(ctx context.Context, req *proto.Empty) (*proto
...
+			ips := strings.Split(value[0], ",")
+			ipStr := strings.TrimSpace(ips[len(ips)-1])
+			ip, err := netip.ParseAddr(ipStr)
+			if err == nil {
+				return ip.String(), peerAddr
+			}
+		}
+
+		if grcpPeerExtracted {
+			ip, err := net.ResolveTCPAddr("", grpcPeer.Addr.String())
```
mlsmaycon: Using netip.ParseAddrPort() might be better for this case as it works with the assumption of IP address x the domain and protocol checks from net.ResolveTCPAddr

management/server/grpcserver.go
```diff
@@ -109,17 +112,52 @@ func (s *GRPCServer) GetServerKey(ctx context.Context, req *proto.Empty) (*proto
 	}, nil
 }
 
+// extractIps returns real peer ip and grpc peer address (ip:port)
+func extractIps(ctx context.Context) (string, string) {
```
mlsmaycon: Do we need 2 addresses here? Maybe just getting from the headers and using gRPCPeer.FromContext as a fallback in case headers are not available?

surik: To keep backward compatibility as previously we were using the `ip:port` of GRPC peer. If the port is not needed I'm happy to keep only one.

mlsmaycon: Got it, we can remove, it was used only in debug logs.

management/server/grpcserver.go
surik: This is much better. Thank you!

management/server/grpcserver.go
```diff
@@ -109,17 +111,45 @@ func (s *GRPCServer) GetServerKey(ctx context.Context, req *proto.Empty) (*proto
 	}, nil
 }
 
+// getRealIP returns real peer IP from load balance or fallback to IP of the GRPC peer
```
lixmal: I think there should also be a config option for how many trusted proxies are in front of our server.
With the current behavior, if it's 0, then `X-Forwarded-For` is either spoofed or wrong (added by the client proxy).
If it's n>1, then we would get the wrong IP address. It should be shifted n times to the left

surik: Very good point. I think I'll use [the real IP middleware](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/interceptors/realip/realip.go) instead of my implementation. See the draft, I only need to make the trusted proxy list configurable.

mlsmaycon (approved):

PR #1514, opened 2024-01-31
Update port, ip choice logic in DNS service

pappz:
## Describe your changes

Ensure we use WG address instead of loopback addresses for eBPF.
- First try to use 53 port
- Try to use 5053 port on WG interface for eBPF
- Try to use 5053 on WG interface or loopback interface

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/dns/service_listener.go
+++ b/client/internal/dns/service_listener.go
@@ -203,7 +203,7 @@ func (s *serviceViaListener) tryToUseeBPF() (ebpfMgr.Manager, uint16, bool) {
 		return nil, 0, false
 	}
 
-	port, err := s.generateFreePort()
+	port, err := s.generateFreePort() //nolint:staticcheck,unused
 	if err != nil {
 		log.Warnf("failed to generate a free port for eBPF DNS forwarder server: %s", err)
 		return nil, 0, false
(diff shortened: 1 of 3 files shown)


client/internal/dns/service_listener.go
```diff
@@ -140,55 +131,89 @@ func (s *serviceViaListener) setListenerStatus(running bool) {
...
+
+	ip, ok := s.testFreePort(defaultPort)
+	if ok {
+		return ip, defaultPort, nil
+	}
+
+	ebpfSrv, ok := s.tryToUseeBPF()
+	if ok {
+		s.ebpfService = ebpfSrv
+		return s.wgInterface.Address().IP.String(), customPort, nil
```
mlsmaycon: we should probe this port or use a dynamic port with the code below:
```Go
	udpAddr := net.UDPAddrFromAddrPort(netip.MustParseAddrPort("0.0.0.0:0"))
	probeListener, err := net.ListenUDP("udp", udpAddr)
	var port uint16
	if err == nil {
		addPort := netip.MustParseAddrPort(probeListener.LocalAddr().String()) // might panic if address is incorrect
		port = addPort.Port()
		probeListener.Close()
		return s.wgInterface.Address().IP.String(),addPort.Port(), nil
	}
```

pappz: In the tryToUseeBPF() function we test the port (I know would better outside but then the result source code more messy). I add the dynamic port.

mlsmaycon (approved):

PR #1516, opened 2024-02-01
Add Locations endpoints

bcmmbaga:
## Describe your changes

Introduced additional endpoints for country and city locations:
- `GET /api/locations/countries`: Provides a list of all countries in 2-letter ISO 3166-1 alpha-2 codes
-  `GET /api/locations/countries/:country/cities`: Retrieves the list of English city names corresponding to a specific country code

Extended the functionality of the `./infrastructure_files/download-geolite2.sh` script to include the download of the geolite cities locations CSV database and the subsequent creation of an SQLite database from the CSV.

```bash
MM_ACCOUNT_ID=... MM_LICENSE_KEY=...  ./infrastructure_files/download-geolite2.sh
```

Both `MM_ACCOUNT_ID` and `MM_LICENSE_KEY` can be obtained from [MaxMind](https://dev.maxmind.com/geoip/geolite2-free-geolocation-data).

Upon successful execution of the script, the resulting geonames.db file should be placed in the designated datadir. Clear instructions for this process are provided in the script output:
```
SQLite database 'geonames.db' created successfully.
Now you can place geonames.db to 'datadir' of management service.
Example:
	docker compose cp geonames.db management:/var/lib/netbird/
```

Note: The new API endpoints will only be accessible if the geonames.db exists in the specified data directory.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


infrastructure_files/download-geolite2.sh
```diff
@@ -22,41 +22,81 @@ then
...
+
+
+download_geolite_csv_and_create_sqlite_db() {
+  DATABASE_URL="https://download.maxmind.com/geoip/databases/GeoLite2-City-CSV/download?suffix=zip"
+
+  # Download the database file
+  echo "Downloading database file..."
+  DATABASE_FILE=$(curl -s -u "$MM_ACCOUNT_ID":"$MM_LICENSE_KEY" -L -O -J "$DATABASE_URL" -w "%{filename_effective}")
+
+  unzip "$DATABASE_FILE"
```
surik: I think this would make sense to check sha256sum similar to mmdb file above

infrastructure_files/download-geolite2.sh
```diff
@@ -22,41 +22,81 @@ then
...
+EOF
+
+
+  # Remove downloaded and extracted files
+  rm -r -r "$EXTRACTION_DIR"
+  rm  "$DATABASE_FILE"
+
+  echo "SQLite database '$DB_NAME' created successfully."
+  echo "Now you can place $DB_NAME to 'datadir' of management service."
+  echo -e "Example:\n\tdocker compose cp $DB_NAME management:/var/lib/netbird/"
```
surik, line 112: does the `store.go` support the change on runtime? For mmdb it works because the DB is fully in memory and we added a special background reloader.

management/server/geolocation/store.go
```diff
@@ -0,0 +1,68 @@
...
+	"gorm.io/gorm/logger"
+)
+
+// SqliteStore represents a location storage backed by a Sqlite DB.
+type SqliteStore struct {
+	db *gorm.DB
+}
+
+func NewSqliteStore(dataDir string) (*SqliteStore, error) {
+	storeStr := "geonames.db?cache=shared"
```
surik: More likely we need `mode=ro` here

management/server/http/api/openapi.yml
```diff
@@ -2733,3 +2733,62 @@ paths:
...
+          description: The 2-letter ISO 3166-1 alpha-2 country code
+      responses:
+        '200':
+          description: List of city names
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: string
```
surik: Now I'm thinking that maybe return arrays of objects will be more future-proof as later we can extend it with `geoname_id` without breaking the client much.

management/cmd/management.go
```diff
@@ -162,6 +163,14 @@ var (
 				}
 			}
 
+			geolocationStore, err := geolocation.NewSqliteStore(config.Datadir)
+			if err != nil {
+				log.Warnf("could not  geo location database, we proceed without location endpoints")
+			} else {
+				log.Infof("geo location database has been initialized from %s", config.Datadir)
+			}
+			geolocationManager := geolocation.NewManager(geolocationStore)
```
surik: Maybe the `geolocationManager` should be a part of `geolocation.NewGeolocation` then we don't need to have two objects with similar scope.

management/server/geolocation/store.go
```diff
@@ -0,0 +1,148 @@
...
+// SqliteStore represents a location storage backed by a Sqlite DB.
+type SqliteStore struct {
+	db       *gorm.DB
+	filePath string
+	mux      *sync.RWMutex
+}
+
+func NewSqliteStore(dataDir string) (*SqliteStore, error) {
+	file := filepath.Join(dataDir, "geonames.db")
+	_, err := fileExists(file)
```
surik: We need to move this inside `connectDB`

management/server/geolocation/geolocation.go
```diff
@@ -185,5 +202,45 @@ func (gl *Geolocation) reload(newSha256sum []byte) error {
 
 	log.Infof("Successfully reloaded '%s'", gl.mmdbPath)
 
-	return nil
+	return gl.locationDB.reload()
```
surik: This should happen independently and not as a follow-up on the mmdb reload. Let's have a similar process with double-checking sum256 to minimize the chance of reading db while it is being copied.

management/server/geolocation/store.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	s.db = db
+
+	log.Infof("Successfully reloaded '%s'", s.filePath)
+	return nil
+}
+
+// connectDB connects to an SQLite database and prepares it by setting up an in-memory database.
+func connectDB(source string) (*gorm.DB, error) {
+	storeStr := ":memory:?cache=shared&mode=ro"
```
surik, line 147: Let's measure how much time this takes.
```go
	start := time.Now()
	defer func() {
		log.Debugf("Time taken: %v", time.Since(start))
	}()
```

management/server/geolocation/store.go
```diff
@@ -0,0 +1,148 @@
...
+		Group("country_name").
+		Scan(&countries)
+	if result.Error != nil {
+		return nil, result.Error
+	}
+	return countries, nil
+}
+
+// GetCitiesByCountry retrieves a list of cities from the store based on the given country ISO code.
+func (s *SqliteStore) GetCitiesByCountry(countryISOCode string) ([]City, error) {
```
surik, line 67: Do we need a read lock here as we lock it when reloading?

management/server/geolocation/store.go
```diff
@@ -0,0 +1,148 @@
...
+	s.mux.Lock()
+	defer s.mux.Unlock()
+
+	log.Infof("Reloading '%s'", s.filePath)
+
+	db, err := connectDB(s.filePath)
+	if err != nil {
+		return err
+	}
+	s.db = db
```
surik: Should we worry about freeing the memory of previously opened db?

management/server/geolocation/store.go
bcmmbaga: Is it time to set up the in-memory database, or for the whole process until connection?

surik: The whole execution of `connectDB`. This may be a helpful metric as I would expect it to take around 100ms.

management/server/geolocation/store.go
bcmmbaga: With the current lock implemented in `reload`, yes, we must consider adding the lock in both database access methods, as we will also add the database closure to free resources before replacing it with a new database connection.

management/server/geolocation/store.go
bcmmbaga: Yes, I missed it, but it's essential because we loaded it into memory.

surik (approved):

PR #1517, opened 2024-02-01
Extend peer meta with location information

surik:
## Describe your changes

This PR uses the geolocation service to resolve IP to location. The lookup happens once on the first connection - when a client calls the Sync procedure.
The location is stored as part of the peer:

```go
type Location struct {
	ConnectionIP net.IP 
	CountryCode  string
	CityName     string
	GeoNameID    uint 
}
```

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/cmd/management.go
+++ b/management/cmd/management.go
@@ -179,7 +179,7 @@ var (
 
 			trustedPeers := config.TrustedHTTPProxies
 			if len(trustedPeers) == 0 {
-				trustedPeers = []netip.Prefix{netip.MustParsePrefix("0.0.0.0/0")}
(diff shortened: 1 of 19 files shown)


pappz:
In the peer.go PeerSystemMeta we should not add the Location to the isEqual function?

bcmmbaga:
> In the peer.go PeerSystemMeta we should not add the Location to the isEqual function?

We shouldn't, as Location isn't included in PeerSystemMeta (it was moved out). It's at the peer level

bcmmbaga (approved):

PR #1537, opened 2024-02-06
Get cache from external cache when refresh fails

mlsmaycon:
## Describe your changes

In some cases, when the refresh cache fails, we should try to get the cache from the external cache obj.

This may happen if the IDP is not responsive between storing metadata and refreshing the cache

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -1223,6 +1223,21 @@ func (am *DefaultAccountManager) lookupUserInCache(userID string, account *Accou
 		}
 	}
 
+	user, err := account.FindUser(userID)
+	if err != nil {
+		log.Errorf("failed finding user %s in account %s", userID, account.Id)
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


management/server/account.go
```diff
@@ -1223,6 +1223,21 @@ func (am *DefaultAccountManager) lookupUserInCache(userID string, account *Accou
 		}
 	}
 
+	user, err := account.FindUser(userID)
+	if err != nil {
+		log.Errorf("failed finding user %s in account %s", userID, account.Id)
+		return nil, err
+	}
+
+	key := user.IntegrationReference.CacheKey(account.Id, userID)
```
surik, line 1232: With this change do we need this pice of code?
https://github.com/netbirdio/netbird/blob/62bacee8dce6fbc0351f5884d90cf991994399b0/management/server/user.go#L893

surik (approved):

PR #1552, opened 2024-02-09
Extend Peer metadata with cloud detection

pascal-fischer:
## Describe your changes
Adding a cloud detection into the peer metadata to identify peers running in the cloud.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/system/info_darwin.go
```diff
@@ -58,6 +59,11 @@ func GetInfo(ctx context.Context) *Info {
 	gio.Hostname = extractDeviceName(ctx, systemHostname)
 	gio.WiretrusteeVersion = version.NetbirdVersion()
 	gio.UIVersion = extractUserAgent(ctx)
+	gio.Cloud = detect_cloud.Detect()
```
lixmal: I'd suggest we add the ctx here as well so we can abort depending on the app life cycle. And pass that one to all `Do` calls as well.

client/system/detect_cloud/detect.go
```diff
@@ -0,0 +1,92 @@
...
+	detectSoftwareFuncs := []func() string{
+		detectOpenStack,
+		detectContainer,
+	}
+
+	vendorResults := make(chan string, len(detectVendorFuncs))
+	softwareResults := make(chan string, len(detectSoftwareFuncs))
+
+	var wg sync.WaitGroup
+	wg.Add(len(detectVendorFuncs) + len(detectSoftwareFuncs))
```
lixmal: I would prefer to put wg.Add(1) inside the loops, so if there are any changes to this function we don't need to remember to modify this line.

client/system/detect_cloud/gce.go
```diff
@@ -0,0 +1,24 @@
...
+	r.Header.Add("Metadata-Flavor", "Google")
+
+	resp, err := hc.Do(r)
+	if err != nil {
+		return ""
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		return "Google Compute Engine"
```
lixmal:
```suggestion
		return "Google Cloud Platform"
```

is this generic GCP better here? Since we also use `aws` and not `ec2`. The whole filename then, really.

client/system/detect_cloud/kube.go
```diff
@@ -0,0 +1,27 @@
+package detect_cloud
+
+import (
+	"os"
+	"strings"
+)
+
+func detectContainer() string {
+	b, err := os.ReadFile("/proc/self/cgroup")
```
lixmal: I believe this one doesn't work with cgroup v2 anymore, it only contains something like `0::/`. Do we want to find an alternative?

lixmal (changes requested):

client/system/detect_cloud/detect.go
pascal-fischer: Don't really have a strong opinion about that so I will move it

lixmal (approved):

lixmal (approved):

lixmal (approved):

Issue #289, opened 2022-03-27
Exit node


FarisZR:
Tailscale/headscale supports using exit nodes for using the network as a full mesh VPN.

This feature is very important for me, as I want to have a single IP, which gives me full speed while I'm in LAN, while being accessible from outside the network.

i would be using it mostly from mobile, related #213

mrbluecoat:
I agree. A mobile client that supports always-on VPN and starts on boot linked to a centralized exit node would be heaven 😇

mrbluecoat:
Dropping by a year later -- Android client and cloud control panel look nice! Still no `0.0.0.0/0` route option though...

himekifee:
+1 for the feature.

codyro:
+1

the-infrequency:
+1

lorenzo95:
+1

morki83:
+1. The only thing missing before switching completely to netbird

tiagogbarbosa:
Any progress?

houpi:
+1

victor-rsibillon:
+1

alexunderboots:
+1

tribor:
exit nodes has been sheduled for Q3 in 2023
https://github.com/netbirdio/netbird/projects/2#card-85699571

still unsure if it is currently worked on. @braginini it would be nice to get some feedback on the progress. thanks

SamB-GB:
+1 for this

j007bond007:
+1 For this as well... it's really needed for mobile devices on untrusted networks, right now I need to run wireguard separately for the full tunnel and NetBird for when I just need overlay mode.

purple-emily:
I would really like this feature.

europacafe:
+1

PavelNiedoba:
+2

PatrickHuetter:
+1

PavelNiedoba:
Actually netbird is using COTURN tunneling, which does something very similar like exit nodes. You can access networks behind NAT, but this is very poorly documented.

PatrickHuetter:
> Actually netbird is using COTURN tunneling, which does something very similar like exit nodes. You can access networks behind NAT, but this is very poorly documented.

How does this work? How did you get this working? I want to have my private nodes connected via ipv6 and have access to ipv4 public internet via one exit node (that has both, ipv4 and ipv6 to public internet).

tribor:
[...]

PR #1726, opened 2024-03-18
Feature/exit nodes - Windows/macOS support

lixmal:
This PR enhances Netbird's routing capabilities on Windows and macOS, allowing us to use the default route on the client.

- Host routes are dynamically added/removed based on excluded IP addresses determined by hooks in custom dialers and listeners that are used for management traffic. This prevents management traffic from being routed through the VPN itself.
- Routes are removed only when their reference count reaches 0, ensuring routes persist while in use. 
-  Hooks into peer added/removed events manage routes based on peer IP addresses. This prevents routing loops and ensures dedicated connections to all peers.

Updates https://github.com/netbirdio/netbird/issues/289

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


vysecurity:
Are exit nodes added? Might switch from another big provider given that Netbird appears to have at least 2-3 times the speed.

yannrapenne:
> Are exit nodes added? Netbird appears to have at least 2-3 times the speed.

Not yet, but I think it's imminent.

pascal-fischer (approved):

PR #1727, opened 2024-03-18
Extend management to sync meta and posture checks with peer

bcmmbaga:
## Describe your changes

Extends the management behavior as follows:

- Upon peer login, the management will also include the posture checks applied on the peer, which are to be evaluated on the client.
- During synchronization, the management will receive the updated peer metadata, update them if there are any changes, and return the currently applied posture checks on the peer.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/grpcserver.go
```diff
@@ -134,7 +134,17 @@ func (s *GRPCServer) Sync(req *proto.EncryptedMessage, srv proto.ManagementServi
 		return err
 	}
 
-	peer, netMap, err := s.accountManager.SyncPeer(PeerSync{WireGuardPubKey: peerKey.String()})
+	if syncReq.GetMeta() == nil {
+		msg := status.Errorf(codes.FailedPrecondition,
+			"peer system meta has to be provided on sync. Peer %s, remote addr %s", peerKey.String(), realIP)
+		log.Warn(msg)
```
pappz: If I understand well the old clients do not send meta on sync request. Does it make sense to log this warning in this case?

Issue #1554, opened 2024-02-09
check latency to remote peers with same routing - so client peer is sent to the best remote peer based on actual ISP network conditions


felixn-unity:
**Is your feature request related to a problem? Please describe.**
When having multiple peers in GEO regions advertising the same routes, the client peer will connect to any of the peers in random/first-in-list (not sure how it is selected)

This can end up in client peer connecting to remote peer in another GEO region, meaning not optimal performance.

**Describe the solution you'd like**
Measure the latency from the client peer to all of the remote peers having the remote route being requested

**Describe alternatives you've considered**
We could maybe have user distribution based on users,  so client peer region and in that way is sent to the "correct/best" region

**Additional context**
No additional contect

nazarewk:
@felixn-unity @pascal-fischer I believe this is already handled and can be closed?

TechHutTV:
Latency-aware route selection landed in PR #1732 ("Add latency based routing"), shipped in v0.27.2. When multiple routing peers advertise the same network, the client now prefers the one with the lowest measured latency, which is exactly the optimisation requested here. Closing as fixed — please reopen if you still see clients picking a sub-optimal routing peer on the latest release.

PR #1732, opened 2024-03-20
Add latency based routing

pascal-fischer:
## Describe your changes
Now that we have the latency between peers available we can use this data to consider when choosing the beste route. This way the route with the routing peer with the lowes latency will be prefered over others with the same target network.

https://github.com/netbirdio/netbird/issues/1554

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/routemanager/client.go
+++ b/client/internal/routemanager/client.go
@@ -115,6 +115,8 @@ func (c *clientNetwork) getBestRouteFromStatuses(routePeerStatuses map[string]ro
 		latency := time.Second
 		if peerStatus.latency != 0 {
 			latency = peerStatus.latency
+		} else {
+			log.Warnf("peer %s has 0 latency", r.Peer)
 		}
 		tempScore += 1 - latency.Seconds()
 
--- a/client/internal/routemanager/client.go
+++ b/client/internal/routemanager/client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/netip"
-	"sort"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


client/internal/routemanager/client.go
```diff
@@ -140,6 +149,39 @@ func (c *clientNetwork) getBestRouteFromStatuses(routePeerStatuses map[string]ro
 	return chosen
 }
 
+// calculateLatencyScores calculates the latency scores for route based on the latency of the peers.
+// The output will return a map of peer ID to latency score. The lower the latency, the lower the score.
+// The highest latency will have a score of 1, the second highest will have a score of 2, and so on.
+// The lowest latency will have the highest score.
+func calculateLatencyScores(routePeerStatuses map[string]routerPeerStatus) map[string]int {
```
mlsmaycon: It might be better to work with a float score and convert the latency to float based on milliseconds.

This will prevent unforeseen cases like a  relayed connection might have a slightly lower latency than P2P

We still want P2P as it will provide top bandwidth

mlsmaycon (changes requested):
The calculation might leave peers in better conditions out of the choice?

E.g. if

client/internal/routemanager/client.go
mlsmaycon: e.g., 3 routing peers:

peer 1 with direct connection and latency of 0.29ms
peer 2 with a relayed remote connection, no proxy, and a latency of 0.26ms
peer 3 with local relayed connection and latency of 0.25ms


scores without account latency would be:

peer 1: 2
peer 2: 1
peer 3 0

with proposed latency calc: 
peer 1: 3
pee2: 3
peer3: 3

The decision will then be made by choosing the first evaluated peer.

pascal-fischer: Yes I see your point I will do a rework of that calculation

lixmal (approved):

mlsmaycon (approved):

PR #1586, opened 2024-02-15
unify Config generation, loading and updating

nazarewk:
## Describe your changes

There are 2 primary changes:
- (misc) allows passing custom system service name (helps https://github.com/NixOS/nixpkgs/pull/287236 )
- unifies configuration loading and updating into a single `Config.apply(ConfigInput)`:
	- has better logging (informs of everything it does)
	- accepts both empty `Config{}` and empty `InputConfig{}` filling in missing parts automatically
	- handling was slightly inconsistent between `createNewConfig()` and `update()`
	- added `InterfaceBlacklist` to the `InputConfig`

Basically when I tried to start the `netbird service run` with just `WgIface` and `WgPort` present in `config.json` it crashed, so I got to fixing stuff :)

## Issue ticket number and link

- partially addresses #1562 by making incomplete configs acceptable (currently it segfaults on missing pointers/values)
- to be used in https://github.com/NixOS/nixpkgs/pull/287236

### Checklist
- [x] Is it a bug fix
- [ ] Is a typo/documentation fix
- [x] Is a feature enhancement
- [x] It is a refactor
- [ ] Created tests that fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

PR #1734, opened 2024-03-21
support to configure extra blacklist of iface in "up" command

hoozecn:
## Describe your changes

we can configure extra blacklist of iface in "up" command instead of editing config file

### Checklist
- [*] Is a feature enhancement


hoozecn:
Yes. we use netbird client as Daemonset in Kubernetes so that the selected nodes will work as gateways of the cluster. the change in PR is enough for us and works well.

For your point, I'm happy to see to make it be configurable too in traditional way when we run `netbird up`  in a "normal" node for instance a bare-metal node.  I would like to check the codebase to see how to make it possible with gRPC.

But before that, do you think it worth a try?

braginini:
> Yes. we use netbird client as Daemonset in Kubernetes so that the selected nodes will work as gateways of the cluster. the change in PR is enough for us and works well.
> 
> For your point, I'm happy to see to make it be configurable too in traditional way when we run `netbird up` in a "normal" node for instance a bare-metal node. I would like to check the codebase to see how to make it possible with gRPC.
> 
> But before that, do you think it worth a try?

I will sync with the team tomorrow morning and get back to you after. Thank you!

dkrhodes:
Update the PR to configure in daemon mode

hoozecn:
@braginini hey man, thanks for you sync. I have update the pr to support the daemon mode. 

Just find another PR not merged #1586. Seems it's going to support to configure InterfaceBlacklist too. 

The ExtraIFaceBlacklist option make help people avoid to provide the default blacklist in the command line. But use InterfaceBlacklist will be more explicit. 

Which do you think is better?

pappz (approved):

PR #1749, opened 2024-03-26
Evaluate process check on client

bcmmbaga:
## Describe your changes

Evaluate the assigned process check paths by verifying their existence and if there is a running process with the specified path in the client, then synchronize metadata with management.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/system/info_linux.go
```diff
@@ -122,3 +124,51 @@ func sysInfo() (serialNumber string, productName string, manufacturer string) {
 	si.GetSysInfo()
 	return si.Product.Version, si.Product.Name, si.Product.Vendor
 }
+
+// getRunningProcesses returns a list of running process paths.
+func getRunningProcesses() ([]string, error) {
```
pappz: Duplicated codes between darwin, windows, Linux. Instead of copy+past, you could move this code into a new file and in the build tags you can define the list of supported operation systems.

pappz (changes requested):

pappz (approved):

PR #1841, opened 2024-04-12
Delete route from system only if added by the client

mlsmaycon:
This adds a flags to check if the addVPNRoute executed without issues before removing a route from the system

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/routemanager/client.go
+++ b/client/internal/routemanager/client.go
@@ -221,6 +221,8 @@ func (c *clientNetwork) removeRouteFromPeerAndSystem() error {
 				return fmt.Errorf("remove route %s from system, err: %v", c.network, err)
 			}
 			c.systemUpdated = false
+		} else {
+			log.Debugf("system is not updated for network %s, skiping removal from system route table", c.network)
 		}
 
 		if err := c.removeRouteFromWireguardPeer(c.chosenRoute.Peer); err != nil {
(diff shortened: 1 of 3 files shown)


client/internal/routemanager/systemops.go
```diff
@@ -256,8 +256,7 @@ func addNonExistingRoute(prefix netip.Prefix, intf string) error {
 		return fmt.Errorf("exists in route table: %w", err)
 	}
 	if ok {
-		log.Warnf("Skipping adding a new route for network %s because it already exists", prefix)
-		return nil
+		return fmt.Errorf("Skipping adding a new route for network %s because it already exists", prefix)
```
lixmal:
```suggestion
		return fmt.Errorf("skipping adding a new route for network %s because it already exists", prefix)
```

client/internal/routemanager/client.go
```diff
@@ -215,8 +216,11 @@ func (c *clientNetwork) removeRouteFromWireguardPeer(peerKey string) error {
 
 func (c *clientNetwork) removeRouteFromPeerAndSystem() error {
 	if c.chosenRoute != nil {
-		if err := removeVPNRoute(c.network, c.wgInterface.Name()); err != nil {
-			return fmt.Errorf("remove route %s from system, err: %v", c.network, err)
+		if c.systemUpdated {
```
lixmal, line 219: Can we add a debug log if this is false

lixmal:
This is most likely superseded by changes in https://github.com/netbirdio/netbird/pull/1943, but needs tests

mlsmaycon:
> This is most likely superseded by changes in #1943, but needs tests

Ok, let's close this one, as it would need a better approach anyway.

PR #1861, opened 2024-04-17
Redeem invite only when incoming user was invited

mlsmaycon:
## Describe your changes
This will reduce calls to cache and external IDP managers by reducing the redeem invite calls to only when the invited claim is received.
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -1379,7 +1379,7 @@ func (am *DefaultAccountManager) lookupCache(accountUsers map[string]userLoggedI
 	}
 
 	// if we know users that are not yet in cache more likely cache is outdated
-	if knownUsersCount > 0 && mustRefreshInviteStatus {
+	if knownUsersCount > 0 || mustRefreshInviteStatus {
 		log.Debugf("reloading cache with IDP manager. Users unknown to the cache: %d, Must refresh invite status? %t", knownUsersCount, mustRefreshInviteStatus)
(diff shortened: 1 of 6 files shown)


management/server/account.go
```diff
@@ -1345,31 +1349,38 @@ func (am *DefaultAccountManager) getAccountFromCache(accountID string, forceRelo
...
 			continue
 		}
 		log.Debugf("cache doesn't know about %s user", user)
 	}
 
 	// if we know users that are not yet in cache more likely cache is outdated
-	if knownUsersCount > 0 {
-		log.Debugf("cache doesn't know about %d users from store, reloading", knownUsersCount)
+	if knownUsersCount > 0 && mustRefreshInviteStatus {
+		log.Debugf("reloading cache with IDP manager. Users unkown to the cache: %d, Must refresh invite status? %t", knownUsersCount, mustRefreshInviteStatus)
```
pascal-fischer: Shouldn't this be an OR operator? otherwise, either `mustRefreshInviteStatus` or `knownUsersCount` is unnecessary as the break will make sure it is > 0. Also the break will falsify the log message message becasue `knownUsersCount` doesn't finish the loop

pascal-fischer (changes requested):

pascal-fischer (approved):

PR #2031, opened 2024-05-22
Fix checkpid on windows

hurricanehrndz:
## Describe your changes
[syscall.signal 0 does not work on windows as it is not implemented](https://pkg.go.dev/os?utm_source=godoc#Process.Signal)

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/ui/client_ui.go
+++ b/client/ui/client_ui.go
@@ -716,7 +716,7 @@ func openURL(url string) error {
 
 // checkPIDFile exists and return error, or write new.
 func checkPIDFile() error {
-	pidFile := path.Join(os.TempDir(), "wiretrustee-ui.pid")
+	pidFile := path.Join(os.TempDir(), "netbrid-ui.pid")
 	if piddata, err := os.ReadFile(pidFile); err == nil {
 		if pid, err := strconv.Atoi(string(piddata)); err == nil {
 			if pidExists(pid) {
--- a/client/ui/client_ui.go
+++ b/client/ui/client_ui.go
@@ -3,6 +3,7 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	_ "embed"
 	"flag"
@@ -473,7 +474,6 @@ func (s *serviceClient) updateStatus() error {
 		Stop:                backoff.Stop,
 		Clock:               backoff.SystemClock,
 	})
-
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


client/ui/client_ui.go
```diff
@@ -716,16 +716,43 @@ func openURL(url string) error {
...
+func pidExists(pid int) bool {
+	if runtime.GOOS != "windows" {
+		if process, err := os.FindProcess(pid); err == nil {
+			if err := process.Signal(syscall.Signal(0)); err == nil {
+				return true
+			}
+		}
+	}
+	// Prepare the command
+	cmd := exec.Command("tasklist", "/FI", fmt.Sprintf("PID eq %d", pid))
```
mlsmaycon, line 740: Can you use a function similar to [GetSystem32Command](https://github.com/netbirdio/netbird/blob/main/client/firewall/uspfilter/allow_netbird_windows.go#L103) to avoid path issues?

mlsmaycon (changes requested):
Thanks @hurricanehrndz for the contribution. Left one request.

client/ui/client_ui.go
```diff
@@ -716,16 +716,43 @@ func openURL(url string) error {
...
+
+	// Get the output
+	output, err := cmd.Output()
+	if err != nil {
+		return false
```
lixmal, line 746: Please `log.Tracef` the err and output here
```suggestion
        log.Tracef("Failed to list tasks: %v, output: %s", err, output)
		return false
```

client/ui/client_ui.go
```diff
@@ -716,16 +716,43 @@ func openURL(url string) error {
...
+	}
+
+	// Parse the output
+	scanner := bufio.NewScanner(strings.NewReader(string(output)))
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.Contains(line, fmt.Sprintf("%d", pid)) {
+			return true
+		}
+	}
```
lixmal, line 756: Can you handle `scanner.Err()` here?

client/ui/client_ui.go
```diff
@@ -716,16 +716,43 @@ func openURL(url string) error {
 
 // checkPIDFile exists and return error, or write new.
 func checkPIDFile() error {
-	pidFile := path.Join(os.TempDir(), "wiretrustee-ui.pid")
+	pidFile := path.Join(os.TempDir(), "netbrid-ui.pid")
```
lixmal:
```suggestion
	pidFile := path.Join(os.TempDir(), "netbird-ui.pid")
```

lixmal (changes requested):

client/ui/client_ui.go
hurricanehrndz: If scanner.Err actually returns an error the overall return of the function would be false, the only benefit to handling scanner.Err is to log an issue with the scanning, but it is not like the log is written to disk or anything so ...

lixmal: Right, but if there's an issue that would at least give us a hint when debugging the application from CLI

hurricanehrndz:
Sounds good, I will updates it accordingly

hurricanehrndz:
Although this patch somewhat works, I am not entirely happy, since a miss click will run multiple apps. I

PR #2060, opened 2024-05-27
Add mTLS support for SSO

Foosec:
## Describe your changes
Adds ClientCertPath and ClientCertKey to the config, to provide mTLS credentials for IDP authentication.
A usecase is having your IDP behind a mTLS terminating reverse proxy, thus reducing the risk of 0 day exploits against your IDP and adding an extra step of authentication. 
This is a feature that could be extended to cover netbird management as well.

I would be very happy to have someone more used to the code point out if something is miss placed. 
Ideally this could be added to the UI and possibly to mobile clients as well, but i did not want to go poking in the UI code without even an idea if this would be approved.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


Foosec:
@mlsmaycon Sorry for bothering you, its unclear to me whether this was closed due to the deletion of netbirdio:0.28.0, or is it not a wanted feature? Thank you!

mlsmaycon:
Hello @Foosec this was closed because of the other PR. My bad, I should've updated it before that. 

Can you reopen it against main? We should have more capacity to evaluate it in the coming weeks.

PR #2061, opened 2024-05-27
Improve login performance

pascal-fischer:
## Describe your changes
This PR reduced database reads by adding getAccountIDBy... to the store and adjusting the database locks.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/management/server/peer.go
+++ b/management/server/peer.go
@@ -556,6 +556,7 @@ func (am *DefaultAccountManager) SyncPeer(sync PeerSync, account *Account) (*nbp
 func (am *DefaultAccountManager) LoginPeer(login PeerLogin) (*nbpeer.Peer, *NetworkMap, error) {
 	accountID, err := am.Store.GetAccountIDByPeerPubKey(login.WireGuardPubKey)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


management/server/peer.go
```diff
@@ -574,19 +583,51 @@ func (am *DefaultAccountManager) LoginPeer(login PeerLogin) (*nbpeer.Peer, *Netw
...
+	var unlock func()
+	// early detection of what type of lock we need
+	if peer.UpdateMetaIfNew(login.Meta) || expired {
+		unlock = am.Store.AcquireAccountWriteLock(accountID)
+		isWriteLock = true
+	} else {
+		unlock = am.Store.AcquireAccountReadLock(accountID)
+		isWriteLock = false
+	}
+	log.Debugf("peer changed meta or login expired, acquiring write lock: %v", isWriteLock)
```
mlsmaycon: this log should be in line 609

pascal-fischer: No, the log is true/false so it will be logged either way. I could do 2 logs then it is more explicit

mlsmaycon (approved):

PR #2068, opened 2024-05-28
Generate network maps for routed network

bcmmbaga:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/proto/management.proto
```diff
@@ -342,32 +348,78 @@ message NameServer {
...
+  RuleAction action = 3;
+
+  // NetworkType of the routed network.
+  NetworkType networkType = 4;
+
+  // Network prefix for the routed network.
+  string destination = 5;
+
+  // Protocol of the routed network.
+  RuleProtocol protocol = 6;
```
lixmal, line 452: Do you think we can make this an int? It would require a slightly different implementation but offers more flexibility for the users

management/server/route.go
```diff
@@ -2,16 +2,53 @@ package server
...
+	// Action of the traffic when the rule is applicable
+	Action string
+
+	// Destination a network prefix for the routed traffic
+	Destination string
+
+	// Protocol of the traffic
+	Protocol string
+
+	// NetworkFamily string
```
lixmal: Comment doesn't match

management/server/route.go
```diff
@@ -323,6 +360,84 @@ func (am *DefaultAccountManager) ListRoutes(accountID, userID string) ([]*route.
...
+			rules = append(rules, &rr)
+			continue
+		}
+
+		for _, port := range rule.Ports {
+			pr := rr // clone rule and add set new port
+
+			p, err := strconv.ParseUint(port, 10, 16)
+			if err != nil {
+				log.Error("failed to parse port %s for rule: %s", port, rule.ID)
```
lixmal:
```suggestion
				log.Errorf("failed to parse port %s for rule: %s", port, rule.ID)
```

lixmal (changes requested):

management/proto/management.proto
bcmmbaga: Destination or protocol? If the protocol is already of int32 type

lixmal: Protocol is an enum, isn't it?

```
enum RuleProtocol {
  UNKNOWN = 0;
  ALL = 1;
  TCP = 2;
  UDP = 3;
  ICMP = 4;
}
```

bcmmbaga: Yes, it's an enum

lixmal: I was hoping to be able to use arbitrary protocol numbers from https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml, changing that later might be more difficult

bcmmbaga: RuleProtocol is shared between the normal firewall rules and routes firewall rules. Changing this to arbitrary protocol numbers will cause incompatibility with older clients

lixmal (approved):

management/proto/management.proto
```diff
@@ -342,32 +348,78 @@ message NameServer {
...
+  message Range {
+    uint32 start = 1;
+    uint32 end = 2;
+  }
+}
+
+// RouteFirewallRule signifies a firewall rule applicable for a routed network.
+message RouteFirewallRule {
+  // PeerIP IP address of the routing peer.
+  string peerIP = 1;
```
mlsmaycon: let's use network ranges here instead of IPs

management/server/route.go
```diff
@@ -354,7 +354,7 @@ func generateRouteFirewallRules(route *route.Route, rule *PolicyRule, groupPeers
 		}
 
 		baseRule := RouteFirewallRule{
-			PeerIP:      peer.IP.String(),
+			SourceRange: peer.IP.String(),
```
mlsmaycon: we need to add a /32 here

Issue #2046, opened 2024-05-24
SELF HOST not success, request /api/users get 502 Bad Gateway


shadowofmoo:
**Describe the problem**
use the quick start script `export NETBIRD_DOMAIN=netbird.*****.com; curl -fsSL https://github.com/netbirdio/netbird/releases/latest/download/getting-started-with-zitadel.sh | bash`  success deploy.

and visit https://netbird.****.com/peers is always loading, use F12 look up, find request /api/users get  502 Bad Gateway

I'm deploy netbird in tencent cloud, so maybe is network problem, pls help me

**To Reproduce**

None

**Expected behavior**

**Are you using NetBird Cloud?**

NO, self host in tencent cloud

**NetBird version**

latest version

**NetBird status -d output:**
```
[...]

PR #2085, opened 2024-06-01
Adding geolocation download log message.

keac:
## Describe your changes

Adding geolocation download prompt message.
When the network condition is poor, there are no prompts when downloading the geolocation file.

#2046 
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/geolocation/database.go
+++ b/management/server/geolocation/database.go
@@ -28,8 +28,10 @@ func loadGeolocationDatabases(dataDir string) error {
 		exists, _ := fileExists(path.Join(dataDir, file))
 		if exists {
 			continue
+		}else{
+			log.Infof("geo location file %s not found , file will be downloaded", file)
 		}
-
+		
 		switch file {
 		case MMDBFileName:
 			extractFunc := func(src string, dst string) error {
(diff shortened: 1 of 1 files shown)


management/server/geolocation/database.go
bcmmbaga, line 1: Thank you for your contribution. Everything looks great overall, but I noticed that the import statement for `log "github.com/sirupsen/logrus"` is missing. This omission is causing the tests to fail. Could you please add this import statement to ensure that the code passes all tests?

bcmmbaga (changes requested):

mlsmaycon:
@keac, any feedback on the requested change?

Anbool:
I don't think adding an IP library download prompt will solve this problem, it's worth mentioning that it would be nice if users could customize the domain pkgs.netbird.io to avoid not being able to access areas of the pkgs.netbird.io domain.

mlsmaycon (approved):

mlsmaycon:
> I don't think adding an IP library download prompt will solve this problem, it's worth mentioning that it would be nice if users could customize the domain pkgs.netbird.io to avoid not being able to access areas of the pkgs.netbird.io domain.

thanks for the feedback, @Anbool; this PR just adds logs to the download process. Can you open a feature request for the custom domain?

bcmmbaga (approved):

PR #2101, opened 2024-06-06
The netbird management server was not handling signed id_tokens other that RSA type

HarryKodden:
## Describe your changes

Added support for Elliptic Curve signed id_tokens

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


mlsmaycon:
Hello @HarryKodden could you provide an example of IDP that is not using RSA keys for us to test?

mlsmaycon:
@HarryKodden let us know if you need some help with the conflicts. 

Also, could you provide an example of IDP that is not using RSA keys for us to test?

HarryKodden:
> Hello @HarryKodden could you provide an example of IDP that is not using RSA keys for us to test?

Sorry for long time of silence!

To answer your question: I do not know of iDP's that are not using RSA keys, I do know of iDP's that offer both RSA and EC
For example: https://proxy.sram.surf.nl/OIDC/jwks

The point is, the current code is not in is not iterating over the offered keys until it succesfully finds a RSA public key. 
It always picks key[0] and either picks the X5c certificate value (if available) of creates a RSA PEM from the element[0]
https://github.com/netbirdio/netbird/blob/4bbedb5193f13f1e033acd4657a9c4f41d048bde/management/server/jwtclaims/jwtValidator.go#L222C10-L222C28

and this function treats that public key as a RSA key, no introspection to the key type.

This code alternative does look at the key type:
```
func getPublicKey(token *jwt.Token, jwks *Jwks) (interface{}, error) {
	// todo as we load the jkws when the server is starting, we should build a JKS map with the pem cert at the boot time

	for k := range jwks.Keys {
		if token.Header["kid"] != jwks.Keys[k].Kid {
			continue
		}

		if len(jwks.Keys[k].X5c) != 0 {
			cert := "-----BEGIN CERTIFICATE-----\n" + jwks.Keys[k].X5c[0] + "\n-----END CERTIFICATE-----"
			return jwt.ParseRSAPublicKeyFromPEM([]byte(cert))
		}

		if jwks.Keys[k].Kty == "RSA" {
			log.Debugf("generating PublicKey from RSA JWK")
			return getPublicKeyFromJWK_RSA(jwks.Keys[k])
		}
		if jwks.Keys[k].Kty == "EC" {
			log.Debugf("generating PublicKey from ECDSA JWK")
			return getPublicKeyFromJWK_ECDSA(jwks.Keys[k])
		}

		log.Debugf("Key Type: %s not yet supported, please raise ticket !", jwks.Keys[k].Kty)
	}

	return nil, errors.New("unable to find appropriate key")
}
```

mlsmaycon:
Thanks for the feedback @HarryKodden. If you have the chance, can you address the linter issue and the conflicting files? If you need support with that, let us know, and we will continue development.

PR #2119, opened 2024-06-11
Deprecate FileStore engine support

bcmmbaga:
## Describe your changes

Deprecates support for JSON file storage. 

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/store.go
```diff
@@ -76,50 +79,74 @@ func getStoreEngineFromEnv() StoreEngine {
...
+}
+
+// NewStore creates a new store based on the provided engine type, data directory, and telemetry metrics
+func NewStore(kind StoreEngine, dataDir string, metrics telemetry.AppMetrics) (Store, error) {
+	kind = getStoreEngine(kind)
+
+	if err := checkFileStoreEngine(kind, dataDir); err != nil {
+		return nil, err
+	}
+
```
pappz, line 107: It is not clear the `case handleUnsupportedStoreEngine` and the `checkFileStoreEngine`
The `handleUnsupportedStoreEngine` tries to migrate the json file to SQL. But before that, the checkFileStoreEngine will return with error because the JSON file exists.

management/server/store.go
```diff
@@ -160,25 +187,19 @@ func NewTestStoreFromJson(dataDir string) (Store, func(), error) {
...
 	}
 
-	cleanUp := func() {}
-
 	// if store engine is not set in the config we first try to evaluate NETBIRD_STORE_ENGINE
 	kind := getStoreEngineFromEnv()
 	if kind == "" {
 		// NETBIRD_STORE_ENGINE is not set we evaluate default based on dataDir
-		kind = getStoreEngineFromDatadir(dataDir)
+		kind = SqliteStoreEngine
```
pappz, line 193: The comment is not valid any more above this line

management/server/store.go
bcmmbaga: `checkFileStoreEngine` only checks if  `store.json` exists and  `jsonfile` is set as the store engine in `management.json`. It then returns an error with a link pointing to the steps for migrating from jsonfile to SQLite.

`handleUnsupportedStoreEngine` handles the case when the store engine is not supported. If there is an existing `store.json`, it automatically migrates to SQLite, but the user will need to update the store engine to SQLite. If there is no existing store.json, it returns an error indicating an unsupported store engine

pappz (approved):

PR #2144, opened 2024-06-18
Context creation and logging in management

pascal-fischer:
## Describe your changes
This PR creates 3 root contexts. It adds a logSource parameter to each context: 
- GRPC: For all requests coming via grpc api. For this case, the context will be extended with a requestID, the deviceID of the peer, and the accountID the peer belongs to
- HTTP: For all requests coming from REST api. For this case, the context will be extended with a requestID, the userID of the user querying the api, and the accountID the user is accessing (needs updating once we implement multitenancy).
- SYSTEM: which logs general system logs that occur during management operation e.g. startup, expiration trigger...

The PR updates the logger with a custom implementation that logs the additional fields.

> [!CAUTION]
> This PR will cause both management and client logger to lose their formatting. The formatting will be created in a separate PR before merging the feature branch.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


formatter/set.go
```diff
@@ -1,10 +1,20 @@
...
+	"os"
+
+	"github.com/sirupsen/logrus"
+)
 
 // SetTextFormatter set the text formatter for given logger.
 func SetTextFormatter(logger *logrus.Logger) {
-	logger.Formatter = NewTextFormatter()
+	var formatter logrus.Formatter
+	if os.Getenv("NB_LOG_FORMAT") == "json" {
```
pappz: The logic does not match with the function name. If it is a text formatter then do not provide json formatter.
Please create a new SetJsonFormatter and call the proper function based on env in the util/log.go

go.mod
```diff
@@ -47,6 +47,7 @@ require (
 	github.com/google/martian/v3 v3.0.0
 	github.com/google/nftables v0.0.0-20220808154552-2eca00135732
 	github.com/gopacket/gopacket v1.1.1
+	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
```
pappz: Is this package necessary? The v2 is not better?

formatter/set.go
pascal-fischer: Good point, I forgot about that.

go.mod
pascal-fischer: this was imported by the IDE. I fixed the imports and use the same version now

formatter/set.go
```diff
@@ -1,10 +1,21 @@
...
 
-import "github.com/sirupsen/logrus"
+import (
+	"github.com/sirupsen/logrus"
+)
 
 // SetTextFormatter set the text formatter for given logger.
 func SetTextFormatter(logger *logrus.Logger) {
-	logger.Formatter = NewTextFormatter()
+	formatter := NewTextFormatter()
```
pappz: Do we need to `formatter` variable?

pascal-fischer: not necessarily, I can remove it

pappz (approved):

PR #2149, opened 2024-06-18
Fix store migration on empty string

mlsmaycon:
When fetching empty values from the database to check for migration our parser failed to handle null strings preventing the service from start

This uses sql.NullString to handle that and check for empty string resulted from null data

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/migration/migration.go
+++ b/management/server/migration/migration.go
@@ -35,8 +35,8 @@ func MigrateFieldFromGobToJSON[T any, S any](db *gorm.DB, fieldName string) erro
 	}
 	tableName := stmt.Schema.Table
 
-	var item string
-	if err := db.Model(model).Select(oldColumnName).First(&item).Error; err != nil {
+	var sqliteItem sql.NullString
+	if err := db.Model(model).Select(oldColumnName).First(&sqliteItem).Order("id").Error; err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
 			log.Debugf("No records in table %s, no migration needed", tableName)
 			return nil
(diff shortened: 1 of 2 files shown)


management/server/sql_store_test.go
```diff
@@ -569,6 +570,26 @@ func TestMigrate(t *testing.T) {
...
+
+	prefix = netip.MustParsePrefix("12.0.0.0/24")
+	nRT := &route2.Route{
+		Network: prefix,
+		ID:      "route2",
+		Peer:    "peer-id",
+	}
+
+	err = store.db.Save(nRT).Error
+	require.NoError(t, err, "Failed to insert Gob data")
```
lixmal:
```suggestion
	require.NoError(t, err, "Failed to insert json nil slice data")
```

management/server/sql_store_test.go
```diff
@@ -569,6 +570,26 @@ func TestMigrate(t *testing.T) {
...
+
+	err = migrate(store.db)
+	require.NoError(t, err, "Migration should not fail on gob populated db")
```
lixmal:
```suggestion
	require.NoError(t, err, "Migration should not fail on json nil slice populated db")
```

management/server/migration/migration.go
```diff
@@ -35,19 +35,22 @@ func MigrateFieldFromGobToJSON[T any, S any](db *gorm.DB, fieldName string) erro
 	}
 	tableName := stmt.Schema.Table
 
-	var item string
-	if err := db.Model(model).Select(oldColumnName).First(&item).Error; err != nil {
+	var sqliteItem sql.NullString
+	if err := db.Model(model).Select(oldColumnName).First(&sqliteItem).Order("id").Error; err != nil {
```
lixmal:
```suggestion
	if err := db.Model(model).Select(oldColumnName).First(&sqliteItem).Error; err != nil {
```

lixmal (approved):

PR #2188, opened 2024-06-24
[client] Mtls support

Foosec:
## Describe your changes
Add support for mTLS to the SSO idp, this adds extra security and allows us to hide the identity provider behind a mTLS terminating proxy, reducing the effect of potential zero day exploits and adding another factor of authentication.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


bcmmbaga:
Thank you for your contribution! The code looks good and is a solid first step toward supporting mTLS, but we might adapt it in the future when we start to work on more complete mTLS support.

PR #2191, opened 2024-06-24
macOS HostDNS refactor

hurricanehrndz:
On macOS use the recommended settings for providing split DNS. As per the docs an empty string will force the configuration to be the default. In order to to support split DNS an additional service config is added for the local server and search domain settings.

see: https://developer.apple.com/documentation/devicemanagement/vpn/dns

## Describe your changes

This patch refactors split DNS on macOS devices. Current code tries to force mDNSResponder to use a specific search order for search domains. Unfortunately macOS will randomly choose which DNS server to forward queries to. This leads to unexpected behavior and results.

This patch does create an issue though. When Management and Signal domains are part of the search domains and/or all DNS is being forwarded to servers only available via the Wireguard tunnel, resuming connections after sleep can be a little unpredictable.

In order to increase stability a strategy needs to be employed that ensures Management/Admin/TURN URLs are resolvable via the local DNS servers. There are two means to go about this, in the upstream resolver code force lookup to Admin/Signal/TURN URLs to the the local resolver or you can apply a system configuration that forwards lookup for said URLs to the local resolver. I tested with the latter with hardcoded entries and it worked well. I am incline to provide another patch to achieve a similar outcome but wanted your opinion. 

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/client/internal/dns/host_darwin.go
+++ b/client/internal/dns/host_darwin.go
@@ -7,8 +7,10 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"net"
 	"net/netip"
 	"os/exec"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


hurricanehrndz:
will reopen soon

Issue #2196, opened 2024-06-25
Unable to connect to overlay network after hibernation or changing the docking status on macos


christian-schlichtherle:
**Describe the problem**

After waking up my MacBook Pro from hibernation, Netbird fails to connect to the overlay network again:

```shell
$ sudo netbird status
Error: status failed: create wg interface: resource busy
$ sudo ifconfig utun100
utun100: flags=8051<UP,POINTOPOINT,RUNNING,MULTICAST> mtu 1280
	inet 100.90.208.249 --> 100.90.208.249 netmask 0xff000000
	inet6 fe80::f22f:4bff:fe13:efad%utun100 prefixlen 64 scopeid 0x20 
	inet6 fe80::%utun100 prefixlen 64 scopeid 0x20 
	nd6 options=201<PERFORMNUD,DAD>
$ sudo netbird up
Connected
$ sudo netbird status
Error: status failed: create wg interface: resource busy
$ sudo ifconfig utun100
utun100: flags=8051<UP,POINTOPOINT,RUNNING,MULTICAST> mtu 1280
	inet 100.90.208.249 --> 100.90.208.249 netmask 0xff000000
	inet6 fe80::f22f:4bff:fe13:efad%utun100 prefixlen 64 scopeid 0x20 
	inet6 fe80::%utun100 prefixlen 64 scopeid 0x20 
	nd6 options=201<PERFORMNUD,DAD>
```

I have Netbird UI 0.28.2 installed. My machine has multiple network interfaces: When undocked, it's only connected to my WiFi. When being docked, it's also connected via a Thunderbolt Ethernet adapter with 10Gbps, so when docking/undocking my notebook the OS is actually roaming. This may or may not be related to the problem.

I can do `netbird down`, `netbird up`, but it doesn't reconnect again. Wireguard stays unconnected:

```shell
$ sudo wg show
interface: utun100
  public key: (not shown)
  private key: (hidden)
  listening port: 51820
```

The only way to reconnect is to reboot the OS, which is very annoying. Is there another workaround at least?

**Expected behavior**

Netbird should automatically reconnect to the overlay network after waking up the machine from hibernation or undocking/docking it.

**Are you using NetBird Cloud?**

Yes.

**NetBird version**

0.28.2

**NetBird status -d output:**

```shell
$ sudo netbird status -d
Error: status failed: create wg interface: resource busy
```

mlsmaycon:
[...]

Issue #2130, opened 2024-06-13
Network Monitor causes race conditions when multiple interfaces are connected


hurricanehrndz:
**Describe the problem**

When more than one network interface is connected to the same network, the network monitor will try and restart the engine multiple times in succession

This error will occur
```
2024-05-29T17:14:13-04:00 INFO client/internal/connect.go:339: using 43196 as wireguard port: 43195 is in use
2024-05-29T17:14:13-04:00 INFO client/internal/routemanager/manager.go:93: Routing setup complete
2024-05-29T17:14:13-04:00 ERRO client/internal/engine.go:313: failed creating tunnel interface utun4319: [resource busy]
2024-05-29T17:14:13-04:00 INFO client/internal/routemanager/manager.go:117: Routing cleanup complete
2024-05-29T17:14:13-04:00 ERRO client/internal/connect.go:261: error while starting Netbird Connection Engine: create wg interface: resource busy
```

**To Reproduce**

Using macOS device, connected via ethernet and wifi, wait, disconnect, watch netbird fail

**Expected behavior**

For a resource conflict and race condition not to occur

**Are you using NetBird Cloud?**

No

**NetBird version**

0.28/master

** Possible Solution **

Add a de-bounce mechanism to avoid multiple engine restarts within 3s of a restart 
```
From a1f8b0ed4c5d49bb0c00243f5b5a11becae3cf1e Mon Sep 17 00:00:00 2001
From: Carlos Hernandez <carlos@hrndz.ca>
Date: Fri, 31 May 2024 08:16:30 -0600
Subject: [PATCH] Ignore route changes if last was less than 2s ago

---
 client/internal/engine.go | 71 ++++++++++++++++++++++++++-------------
 1 file changed, 48 insertions(+), 23 deletions(-)

diff --git a/client/internal/engine.go b/client/internal/engine.go
index 2cbde13a..915e3ea0 100644
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -152,6 +152,8 @@ type Engine struct {
 	wgProbe     *Probe
 
 	wgConnWorker sync.WaitGroup
+
+	lastNetworkChangeTimestamp *NetworkChangeTimestamp
 }
 
 // Peer is an instance of the Connection Peer
@@ -160,6 +162,17 @@ type Peer struct {
 	WgAllowedIps string
 }
 
+// NetworkChangeTimestamp is a thread-safe structure that holds the timestamp of the last network change
[...]

PR #2197, opened 2024-06-25
Debounce network changes

hurricanehrndz:
## Describe your changes
Block restarting netbird engine for 3s when network change detected. This will help prevent Netbird from attempting to create multiple wiregurad interfaces when restarting the engine in rapid succession

#2196 
#2130 

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -160,7 +160,8 @@ type Engine struct {
 	relayProbe  *Probe
 	wgProbe     *Probe
 
-	wgConnWorker sync.WaitGroup
+	wgConnWorker               sync.WaitGroup
+	lastNetworkChangeTimestamp *NetworkChangeTimestamp
 
 	// checks are the client-applied posture checks that need to be evaluated on the client
 	checks []*mgmProto.Checks
@@ -170,6 +171,12 @@ type Engine struct {
 type Peer struct {
 	WgPubKey     string
 	WgAllowedIps string
+}
+
+// NetworkChangeTimestamp is a thread-safe structure that holds the timestamp of the last network change
+type NetworkChangeTimestamp struct {
+	mu    sync.Mutex
+	value int64
 }
 
 // NewEngine creates a new Connection Engine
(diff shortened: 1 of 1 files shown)


hurricanehrndz:
Please feel free to make changes as you fit

client/internal/engine.go
```diff
@@ -1474,7 +1482,20 @@ func (e *Engine) startNetworkMonitor() {
 	e.networkMonitor = networkmonitor.New()
 	go func() {
 		err := e.networkMonitor.Start(e.ctx, func() {
-			log.Infof("Network monitor detected network change, restarting engine")
+			log.Infof("Network monitor: detected network change")
+			e.lastNetworkChangeTimestamp.mu.Lock()
+			defer e.lastNetworkChangeTimestamp.mu.Unlock()
+			now := time.Now().Unix()
+			if e.lastNetworkChangeTimestamp.value+3 > now || e.cancel == nil {
```
christian-schlichtherle, line 1489: Maybe one second is enough?

hurricanehrndz: This doesn't add any additional delay. What it does is is stop restarting the engine if the last time it restarted the engine was less than 3 second. You could decrease it but it will not make bringing up the connection any faster

christian-schlichtherle (approved):
Reviewed and tested - thanks for this critical fix!

pappz:
@hurricanehrndz Thank you for the PR! I have a question.
You made a new, custom struct for locking the time value.
In your change, you are doing multiple operations under the locked state. I am not sure if that is necessary or not. If not then it could be better to use an atomic store. The atomic provides options to store value in atomic way. https://pkg.go.dev/sync/atomic

hurricanehrndz:
> @hurricanehrndz Thank you for the PR! I have a question. You made a new, custom struct for locking the time value. In your change, you are doing multiple operations under the locked state. I am not sure if that is necessary or not. If not then it could be better to use an atomic store. The atomic provides options to store value in atomic way. https://pkg.go.dev/sync/atomic

I think it is necessary because the bug is caused by a race condition. When a device has multiple network interfaces all with a default route to the internet, the callback function will be called multiple times simultaneously.

lixmal:
Hi @hurricanehrndz,

we want to go with a slightly different solution that doesn't ignore potential changes for 3 seconds.
I'll close this PR and create a new one. Thank you for bringing the issue to our attention and the inspiration.

hurricanehrndz:
Please tag me on the new PR and I can test it for you.

PR #2204, opened 2024-06-26
Postgres getting started script Update

r0b2g1t:
The PR includes the option to set the Zitadel Database variable and also set Postgres as default if the variable is unset.

### Checklist
- [x] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/infrastructure_files/getting-started-with-zitadel.sh
+++ b/infrastructure_files/getting-started-with-zitadel.sh
@@ -464,11 +464,14 @@ initEnvironment() {
   fi
 
   if [[ $ZITADEL_DATABASE == "" ]]; then
-    CRDB=$(renderDockerComposeCockroachDB)
-    ZITADEL_DB_ENV=$(renderZidatelCockroachDBEnv)
-  elif [[ $DATABASE == "postgres" ]]; then
+    echo "Use Postgres as default Zitadel database."
+    echo "For using CockroachDB please the environment variable 'export ZITADEL_DATABASE=cockroach'."
     CRDB=$(renderDockerComposePostgres)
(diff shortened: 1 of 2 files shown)


infrastructure_files/getting-started-with-zitadel.sh
```diff
@@ -430,6 +430,12 @@ initEnvironment() {
   TURN_MAX_PORT=65535
   TURN_EXTERNAL_IP_CONFIG=$(get_turn_external_ip)
 
+  if [ ! "$ZITADEL_DATABASE" ]; then
```
mlsmaycon, line 433: Hello @r0b2g1t this was resolved with the change we pushed in https://github.com/netbirdio/netbird/pull/2194

r0b2g1t: I saw the merges, but with the set of env variable ZITADEL_DATABASE to postgres it was not able to for to us it, which was discussed in my last PR. with checking the the variable explicit of the content postgres you are more flexible for further changes and the use of other databases and there differentiation.

mlsmaycon: The effect is similar since we want to default to Postgres; we can initialize the ZITADEL_DATABASE if it doesn't match the cockroach with Postgres, and, in the future, do the same when there are other engines supported. So we would just add another condition before the final else for each new engine.

r0b2g1t: Yes, thats true it works also in this way. Thanks for your explanation. :-)

r0b2g1t:
The current solution works also in that way.

PR #2229, opened 2024-07-02
Avoid using custom dialers for proxy when not mobile

mlsmaycon:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/wgproxy/proxy_ebpf.go
+++ b/client/internal/wgproxy/proxy_ebpf.go
@@ -18,8 +18,9 @@ import (
 
 	"github.com/netbirdio/netbird/client/internal/ebpf"
 	ebpfMgr "github.com/netbirdio/netbird/client/internal/ebpf/manager"
-	nbnet "github.com/netbirdio/netbird/util/net"
 )
+
+const loopbackIPv4Addres = "127.0.0.1"
 
 // WGEBPFProxy definition for proxy with EBPF support
 type WGEBPFProxy struct {
@@ -72,10 +73,10 @@ func (p *WGEBPFProxy) listen() error {
 
 	addr := net.UDPAddr{
 		Port: wgPorxyPort,
-		IP:   net.ParseIP("127.0.0.1"),
+		IP:   net.ParseIP(loopbackIPv4Addres),
 	}
 
-	conn, err := nbnet.ListenUDP("udp", &addr)
+	conn, err := net.ListenUDP("udp", &addr)
 	if err != nil {
 		cErr := p.Free()
 		if cErr != nil {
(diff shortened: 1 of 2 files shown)


client/internal/wgproxy/proxy_userspace.go
```diff
@@ -35,7 +36,11 @@ func (p *WGUserSpaceProxy) AddTurnConn(turnConn net.Conn) (net.Addr, error) {
 	p.remoteConn = turnConn
 
 	var err error
-	p.localConn, err = nbnet.NewDialer().DialContext(p.ctx, "udp", fmt.Sprintf(":%d", p.localWGListenPort))
+	if runtime.GOOS == "ios" || runtime.GOOS == "android" {
+		p.localConn, err = nbnet.NewDialer().DialContext(p.ctx, "udp", fmt.Sprintf(":%d", p.localWGListenPort))
+	} else {
+		p.localConn, err = net.Dial("udp", fmt.Sprintf(":%d", p.localWGListenPort))
```
lixmal, line 42: Can we use DialContext here regardless?

client/internal/wgproxy/proxy_ebpf.go
```diff
@@ -72,10 +73,10 @@ func (p *WGEBPFProxy) listen() error {
 
 	addr := net.UDPAddr{
 		Port: wgPorxyPort,
-		IP:   net.ParseIP("127.0.0.1"),
+		IP:   net.ParseIP(loopbackIPv4Addres),
```
lixmal, line 76: Maybe something like `net.IP{127, 0, 0, 1}` is a bit better here.

client/internal/wgproxy/proxy_ebpf.go
```diff
@@ -249,27 +250,11 @@ func (p *WGEBPFProxy) prepareSenderRawSocket() (net.PacketConn, error) {
 		return nil, fmt.Errorf("binding to lo interface failed: %w", err)
 	}
 
-	// Set the fwmark on the socket.
-	err = nbnet.SetSocketOpt(fd)
-	if err != nil {
-		return nil, fmt.Errorf("setting fwmark failed: %w", err)
-	}
```
lixmal, line 256: This is the only block that should be removed. The other stuff should stay, we should still check if os.NewFile returns nil.

client/internal/wgproxy/proxy_userspace.go
```diff
@@ -35,7 +36,11 @@ func (p *WGUserSpaceProxy) AddTurnConn(turnConn net.Conn) (net.Addr, error) {
 	p.remoteConn = turnConn
 
 	var err error
-	p.localConn, err = nbnet.NewDialer().DialContext(p.ctx, "udp", fmt.Sprintf(":%d", p.localWGListenPort))
+	if runtime.GOOS == "ios" || runtime.GOOS == "android" {
```
lixmal, line 39: This should be android only, iOS already calls `net.Dial` underneath anyway. I'm not sure about Android either, it likely has `127.0.0.0/8` in the `local` table as well, so there might be no need to protect the socket there.

PR #2244, opened 2024-07-08
[relay] Feature/relay integration

pappz:
## Describe your changes

- Add Relay server implementation
- Integrate the Relay server
- Refactor the conn worker logic
- Refactor the offer/answer sending logic
- Add fallback logic to Relay connection in case failed to use ICE p2p connection

### Relay package
- Authentication (client/server)
- Helathcheck and reconnection (guard.go). The health check is in a separated package but the usage of it is also relevant.
- Protocol life circle betwen the Relay server - Relay client. Review to process the message types (messages package).
	- Hello (auth msg),
	- package transfer logic and internal store of the peers
	- close procedure
- Roaming between the server instances in case if the peers are connected on different servers. client/manager.go
- Review the command line (cmd) package what boot TLS, listeners, metrics components
- TLS managment code review. It is relevant for the signal and managmenet code also because I reuse these function in all server services.
- Metrics

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


relay/cmd/main.go
```diff
@@ -0,0 +1,196 @@
...
+
+func execute(cmd *cobra.Command, args []string) error {
+	cfg, err := loadConfig(cfgFile)
+	if err != nil {
+		return fmt.Errorf("failed to load config: %s", err)
+	}
+
+	err = cfg.Validate()
+	if err != nil {
+		log.Errorf("invalid config: %s", err)
```
lixmal: This should return the error instead and drop the os.Exit

relay/server/server.go
```diff
@@ -0,0 +1,80 @@
...
+	wslErr := r.wSListener.Listen(r.relay.Accept)
+	if wslErr != nil {
+		log.Errorf("failed to bind ws server: %s", wslErr)
+	}
+
+	return wslErr
+}
+
+// Close stops the relay server. If there are active connections, they will be closed gracefully. In case of a timeout,
+// the connections will be forcefully closed.
```
lixmal, line 62: Since this is shutting down the server gracefully I suggest to name this Shutdown and accept the context (instead of creating one inside) to align with how the http package works.

The same approach might be applicable to the listener Close method in line 67.

relay/server/peer.go
```diff
@@ -0,0 +1,181 @@
...
+	defer cancel()
+
+	writeDone := make(chan struct{})
+	var (
+		n   int
+		err error
+	)
+
+	go func() {
+		_, err = p.conn.Write(buf)
```
lixmal, line 143: This is not setting n.

Also consider using p.conn.SetWriteDeadline additionally, so we don't waste resources waiting or writing if the timeout has already elapsed. It might also be undesired to let the write finish if we've already abandoned it.

relay/server/peer.go
```diff
@@ -0,0 +1,181 @@
...
+		close(writeDone)
+	}()
+
+	select {
+	case <-ctx.Done():
+		return 0, fmt.Errorf("write operation timed out")
```
lixmal: We can return ctx.Err() to simpify

relay/server/relay.go
```diff
@@ -0,0 +1,163 @@
...
+	peers := r.store.Peers()
+	for _, peer := range peers {
+		wg.Add(1)
+		go func(p *Peer) {
+			p.CloseGracefully(ctx)
+			wg.Done()
+		}(peer)
+	}
+	wg.Wait()
+	r.metricsCancel()
```
lixmal, line 145: Would it make sense to wait for the metrics to finish as well? I assume the scraper won't have the opportunity to scrape the metrics anymore anyway.

relay/server/relay.go
```diff
@@ -0,0 +1,163 @@
...
+// It closes the connection with all peers in gracefully and stops accepting new connections.
+func (r *Relay) Close(ctx context.Context) {
+	log.Infof("close connection with all peers")
+	r.closeMu.Lock()
+	wg := sync.WaitGroup{}
+	peers := r.store.Peers()
+	for _, peer := range peers {
+		wg.Add(1)
+		go func(p *Peer) {
+			p.CloseGracefully(ctx)
```
lixmal, line 140: It might be better to accumulate the errors inside this method (e.g. using github.com/hashicorp/go-multierror), collect them here and return an error from this `Close` method (that should probably also be name Shutdown).
That way the logs are not scattered, but it might impact memory if there are a ton of clients. What do you think?

relay/cmd/main.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed creating LetsEncrypt cert manager: %v", err)
+	}
+	return certManager.TLSConfig(), nil
+}
+
+func main() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
```
lixmal: Should we print the error here, e.g. `log.Fatalf`?

relay/cmd/main.go
```diff
@@ -0,0 +1,196 @@
...
+		return fmt.Errorf("failed to bind server: %s", err)
+	}
+
+	waitForExitSignal()
+
+	err = srv.Close()
+	if err != nil {
+		return fmt.Errorf("failed to close server: %s", err)
+	}
+	return nil
```
lixmal: We should shutdown metricsServer somewhere here as well, right?

relay/server/listener/ws/listener.go
```diff
@@ -0,0 +1,88 @@
...
+	l.server = &http.Server{
+		Addr:      l.Address,
+		Handler:   mux,
+		TLSConfig: l.TLSConfig,
+	}
+
+	log.Infof("WS server is listening on address: %s", l.Address)
+	var err error
+	if l.TLSConfig != nil {
+		err = l.server.ListenAndServeTLS("", "")
```
lixmal, line 42: Should this pass the cert and key files?

relay/cmd/main.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create relay server: %v", err)
+	}
+	log.Infof("server will be available on: %s", srv.InstanceURL())
+	err = srv.Listen(srvListenerCfg)
+	if err != nil {
+		return fmt.Errorf("failed to bind server: %s", err)
+	}
+
+	waitForExitSignal()
```
lixmal: This here is ineffective because it only starts intercepting signals after srv.Listen has already returned.

Something like this would work

```
	osSigs := make(chan os.Signal, 1)
	signal.Notify(osSigs, syscall.SIGINT, syscall.SIGTERM)
	
	go func() {
		err = srv.Listen(srvListenerCfg)
		if err != nil {
			log.Fatalf("failed to start server: %v", err)
		}
	}()

	<-osSigs
	err = srv.Close()
	if err != nil {
		return fmt.Errorf("failed to close server: %s", err)
	}
```

Or move srv.Close in a goroutine and leave srv.Listen in the main goroutine.
(Plus improve reporting errors)

relay/cmd/main.go
pappz: It has been fixed in a child PR https://github.com/netbirdio/netbird/pull/2456

relay/server/listener/ws/listener.go
pappz: No. We manage the TLS certs in the TLSConfig.GetCertificate function. In this case, the empty strings are the expected.

relay/server/relay.go
pappz: It does not wait, it is just a context cancel. It returns back immediately.

lixmal: I'm aware. I'm asking if we should wait for the metrisc to finish their work e.g. with a done channel

relay/server/peer.go
pappz: The WS library always returns back with the size of the buf. So makes no sense to handle n.
```
	err := c.Conn.Write(ctx, websocket.MessageBinary, b)
	return len(b), err
```

pappz: The SetWriteDeadline is not implemented. This function is a workaround for it. The Websocket library that we are using provides an option to wrap the Webscoket into net.Conn type but it provides less good performance in the write operation. https://github.com/coder/websocket/blob/3dd723ae69f00f042c4fd6961ed6cee3e9e11659/netconn.go#L208

So because we know this is a read/write heavy  I choose a custom write deadline function.

relay/server/relay.go
pappz: In the peer, we have log with tag context (id of the peer). If we extend the logs here then we lose this feature.

relay/auth/validator.go
```diff
@@ -0,0 +1,6 @@
+package auth
+
+// Validator is an interface that defines the Validate method.
+type Validator interface {
```
lixmal, line 6: Best practice in Go is to declare interfaces close to the consumer.
This way the consumer (package) doesn't need to import the implmenting package(s) which also helps with avoiding cyclic dependencies.
So in this case best to move it to the `server` package

relay/auth/allow_all.go
```diff
@@ -0,0 +1,10 @@
+package auth
+
+// AllowAllAuth is a Validator that allows all connections.
+// Used this for testing purposes only.
+type AllowAllAuth struct {
```
lixmal: This should probably live in its own package

relay/auth/hmac/store.go
```diff
@@ -0,0 +1,34 @@
...
+func (a *TokenStore) UpdateToken(token *Token) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	if token == nil {
+		return
+	}
+
+	t, err := marshalToken(*token)
+	if err != nil {
+		log.Errorf("failed to marshal token: %s", err)
```
lixmal: Better to return this error instead

relay/auth/hmac/validator.go
```diff
@@ -0,0 +1,27 @@
...
+	return &TimedHMACValidator{
+		ta,
+	}
+}
+
+func (a *TimedHMACValidator) Validate(credentials any) error {
+	b := credentials.([]byte)
+	c, err := unmarshalToken(b)
+	if err != nil {
+		log.Errorf("failed to unmarshal token: %s", err)
```
lixmal: This error should be returned instead of being logged using fmt.Errorf("... :%w", err)

relay/auth/hmac/token.go
```diff
@@ -0,0 +1,105 @@
...
+	Payload   string
+	Signature string
+}
+
+func marshalToken(token Token) ([]byte, error) {
+	buffer := bytes.NewBuffer([]byte{})
+	encoder := gob.NewEncoder(buffer)
+	err := encoder.Encode(token)
+	if err != nil {
+		log.Errorf("failed to marshal token: %s", err)
```
lixmal: This error should be wrapped and returned instead of being logged additionally

relay/auth/hmac/token.go
```diff
@@ -0,0 +1,105 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type Token struct {
+	Payload   string
+	Signature string
+}
+
+func marshalToken(token Token) ([]byte, error) {
+	buffer := bytes.NewBuffer([]byte{})
```
lixmal: This can simply be `var buffer bytes.Buffer`

relay/auth/hmac/token.go
```diff
@@ -0,0 +1,105 @@
...
+		secret:     secret,
+		timeToLive: timeToLive,
+	}
+}
+
+// GenerateToken generates new time-based secret token - basically Payload is a unix timestamp and Signature is a HMAC
+// hash of a timestamp with a preshared TURN secret
+func (m *TimedHMAC) GenerateToken() (*Token, error) {
+	timeAuth := time.Now().Add(m.timeToLive).Unix()
+	timeStamp := fmt.Sprint(timeAuth)
```
lixmal: this can be 	`strconv.FormatInt(timeAuth, 10)`, probably a bit more efficient

relay/auth/hmac/token.go
```diff
@@ -0,0 +1,105 @@
...
+func (m *TimedHMAC) Validate(token Token) error {
+	expectedMAC, err := m.generate(token.Payload)
+	if err != nil {
+		return err
+	}
+
+	expectedSignature := base64.StdEncoding.EncodeToString(expectedMAC)
+
+	if !hmac.Equal([]byte(expectedSignature), []byte(token.Signature)) {
+		return fmt.Errorf("signature mismatch")
```
lixmal, line 81: Probably better to declare this as package var: `var ErrSignatureMismatch = errors.New("signature mismatch")` so it can be used to determine the error from the caller

relay/auth/hmac/token.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	timeAuthInt, err := strconv.ParseInt(token.Payload, 10, 64)
+	if err != nil {
+		return fmt.Errorf("invalid payload: %s", err)
```
lixmal: use %w to wrap the error

relay/auth/hmac/token.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	if time.Now().Unix() > timeAuthInt {
+		return fmt.Errorf("expired token")
```
lixmal, line 90: Same here with `var ...`

relay/auth/hmac/validator.go
```diff
@@ -0,0 +1,27 @@
...
+
+func NewTimedHMACValidator(secret string, duration time.Duration) *TimedHMACValidator {
+	ta := NewTimedHMAC(secret, duration)
+	return &TimedHMACValidator{
+		ta,
+	}
+}
+
+func (a *TimedHMACValidator) Validate(credentials any) error {
+	b := credentials.([]byte)
```
lixmal: This is not safe.

we can do any of the following to fix it

- type assert with `b, ok := ...` and return an error
- let unmarshalToken handle the `any` with its gob encoder
- only accept `credentials []byte` (or `Validate[T ~[]byte](credentials T)` if derived types are involved)

relay/auth/hmac/token.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	return nil
+}
+
+func (m *TimedHMAC) generate(payload string) ([]byte, error) {
+	mac := hmac.New(sha1.New, []byte(m.secret))
```
lixmal: Even though hmac-sha1 is unbroken, can we use something newer?

relay/auth/hmac/token.go
```diff
@@ -0,0 +1,105 @@
...
+	_, err := mac.Write([]byte(payload))
+	if err != nil {
+		log.Errorf("failed to generate token: %s", err)
```
lixmal: Return this error wrapped as well

relay/auth/hmac/token.go
```diff
@@ -0,0 +1,105 @@
...
+		Payload:   timeStamp,
+		Signature: base64.StdEncoding.EncodeToString(checksum),
+	}, nil
+}
+
+// Validate checks if the token is valid
+func (m *TimedHMAC) Validate(token Token) error {
+	expectedMAC, err := m.generate(token.Payload)
+	if err != nil {
+		return err
```
lixmal, line 75: let's add some context to this error like "generate: %w"

relay/auth/hmac/token.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	expectedSignature := base64.StdEncoding.EncodeToString(expectedMAC)
```
lixmal, line 78: Do we need to store these signatures as strings or can we use []byte instead?
If we need strings maybe it's better to decode the existing signature instead and compare the raw hmac values.

relay/auth/hmac/store.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"sync"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// TokenStore is a simple in-memory store for token
+// With this can update the token in thread safe way
+type TokenStore struct {
```
lixmal, line 11: This should probably live in a different package than hmac (together with `Token`) since it's accessed outside the hmac package

relay/healthcheck/receiver.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+var (
+	heartbeatTimeout = healthCheckInterval + 3*time.Second
+)
+
+// Receiver is a healthcheck receiver
+// It will listen for heartbeat and check if the heartbeat is not received in a certain time
+// If the heartbeat is not received in a certain time, it will send a timeout signal and stop to work
+// It will also stop if the context is canceled
```
lixmal: This line can be removed, ctx is internal only (replace with Stop method instruction)

relay/messages/message.go
```diff
@@ -0,0 +1,221 @@
...
+import (
+	"bytes"
+	"encoding/gob"
+	"fmt"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	MsgTypeHello         MsgType = 0
```
lixmal: Probably better to start with `Unknown = 0`, to be able to distinguish the types from the default value

relay/messages/message.go
```diff
@@ -0,0 +1,221 @@
...
+	MsgTypeHelloResponse MsgType = 1
+	MsgTypeTransport     MsgType = 2
+	MsgTypeClose         MsgType = 3
+	MsgTypeHealthCheck   MsgType = 4
+
+	sizeOfMsgType       = 1
+	sizeOfMagicBye      = 4
```
lixmal: sizeOfMagicBy**t**e?

relay/messages/message.go
```diff
@@ -0,0 +1,221 @@
...
+	headerSizeTransport = sizeOfMsgType + IDSize                  // 1 byte for msg type, IDSize for peerID
+	headerSizeHello     = sizeOfMsgType + sizeOfMagicBye + IDSize // 1 byte for msg type, 4 byte for magic header, IDSize for peerID
+
+	MaxHandshakeSize = 8192
+)
+
+var (
+	ErrInvalidMessageLength = fmt.Errorf("invalid message length")
```
lixmal: errors.New might be preferable here

relay/server/peer.go
```diff
@@ -0,0 +1,180 @@
...
+	for {
+		n, err := p.conn.Read(buf)
+		if err != nil {
+			if err != io.EOF {
+				p.log.Errorf("failed to read message: %s", err)
+			}
+			return
+		}
+
+		msg := buf[:n]
```
lixmal, line 71: Any chance we end up with len(msg) == 0 here because there was no payload? That would make `DetermineClientMsgType` panic

relay/server/peer.go
```diff
@@ -0,0 +1,180 @@
...
+		select {
+		case <-hc.HealthCheck:
+			_, err := p.Write(messages.MarshalHealthcheck())
+			if err != nil {
+				p.log.Errorf("failed to send healthcheck message: %s", err)
+				return
+			}
+		case <-hc.Timeout:
+			p.log.Errorf("peer healthcheck timeout")
+			_ = p.conn.Close()
```
lixmal: Should we log this error at least?

relay/server/peer.go
```diff
@@ -0,0 +1,180 @@
...
+		switch msgType {
+		case messages.MsgTypeHealthCheck:
+			hc.OnHCResponse()
+		case messages.MsgTypeTransport:
+			p.metrics.TransferBytesRecv.Add(ctx, int64(n))
+			p.metrics.PeerActivity(p.String())
+			p.handleTransportMsg(msg)
+		case messages.MsgTypeClose:
+			p.log.Infof("peer exited gracefully")
+			_ = p.conn.Close()
```
lixmal: Should we log the error?

relay/server/peer.go
```diff
@@ -0,0 +1,180 @@
...
+		if err != nil {
+			if err != io.EOF {
+				p.log.Errorf("failed to read message: %s", err)
+			}
+			return
+		}
+
+		msg := buf[:n]
+
+		msgType, err := messages.DetermineClientMsgType(msg)
```
lixmal: I suggest also adding the protocol version in the first byte(s) before anything else

relay/healthcheck/sender.go
```diff
@@ -0,0 +1,71 @@
...
+
+	for {
+		select {
+		case <-ticker.C:
+			hc.HealthCheck <- struct{}{}
+		case <-timeoutTimer.C:
+			hc.Timeout <- struct{}{}
+			return
+		case <-hc.ack:
+			timeoutTimer.Stop()
```
lixmal: Should this reset the timer instead? Otherwise we never timeout after the first ack

relay/healthcheck/sender.go
```diff
@@ -0,0 +1,71 @@
+package healthcheck
+
+import (
+	"context"
+	"time"
+)
+
+var (
+	healthCheckInterval = 25 * time.Second
```
lixmal, line 9: these can be consts

relay/server/peer.go
lixmal: >The WS library always returns back with the size of the buf. So makes no sense to handle n.

My point is we're returning `n, err` but never set n or use something else

relay/server/listener/ws/listener.go
```diff
@@ -0,0 +1,83 @@
...
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", l.onAccept)
+
+	l.server = &http.Server{
+		Addr:      l.Address,
+		Handler:   mux,
+		TLSConfig: l.TLSConfig,
+	}
+
+	log.Infof("WS server is listening on address: %s", l.Address)
```
lixmal: This should be moved further down if there was no error

relay/server/listener/ws/listener.go
```diff
@@ -0,0 +1,83 @@
...
+func (l *Listener) onAccept(w http.ResponseWriter, r *http.Request) {
+	wsConn, err := websocket.Accept(w, r, nil)
+	if err != nil {
+		log.Errorf("failed to accept ws connection from %s: %s", r.RemoteAddr, err)
+		return
+	}
+
+	rAddr, err := net.ResolveTCPAddr("tcp", r.RemoteAddr)
+	if err != nil {
+		_ = wsConn.Close(websocket.StatusInternalError, "internal error")
```
lixmal: Should we ignore these errors?

pappz: It is meaning less information. The code never touches again this wsConn instance in case of error.

relay/server/listener/ws/listener.go
pappz: The next line (listen) will block the running

relay/server/peer.go
pappz: The p.conn is a net.Conn type. It expect the `n, err`. We would like to grant an option to replace or extend the protocols with more than WebScoket. Like raw UDP for example or another WS library. In this case it I a good thing to handle `n, err`. So the current usage make no sense to keep it but we know in the future needed, so easier to keep the general net.Conn interface.

relay/auth/hmac/validator.go
pappz: Personally, I like to log the error where it happened. The reason is, in the logs the error will print out the log line number, so I do not need to look for where come from the error.
Nevertheless I update the code.

relay/healthcheck/sender.go
pappz: We modify it from the test code to limit the running time for the tests. Do you have better patter than this?

relay/messages/message.go
pappz: I do not understand. If we add a new message type in the future, the current switch/case (line 64.) can handle the "new, unknown" type.

relay/server/peer.go
pappz: I think no because in this case should return with err but who knows how works the external libraries. Good point. Fix it.

relay/auth/hmac/token.go
pappz: Good point. From the Relay perspective could, but we reuse this code for the TURN authentication. I would like to be backward compatible.

relay/auth/hmac/token.go
pappz: In the line 100. already print out

relay/auth/hmac/store.go
pappz: The problem with it the unmarshal function. It is not exported. In the store, we marshal the token  but in the hmac pkg unmarshal it. So the hmac will depend from the token and the token will depend from the hmac. Circle dependency

pappz: I will do something in a separated PR

relay/client/client.go
```diff
@@ -0,0 +1,523 @@
...
+	return len(c.conns) > 0
+}
+
+// Close closes the connection to the relay server and all connections to other peers.
+func (c *Client) Close() error {
+	return c.close(true)
+}
+
+func (c *Client) connect() error {
+	conn, err := ws.Dial(c.connectionURL)
```
lixmal, line 224: If this fails (e.g. relay srv is not available) then the client nerver starts up and even `netbird status` is not available. Is that desired?

relay/client/dialer/ws/ws.go
```diff
@@ -0,0 +1,59 @@
...
+	if resp.Body != nil {
+		_ = resp.Body.Close()
+	}
+
+	conn := NewConn(wsConn, address)
+	return conn, nil
+}
+
+func prepareURL(address string) (string, error) {
+	if !strings.HasPrefix(address, "rel") {
```
lixmal: Should probably test for `rel:` or `rel://` to avoid allowing something like `relay.example.org:8080`

management/server/token_mgr.go
```diff
@@ -0,0 +1,132 @@
...
+	var relayAddr string
+	if relayConfig != nil {
+		relayAddr = relayConfig.Address
+	}
+	return &TimeBasedAuthSecretsManager{
+		mux:           sync.Mutex{},
+		updateManager: updateManager,
+		turnCfg:       turnCfg,
+		relayAddr:     relayAddr,
+		hmacToken:     auth.NewTimedHMAC(turnCfg.Secret, turnCfg.CredentialsTTL.Duration),
```
lixmal: This is using the turn config secret for the new relay, we might want to change that to something separate

relay/client/client.go
```diff
@@ -0,0 +1,523 @@
...
+	c.conns[hashedStringID] = newConnContainer(conn, msgChannel)
+	return conn, nil
+}
+
+// ServerInstanceURL returns the address of the relay server. It could change after the close and reopen the connection.
+func (c *Client) ServerInstanceURL() (string, error) {
+	c.muInstanceURL.Lock()
+	defer c.muInstanceURL.Unlock()
+	if c.instanceURL == nil {
+		return "", fmt.Errorf("relay connection is not established")
```
lixmal, line 199: Can we use errRelayClientNotConnected here as well?

client/internal/peer/handshaker.go
```diff
@@ -0,0 +1,191 @@
...
+func (h *Handshaker) AddOnNewOfferListener(offer func(remoteOfferAnswer *OfferAnswer)) {
+	h.onNewOfferListeners = append(h.onNewOfferListeners, offer)
+}
+
+func (h *Handshaker) Listen() {
+	for {
+		h.log.Debugf("wait for remote offer confirmation")
+		remoteOfferAnswer, err := h.waitForRemoteOfferConfirmation()
+		if err != nil {
+			if _, ok := err.(*ConnectionClosedError); ok {
```
lixmal: This should use errors.Is for chained errors

relay/server/listener/ws/listener.go
lixmal: If there was an error closing for whatever reason we might not notice that we have leaked resources somewhere. Maybe not in this implementation but in the next?
Anyway, I'm okay if we leave it like this for now and fix it another time.

relay/server/listener/ws/listener.go
lixmal: True, in this case better rephrase to something like

```suggestion
	log.Infof("WS server listening address: %s", l.Address)
```

Because I had the case and didn't immediately notice that the container has exited

relay/healthcheck/sender.go
lixmal: Ah I didn't see that. Let's leave it as is for the time being

relay/messages/message.go
lixmal: The default value in go would be 0 which is not optimal if you want to check for an unset value.
The code you mentioned can return unknown for either, no problem.

relay/server/relay.go
lixmal: Understood.

relay/server/peer.go
lixmal: To be clear we declare a `n` var further up ahead and return it without ever assigning it, if there's no n, just return 0

relay/messages/message.go
mlsmaycon: is a good convention and follows the rest of the code where we have unknown types

management/server/token_mgr.go
pappz: Yes, I have a task for it. It will be a bigger change.

relay/client/client.go
pappz: Yes, it is the expected. Same as the signal connection.

relay/server/listener/ws/listener.go
pappz: I will add a log line

relay/server/peer.go
pappz: I see now. Fixed

relay/client/client.go
pappz: The two errors is a little bit different (but the message is almost the same). And I want to keep a hierarchy between the manager and the client. If we define this error in the client code and the manager starts to using it, it is fine but the opposite way is not.

relay/messages/message.go
pappz: Here the MsgType is not just a list of the possible options in GO. The 0, 1, 2, 3, 4 is a byte in the relay protocol header what will be appear on the network. Do you still think we should use the 0 for nothing?

relay/auth/hmac/token.go
pappz: I define error var only in that case if we want to determine the error type. In this case, I do not see the reason.

relay/auth/hmac/validator.go
pappz: Fixed the assert error handling.
This implements the Validator interface.
```
type Validator interface {
	Validate(any) error
}
```
So it must to handle any type.

relay/server/peer.go
pappz: If we want we can do a version check in the Hello message. In the **additions []byte** we can extend it with any new information in the future. Right now it contains the auth token only

pappz: Maybe the current usage of the additions does not grant enough flexibility in the future. Discuss about it before the release.

relay/healthcheck/sender.go
pappz: Fix in separate PR

mlsmaycon (approved):

PR #2245, opened 2024-07-08
Add batch save/update for groups and users

bcmmbaga:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/group.go
+++ b/management/server/group.go
@@ -163,7 +163,7 @@ func (am *DefaultAccountManager) SaveGroups(ctx context.Context, accountID, user
 	}
 
 	account.Network.IncSerial()
-	if err = am.Store.SaveAccount(ctx, account); err != nil {
+	if err = am.Store.SaveGroups(account); err != nil {
 		return err
 	}
 
--- a/management/server/group.go
+++ b/management/server/group.go
@@ -163,7 +163,7 @@ func (am *DefaultAccountManager) SaveGroups(ctx context.Context, accountID, user
 	}
 
 	account.Network.IncSerial()
-	if err = am.Store.SaveGroups(account); err != nil {
+	if err = am.Store.SaveGroups(account.Id, account.Groups); err != nil {
 		return err
 	}
 
--- a/management/server/group.go
+++ b/management/server/group.go
@@ -112,43 +112,55 @@ func (am *DefaultAccountManager) GetGroupByName(ctx context.Context, groupName, 
 func (am *DefaultAccountManager) SaveGroup(ctx context.Context, accountID, userID string, newGroup *nbgroup.Group) error {
 	unlock := am.Store.AcquireAccountWriteLock(ctx, accountID)
 	defer unlock()
+	return am.SaveGroups(ctx, accountID, userID, []*nbgroup.Group{newGroup})
+}
 
+// SaveGroups adds new groups to the account.
+// Note: This function does not acquire the global lock.
+// It is the caller's responsibility to ensure proper locking is in place before invoking this method.
+func (am *DefaultAccountManager) SaveGroups(ctx context.Context, accountID, userID string, newGroups []*nbgroup.Group) error {
 	account, err := am.Store.GetAccount(ctx, accountID)
 	if err != nil {
 		return err
 	}
 
-	if newGroup.ID == "" && newGroup.Issued != nbgroup.GroupIssuedAPI {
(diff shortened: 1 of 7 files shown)


management/server/group.go
```diff
@@ -112,61 +112,85 @@ func (am *DefaultAccountManager) GetGroupByName(ctx context.Context, groupName,
 func (am *DefaultAccountManager) SaveGroup(ctx context.Context, accountID, userID string, newGroup *nbgroup.Group) error {
 	unlock := am.Store.AcquireAccountWriteLock(ctx, accountID)
 	defer unlock()
+	return am.SaveGroups(ctx, accountID, userID, []*nbgroup.Group{newGroup})
+}
 
+// SaveGroups adds new groups to the account.
+// Note: This function does not acquire the global lock.
+// It is the caller's responsibility to ensure proper locking is in place before invoking this method.
```
pappz, line 120: Why do not move the lock from SaveGroup to here?

management/server/sql_store.go
```diff
@@ -311,6 +311,32 @@ func (s *SqlStore) SavePeerLocation(accountID string, peerWithLocation *nbpeer.P
...
+			user.PATsG = append(user.PATsG, *pat)
+		}
+		account.UsersG = append(account.UsersG, *user)
+	}
+	return s.db.Clauses(clause.OnConflict{UpdateAll: true}).Create(&account.UsersG).Error
+}
+
+// SaveGroups saves the given list of groups to the database.
+// It updates existing groups if a conflict occurs.
+func (s *SqlStore) SaveGroups(account *Account) error {
```
pappz: I think it is not memory efficient. If I understand well the reason of the "G" postfix it is just used by the Gorm. So after that, we always need to clean up because the garbage collector will not do it. Nevertheless, we should not mix the layers. The account structure holds the account info and the SQL-related things. We should create separate struct definition just for the SQL. It is more code but at the end, it will be safer. 

Now I recommend to:
- change the parameters of the function **SaveGroups(accountID string, groups  map[string]*nbgroup.Group)**
- create a new slice []nbgroup.Group  
- copy the relevant data to there
- change the db.clauses to use the new slice

So after it the GroupsG will be empty again

pappz: Maybe the "**&gorm.Session{FullSaveAssociations: true}**" keyword help to you

management/server/sql_store.go
```diff
@@ -311,6 +311,32 @@ func (s *SqlStore) SavePeerLocation(accountID string, peerWithLocation *nbpeer.P
 	return nil
 }
 
+// SaveUsers saves the given list of users to the database.
+// It updates existing users if a conflict occurs.
+func (s *SqlStore) SaveUsers(account *Account) error {
```
pappz: Same like groups

pappz (changes requested):

management/server/sql_store.go
```diff
@@ -311,6 +311,36 @@ func (s *SqlStore) SavePeerLocation(accountID string, peerWithLocation *nbpeer.P
...
+		Clauses(clause.OnConflict{UpdateAll: true}).
+		Create(&usersToSave).Error
+}
+
+// SaveGroups saves the given list of groups to the database.
+// It updates existing groups if a conflict occurs.
+func (s *SqlStore) SaveGroups(accountID string, groups map[string]*nbgroup.Group) error {
+	groupsToSave := make([]nbgroup.Group, 0, len(groups))
+	for id, group := range groups {
+		group.ID = id
```
pappz: Do we really need to do this operation? In [this](https://github.com/netbirdio/netbird/pull/2245/files#diff-ad08fa670a36c3eed29a084065544822057a761069e7e4b94dd1613697bdf399R149) line I think we fill well the variable.

management/server/sql_store.go
```diff
@@ -653,6 +683,8 @@ func (s *SqlStore) GetStoreEngine() StoreEngine {
 	return s.storeEngine
 }
 
+//func (s *SqlStore) SaveGroups()
```
pappz: Unnecessary comment

management/server/sql_store.go
bcmmbaga: Not necessarily needed, but it was to make sure the group ID is the same as the index

pappz (approved):

Issue #2110, opened 2024-06-08
Connected field is true after netbird down


ctrl-zzz:
**Describe the problem**

I've been trying to implement a session-expire feature that checks the `connected` status of a peer; when a peer gets disconnected its login is marked as expired.
The problem is that after `netbird down` the `connected` status seems to be stuck at `true`, so the session can never expire.

I've made sure my code doesn't interfere with the expected behavior and double checked without applying my changes that the problem persists.

**To Reproduce**

1. In the `management/server/account.go` file print the `peer.Status.Connected` value immediately after the [MarkPeerConnected call](https://github.com/netbirdio/netbird/blob/main/management/server/account.go#L1897C53-L1897C59)
2. Deploy code
3. Get NetBird started
4. Perform `netbird up` and `netbird down`
5. Check printed value of the `connected` field

**Expected behavior**

The `connected` field should be set to false.

**Are you using NetBird Cloud?**

No, I'm selfhosting it.

**Additional context**

When GETting a peer using the NetBird REST API its `connected` field behaves normally, being set at `false` after `netbird down`.

nazarewk:
@ctrl-zzz @bcmmbaga Any idea whether this still happens with the latest NetBird version? There seems to be a merged PR referencing this issue

ctrl-zzz:
> @ctrl-zzz @bcmmbaga Any idea whether this still happens with the latest NetBird version? There seems to be a merged PR referencing this issue

I implemented a fix that resolved this issue, and it was merged with my PR. At the time, the problem no longer occurred - at least up until the last time I tested it. However, I'm not sure if the issue may have resurfaced with more recent versions :)

bcmmbaga:
> [@ctrl-zzz](https://github.com/ctrl-zzz) [@bcmmbaga](https://github.com/bcmmbaga) Any idea whether this still happens with the latest NetBird version? There seems to be a merged PR referencing this issue

The fix has been merged and this is no longer an issue with the latest version

TechHutTV:
Fixed in PR #2262 ("fix: save peer status correctly in sqlstore"), shipped in v0.28.5 — the GORM zero-value quirk that left `connected=true` in the DB after `netbird down` is gone. You confirmed it's resolved last April; closing as fixed.

PR #2262, opened 2024-07-11
fix: save peer status correctly in sqlstore

ctrl-zzz:
## Describe your changes
This change fixes an inconsistency between the value of the boolean fields in the peer status struct, and those same values in the db table.

The problem should be in how the fields get updated in the `SavePeerStatus` function in the `management/server/sql_store.go` file, particularly because of how GORM manages the update itself. There's an [explanation on the GORM documentation](https://gorm.io/docs/update.html#Updates-multiple-columns) on why this happens.

This could be an issue for other boolean fields updates as well.

https://github.com/netbirdio/netbird/issues/2110

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/sql_store.go
+++ b/management/server/sql_store.go
@@ -274,9 +274,23 @@ func (s *SqlStore) GetInstallationID() string {
 func (s *SqlStore) SavePeerStatus(accountID, peerID string, peerStatus nbpeer.PeerStatus) error {
 	var peerCopy nbpeer.Peer
 	peerCopy.Status = &peerStatus
+
+	peerStatusJson, err := json.Marshal(peerCopy.Status)
+	if err != nil {
+		log.Errorf("cannot marshal peer %s status", peerID)
+		return err
+	}
+
+	parsedPeerStatus := make(map[string]interface{})
+	err = json.Unmarshal(peerStatusJson, &parsedPeerStatus)
+	if err != nil {
+		log.Errorf("cannot unmarshal peer %s status", peerID)
+		return err
+	}
+
 	result := s.db.Model(&nbpeer.Peer{}).
 		Where("account_id = ? AND id = ?", accountID, peerID).
-		Updates(peerCopy)
+		Updates(parsedPeerStatus)
 
 	if result.Error != nil {
 		return result.Error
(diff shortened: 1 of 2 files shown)


management/server/sql_store.go
```diff
@@ -274,9 +274,23 @@ func (s *SqlStore) GetInstallationID() string {
...
+	err = json.Unmarshal(peerStatusJson, &parsedPeerStatus)
+	if err != nil {
+		log.Errorf("cannot unmarshal peer %s status", peerID)
+		return err
+	}
+
 	result := s.db.Model(&nbpeer.Peer{}).
 		Where("account_id = ? AND id = ?", accountID, peerID).
-		Updates(peerCopy)
+		Updates(parsedPeerStatus)
```
bcmmbaga: Thank you for your work on fixing this issue. Your approach using map updates works well. However, we can simplify it further.

Instead of marshaling and unmarshaling, we can directly select and update only the fields we need. By passing the peerCopy struct and ignoring the fields we don't want to update, we can streamline the process and avoid the additional steps of marshaling and unmarshaling

```
	fieldsToUpdate := []string{
		"peer_status_last_seen", "peer_status_connected",
		"peer_status_login_expired", "peer_status_required_approval",
	}
	result := s.db.Model(&nbpeer.Peer{}).Select(fieldsToUpdate).
		Where("account_id = ? AND id = ?", accountID, peerID).
		Updates(&peerCopy)
	if result.Error != nil {
		return result.Error
	}

```

bcmmbaga (changes requested):
To ensure we fully cover this case in our tests, we should update the existing  [TestSqlite_SavePeerStatus](https://github.com/netbirdio/netbird/blob/58fbc1249cc618ba3544942c31c2f3139e0c4729/management/server/sql_store_test.go#L365). Here are the specific changes I suggest:

https://github.com/netbirdio/netbird/blob/58fbc1249cc618ba3544942c31c2f3139e0c4729/management/server/sql_store_test.go#L376

```
	newStatus := nbpeer.PeerStatus{Connected: false, LastSeen: time.Now().UTC()}
```

https://github.com/netbirdio/netbird/blob/58fbc1249cc618ba3544942c31c2f3139e0c4729/management/server/sql_store_test.go#L391

```
Status:   &nbpeer.PeerStatus{Connected: true, LastSeen: time.Now().UTC()},
```

These changes will test updating the `Connected` status from `true` to `false`, which directly addresses the case we're fixing. This will verify that our new approach correctly handles boolean field updates in the database.

ctrl-zzz:
Hello @bcmmbaga, I updated the code according to your suggestions. I also tested it and it seems to be working :)

bcmmbaga (approved):

Issue #2260, opened 2024-07-11
client raising a "failed marshalling message string field contains invalid UTF-8" when trying to register the peer due to invalid SerialNumber 


EdouardVanbelle:
Hello I am getting an error: `encryption/message.go:14: failed marshalling message string field contains invalid UTF-8`
during a `netbird up --key XXX --management-url YYY`

**NetBird version**

0.28.4

**More details on the reason**

I have rebuilt netbird to have a better understanding of the isssue, 
here is the reason: my board do not have any serialnumber !

For better understanding: 

```diff
$ git diff
diff --git a/encryption/message.go b/encryption/message.go
index a646fa67..8fe76f7a 100644
--- a/encryption/message.go
+++ b/encryption/message.go
@@ -9,6 +9,7 @@ import (
 // EncryptMessage encrypts a body of the given protobuf Message
 func EncryptMessage(remotePubKey wgtypes.Key, ourPrivateKey wgtypes.Key, message pb.Message) ([]byte, error) {
        byteResp, err := pb.Marshal(message)
+       log.Debugf("DEBUG marshalling message %v", message)
        if err != nil {
                log.Errorf("failed marshalling message %v", err)
                return nil, err
```

which raises me:
```
/home/ed/netbird/client/client up --log-level debug -F --log-file console --management-url https://MYSERVER:443 --setup-key MYKEY

2024-07-11T18:35:09+02:00 DEBG client/internal/login.go:93: connecting to the Management service https://MYSERVER:443
2024-07-11T18:35:09+02:00 DEBG client/internal/login.go:63: connected to the Management service https://MYSERVER:443
[...]

PR #2263, opened 2024-07-11
fix 2260: fallback serial to Board

EdouardVanbelle:
## Describe your changes
My device does not have any serial, this patch fallbacks to board serial to avoid marshall error during message encryption
note: this patch is arguable, I should also check that Board has a valid serial

my case is this one:
```bash
# cat /sys/class/dmi/id/chassis_serial
���������������������������������
# cat /sys/class/dmi/id/product_serial
���������������������������������
# cat /sys/class/dmi/id/board_serial
GETY419003GB

# cat /sys/class/dmi/id/product_name
���������������������������������
# cat /sys/class/dmi/id/board_name
DE3815TYKH
```

fix https://github.com/netbirdio/netbird/issues/2260

### Checklist
- [X] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/system/info_linux.go
+++ b/client/system/info_linux.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"os"
 	"os/exec"
+	"regexp"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


client/system/info_linux.go
```diff
@@ -89,9 +90,17 @@ func _getInfo() string {
 func sysInfo() (serialNumber string, productName string, manufacturer string) {
 	var si sysinfo.SysInfo
 	si.GetSysInfo()
+	isascii, _ := regexp.Compile("^[[:ascii:]]+$")
```
pappz: Please handle the error or use the regexp.Mustcompile

EdouardVanbelle: fixed via regexp.Mustcompile

pappz (approved):

PR #2273, opened 2024-07-15
fix forwarded metrics

mlsmaycon:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/signal/server/signal.go
+++ b/signal/server/signal.go
@@ -23,6 +23,8 @@ const (
 	labelTypeError         = "error"
 	labelTypeNotConnected  = "not_connected"
 	labelTypeNotRegistered = "not_registered"
+	labelTypeStream        = "stream"
+	labelTypeMessage       = "message"
 
 	labelError             = "error"
 	labelErrorMissingId    = "missing_id"
@@ -62,6 +64,7 @@ func (s *Server) Send(ctx context.Context, msg *proto.EncryptedMessage) (*proto.
 	}
 
 	if dstPeer, found := s.registry.Get(msg.RemoteKey); found {
+		start := time.Now()
 		//forward the message to the target peer
 		if err := dstPeer.Stream.Send(msg); err != nil {
 			log.Errorf("error while forwarding message from peer [%s] to peer [%s] %v", msg.Key, msg.RemoteKey, err)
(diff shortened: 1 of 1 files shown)


signal/server/signal.go
```diff
@@ -62,13 +64,15 @@ func (s *Server) Send(ctx context.Context, msg *proto.EncryptedMessage) (*proto.
...
 	if dstPeer, found := s.registry.Get(msg.RemoteKey); found {
+		start := time.Now()
 		//forward the message to the target peer
 		if err := dstPeer.Stream.Send(msg); err != nil {
 			log.Errorf("error while forwarding message from peer [%s] to peer [%s] %v", msg.Key, msg.RemoteKey, err)
 			//todo respond to the sender?
 
 			s.metrics.MessageForwardFailures.Add(ctx, 1, metric.WithAttributes(attribute.String(labelType, labelTypeError)))
 		} else {
+			s.metrics.MessageForwardLatency.Record(ctx, float64(time.Since(start).Nanoseconds())/1e6, metric.WithAttributes(attribute.String(labelType, labelTypeMessage)))
```
pappz, line 75: Instead of /1e6 we could use time.Since(start).Milliseconds(), I think.

signal/server/signal.go
```diff
@@ -118,22 +122,21 @@ func (s *Server) ConnectStream(stream proto.SignalExchange_ConnectStreamServer)
 		} else if err != nil {
 			return err
 		}
-		start := time.Now()
 
 		log.Debugf("received a new message from peer [%s] to peer [%s]", p.Id, msg.RemoteKey)
 
 		// lookup the target peer where the message is going to
 		if dstPeer, found := s.registry.Get(msg.RemoteKey); found {
+			start := time.Now()
```
lixmal, line 130: Any reason you don't want to include the map lookup in the latency anymore? My idea was to include the whole operation

signal/server/signal.go
lixmal: If you do that we'll lose precision, as `Milliseconds` returns an int, but we have fractional milliseconds in the histogram buckets: 


https://github.com/netbirdio/netbird/blob/main/signal%2Fmetrics%2Fapp.go#L90-L104

But maybe we don't need those small buckets in the first place. We should take a look at the actual data.

signal/server/signal.go
mlsmaycon: I would rather keep the map lookup as a separate metric and have the latency only linked to the send operation

signal/server/signal.go
mlsmaycon: let's keep them for now

PR #2279, opened 2024-07-16
Wait on daemon down

pascal-fischer:
## Describe your changes
This PR awaits the engine shutdown before sending the confirmation back to the cli. It also fixes an issue where there was a race condition when running multiple up and down commands in sequence.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -268,7 +268,7 @@ func (e *Engine) Stop() error {
 	e.wgConnWorker.Wait()
 
 	maxWaitTime := 5 * time.Second
-	startTime := time.Now()
+	timeout := time.After(maxWaitTime)
 
 	for {
 		if !e.IsWGIfaceUp() {
@@ -276,11 +276,12 @@ func (e *Engine) Stop() error {
 			return nil
 		}
 
-		if time.Since(startTime) > maxWaitTime {
+		select {
+		case <-timeout:
 			return fmt.Errorf("timeout when waiting for interface shutdown")
+		default:
+			time.Sleep(100 * time.Millisecond)
 		}
-
-		time.Sleep(100 * time.Millisecond)
 	}
 }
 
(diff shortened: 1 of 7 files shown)


client/internal/networkmonitor/monitor_bsd.go
```diff
@@ -21,19 +22,28 @@ func checkChange(ctx context.Context, nexthopv4, nexthopv6 systemops.Nexthop, ca
...
+
 	for {
 		select {
 		case <-ctx.Done():
 			return ErrStopped
 		default:
 			buf := make([]byte, 2048)
 			n, err := unix.Read(fd, buf)
-			if err != nil {
+			if err != nil && !errors.Is(err, unix.EBADF) && !errors.Is(err, unix.EINVAL) {
```
lixmal: I think we need to `continue` on the ignored errors as well, just don't log them, or leg them on debug. Apologies for the previous suggestion

pascal-fischer: fixed

client/server/server.go
```diff
@@ -593,7 +593,27 @@ func (s *Server) Down(_ context.Context, _ *proto.DownRequest) (*proto.DownRespo
...
+	for {
+		select {
+		case <-ctx.Done():
+			return &proto.DownResponse{}, nil
+		default:
+			if !engine.IsWGIfaceUp() {
+				return &proto.DownResponse{}, nil
+			}
+
+			if time.Since(startTime) > maxWaitTime {
```
lixmal: I think it's more idomatic to use `time.After` here and include the returned channel in the `select`, unless you still need to check `if !engine.IsWGIfaceUp()` in any case?

client/internal/engine.go
```diff
@@ -1533,3 +1533,19 @@ func isChecksEqual(checks []*mgmProto.Checks, oChecks []*mgmProto.Checks) bool {
...
 	})
 }
+
+func (e *Engine) IsWGIfaceUp() bool {
+	if e == nil || e.wgInterface == nil {
+		return false
+	}
+	iface, err := net.InterfaceByName(e.wgInterface.Name())
+	if err != nil {
+		return false
```
lixmal, line 1559: Please log the error to ease debugging, at least on `debug` level

lixmal (approved):

PR #2280, opened 2024-07-16
Add faster availability DNS probe and update test domain to .com

mlsmaycon:
## Describe your changes
- Count success queries and compare it before doing after network map probes.

- Reduce the first dns probe to 500ms

- Updated test domain with com instead of . due to Palo alto DNS proxy server issues
### Checklist
- [x] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/client/internal/dns/upstream.go
+++ b/client/internal/dns/upstream.go
@@ -175,7 +175,7 @@ func (u *upstreamResolverBase) probeAvailability() {
 	}
 
 	// avoid probe if upstreams could resolve at least one query and fails count is less than failsTillDeact
-	if u.successCount.Load() > 0 && u.failsCount.Load() > u.failsTillDeact {
+	if u.successCount.Load() > 0 && u.failsCount.Load() < u.failsTillDeact {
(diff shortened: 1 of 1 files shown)


client/internal/dns/upstream.go
```diff
@@ -172,6 +174,11 @@ func (u *upstreamResolverBase) probeAvailability() {
 	default:
 	}
 
+	// avoid probe if upstreams could resolve at least one query and fails count is less than failsTillDeact
+	if u.successCount.Load() > 0 && u.failsCount.Load() > u.failsTillDeact {
```
lixmal: shouldn't this be
```suggestion
	if u.successCount.Load() > 0 && u.failsCount.Load() < u.failsTillDeact {
```
?

Besides, that would prevent us from failing that upstream dns server if it became unavaible later, correct?

mlsmaycon: No, as the counters are updated according to their state

lixmal (approved):

PR #2284, opened 2024-07-17
Refactor macOS system DNS configuration

hurricanehrndz:
On macOS use the recommended settings for providing split DNS. As per the docs an empty string will force the configuration to be the default. In order to to support split DNS an additional service config is added for the local server and search domain settings.

see: https://developer.apple.com/documentation/devicemanagement/vpn/dns

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


mlsmaycon (approved):

mlsmaycon:
Thanks for the contribution @hurricanehrndz, the configuration is much simpler now.

PR #2285, opened 2024-07-18
Remove copy functions from signal

mlsmaycon:
## Describe your changes
remove migration function for wiretrustee directories to netbird

refactor main run function
### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/signal/cmd/run.go
+++ b/signal/cmd/run.go
@@ -113,7 +113,9 @@ var (
 
 			grpcRootHandler := grpcHandlerFunc(grpcServer)
 
-			startServer(certManager, grpcRootHandler)
+			if certManager != nil {
+				startServerWithCertManager(certManager, grpcRootHandler)
+			}
 
 			var compatListener net.Listener
 			var grpcListener net.Listener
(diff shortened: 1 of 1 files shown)


signal/cmd/run.go
```diff
@@ -211,6 +171,60 @@ var (
...
+func getTLSConfigurations() ([]grpc.ServerOption, *autocert.Manager, error) {
+	var (
+		err         error
+		certManager *autocert.Manager
+		tlsConfig   *tls.Config
+	)
+
+	if signalLetsencryptDomain == "" && signalCertFile == "" && signalCertKey == "" {
+		log.Infof("running without TLS")
+		return nil, certManager, nil
```
pappz: It makes no sense to return with certManager if it is nil.

signal/cmd/run.go
```diff
@@ -141,24 +112,13 @@ var (
 			proto.RegisterSignalExchangeServer(grpcServer, srv)
 
 			grpcRootHandler := grpcHandlerFunc(grpcServer)
+
+			startServer(certManager, grpcRootHandler)
```
pappz: It would be more understandable if you put the `if certManager != nil {` outside of the start function. Because now has serve function inside the start and outside too and is not trivial when will be used for.

bcmmbaga (approved):

PR #2291, opened 2024-07-19
Apply host DNS settings on peer state change

### LINKED ISSUE #2002: Netbird should connect to peers before setting up DNS
**Describe the problem**

In the latest version, netbird tries to resolve DNS before connecting to the peers. This causes the DNS resolution to fail in case where the DNS being used is a private DNS behind a routing peer. This causes netbird to wait for the DNS resolution to timeout before connecting to peers on the network, and hence, it takes at least 15 seconds in our case to connect to the first peer.

After the peer connects, the DNS resolution works perfectly fine. But this delay in most cases is unbearable and causes usability issues for a lot of people.

Logs:

```
May 16 22:31:13 hades netbird[1405]: 2024-05-16T22:31:13+05:30 INFO signal/client/grpc.go:158: connected to the Signal Service stream
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 WARN [error: read udp 192.168.69.100:59667->192.168.0.2:53: i/o timeout, upstream: 192.168.0.2:53] client/internal/dns/upstream.go:102: got an error while connecting to upstream
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 ERRO client/internal/dns/upstream.go:134: all queries to the upstream nameservers failed with timeout
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 WARN [error: read udp 192.168.69.100:41074->192.168.0.2:53: i/o timeout, upstream: 192.168.0.2:53] client/internal/dns/upstream.go:102: got an error while connecting to upstream
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 ERRO client/internal/dns/upstream.go:134: all queries to the upstream nameservers failed with timeout
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 WARN [error: read udp 192.168.69.100:37390->192.168.0.2:53: i/o timeout, upstream: 192.168.0.2:53] client/internal/dns/upstream.go:102: got an error while connecting to upstream
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 ERRO client/internal/dns/upstream.go:134: all queries to the upstream nameservers failed with timeout
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 WARN client/internal/dns/upstream.go:265: Upstream resolving is Disabled for 30s
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 INFO [nameservers: [{192.168.0.2 udp 53}]] client/internal/dns/server.go:504: Temporarily deactivating nameservers group due to timeout
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 WARN [error: read udp 192.168.69.100:59543->192.168.0.2:53: i/o timeout, upstream: 192.168.0.2:53] client/internal/dns/upstream.go:102: got an error while connecting to upstream
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 ERRO client/internal/dns/upstream.go:134: all queries to the upstream nameservers failed with timeout
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 WARN [error: read udp 192.168.69.100:56609->192.168.0.2:53: i/o timeout, upstream: 192.168.0.2:53] client/internal/dns/upstream.go:102: got an error while connecting to upstream
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 ERRO client/internal/dns/upstream.go:134: all queries to the upstream nameservers failed with timeout
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 WARN [upstream: 192.168.0.2:53, error: read udp 192.168.69.100:45333->192.168.0.2:53: i/o timeout] client/internal/dns/upstream.go:102: got an error while connecting to upstream
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 ERRO client/internal/dns/upstream.go:134: all queries to the upstream nameservers failed with timeout
May 16 22:31:15 hades netbird[1405]: 2024-05-16T22:31:15+05:30 INFO client/internal/dns/resolvconf_linux.go:73: added 2 search domains. Search list: [local.netbird lan]
May 16 22:31:16 hades netbird[1405]: 2024-05-16T22:31:16+05:30 WARN [error: read udp 192.168.69.100:46766->192.168.0.2:53: i/o timeout, upstream: 192.168.0.2:53] client/internal/dns/upstream.go:102: got an error while connecting to upstream
May 16 22:31:16 hades netbird[1405]: 2024-05-16T22:31:16+05:30 ERRO client/internal/dns/upstream.go:134: all queries to the upstream nameservers failed with timeout
May 16 22:31:16 hades netbird[1405]: 2024-05-16T22:31:16+05:30 WARN [error: read udp 192.168.69.100:57562->192.168.0.2:53: i/o timeout, upstream: 192.168.0.2:53] client/internal/dns/upstream.go:102: got an error while connecting to upstream
May 16 22:31:16 hades netbird[1405]: 2024-05-16T22:31:16+05:30 ERRO client/internal/dns/upstream.go:134: all queries to the upstream nameservers failed with timeout
May 16 22:31:17 hades netbird[1405]: 2024-05-16T22:31:17+05:30 INFO management/client/grpc.go:147: connected to the Management Service stream
May 16 22:31:17 hades netbird[1405]: 2024-05-16T22:31:17+05:30 WARN client/internal/routemanager/client.go:154: the network 192.168.0.0/19 has not been assigned a routing peer as no peers from the list [<LIST>] are currently connected
May 16 22:31:18 hades netbird[1405]: 2024-05-16T22:31:18+05:30 INFO client/internal/routemanager/client.go:165: new chosen route is <ROUTE> with peer <PEER-ID> with score 2.974409 for network 192.168.0.2/32
May 16 22:31:18 hades netbird[1405]: 2024-05-16T22:31:18+05:30 INFO client/internal/dns/upstream.go:241: upstreams [192.168.0.2:53] are responsive again. Adding them back to system
```

In the logs it took a few seconds to connect, but usually on `netbird up`, this takes at least 10-15 seconds to connect.

**To Reproduce**

1. Set up a private DNS on netbird using routing peer.
2. Connect to netbird.
3. Notice that netbird tries to resolve DNS and fails before trying to connect to the peers.
4. See error.

**Expected behavior**

The DNS resolution should take place after the peer connections are initialized. There's no need for netbird to replace and resolve DNS before connecting to peers.

Self-hosted NetBird's control plane.

**NetBird version**

`netbird version`: `0.27.7`

hurricanehrndz:
## Describe your changes

This is a PoC and a potential fix #2002. This patch is not full fleshed out and their are some unattended side effects, but I wanted to get your thoughts and opinions @mlsmaycon and @pascal-fischer and @lixmal. 

More Info:

This essential tries to ensure that host dns changes aren't applied til at least one routing peer is connected and DNS servers are reachable. It does so by tying deactivation and reactivation of upstream servers to Peerlist state changes in the status recorder.  

As it stands, now `status -d` doesn't accurately reflect the status on first connection. That is because I haven't figure out an effective method to deactivate the servers without more intense changes. Primarily because of how the deactivate and reactivate callbacks are structure and their dependence of the removeIndex. One potential workaround is to modify deactivate to skip applying host config, so that we can fix this side effect.

Thoughts and discussion please

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


lixmal:
>This essential tries to ensure that host dns changes aren't applied til at least one routing peer is connected and DNS servers are reachable.

This approach would break non-routed dns servers (e.g. 8.8.8.8 without exit node), correct?

hurricanehrndz:
> > This essential tries to ensure that host dns changes aren't applied til at least one routing peer is connected and DNS servers are reachable.
> 
> This approach would break non-routed dns servers (e.g. 8.8.8.8 without exit node), correct?

Not at all, it would just not apply settings until the client is connected to one peer. I am going to try and actually formulate this to PR

I may update the activate and deactivate upstream handlers to take a flag to skip applying system settings so as fix the DNS updates from management.

hurricanehrndz:
@lixmal this is ready not sure if perhaps this should be gated under a config option and the default would be maintain the current behavior. Test and let me know what you think

hurricanehrndz:
Actually I got an idea to make hybrid model work, let me get that going instead

LeszekBlazewski:
Looking forward for this fix. 

If I could help in any way to test this (for example on a setup I described in https://github.com/netbirdio/netbird/issues/2002#issuecomment-2293128357) let me know.

hurricanehrndz:
Okay so this will waitforresponse and trigger probes base when a peer connstatus changes

hurricanehrndz:
@lixmal and @mlsmaycon can we get a build so @LeszekBlazewski can provide some feedback

lixmal:
You can find the binaries in the artifacts in this PR: https://github.com/netbirdio/netbird/actions/runs/10422794281

LeszekBlazewski:
Hey guys,

I have run some tests with the latest binaries built in this PR.

My setup:

1. private R53 hosted zone with a configured inbound resolver (3 ips in range `10.3.X.X`, only configured to match specific domains). Those 3 IP addresses are only accessible within VPC which runs a netbird peer. For the whole 10.3.0.0/16 range I have a network route configured which points all that traffic to the peer which runs inside that VPC and therefore is allowed to use the resolver. 
2. Imaginary DNS nameserver (ip 1.2.3.4, match all domains) that won't ever resolve
3. Just for testing sake, the cloudflare and google DNS resolver (match all domains)
4. Exit node setup

After comparing the current behaviour (netbird 0.28.7) vs the built binaries, I have found out that indeed the introduced changes postpone DNS servers evaluation until a routing peer is connected which does potentially address the issue mentioned in https://github.com/netbirdio/netbird/issues/2002 ... unless the nameserver is not supposed to be handled by that connected peer. Based on the logs, I have observed that nameserver probing is triggered regardless of the peer that got just connected ( I guess it's assumed that all configured netbird DNS servers can be handled by all connected peers or there are other technical limitations which prevent figuring out such information, as described below).

For example in my described setup, it would make sense to probe the nameservers from point 1 **only** if the routing peer which handles `10.3.X.X/16` would be connected. I am aware that such condition might be challenging to meet since there is no peer assignment for DNS resolution handling (only for distribution). Also the public DNS resolution (Google and Cloudflare) make the decision wether to probe / don't probe DNS even more harder since no one configures network routes for those DNS servers (there is a reason why they are public).

But still, I think that not applying the config when the DNS server is not reachable is key here and that seems to work. Moreover the backoff over time  will fix the above mentioned issue and all the DNS servers who are reachable will be connected.

hurricanehrndz:
> Hey guys,
> 
> I have run some tests with the latest binaries built in this PR.
> 
> My setup:
> 
>     1. private R53 hosted zone with a configured inbound resolver (3 ips in range `10.3.X.X`, only configured to match specific domains). Those 3 IP addresses are only accessible within VPC which runs a netbird peer. For the whole 10.3.0.0/16 range I have a network route configured which points all that traffic to the peer which runs inside that VPC and therefore is allowed to use the resolver.
> 
>     2. Imaginary DNS nameserver (ip 1.2.3.4, match all domains) that won't ever resolve
> 
>     3. Just for testing sake, the cloudflare and google DNS resolver (match all domains)
> 
>     4. Exit node setup
> 
> 
> After comparing the current behaviour (netbird 0.28.7) vs the built binaries, I have found out that indeed the introduced changes postpone DNS servers evaluation until a routing peer is connected which does potentially address the issue mentioned in #2002 ... unless the nameserver is not supposed to be handled by that connected peer. Based on the logs, I have observed that nameserver probing is triggered regardless of the peer that got just connected ( I guess it's assumed that all configured netbird DNS servers can be handled by all connected peers or there are other technical limitations which prevent figuring out such information, as described below).
> 
> For example in my described setup, it would make sense to probe the nameservers from point 1 **only** if the routing peer which handles `10.3.X.X/16` would be connected. I am aware that such condition might be challenging to meet since there is no peer assignment for DNS resolution handling (only for distribution). Also the public DNS resolution (Google and Cloudflare) make the decision wether to probe / don't probe DNS even more harder since no one configures network routes for those DNS servers (there is a reason why they are public).
> 
> But still, I think that not applying the config when the DNS server is not reachable is key here and that seems to work. Moreover the backoff over time will fix the above mentioned issue and all the DNS servers who are reachable will be connected.

I made this patch to show what is one potential fix, probing base on routes when a peer connects will get more complicated, but it is possible. 

This patch still includes the old logic as fallback, so triggering if public DNS servers are configured they should get apply very early on

hurricanehrndz:
> Hey guys,
> 
> I have run some tests with the latest binaries built in this PR.
> 
> My setup:
> 
>     1. private R53 hosted zone with a configured inbound resolver (3 ips in range `10.3.X.X`, only configured to match specific domains). Those 3 IP addresses are only accessible within VPC which runs a netbird peer. For the whole 10.3.0.0/16 range I have a network route configured which points all that traffic to the peer which runs inside that VPC and therefore is allowed to use the resolver.
> 
>     2. Imaginary DNS nameserver (ip 1.2.3.4, match all domains) that won't ever resolve
> 
>     3. Just for testing sake, the cloudflare and google DNS resolver (match all domains)
> 
>     4. Exit node setup
> 
> 
> After comparing the current behaviour (netbird 0.28.7) vs the built binaries, I have found out that indeed the introduced changes postpone DNS servers evaluation until a routing peer is connected which does potentially address the issue mentioned in #2002 ... unless the nameserver is not supposed to be handled by that connected peer. Based on the logs, I have observed that nameserver probing is triggered regardless of the peer that got just connected ( I guess it's assumed that all configured netbird DNS servers can be handled by all connected peers or there are other technical limitations which prevent figuring out such information, as described below).
> 
> For example in my described setup, it would make sense to probe the nameservers from point 1 **only** if the routing peer which handles `10.3.X.X/16` would be connected. I am aware that such condition might be challenging to meet since there is no peer assignment for DNS resolution handling (only for distribution). Also the public DNS resolution (Google and Cloudflare) make the decision wether to probe / don't probe DNS even more harder since no one configures network routes for those DNS servers (there is a reason why they are public).
> 
> But still, I think that not applying the config when the DNS server is not reachable is key here and that seems to work. Moreover the backoff over time will fix the above mentioned issue and all the DNS servers who are reachable will be connected.

I will try and add the ability to trigger a peer with route. What I gather though was that the App was performing and behaving better apply DNS changes base on when peers connect, is that right?

LeszekBlazewski:
Yeah, for sure. With the patch from this PR the amount of timed out calls to private NS is minimised. 

One thing I wanted to figure out and I am wondering if this could be related to my setup somehow is: Based on the logs, even after the peer which handles 10.3.X.X traffic connects (logs indicate that), the first few calls to probe the private NS still fail and only after few seconds they come back online. I will try to investigate this since I am confident that those NS are available at all times and it could be related to the exit node node or the routing peer connecting as relay.

hurricanehrndz:
> Yeah, for sure. With the patch from this PR the amount of timed out calls to private NS is minimised.
> 
> One thing I wanted to figure out and I am wondering if this could be related to my setup somehow is: Based on the logs, even after the peer which handles 10.3.X.X traffic connects (logs indicate that), the first few calls to probe the private NS still fail and only after few seconds they come back online. I will try to investigate this since I am confident that those NS are available at all times and it could be related to the exit node node or the routing peer connecting as relay.

Do you have any system extensions that filter DNS requests, such as Cisco Umbrella? Assuming this is a macOS device

LeszekBlazewski:
The only thing running in the background is eset with the Network Access Protection enabled which might do some DNS filtering (I am not sure since I don't have access to the management of this antivirus).

What is interesting is that the NS requests fail only initially and then just work without me doing any changes.

hurricanehrndz:
> The only thing running in the background is eset with the Network Access Protection enabled which might do some DNS filtering (I am not sure since I don't have access to the management of this antivirus but I was told it does not). I will make sure to disable it when testing.
> 
> What is interesting is that the NS requests fail only initially and then just work without me doing any changes.

This patch now support enabling publicly accessible servers at processing the DNS update time and/or start

hurricanehrndz:
@LeszekBlazewski 

Minor updates to behavior in order to ensure publicly accessible servers work from the start without a peer. I unfortunately didn't filter probing b

LeszekBlazewski:
Indeed, did a quick test and the public nameservers were enabled almost instantly whereas the rest got processed later on. I am doing some checks on why the first initial few nameserver requests after the connection to the peer which is supposed to handle that traffic still fail.

hurricanehrndz:
So this is ready for review, we have been running this for a two weeks with some excellent results. @LeszekBlazewski  I will need to fix some golang bugs before we can have DNS probing on routes. After fixing that refactoring of the networkmonitor would be necessary. I am not sure if all these changes can be implemented in a timely manner, so for now this patch is a good stop gap. I am going to go fix 44740 for now 

https://github.com/golang/go/issues/44740

hurricanehrndz:
@LeszekBlazewski can you test the latest build, the patch has been updated to work with the latest 29.x releases. I believe @mlsmaycon will be reviewing this soon

https://github.com/netbirdio/netbird/actions/runs/10997072510?pr=2291

hurricanehrndz:
@LeszekBlazewski this now applies settings base on routing table

Eu-Arthur:
i have same problem, like #2002, and  this PR fix my problem.

But, it's possible to be rebase to last version ?

hurricanehrndz:
Yeah, I will rebase it and see if I can fix some tests.

hurricanehrndz:
> i have same problem, like #2002, and this PR fix my problem.
> 
> But, it's possible to be rebase to last version ?

I have updated this PR, but it is a WIP, please test. I need update this PR to support other platforms

Eu-Arthur:
Yes, i don't see problem, for me that work on latest version

PR #2296, opened 2024-07-21
Add SavePeer method to prevent a possible account inconsistency

mlsmaycon:
## Describe your changes
SyncPeer was storing the account with a simple read lock

This change introduces the SavePeer method to the store to be used in these cases
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/peer.go
+++ b/management/server/peer.go
@@ -7,9 +7,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/netbirdio/netbird/management/server/posture"
 	"github.com/rs/xid"
 	log "github.com/sirupsen/logrus"
+
+	"github.com/netbirdio/netbird/management/server/posture"
 
 	"github.com/netbirdio/netbird/management/proto"
 	"github.com/netbirdio/netbird/management/server/activity"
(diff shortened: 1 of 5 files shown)


management/server/peer.go
```diff
@@ -539,7 +540,7 @@ func (am *DefaultAccountManager) SyncPeer(ctx context.Context, sync PeerSync, ac
 
 	peer, updated := updatePeerMeta(peer, sync.Meta, account)
 	if updated {
-		err = am.Store.SaveAccount(ctx, account)
+		err = am.Store.SavePeer(ctx, account.Id, peer)
```
pappz, line 543: If we just update the Meta why do not implement a SaveMeta instead of writing out the full Peer structure?

mlsmaycon: We can't update only the meta since it is flattened by the tag `gorm:"embedded;embeddedPrefix:meta_"`, which creates a column per field. Plus, this method will be used more often.

pappz: I think the gorm can handle the embedded structures well in case of an update just need pass it in the peer structure. Peer{Meta: peerSystemMeta}. But ok.

pappz (approved):

PR #2298, opened 2024-07-22
Add network route access control on client

lixmal:
## Describe your changes

Add:
- Methods to set up forward rules for routed traffic (network ACLs) according to the network map
Modify:
- Apply ACLs in engine as early as possible to avoid security issues
- Insert instead of append all netbird rules in existing input/output/forward chains
- Rename existing ACLs to "peer ACLs"
- Use dedicated type for route ID where possible
- Use `netip.Prefix` instead of `string` where possible

Fix:
- Management proto and network map:
  - Add default return traffic permit when there's no ACL applied to the network route
  - Fix missing protocol in default rules
  - Partically add IPv6 rules to DNS routes

Remove:
- Default accept in forward chains
- Rules in prerouting chains
- Fwmark accept rules in output chains
- Src addresses in output chains
- Dst addresses in input chains
- Dysfunctional rule in output chains
- Accept traffic initiated by routed networks behind routing peers in input chains

The prerouting/fwmark parts were used as substitutes for route ACLs to prevent locally routed networks (e.g. containers) from bypassing the input rules (because forward rules match the traffic).

### Checklist
- [x] Is it a bug fix
- [x] It is a refactor


client/firewall/nftables/acl_linux.go
```diff
@@ -199,8 +188,7 @@ func (m *AclManager) DeleteRule(rule firewall.Rule) error {
 	return nil
 }
 
-// createDefaultAllowRules In case if the USP firewall manager can use the native firewall manager we must to create allow rules for
-// input and output chains
+// creeateDefaultAllowRules creates default allow rules for the input and output chains
```
pascal-fischer: typo in creeateDefaultAllowRules

lixmal:
```suggestion
// createDefaultAllowRules creates default allow rules for the input and output chains
```

pascal-fischer (approved):

Issue #2114, opened 2024-06-11
New peers are not propagated to already existing peers


glaeqen:
**Describe the problem**

After adding a new peer to the network, this peer is not getting propagated to the already existing peers. At best only peers that were added very recently (last 24h?) seem to get the list of peers updated. The only thing that helps is to `netbird down` and `netbird up` on all the affected, exisiting peers machines which forces a peer refresh. Existing peers seem to show a following error after a new peer was added

```
2024-06-10T15:57:29+02:00 ERRO signal/client/grpc.go:410: error while handling message of Peer [key: <KEY>] error: [wrongly addressed message <KEY>]
```

This feels so fundamentally wrong that it's hard to believe that it's a bug and not just my fault. I doubt it's a networking issue, affected peers are located both in the cloud and self-hosted networks so it does not seem probable.

**Expected behavior**

Automatic peer propagation

**Are you using NetBird Cloud?**

self-hosted NetBird's control plane

**NetBird version**

0.27.10

mlsmaycon:
This is a very odd indeed. Assuming your connection with the management is fine this should never happen unless there is an manual update of the network map in the database or a async update causing the serial number to be lower. 

If you are running on 0.27.5+, can you enable debug logs on one of the existing clients with: `netbird debug log level debug`

then add a new peer and check if there is a log similar to:
```
received outdated NetworkMap with serial XXXX, ignoring
```

glaeqen:
It occured twice earlier this week but I cannot reproduce it since 🙄 When it happened I happened to test that when I updated ACL (I added an affected peer to some new, arbitrary group), netbird client of that peer noticed the ACL update request and as a side-effect it also did update the peer list. If it happens again I'll try to gather more detailed logs of all parties involved.

mlsmaycon:
sounds great, thanks @glaeqen

mohamed-essam:
@mlsmaycon Yes, the logs contain outdated NetworkMap, excerpt from the logs here:

```
2024-07-24T08:22:33Z DEBG client/internal/peer/conn.go:288: connection offer sent to peer V1uRp/w/p/b5eB8ulqlkzbPLaxzf8C2CyLkx1EkTzG0=, waiting for the confirmation
2024-07-24T08:22:33Z DEBG client/internal/peer/conn.go:256: trying to connect to peer XD5VjsfLgs3RNijJ7D+yqJFeA07sEaJ24MCfk10m3no=
[...]

PR #2324, opened 2024-07-25
client: Restart mgmt connection if networkMap serial is out of order

mohamed-essam:
## Describe your changes
Restart the connection to management service when serial received in networkMap is out of order. As well as adding a bit of logging around the serial check in networkMap updating client-side.

This is a workaround for the issue mentioned, the root cause is most likely a race condition in updating the networkMap serial in the management service itself, this PR will allow the issue to be non-apparent for users.

https://github.com/netbirdio/netbird/issues/2114

### Checklist
- [X] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -131,6 +131,9 @@ type Engine struct {
 	clientCtx    context.Context
 	clientCancel context.CancelFunc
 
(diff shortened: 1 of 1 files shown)


mohamed-essam:
Needs further testing

PR #2326, opened 2024-07-25
Add session expire functionality based on inactivity

ctrl-zzz:
## Describe your changes

I have implemented a session timeout feature based on inactivity, forcing re-authentication with every `netbird up`. This was necessary because, while evaluating if Netbird was suitable for my use case, I noticed that authentication is only required after the login expiration period has elapsed. For security reasons, I wanted access to be required with every connection: this way, the authentication process with the identity provider is triggered and, by using a session timeout setting in Keycloak, credentials are requested each time.

Building on the existing approach for login expiration, I implemented inactivity expiration by checking the status of a peer: after a configurable period of time following `netbird down`, the peer shows `login required`.
I also added a setting, `inactivityExpirationEnabled`, that can  enable or disable the inactivity expiration feature through the APIs.

I needed this functionality due to a business requirement and thought it might be useful to you as well.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


mlsmaycon:
Hello @ctrl-zzz can you check the failing tests and Sonar checks? 

We will review the feature and give you a feedback ASAP

mlsmaycon:
> Hello, I believe they are about the Cognitive Complexity of some methods. Since it's a new feature I added some new properties that had to be checked as well for some functionalities. I guess there needs to be less conditions to be checked, but not checking them wouldn't make the methods work as intended

thanks, there are some optimizations that can done, see some suggestions below:

For `func (am *DefaultAccountManager) UpdateAccountSettings`, you can move some of the new checks into a separate method or function, and that should reduce complexity. 

For `func (am *DefaultAccountManager) MarkPeerConnected` you can rewrite it like this:

```Go
	
if peer.AddedWithSSOLogin() {
       if  peer.LoginExpirationEnabled && account.Settings.PeerLoginExpirationEnabled {
		am.checkAndSchedulePeerLoginExpiration(ctx, account)
	}
	
        if peer.InactivityExpirationEnabled && account.Settings.PeerInactivityExpirationEnabled {
		am.checkAndSchedulePeerInactivityExpiration(ctx, account)
	}
}
```

management/server/account.go
```diff
@@ -50,6 +50,8 @@ const (
 	CacheExpirationMax         = 7 * 24 * 3600 * time.Second // 7 days
 	CacheExpirationMin         = 3 * 24 * 3600 * time.Second // 3 days
 	DefaultPeerLoginExpiration = 24 * time.Hour
+
+	DefaultPeerInactivityExpiration = 1 * time.Second
```
mlsmaycon: Because of various reasons, a peer might be online while the management service is down. We should have a reasonable default to avoid issues with most folks who decide to enable this setting. 

As we have validated so far with ephemeral peers, a 10-minute timeout could be a good default and cover most cases.

We do have plans to add a logout option for the client when running `netbird down` or `netbird logout`

pascal-fischer (approved):

mlsmaycon (dismissed):

mlsmaycon (approved):

PR #2337, opened 2024-07-27
Allow setup keys to be provided in a file

moosetheory:
## Describe your changes
Adds a flag and a bit of logic to allow a setup key to be passed in using a file. The flag should be exclusive with the standard `--setup-key` flag.

N/A

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

#### Why?
I am attempting to set up netbird automatically in a Lix environment, but the secrets library I use requires that I write the secrets to files and ask programs to use those at runtime. I believe this change will also allow someone to use docker secrets to provide a setup key.

I am unsure what/if changes should be made to any documentation or to the README file.

(I apologize if I have missed anything, this is my first time contributing to an open source project.)

### DIFF
--- a/client/cmd/login.go
+++ b/client/cmd/login.go
@@ -39,6 +39,13 @@ var loginCmd = &cobra.Command{
 			ctx = context.WithValue(ctx, system.DeviceNameCtxKey, hostName)
 		}
 
+		if setupKeyPath != "" && setupKey == "" {
+			setupKey, err = getSetupKeyFromFile(setupKeyPath)
+			if err != nil {
+				return err
+			}
+		}
+
 		// workaround to run without service
 		if logFile == "console" {
 			err = handleRebrand(cmd)
--- a/client/cmd/root.go
+++ b/client/cmd/root.go
@@ -55,6 +55,7 @@ var (
 	managementURL           string
 	adminURL                string
 	setupKey                string
+	setupKeyPath            string
 	hostName                string
 	preSharedKey            string
 	natExternalIPs          []string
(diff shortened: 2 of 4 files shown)


moosetheory:
Looking at the output from the Windows testing, it seems to be in the server code, and I don't know how my tweaks could have caused it to fail, but if someone could point me in the correct direction for that error, I will take a look at trying to figure it out.

client/cmd/login.go
```diff
@@ -39,6 +39,13 @@ var loginCmd = &cobra.Command{
 			ctx = context.WithValue(ctx, system.DeviceNameCtxKey, hostName)
 		}
 
+		if setupKeyPath != "" && setupKey == "" {
+			setupKey, err = getSetupKeyFromFile(setupKeyPath)
```
mlsmaycon, line 43: Better to use a new variable instead of overwriting the global used in the flags.

client/cmd/up.go
```diff
@@ -73,6 +73,13 @@ func upFunc(cmd *cobra.Command, args []string) error {
 		ctx = context.WithValue(ctx, system.DeviceNameCtxKey, hostName)
 	}
 
+	if setupKeyPath != "" && setupKey == "" {
```
mlsmaycon, line 76: can you move this to a single function?

client/cmd/root.go
```diff
@@ -246,6 +249,11 @@ var CLIBackOffSettings = &backoff.ExponentialBackOff{
 	Clock:               backoff.SystemClock,
 }
 
+func getSetupKeyFromFile(setupKeyPath string) (string, error) {
```
mlsmaycon, line 252: let's use strings.TrimSpace and strings.Trim(string(data),"\n") to avoid issues

mlsmaycon (changes requested):
Thanks for the PR @moosetheory. I've left some comments to it.

mlsmaycon (approved):
Thanks for the contribution @moosetheory. We will take the requested changes internally.

moosetheory:
Oh, thanks! I've been running into some test issues and have not been able to figure them out. I'm glad you are able to!

Thanks a ton!

mlsmaycon:
no worries, here is the PR in case you want to have a look https://github.com/netbirdio/netbird/pull/2411

PR #2359, opened 2024-07-31
Add write lock for peer when saving its connection status

mlsmaycon:
## Describe your changes
This pull request adds a write lock mechanism for peers when saving their connection status. The primary goal is to ensure thread safety and prevent race conditions during peer status updates.

Key Changes

- Implemented write locks for peers by their key.
- Utilized read account locks for methods used by grpcServer.Sync.
- Renamed lock methods for clarity.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -1858,6 +1858,8 @@ func (am *DefaultAccountManager) getAccountWithAuthorizationClaims(ctx context.C
 }
 
 func (am *DefaultAccountManager) SyncAndMarkPeer(ctx context.Context, peerPubKey string, meta nbpeer.PeerSystemMeta, realIP net.IP) (*nbpeer.Peer, *NetworkMap, []*posture.Checks, error) {
+	peerUnlock := am.Store.AcquireWriteLockByUID(ctx, peerPubKey)
+	defer peerUnlock()
 	accountID, err := am.Store.GetAccountIDByPeerPubKey(ctx, peerPubKey)
 	if err != nil {
 		if errStatus, ok := status.FromError(err); ok && errStatus.Type() == status.NotFound {
@@ -1866,8 +1868,8 @@ func (am *DefaultAccountManager) SyncAndMarkPeer(ctx context.Context, peerPubKey
 		return nil, nil, nil, err
 	}
 
-	unlock := am.Store.AcquireReadLockByUID(ctx, accountID)
-	defer unlock()
+	accountUnlock := am.Store.AcquireReadLockByUID(ctx, accountID)
+	defer accountUnlock()
 
 	account, err := am.Store.GetAccount(ctx, accountID)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


management/server/account.go
```diff
@@ -1888,6 +1890,8 @@ func (am *DefaultAccountManager) SyncAndMarkPeer(ctx context.Context, peerPubKey
 }
 
 func (am *DefaultAccountManager) CancelPeerRoutines(ctx context.Context, peer *nbpeer.Peer) error {
+	peerUnlock := am.Store.AcquireWriteLockByUID(ctx, peer.Key)
```
pascal-fischer: We should move the peerLock after the account lock otherwise we might still experience overwrites if the peerLock is called while an accountWrite lock is currently active

management/server/file_store.go
```diff
@@ -281,26 +281,26 @@ func (s *FileStore) AcquireGlobalLock(ctx context.Context) (unlock func()) {
...
 
-// AcquireAccountWriteLock acquires account lock for writing to a resource and returns a function that releases the lock
-func (s *FileStore) AcquireAccountWriteLock(ctx context.Context, accountID string) (unlock func()) {
-	log.WithContext(ctx).Debugf("acquiring lock for account %s", accountID)
+// AcquireWriteLockByUID acquires an ID lock for writing to a resource and returns a function that releases the lock
+func (s *FileStore) AcquireWriteLockByUID(ctx context.Context, uniqueID string) (unlock func()) {
+	log.WithContext(ctx).Debugf("acquiring lock for ID %s", uniqueID)
 	start := time.Now()
-	value, _ := s.accountLocks.LoadOrStore(accountID, &sync.Mutex{})
+	value, _ := s.accountLocks.LoadOrStore(uniqueID, &sync.Mutex{})
```
pascal-fischer: The `s.accountLocks` should be renamed just to account for a types. Maybe call it resourceLocks

management/server/file_store.go
```diff
@@ -281,26 +281,26 @@ func (s *FileStore) AcquireGlobalLock(ctx context.Context) (unlock func()) {
...
 	mtx := value.(*sync.Mutex)
 	mtx.Lock()
 
 	unlock = func() {
 		mtx.Unlock()
-		log.WithContext(ctx).Debugf("released lock for account %s in %v", accountID, time.Since(start))
+		log.WithContext(ctx).Debugf("released lock for ID %s in %v", uniqueID, time.Since(start))
```
pascal-fischer, line 294: An idea would be to have an additional function parameter `resourceType` which can be an enum so that logging can include what we are locking which might help us debugging

management/server/sql_store.go
```diff
@@ -98,33 +98,35 @@ func (s *SqlStore) AcquireGlobalLock(ctx context.Context) (unlock func()) {
...
 
-func (s *SqlStore) AcquireAccountReadLock(ctx context.Context, accountID string) (unlock func()) {
-	log.WithContext(ctx).Tracef("acquiring read lock for account %s", accountID)
+// AcquireReadLockByUID acquires an ID lock for writing to a resource and returns a function that releases the lock
+func (s *SqlStore) AcquireReadLockByUID(ctx context.Context, uniqueID string) (unlock func()) {
+	log.WithContext(ctx).Tracef("acquiring read lock for ID %s", uniqueID)
 
 	start := time.Now()
-	value, _ := s.accountLocks.LoadOrStore(accountID, &sync.RWMutex{})
+	value, _ := s.accountLocks.LoadOrStore(uniqueID, &sync.RWMutex{})
```
pascal-fischer: Same remarks as for the file store

pascal-fischer (changes requested):

management/server/file_store.go
mlsmaycon: let's do it in another PR

pascal-fischer (approved):

PR #2370, opened 2024-08-01
Add batch delete for groups and users

bcmmbaga:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/user.go
```diff
@@ -1190,6 +1158,116 @@ func (am *DefaultAccountManager) getEmailAndNameOfTargetUser(ctx context.Context
 	return "", "", fmt.Errorf("user info not found for user: %s", targetId)
 }
 
+// DeleteRegularUsers deletes regular users from an account.
+// Note: This function does not acquire the global lock.
+// It is the caller's responsibility to ensure proper locking is in place before invoking this method.
+//
+// If an error occurs while deleting the user, the function skips it and continues deleting other users.
+// Errors are collected and returned at the end.
+func (am *DefaultAccountManager) DeleteRegularUsers(ctx context.Context, accountID, initiatorUserID string, targetUserIDs []string) error {
```
pappz, line 1167: Is it possible to pass the service user to this function? If yes then should check the type of the user in the for iteration;

bcmmbaga: No. This will only handle regular users for now.

PR #2378, opened 2024-08-02
Fix DNS resolution for routes on iOS

pascal-fischer:
## Describe your changes
When adding DNS routes that are only resolved by a private nameserver the resolution was failing. This PR uses the custom dailer form the dns package also in dns routes.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/routemanager/dynamic/route.go
+++ b/client/internal/routemanager/dynamic/route.go
@@ -13,9 +13,11 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	nberrors "github.com/netbirdio/netbird/client/errors"
+	nbdns "github.com/netbirdio/netbird/client/internal/dns"
 	"github.com/netbirdio/netbird/client/internal/peer"
 	"github.com/netbirdio/netbird/client/internal/routemanager/refcounter"
 	"github.com/netbirdio/netbird/client/internal/routemanager/util"
+	"github.com/netbirdio/netbird/iface"
 	"github.com/netbirdio/netbird/management/domain"
 	"github.com/netbirdio/netbird/route"
 )
@@ -47,6 +49,8 @@ type Route struct {
 	currentPeerKey       string
 	cancel               context.CancelFunc
 	statusRecorder       *peer.Status
+	wgInterface          *iface.WGIface
+	serviceViaMemory     *nbdns.ServiceViaMemory
 }
 
 func NewRoute(
@@ -55,6 +59,7 @@ func NewRoute(
 	allowedIPsRefCounter *refcounter.AllowedIPsRefCounter,
 	interval time.Duration,
 	statusRecorder *peer.Status,
+	wgInterface *iface.WGIface,
 ) *Route {
 	return &Route{
 		route:                rt,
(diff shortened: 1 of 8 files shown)


client/internal/routemanager/dynamic/route_ios.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/netbirdio/netbird/management/domain"
+)
+
+const dialTimeout = 1 * time.Second
+
+func (r *Route) getIPsFromResolver(domain domain.Domain) ([]net.IP, error) {
+	privateClient, err := nbdns.GetClientPrivate(r.wgInterface.Address().IP, r.wgInterface.Name(), dialTimeout)
+	if err != nil {
+		log.Debugf("DNS query for %s failed: %s", domain.SafeString(), err)
+		return nil, err
```
lixmal: Can we add some context to the errors here for easier location?

client/internal/routemanager/dynamic/route_ios.go
```diff
@@ -0,0 +1,51 @@
...
+
+	log.Debugf("DNS query for %s took %s", domain.SafeString(), time.Since(startTime))
+
+	if response.Rcode != dns.RcodeSuccess {
+		return nil, fmt.Errorf("dns response code: %s", dns.RcodeToString[response.Rcode])
+	}
+
+	for _, ans := range response.Answer {
+		if aRecord, ok := ans.(*dns.A); ok {
+			return []net.IP{aRecord.A}, nil
```
lixmal: We need to return all records, not just one.
Also can we support AAAA as well? So we don't have to touch it at a later point

client/internal/routemanager/dynamic/route_ios.go
```diff
@@ -0,0 +1,51 @@
...
+
+	startTime := time.Now()
+
+	response, _, err := privateClient.Exchange(msg, r.resolverAddr)
+	if err != nil {
+		log.Debugf("DNS query for %s failed after %s: %s ", domain.SafeString(), time.Since(startTime), err)
+		return nil, err
+	}
+
+	log.Debugf("DNS query for %s took %s", domain.SafeString(), time.Since(startTime))
```
lixmal: - This should probably be a `Tracef`
- Maybe put this in line 33 instead and drop the one in line 34 altogether

client/internal/routemanager/dynamic/route_ios.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/miekg/dns"
+	log "github.com/sirupsen/logrus"
+
+	nbdns "github.com/netbirdio/netbird/client/internal/dns"
+
+	"github.com/netbirdio/netbird/management/domain"
+)
+
+const dialTimeout = 1 * time.Second
```
lixmal: Maybe the timeout is a bit low for servers via VPN? Speaking of, we don't use any timeout in the regular LookupIP, so we should probably streamline these two

client/internal/routemanager/dynamic/route.go
```diff
@@ -228,11 +235,16 @@ func (r *Route) resolve(results chan resolveResult) {
 		wg.Add(1)
 		go func(domain domain.Domain) {
 			defer wg.Done()
-			ips, err := net.LookupIP(string(domain))
+
+			ips, err := r.getIPsFromResolver(domain)
 			if err != nil {
```
lixmal, line 240: We should log this error, and probably drop the `log.Debugf`s within `getIPsFromResolver` in the `err != nil` clauses. Alternatively also log that one error in `getIPsFromResolver`

lixmal (dismissed):

lixmal (approved):

PR #2399, opened 2024-08-06
[relay] Feature/relay integration r53

pappz:
## Describe your changes

Support route53 for TLS handling

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


relay/cmd/main.go
```diff
@@ -54,6 +61,22 @@ func (c Config) HasLetsEncrypt() bool {
 	return c.LetsencryptDataDir != "" && c.LetsencryptDomains != nil && len(c.LetsencryptDomains) > 0
 }
 
+func (c Config) HasRoute54Credentials() bool {
```
mlsmaycon:
```suggestion
func (c Config) HasRoute53Credentials() bool {
```

relay/cmd/main.go
```diff
@@ -73,6 +81,8 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&cobraConfig.ExposedAddress, "exposed-address", "e", "", "instance domain address (or ip) and port, it will be distributes between peers")
 	rootCmd.PersistentFlags().StringVarP(&cobraConfig.LetsencryptDataDir, "letsencrypt-data-dir", "d", "", "a directory to store Let's Encrypt data. Required if Let's Encrypt is enabled.")
 	rootCmd.PersistentFlags().StringArrayVarP(&cobraConfig.LetsencryptDomains, "letsencrypt-domains", "a", nil, "list of domains to issue Let's Encrypt certificate for. Enables TLS using Let's Encrypt. Will fetch and renew certificate, and run the server with TLS")
```
mlsmaycon:
```suggestion
	rootCmd.PersistentFlags().StringSliceVarP(&cobraConfig.LetsencryptDomains, "letsencrypt-domains", "a", nil, "list of domains to issue Let's Encrypt certificate for. Enables TLS using Let's Encrypt. Will fetch and renew certificate, and run the server with TLS")
```

This allows strings like: domain1.com,domain2.com

encryption/route53.go
```diff
@@ -0,0 +1,61 @@
...
+	CA      string
+}
+
+func (r *Route53TLS) GetCertificate() (*tls.Config, error) {
+	certmagic.Default.Logger = logger()
+	certmagic.Default.Storage = &certmagic.FileStorage{Path: r.DataDir}
+	certmagic.DefaultACME.Agreed = true
+	certmagic.DefaultACME.Email = r.Email
+	if r.CA == "" {
+		certmagic.DefaultACME.CA = certmagic.LetsEncryptStagingCA
```
mlsmaycon:
```suggestion
		certmagic.DefaultACME.CA =certmagic.LetsEncryptProductionCA
```

mlsmaycon (changes requested):
Can you please check how we can avoid this case:

```
Your sites will be served over HTTPS automatically using an automated CA.
By continuing, you agree to the CA's terms of service.
Please enter your email address to signify agreement and to be notified
in case of issues. You can leave it blank, but we don't recommend it.
  Email address: 
```

We should not expect an interactive input when deploying the relay on containers. So we should always skip the email if the user didn't provide it

pappz:
> Can you please check how we can avoid this case:
> 
> ```
> Your sites will be served over HTTPS automatically using an automated CA.
> By continuing, you agree to the CA's terms of service.
> Please enter your email address to signify agreement and to be notified
> in case of issues. You can leave it blank, but we don't recommend it.
>   Email address: 
> ```
> 
> We should not expect an interactive input when deploying the relay on containers. So we should always skip the email if the user didn't provide it

Actually, I can not reproduce the issue on my machine (I tried with the example command that you provided), but I pushed a commit what could fix it.

mlsmaycon (approved):

Issue #2011, opened 2024-05-18
netbird service / connection is down when returning from sleep / hibernation


ez1976:
Hi 
on windows it seems that when the computer is resumed from sleep or Hibernation, although the service is up and the netbird-gui display as connected, the user gets:
"dial service - context deadline exceeded"

of course reboot / restart of the service / running "netbird down && nebird up" or "netbird service restart"
will resume the connection just fine.
attached one of the users client log
at line 10980 the user restarted the computer (or the service)
[Uploading netbird-client.log…]()

mlsmaycon:
Can you upgrade the client to 0.27.7 and enable the network monitor? 

You can do that by:
```
netbird down
netbird up -N
```

ez1976:
On it

hurricanehrndz:
I can also create a PR to update the UI on the full-status @mlsmaycon if you all want. The reason again is because of the internal state never gets updated when signal and management endpoints become unavailable.

hurricanehrndz:
```
2024-05-08T14:28:50+01:00 DEBG client/internal/engine.go:1135: removing Netbird interface utun100
2024-05-08T14:28:50+01:00 ERRO client/internal/connect.go:245: error while starting Netbird Connection Engine: create wg interface: resource busy
```

hurricanehrndz:
I believe Issue#2006 is related

ez1976:
Try
netbird down
netbird up -N

ez1976:
No.
The -N adds a network monitor that resets the connection if the network
adapter status changes.
It is disabled by default

ez1976:
Or even better, enabled by default.
After the laptop resumes from sleep give it a bit of time and it will
restart the service.
On windows laptops i actually pushed a task schedule to restart the service
every time the laptop restarts or resumes from sleep

clarkmcc:
> Try
netbird down
netbird up -N

This did not work for me. After running both commands I get
```shell
$ netbird status
Error: status failed: create wg interface: resource busy
```

ez1976:
Do you have virtualbox or any other virtualization on your computer?

clarkmcc:
I have Parallels. I was able to resolve it by killing the NetBird service and restarting it.

mlsmaycon:
Hello folks, recently we released a new version with a new handler. Can you run a test with the latest version?

PR #2406, opened 2024-08-07
Engine should restart and block more restarts

hurricanehrndz:
## Describe your changes
Block rapid restarts in succession for 3s.

Possible fix for #2196 #2011
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -13,6 +13,7 @@ import (
 	"slices"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/pion/ice/v3"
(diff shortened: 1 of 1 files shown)


lixmal:
What problem is this solving? Would increasing the timeout suffice?

hurricanehrndz:
> What problem is this solving? Would increasing the timeout suffice?

Increasing the timeout might. The only difference between your approach and this approach is that engine restart is initiate and then blocked, meanwhile the current code base waits to restart. Either way as long as utun100 resource busy is no longer happening this can be closed

We should wait until @christian-schlichtherle or @fadyHemaya have responded to the issues to confirm 28.5+ and the current code base doesn't result in utun100 resource busy.

lixmal:
> > What problem is this solving? Would increasing the timeout suffice?
> 
> The only difference between your approach and this approach is that engine restart is initiate and then blocked, meanwhile the current code base waits to restart.

We need to pick up the latest network state (routes), that's why we cannot just ignore changes for X seconds.
If we miss the latest change in a row of route changes we might set up routes pointing to the wrong thing and netbird might be broken. I don't might increasing the timeout, though.

hurricanehrndz:
Good point, thank you for the clarification

PR #2410, opened 2024-08-09
[management] Prevent removal of All group from peers during user groups propagation 

bcmmbaga:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/setupkey.go
+++ b/management/server/setupkey.go
@@ -224,8 +224,12 @@ func (am *DefaultAccountManager) CreateSetupKey(ctx context.Context, accountID s
 	}
 
 	for _, group := range autoGroups {
-		if _, ok := account.Groups[group]; !ok {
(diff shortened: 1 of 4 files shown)


management/server/setupkey.go
```diff
@@ -279,6 +283,16 @@ func (am *DefaultAccountManager) SaveSetupKey(ctx context.Context, accountID str
 		return nil, status.Errorf(status.NotFound, "setup key not found")
 	}
 
+	for _, group := range keyToSave.AutoGroups {
+		g, ok := account.Groups[group]
```
mlsmaycon: let's make these two checks a function check for both, create and save setup key

mlsmaycon (changes requested):

mlsmaycon (approved):

PR #2430, opened 2024-08-14
Handle route expansion on client

lixmal:
## Describe your changes

- Handle mgmt server handing out sources instead of single source
- Adds ipset to iptables and sets for nftables when `len(sources) > 1`
- Make refcounter more generic for that purpose

### Checklist
- [x] Is it a bug fix
- [x] It is a refactor


pappz:
After I stop the agent the **table ip netbird** still exists. I tested with nft.

PR #2435, opened 2024-08-15
[client] Destory WG interface on down timeout

pascal-fischer:
## Describe your changes
This PR updates the wait on engine down to not only wait for the interface to be down but completely removed. If the waiting loop reaches the timeout we will trigger an interface destroy. On the up command, it now waits until the engine is fully running before sending the response to the CLI. Includes a small refactor of probes to comply with sonar rules about parameter count in the function call.

### Checklist
- [x] Is it a bug fix
- [x] It is a refactor


iface/iface_destroy_linux.go
```diff
@@ -0,0 +1,22 @@
+//go:build linux
```
lixmal: This needs a no android flag

iface/iface.go
```diff
@@ -163,3 +179,29 @@ func (w *WGIface) GetDevice() *DeviceWrapper {
 func (w *WGIface) GetStats(peerKey string) (WGStats, error) {
 	return w.configurer.getStats(peerKey)
 }
+
+func (w *WGIface) waitUntilRemoved() error {
+	maxWaitTime := 5 * time.Second
+	timeout := time.After(maxWaitTime)
```
lixmal: We should free this timer to avoid leaks

client/server/server.go
```diff
@@ -589,29 +588,19 @@ func (s *Server) Down(ctx context.Context, _ *proto.DownRequest) (*proto.DownRes
 	if s.actCancel == nil {
 		return nil, fmt.Errorf("service is not up")
 	}
+	log.Debugf("Should continue doing stuff")
```
lixmal: debug

PR #2446, opened 2024-08-18
[client] Feature/relay integration fix wgstate

pappz:
## Describe your changes

- The agent checks the state of the connection with the WireGuard last handshake information. It could cause an inconsistent state with the ICE connection if we open the Relay connection too. In this PR the WireGuard check logic will be activated when the Relay connection is active.
- After WireGuard check activation we do not need to wait the full timeout period because for any peer configuration, the system sends a new handshake. 

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/internal/peer/worker_relay.go
```diff
@@ -14,7 +15,7 @@ import (
 
 var (
 	wgHandshakePeriod   = 2 * time.Minute
-	wgHandshakeOvertime = 30000 * time.Millisecond
+	wgHandshakeOvertime = 30 * time.Second //
```
lixmal:
```suggestion
	wgHandshakeOvertime = 30 * time.Second
```

lixmal (approved):

PR #2449, opened 2024-08-19
[management] Add buffering for getAccount requests during login

pascal-fischer:
## Describe your changes
This PR adds a buffer for the getAccount operation during login to collect requests over a specific time inteval and only call the database once. This way reducing overall load on the DB.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -9,11 +9,9 @@ import (
 	"math/rand"
 	"net"
 	"net/netip"
-	"os"
 	"reflect"
 	"regexp"
 	"slices"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -163,9 +161,7 @@ type DefaultAccountManager struct {
 	eventStore           activity.Store
 	geo                  *geolocation.Geolocation
 
-	getAccountRequests  map[string][]*AccountRequest
-	mu                  sync.Mutex
-	getAccountRequestCh chan *AccountRequest
+	cache *AccountCache
 
 	// singleAccountMode indicates whether the instance has a single account.
 	// If true, then every new user will end up under the same account.
@@ -285,18 +281,6 @@ type UserInfo struct {
 	Issued               string                                     `json:"issued"`
 	IntegrationReference integration_reference.IntegrationReference `json:"-"`
 	Permissions          UserPermissions                            `json:"permissions"`
-}
-
-// AccountRequest holds the result channel to return the requested account.
-type AccountRequest struct {
-	AccountID  string
-	ResultChan chan *AccountResult
-}
-
-// AccountResult holds the account data or an error.
-type AccountResult struct {
-	Account *Account
-	Err     error
 }
 
 // getRoutesToSync returns the enabled routes for the peer ID and the routes
@@ -982,11 +966,10 @@ func BuildManager(
 		dnsDomain:                dnsDomain,
 		eventStore:               eventStore,
 		peerLoginExpiry:          NewDefaultScheduler(),
-		getAccountRequests:       make(map[string][]*AccountRequest),
-		getAccountRequestCh:      make(chan *AccountRequest),
 		userDeleteFromIDPEnabled: userDeleteFromIDPEnabled,
 		integratedPeerValidator:  integratedPeerValidator,
 		metrics:                  metrics,
+		cache:                    NewAccountCache(ctx, store),
 	}
 	allAccounts := store.GetAllAccounts(ctx)
 	// enable single account mode only if configured by user and number of existing accounts is not grater than 1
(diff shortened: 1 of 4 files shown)


management/server/account.go
```diff
@@ -2206,3 +2226,35 @@ func separateGroups(autoGroups []string, allGroups map[string]*nbgroup.Group) ([
...
+func (am *DefaultAccountManager) GetAccountWithBackpressure(ctx context.Context, accountID string) (*Account, error) {
+	req := &Request{
+		AccountID:  accountID,
+		ResultChan: make(chan *AccountResult, 1),
+	}
+
+	am.requestCh <- req
+
+	result := <-req.ResultChan
+	close(req.ResultChan)
```
skillcoder: I just want to highlight, and will be appreciate if you point me why I'm wrong in this case.
As far as I know 
The sender should close the channel in Go because the sender knows when it's done sending data. 
Sending data to closed channel will lead to panic.

pascal-fischer: You are right, I played with the channel a bit and on the last commit I changed it. I've moved it back to the sender now

management/server/account.go
```diff
@@ -2206,3 +2226,39 @@ func separateGroups(autoGroups []string, allGroups map[string]*nbgroup.Group) ([
 	}
 	return newAutoGroups, jwtAutoGroups
 }
+
```
pappz: The AM is already more than 2200 lines monster. New features, and logic should move separated structure. If I see well the new functions are deepened from the store only. 
```
type AccountCache struct {
        store Store
  	getAccountRequests  map[string][]*AccountRequest
	mu                  sync.Mutex
	getAccountRequestCh chan *AccountRequest
}
func NewAccountCache(store Store) *AccountCache{
...
}
func (ac *AccountCache) GetAccountWithBackpressure(ctx context.Context, accountID string) (*Account, error) {
...
}
func (am *AccountCache) processGetAccountBatch(ctx context.Context, accountID string) {
...
}
func (am *AccountCache) processGetAccountRequests(ctx context.Context) {
...
}
```

management/server/peer.go
```diff
@@ -971,3 +973,28 @@ func (am *DefaultAccountManager) updateAccountPeers(ctx context.Context, account
...
+
+func (am *DefaultAccountManager) processGetAccountRequests(ctx context.Context) {
+	for {
+		select {
+		case req := <-am.getAccountRequestCh:
+			am.mu.Lock()
+			am.getAccountRequests[req.AccountID] = append(am.getAccountRequests[req.AccountID], req)
+			if len(am.getAccountRequests[req.AccountID]) == 1 {
+				go func(ctx context.Context, accountID string) {
+					bufferIntervalStr := os.Getenv("NB_GET_ACCOUNT_BUFFER_INTERVAL")
```
pappz: The env vars will not change. Move the bufferInterval initialization to an init() function or into a constructor.

management/server/account.go
```diff
@@ -2206,3 +2226,39 @@ func separateGroups(autoGroups []string, allGroups map[string]*nbgroup.Group) ([
...
+		AccountID:  accountID,
+		ResultChan: make(chan *AccountResult, 1),
+	}
+
+	log.WithContext(ctx).Debugf("requesting account with backpressure: %s", accountID)
+	startTime := time.Now()
+	am.getAccountRequestCh <- req
+
+	result := <-req.ResultChan
+	log.WithContext(ctx).Debugf("got account with backpressure after %s", time.Since(startTime))
```
pappz: If this measurement is essential then set to Trace level.

management/server/account.go
```diff
@@ -2206,3 +2226,39 @@ func separateGroups(autoGroups []string, allGroups map[string]*nbgroup.Group) ([
...
+
+func (am *DefaultAccountManager) GetAccountWithBackpressure(ctx context.Context, accountID string) (*Account, error) {
+	req := &AccountRequest{
+		AccountID:  accountID,
+		ResultChan: make(chan *AccountResult, 1),
+	}
+
+	log.WithContext(ctx).Debugf("requesting account with backpressure: %s", accountID)
+	startTime := time.Now()
+	am.getAccountRequestCh <- req
```
pappz: I am not sure about the teardown procedure of the Management service but if you handle the cancel context in the `processGetAccountRequests` then should handle the here too.

management/server/account_cache.go
```diff
@@ -0,0 +1,105 @@
...
+	mu                  sync.Mutex
+	getAccountRequestCh chan *AccountRequest
+	bufferInterval      time.Duration
+}
+
+func NewAccountCache(ctx context.Context, store Store) *AccountCache {
+	bufferIntervalStr := os.Getenv("NB_GET_ACCOUNT_BUFFER_INTERVAL")
+	bufferInterval, err := time.ParseDuration(bufferIntervalStr)
+	if err != nil {
+		bufferInterval = 300 * time.Millisecond
```
pappz, line 37: An error log could help to find an accidental typo

pappz (approved):

PR #2456, opened 2024-08-20
[relay] Feature/relay remove cfg file

pappz:
## Describe your changes

- Remove config file support
- Add ENV variable support

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/encryption/route53.go
+++ b/encryption/route53.go
@@ -32,10 +32,10 @@ func (r *Route53TLS) GetCertificate() (*tls.Config, error) {
 	certmagic.Default.Logger = logger()
 	certmagic.Default.Storage = &certmagic.FileStorage{Path: r.DataDir}
 	certmagic.DefaultACME.Agreed = true
-	if r.Email == "" {
-		certmagic.DefaultACME.Email = emailFromDomain(r.Domains[0])
+	if r.Email != "" {
+		certmagic.DefaultACME.Email = r.Email
 	} else {
-		certmagic.DefaultACME.Email = ""
+		certmagic.DefaultACME.Email = emailFromDomain(r.Domains[0])
 	}
 
 	if r.CA == "" {
--- a/encryption/route53.go
+++ b/encryption/route53.go
@@ -3,7 +3,9 @@ package encryption
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
 	"os"
+	"strings"
 
 	"github.com/caddyserver/certmagic"
 	"github.com/libdns/route53"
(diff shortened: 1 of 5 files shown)


relay/cmd/main.go
```diff
@@ -76,7 +73,6 @@ var (
 func init() {
 	_ = util.InitLog("trace", "console")
```
lixmal, line 75: Do we still need this?

relay/cmd/main.go
```diff
@@ -85,7 +81,11 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&cobraConfig.LetsencryptAWSRoute53, "letsencrypt-aws-route53", false, "use AWS Route 53 for Let's Encrypt DNS challenge")
 	rootCmd.PersistentFlags().StringVarP(&cobraConfig.TlsCertFile, "tls-cert-file", "c", "", "")
 	rootCmd.PersistentFlags().StringVarP(&cobraConfig.TlsKeyFile, "tls-key-file", "k", "", "")
-	rootCmd.PersistentFlags().StringVarP(&cobraConfig.AuthSecret, "auth-secret", "s", "", "log level")
+	rootCmd.PersistentFlags().StringVarP(&cobraConfig.AuthSecret, "auth-secret", "s", "", "auth secret")
+	rootCmd.PersistentFlags().StringVar(&cobraConfig.LogLevel, "log-level", "info", "log level")
+	rootCmd.PersistentFlags().StringVar(&cobraConfig.LogFile, "log-path", "console", "log path")
```
lixmal: should this be `log-file` to be consistent with mgmt server etc?

relay/cmd/main.go
```diff
@@ -85,7 +81,11 @@ func init() {
...
+
+	setFlagsFromEnvVars(rootCmd)
```
lixmal, line 89: The order usually is env -> flags, i.e. cli flags override any env vars. At least that's how aws & friends handle it. Now I must admit I'm unsure how we handle it in our other components, maybe that should take precedence.

pappz: You are right but this code does what you expect.
In this example the listen address will be :1234
```
NB_LISTEN_ADDRESS=:654 ./cmd --listen-address=:1234 
```
But fix me if I mixed something.

relay/cmd/main.go
pappz: I think so. If we have any issues before line 104. (where we process the log parameters) then I still want to use our logging format.

relay/cmd/main.go
lixmal: Apologies, indeed you are right. I misinterpreted how we visit the flags

lixmal (approved):

encryption/route53.go
```diff
@@ -23,14 +25,17 @@ type Route53TLS struct {
...
 	certmagic.Default.Logger = logger()
 	certmagic.Default.Storage = &certmagic.FileStorage{Path: r.DataDir}
 	certmagic.DefaultACME.Agreed = true
-
 	if r.Email == "" {
-		certmagic.DefaultACME.Email = "myemail@example.com"
+		certmagic.DefaultACME.Email = emailFromDomain(r.Domains[0])
 	} else {
-		certmagic.DefaultACME.Email = r.Email
+		certmagic.DefaultACME.Email = ""
```
mlsmaycon: should be `= r.Email`

mlsmaycon (approved):

PR #2461, opened 2024-08-21
[management] Add support to ECDSA public Keys

HarryKodden:
### Checklist
- [ ] Is it a bug fix
- [X] It is a refactor

### DIFF
--- a/management/server/jwtclaims/jwtValidator.go
+++ b/management/server/jwtclaims/jwtValidator.go
@@ -39,11 +39,6 @@ type Options struct {
 	// When set, all requests with the OPTIONS method will use authentication
 	// Default: false
 	EnableAuthOnOptions bool
-	// When set, the middelware verifies that tokens are signed with the specific signing algorithm
-	// If the signing method is not constant the ValidationKeyGetter callback can be used to implement additional checks
-	// Important to avoid security issues described here: https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/
-	// Default: nil
-	SigningMethod jwt.SigningMethod
 }
 
 // Jwks is a collection of JSONWebKey obtained from Config.HttpServerConfig.AuthKeysLocation
@@ -136,8 +131,6 @@ func NewJWTValidator(ctx context.Context, issuer string, audienceList []string, 
 
 			return publicKey, nil
 		},
-		// Canged SigningMethod to 'nil', since we can't tell yet...
-		SigningMethod:       nil, // jwt.SigningMethodRS256,
 		EnableAuthOnOptions: false,
 	}
 
@@ -174,14 +167,6 @@ func (m *JWTValidator) ValidateAndParse(ctx context.Context, token string) (*jwt
 	if err != nil {
 		log.WithContext(ctx).Errorf("error parsing token: %v", err)
 		return nil, fmt.Errorf("error parsing token: %w", err)
-	}
-
-	if m.options.SigningMethod != nil && m.options.SigningMethod.Alg() != parsedToken.Header["alg"] {
-		errorMsg := fmt.Sprintf("Expected %s signing method but token specified %s",
-			m.options.SigningMethod.Alg(),
-			parsedToken.Header["alg"])
-		log.WithContext(ctx).Debugf("error validating token algorithm: %s", errorMsg)
-		return nil, fmt.Errorf("error validating token algorithm: %s", errorMsg)
 	}
 
 	// Check if the parsed token is valid...
@@ -284,7 +269,7 @@ func getPublicKeyFromECDSA(jwk JSONWebKey) (publicKey *ecdsa.PublicKey, err erro
 
 func getPublicKeyFromRSA(jwk JSONWebKey) (*rsa.PublicKey, error) {
 
-  decodedE, err := base64.RawURLEncoding.DecodeString(jwk.E)
+	decodedE, err := base64.RawURLEncoding.DecodeString(jwk.E)
 	if err != nil {
 		return nil, err
 	}
--- a/management/server/jwtclaims/jwtValidator.go
+++ b/management/server/jwtclaims/jwtValidator.go
@@ -130,7 +130,7 @@ func NewJWTValidator(ctx context.Context, issuer string, audienceList []string, 
 
 			publicKey, err := getPublicKey(token, keys)
 			if err != nil {
-				log.Errorf("getPublicKey error: %s", err)
+				log.WithContext(ctx).Errorf("getPublicKey error: %s", err)
 				return nil, err
 			}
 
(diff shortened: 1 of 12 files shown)


HarryKodden:
@mlsmaycon Hi Maycon, I made a new PR and this is having no issues. Is this OK to proceed with ?

mlsmaycon:
Thanks, @HarryKodden. Yes, can you reference it here and close the other one?

HarryKodden:
This PR replaces https://github.com/netbirdio/netbird/pull/2101 since that was having merging issues.
2101 will be closed.

management/server/jwtclaims/jwtValidator.go
```diff
@@ -205,83 +219,88 @@ func getPemKeys(ctx context.Context, keysLocation string) (*Jwks, error) {
...
-	if err != nil {
-		return "", fmt.Errorf("unable to decode JWK exponent, error: %s", err)
-	}
+func getPublicKeyFromECDSA(jwk JSONWebKey) (publicKey *ecdsa.PublicKey, err error) {
 
-	publicKey := &rsa.PublicKey{
-		N: intModules,
-		E: exponent,
+	if jwk.X == "" || jwk.Y == "" || jwk.Crv == "" {
+		return nil, fmt.Errorf("ecdsa key incompleet")
```
bcmmbaga:
```suggestion
		return nil, fmt.Errorf("ecdsa key incomplete")
```

management/server/jwtclaims/jwtValidator.go
```diff
@@ -115,15 +128,16 @@ func NewJWTValidator(ctx context.Context, issuer string, audienceList []string,
 				}
 			}
 
-			cert, err := getPemCert(ctx, token, keys)
+			publicKey, err := getPublicKey(token, keys)
 			if err != nil {
+				log.Errorf("getPublicKey error: %s", err)
```
bcmmbaga: Let's log the error with context

```suggestion
				log.WithContext(ctx).Errorf("cannot get PublicKey: %s", err)
```

management/server/jwtclaims/jwtValidator.go
```diff
@@ -115,15 +128,16 @@ func NewJWTValidator(ctx context.Context, issuer string, audienceList []string,
...
 				return nil, err
 			}
 
-			result, _ := jwt.ParseRSAPublicKeyFromPEM([]byte(cert))
-			return result, nil
+			return publicKey, nil
 		},
-		SigningMethod:       jwt.SigningMethodRS256,
+		// Canged SigningMethod to 'nil', since we can't tell yet...
+		SigningMethod:       nil, // jwt.SigningMethodRS256,
```
bcmmbaga: We still needs to add the correct signing method here. As `nil` will invalidate this check https://github.com/netbirdio/netbird/blob/ddea0011709f091acd902d871822187f75fb091d/management/server/jwtclaims/jwtValidator.go#L165-L171

management/server/jwtclaims/jwtValidator.go
```diff
@@ -205,83 +219,88 @@ func getPemKeys(ctx context.Context, keysLocation string) (*Jwks, error) {
...
+		if jwks.Keys[k].Kty == "EC" {
+			log.Debugf("generating PublicKey from ECDSA JWK")
+			return getPublicKeyFromECDSA(jwks.Keys[k])
+		}
 
-func generatePemFromJWK(jwk JSONWebKey) (string, error) {
-	decodedModulus, err := base64.RawURLEncoding.DecodeString(jwk.N)
-	if err != nil {
-		return "", fmt.Errorf("unable to decode JWK modulus, error: %s", err)
+		log.Debugf("Key Type: %s not yet supported, please raise ticket !", jwks.Keys[k].Kty)
```
bcmmbaga: Let's log the error with context

```suggestion
		log.WithContext(ctx).Debugf("Key Type: %s not yet supported, please raise ticket!", jwks.Keys[k].Kty)
```

management/server/jwtclaims/jwtValidator.go
```diff
@@ -205,83 +219,88 @@ func getPemKeys(ctx context.Context, keysLocation string) (*Jwks, error) {
...
+			return jwt.ParseRSAPublicKeyFromPEM([]byte(cert))
 		}
-		log.WithContext(ctx).Debugf("generating validation pem from JWK")
-		return generatePemFromJWK(jwks.Keys[k])
-	}
 
-	return cert, errors.New("unable to find appropriate key")
-}
+		if jwks.Keys[k].Kty == "RSA" {
+			log.Debugf("generating PublicKey from RSA JWK")
```
bcmmbaga: Same here. Lets add log with context

management/server/jwtclaims/jwtValidator.go
```diff
@@ -205,83 +219,88 @@ func getPemKeys(ctx context.Context, keysLocation string) (*Jwks, error) {
...
+			return getPublicKeyFromRSA(jwks.Keys[k])
+		}
+		if jwks.Keys[k].Kty == "EC" {
+			log.Debugf("generating PublicKey from ECDSA JWK")
```
bcmmbaga: Same here. Lets add log with context

bcmmbaga (changes requested):

management/server/jwtclaims/jwtValidator.go
HarryKodden: That is true, But the point is, You can't expect that the token is signed with a specific algorithm unless the OIDC Client is registered specifically with the instruction to use that specific algorithms for token signing. If that was not explicitly instructed during client registration, then the OIDC Provider just picks one algorithm and you can see in the header the algorithm which one was used. My 2-cents: The code is unnecessary strict to enforce a specific algorithm.

bcmmbaga: we're not hardcoding the algorithm but rather extracting it dynamically from the JWK that matches the token.

For instance, we can update the `getPublicKey`  to return both the public key and the signing method:
```go
func getPublicKey(token *jwt.Token, jwks *Jwks) (interface{}, jwt.SigningMethod, error) {}
```
This allows us to validate the token using the correct signing method in `ValidationKeyGetter`, ensuring that the process is flexible and aligns with the algorithm specified in the token header.

HarryKodden: @bcmmbaga I am afraid I do not see your point. We could off course return the SigingMethod to look at the alg value in the jwt token. But then the ValidateKeyGetter is comparing the same 'alg' value (which is off course equal...)
Are we on the same page ?

bcmmbaga: I understand your point now. In that case, let's go ahead and remove the `SigningMethod` as well as the check for the algo. Then we can merge this.

HarryKodden: Agreed, ~~Will do !~~ Done !

PR #2462, opened 2024-08-21
[misc] Add support for NETBIRD_STORE_ENGINE_POSTGRES_DSN environment variable in setup.env

arosberg:
## Describe your changes

Updated the configure script and docker-compose templates to better support deployment with postgres. Prevents the need to manually add the NETBIRD_STORE_ENGINE_POSTGRES_DSN environment variable to the docker-compose file after the configuration script is run (https://docs.netbird.io/selfhosted/postgres-store). Also provides warning to the user if the NETBIRD_STORE_CONFIG_ENGINE=postgres and NETBIRD_STORE_ENGINE_POSTGRES_DSN is not set.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/infrastructure_files/configure.sh
+++ b/infrastructure_files/configure.sh
@@ -41,6 +41,18 @@ if [[ "x-$NETBIRD_DOMAIN" == "x-" ]]; then
   exit 1
 fi
 
+# Check if PostgreSQL is set as the store engine
+if [[ "$NETBIRD_STORE_CONFIG_ENGINE" == "postgres" ]]; then
+  # Exit if 'NETBIRD_STORE_ENGINE_POSTGRES_DSN' is not set
+  if [[ -z "$NETBIRD_STORE_ENGINE_POSTGRES_DSN" ]]; then
+    echo "Warning: NETBIRD_STORE_CONFIG_ENGINE=postgres but NETBIRD_STORE_ENGINE_POSTGRES_DSN is not set."
+    echo "Please add the following line to your setup.env file:"
+    echo 'NETBIRD_STORE_ENGINE_POSTGRES_DSN="host=<PG_HOST> user=<PG_USER> password=<PG_PASSWORD> dbname=<PG_DB_NAME> port=<PG_PORT>"'
+    exit 1
+  fi
+  export NETBIRD_STORE_ENGINE_POSTGRES_DSN
+fi
+
 # local development or tests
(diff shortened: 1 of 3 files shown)


mlsmaycon:
Hello @arosberg, thanks for the contribution. Can you add tests to validate the environment variable setting?

You can check these files for that:
https://github.com/netbirdio/netbird/blob/c61cb00f40a2b94c27932312535bfd8498f4757a/infrastructure_files/tests/setup.env
https://github.com/netbirdio/netbird/blob/059fc7c3a2182756ac8830353841b2e3747e22f8/.github/workflows/test-infrastructure-files.yml#L19

arosberg:
Hello @mlsmaycon 

The way I wrote the script the environment variable will only be exported if NETBIRD_STORE_CONFIG_ENGINE=postgres and in the test cases NETBIRD_STORE_CONFIG_ENGINE=sqlite so it will not be exported by the script (And therefore not substituted in the compose template). One thing I could do is _always_ export the NETBIRD_STORE_ENGINE_POSTGRES_DSN variable by adding it to the base.setup.env exports
https://github.com/netbirdio/netbird/blob/059fc7c3a2182756ac8830353841b2e3747e22f8/infrastructure_files/base.setup.env#L74-L78

Then it would substitute even if NETBIRD_STORE_CONFIG_ENGINE=sqlite (Then I would be able to add test cases). Another option is adding a new test just for postgres but this seems a bit heavy to me. What do you think?

mlsmaycon:
> Hello @mlsmaycon
> 
> The way I wrote the script the environment variable will only be exported if NETBIRD_STORE_CONFIG_ENGINE=postgres and in the test cases NETBIRD_STORE_CONFIG_ENGINE=sqlite so it will not be exported by the script (And therefore not substituted in the compose template). One thing I could do is _always_ export the NETBIRD_STORE_ENGINE_POSTGRES_DSN variable by adding it to the base.setup.env exports
> 
> https://github.com/netbirdio/netbird/blob/059fc7c3a2182756ac8830353841b2e3747e22f8/infrastructure_files/base.setup.env#L74-L78
> 
> Then it would substitute even if NETBIRD_STORE_CONFIG_ENGINE=sqlite (Then I would be able to add test cases). Another option is adding a new test just for postgres but this seems a bit heavy to me. What do you think?

that's true, we will add a matrix and a service for that.

mlsmaycon (approved):

PR #2484, opened 2024-08-26
[relay] Feature/relay integration docker

pappz:
## Describe your changes

Create Dockerfile

Refactor the cmd package to the same ways like signal and management services

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/relay/Dockerfile
+++ b/relay/Dockerfile
@@ -1,4 +1,3 @@
-# sonar-ignore: docker:S6471
 FROM gcr.io/distroless/base:debug
 ENTRYPOINT [ "/go/bin/netbird-relay" ]
 COPY netbird-relay /go/bin/netbird-relay
--- a/relay/Dockerfile
+++ b/relay/Dockerfile
@@ -1,3 +1,3 @@
-FROM gcr.io/distroless/base:latest
+FROM gcr.io/distroless/base@sha256:1aae189e3baecbb4044c648d356ddb75025b2ba8d14cdc9c2a19ba784c90bfb9
 ENTRYPOINT [ "/go/bin/netbird-relay" ]
 COPY netbird-relay /go/bin/netbird-relay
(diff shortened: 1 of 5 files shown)


relay/main.go
```diff
@@ -0,0 +1,16 @@
...
+	"os"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/netbirdio/netbird/relay/cmd"
+)
+
+func main() {
+	if err := cmd.Execute(); err != nil {
+		log.Errorf("failed to execute command: %v", err)
```
lixmal: Let's  use `log.Fatalf` and drop the `os.Exit`

relay/cmd/root.go
```diff
@@ -101,16 +105,19 @@ func waitForExitSignal() {
 func execute(cmd *cobra.Command, args []string) error {
 	err := cobraConfig.Validate()
 	if err != nil {
+		log.Debugf("invalid config: %s", err)
```
lixmal, line 108: Do we need these extra logs? Is the returned error not sufficient?

relay/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM gcr.io/distroless/base:debug
```
lixmal: I suggest we also run this as non-root, e.g.

```
FROM gcr.io/distroless/base

COPY netbird-relay /go/bin/netbird-relay
USER 1000:1000
ENTRYPOINT [ "/go/bin/netbird-relay" ]
```

(alternatively use the `nonroot` tag)

We might need to change the default port from 443 to something else or document that we need to ensure cap `NET_BIND_SERVICE` is provided (nowadays default for **docker**: https://github.com/moby/moby/pull/41030)

relay/cmd/root.go
pappz: The **SilenceErrors: true** so I think it is useful

lixmal: Good point, didn't see that. Why do we need it enabled?

pappz: I would like to see logrus formatted lines only in the output. In other case, it will print out the error message itself without timestamp, line of the code, etc...

lixmal: Alright, let's fix it in the upstream Pr if possible

lixmal (approved):

PR #2497, opened 2024-08-29
[relay] Rename RelayConfig to Relay

pappz:
In the config file make no sense to use the config keyword.

Rename RelayConfig to Relay

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/config.go
+++ b/management/server/config.go
@@ -32,10 +32,10 @@ const (
 
 // Config of the Management service
 type Config struct {
-	Stuns       []*Host
-	TURNConfig  *TURNConfig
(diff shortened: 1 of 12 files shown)


management/server/config.go
```diff
@@ -34,7 +34,7 @@ const (
 type Config struct {
 	Stuns       []*Host
 	TURNConfig  *TURNConfig
-	RelayConfig *RelayConfig
+	RelayConfig *RelayConfig `json:"relay"`
```
mlsmaycon: let's rename the field and type too for clarity

Issue #1634, opened 2024-02-27
Installation of netbird fails on Fedora Atomic-based Immutable Systems


M0Rf30:
**Describe the problem**

following the installation details for rpm lead to a failing postinstall execution

The error reported is
```
Post Install of an clean install
open /etc/init.d/netbird: no such file or directory
Error: open /etc/init.d/netbird: no such file or directory
"service" failed: exec: "service": executable file not found in $PATH
Error: "service" failed: exec: "service": executable file not found in $PATH
```
as suggested by the package manager you can view this error with the command
`journalctl -t 'rpm-ostree(netbird.post)'`

**To Reproduce**

1. add the repo under repos.d folder
2. do an `rpm-ostree update` (the repo is evaluated in the right manner)
3. do an `rpm-ostree install netbird`

**Expected behavior**

**NetBird version**

`from 0.25.8 to 0.26.0`

mlsmaycon:
@M0Rf30 can you share your fedora version?

jkrogers9862:
Is there any word on this issue? It is still persistent in 2.7.3 on Fedora Silverblue 39.

jkrogers9862:
I have noted that when I tried to install qemu-kvm and Virtualization tools on Fedora Silverblue 40, this issue appears. I have also had this issue with coreos and fedora iot. This issue continues to be present in netbird 0.27.10. I have been using  use the docker version on systems where this is happening. Please take a new look at this issue.

jkrogers9862:
Is there no fix or workaround discovered for this yet? Why does netbird to install on immutable systems when virtualization tools are installed? The cause give is about as vague as it gets. The same thing happens when just installing network-scripts. It is really frustrating as I want to use native nebird rather than docker because docker seems to lose a lot of functionality and is very unstable.
![Screenshot_20240821_114421](https://github.com/user-attachments/assets/3a732293-0450-4786-be61-3a4ed41144af)
![Screenshot_20240821_114145](https://github.com/user-attachments/assets/9e68b578-2a75-4d64-a9e8-d150cd9d50cb)
![Screenshot_20240821_113837](https://github.com/user-attachments/assets/a85b9977-8f45-4592-98db-c0de8d85e8c2)

TobinHall:
I have just had a brief look at this. It seems that the netbird installer will try to invoke service install which uses the service installer from [github.com/kardianos/service](github.com/kardianos/service). 

[...]

PR #2508, opened 2024-09-01
[misc] Support rpm-ostree based distros in installation script

M0Rf30:
## Describe your changes
This add `rpm-ostree` binary detection in installation script and update `kardianos/service` module to lastest commits. It should close https://github.com/netbirdio/netbird/issues/1634

Latest commits of `kardianos/service` correctly detect and discriminate between `/etc` folders (init.d, systemd/system....)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/golang/protobuf v1.5.4
 	github.com/google/uuid v1.6.0
 	github.com/gorilla/mux v1.8.0
-	github.com/kardianos/service v1.2.1-0.20210728001519-a323c3813bc7
+	github.com/kardianos/service v1.2.3-0.20240613133416-becf2eb62b83
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.23.0
 	github.com/pion/ice/v3 v3.0.2
@@ -205,7 +205,7 @@ require (
 	k8s.io/apimachinery v0.26.2 // indirect
 )
 
-replace github.com/kardianos/service => github.com/netbirdio/service v0.0.0-20230215170314-b923b89432b0
+replace github.com/kardianos/service => github.com/netbirdio/service v0.0.0-20240904111318-17777758453a
 
 replace github.com/getlantern/systray => github.com/netbirdio/systray v0.0.0-20231030152038-ef1ed2a27949
 
--- a/go.sum
+++ b/go.sum
@@ -475,8 +475,8 @@ github.com/netbirdio/ice/v3 v3.0.0-20240315174635-e72a50fcb64e h1:PURA50S8u4mF6R
 github.com/netbirdio/ice/v3 v3.0.0-20240315174635-e72a50fcb64e/go.mod h1:YMLU7qbKfVjmEv7EoZPIVEI+kNYxWCdPK3VS0BU+U4Q=
 github.com/netbirdio/management-integrations/integrations v0.0.0-20240703085513-32605f7ffd8e h1:LYxhAmiEzSldLELHSMVoUnRPq3ztTNQImrD27frrGsI=
 github.com/netbirdio/management-integrations/integrations v0.0.0-20240703085513-32605f7ffd8e/go.mod h1:nykwWZnxb+sJz2Z//CEq45CMRWSHllH8pODKRB8eY7Y=
-github.com/netbirdio/service v0.0.0-20230215170314-b923b89432b0 h1:hirFRfx3grVA/9eEyjME5/z3nxdJlN9kfQpvWWPk32g=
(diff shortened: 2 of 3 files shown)


jkrogers9862:
I am not sure I understand how this will fix [#1634](https://github.com/netbirdio/netbird/issues/1634). 
Does rpm-ostree call this install script when installing? This looks like it just simple fixes allowing install from the install script which was missing until now. The problem is that an error occurs during the post-trans scripts.

M0Rf30:
> I am not sure I understand how this will fix [#1634](https://github.com/netbirdio/netbird/issues/1634). Does rpm-ostree call this install script when installing? This looks like it just simple fixes allowing install from the install script which was missing until now. The problem is that an error occurs during the post-trans scripts.

postinst fails on `netbird service install`.

`service install` fails cause of non-updated service module.

plus, I add rpm-ostree support to install script in order to leverage the rpm packages and repo instead of binary archive

M0Rf30:
> Why was it not a universal problem? On Silverblue, I could still install until I tried to install packages `libvirt` and `network-scripts`. Given the nature of the problem and the proposed solution, I would have thought it would be a 100% failure.
> 
> How can I get a rpm to locally install and test this?

need to generate an rpm with goreleaser, check the documentation

mlsmaycon:
@M0Rf30 we are using a forked version of github.com/kardianos/service; see https://github.com/netbirdio/netbird/blob/92a0092ad5262a3687d64a42edbf0938e3872039/go.mod#L208

Can you make this PR only about adding rpm-ostree, or is the module update a requirement for it? if so, we will update the fork and update the module in a separate PR first.

jkrogers9862:
> @M0Rf30 we are using a forked version of github.com/kardianos/service; see
> 
> https://github.com/netbirdio/netbird/blob/92a0092ad5262a3687d64a42edbf0938e3872039/go.mod#L208
> 
> Can you make this PR only about adding rpm-ostree, or is the module update a requirement for it? if so, we will update the fork and update the module in a separate PR first.

From my understanding the module update is what this Pull request is actually for to fix: #1634. I believe the author decided to also update the install script in addition as it was missing. The problem is that on Fedora Atomic-Based Operating Systems, netbird fails to install when certain packages are installed. On Silverblue, variants and on Fedora IoT and CoreOS, it can't be installed period because those packages are included in the base immutable-images which leaves the Docker variant as the only alternative on those systems.

It seems some updates were made to the kardiano/service which included fixes that solve the issue relating to the issue of installing.

mlsmaycon:
> > @M0Rf30 we are using a forked version of github.com/kardianos/service; see
> > 
> > https://github.com/netbirdio/netbird/blob/92a0092ad5262a3687d64a42edbf0938e3872039/go.mod#L208
> > 
> > Can you make this PR only about adding rpm-ostree, or is the module update a requirement for it? if so, we will update the fork and update the module in a separate PR first.
> 
> From my understanding the module update is what this Pull Request is actually for to fix: #1634. I believe the author decided to also update the install script in addition as it was missing. The problem is that on Fedora Atomic-Based Operating Systems, `netbird` fails to be layered when certain packages are also layered. On Silverblue variants it fails to install when `libvirt` and `network-scripts` is layered and on Fedora IoT and CoreOS, it can't be layered period because parts of those packages are included in the base immutable-images which leaves the Docker variant as the only alternative on those systems.
> 
> It seems some updates were made to the kardiano/service which included fixes that solve the issue relating to the issue of installing. So if something is to be removed from this pull it should be the update to the install script.

I see, in this case it is better for us to upgrade the fork and update the go.mod and go.sum.

I will prepare the PR in a few hours

M0Rf30:
> > > @M0Rf30 we are using a forked version of github.com/kardianos/service; see
> > > https://github.com/netbirdio/netbird/blob/92a0092ad5262a3687d64a42edbf0938e3872039/go.mod#L208
> > > 
> > > Can you make this PR only about adding rpm-ostree, or is the module update a requirement for it? if so, we will update the fork and update the module in a separate PR first.
> > 
> > 
> > From my understanding the module update is what this Pull Request is actually for to fix: #1634. I believe the author decided to also update the install script in addition as it was missing. The problem is that on Fedora Atomic-Based Operating Systems, `netbird` fails to be layered when certain packages are also layered. On Silverblue variants it fails to install when `libvirt` and `network-scripts` is layered and on Fedora IoT and CoreOS, it can't be layered period because parts of those packages are included in the base immutable-images which leaves the Docker variant as the only alternative on those systems.
> > It seems some updates were made to the kardiano/service which included fixes that solve the issue relating to the issue of installing. So if something is to be removed from this pull it should be the update to the install script.
> 
> I see, in this case it is better for us to upgrade the fork and update the go.mod and go.sum.
> 
> I will prepare the PR in a few hours

Already done here in the same time I've opened this

See
https://github.com/netbirdio/service/pull/2

bcmmbaga:
@M0Rf30  The service PR is already merged can you bump the version in go.mod?

bcmmbaga (approved):

PR #2517, opened 2024-09-02
Split relay auth config and handle multiple addresses

lixmal:
### Checklist
- [x] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/client/testdata/management.json
+++ b/client/testdata/management.json
@@ -20,6 +20,11 @@
     "Secret": "c29tZV9wYXNzd29yZA==",
     "TimeBasedCredentials": true
   },
+  "Relay": {
+    "Address": "localhost:0",
+    "CredentialsTTL": "1h",
+    "Secret": "b29tZV9wYXNzd29yZA=="
+  },
   "Signal": {
     "Proto": "http",
     "URI": "signal.wiretrustee.com:10000",
--- a/client/testdata/management.json
+++ b/client/testdata/management.json
@@ -21,7 +21,9 @@
     "TimeBasedCredentials": true
   },
   "Relay": {
-    "Address": "localhost:0",
+    "Addresses": [
+      "localhost:0"
+    ],
     "CredentialsTTL": "1h",
     "Secret": "b29tZV9wYXNzd29yZA=="
   },
--- a/relay/auth/hmac/token.go
+++ b/relay/auth/hmac/token.go
@@ -37,7 +37,7 @@ func unmarshalToken(payload []byte) (Token, error) {
 	return creds, err
 }
 
-// TimedHMAC generates token with TTL and using pre-shared secret known to TURN server
+// TimedHMAC generates token with TTL and using pre-shared secret known to the TURN/relay server
 type TimedHMAC struct {
 	secret     string
 	timeToLive time.Duration
--- a/relay/auth/hmac/token.go
+++ b/relay/auth/hmac/token.go
@@ -3,10 +3,10 @@ package hmac
 import (
 	"bytes"
 	"crypto/hmac"
-	"crypto/sha1"
 	"encoding/base64"
 	"encoding/gob"
 	"fmt"
+	"hash"
 	"strconv"
 	"time"
 
@@ -53,11 +53,11 @@ func NewTimedHMAC(secret string, timeToLive time.Duration) *TimedHMAC {
 
 // GenerateToken generates new time-based secret token - basically Payload is a unix timestamp and Signature is a HMAC
 // hash of a timestamp with a preshared TURN secret
-func (m *TimedHMAC) GenerateToken() (*Token, error) {
+func (m *TimedHMAC) GenerateToken(algo func() hash.Hash) (*Token, error) {
 	timeAuth := time.Now().Add(m.timeToLive).Unix()
 	timeStamp := strconv.FormatInt(timeAuth, 10)
 
-	checksum, err := m.generate(timeStamp)
+	checksum, err := m.generate(algo, timeStamp)
 	if err != nil {
 		return nil, err
 	}
@@ -69,8 +69,8 @@ func (m *TimedHMAC) GenerateToken() (*Token, error) {
 }
 
 // Validate checks if the token is valid
-func (m *TimedHMAC) Validate(token Token) error {
-	expectedMAC, err := m.generate(token.Payload)
+func (m *TimedHMAC) Validate(algo func() hash.Hash, token Token) error {
+	expectedMAC, err := m.generate(algo, token.Payload)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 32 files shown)


relay/client/manager.go
```diff
@@ -86,29 +88,66 @@ func (m *Manager) Serve() error {
...
-	})
-	m.startCleanupLoop()
+	// Wait for the first successful connection or all attempts to fail
+	select {
+	case client := <-successChan:
+		log.Infof("Successfully connected to relay server: %s", client.connectionURL)
+
+		m.relayClient = client
+
+		m.reconnectGuard = NewGuard(m.ctx, m.relayClient)
```
mlsmaycon, line 115: What do you think about refactoring NewGuard to use the URL list as a fallback option?

relay/auth/hmac/token.go
```diff
@@ -37,7 +37,7 @@ func unmarshalToken(payload []byte) (Token, error) {
 	return creds, err
 }
 
-// TimedHMAC generates token with TTL and using pre-shared secret known to TURN server
+// TimedHMAC generates token with TTL and using pre-shared secret known to the TURN/relay server
```
mlsmaycon:
```suggestion
// TimedHMAC generates a token with TTL and uses a pre-shared secret known to the relay server
```
With the new config, the key doesn't necessarily needs to be known by TURN

relay/auth/hmac/token_test.go
```diff
@@ -12,7 +14,7 @@ func TestGenerateCredentials(t *testing.T) {
 	timeToLive := 1 * time.Hour
 	v := NewTimedHMAC(secret, timeToLive)
 
-	creds, err := v.GenerateToken()
+	creds, err := v.GenerateToken(sha1.New)
```
mlsmaycon, line 17: shouldn't we test with `shar256.New`?

relay/auth/hmac/token_test.go
```diff
@@ -97,7 +99,7 @@ func TestInvalidPayload(t *testing.T) {
 		Signature: creds.Signature,
 	}
 
-	if err = v.Validate(*invalidPayloadCreds); err == nil {
+	if err = v.Validate(sha1.New, *invalidPayloadCreds); err == nil {
```
mlsmaycon, line 102: mismatch alg

relay/client/manager.go
```diff
@@ -86,29 +88,66 @@ func (m *Manager) Serve() error {
...
+		}(url)
 	}
 
-	m.reconnectGuard = NewGuard(m.ctx, m.relayClient)
-	m.relayClient.SetOnDisconnectListener(func() {
-		m.onServerDisconnected(m.serverURL)
-	})
-	m.startCleanupLoop()
+	// Wait for the first successful connection or all attempts to fail
+	select {
```
pappz, line 109: What is the goal of this logic?
How will look the server, and load balancer infrastructure? This parallel connection attempt approach will not be ended on the same server with two different URLs? The original idea was that every peer tries to connect to the same URL and the server-side load balancer manages the server instance assigmenet.

pappz, line 109: If we still want to add additional transport methods (WS, UDP, custom TCP based things) to support the wide range of firewalls with the fastest way of communications, then this parallel connection attempt could cause overload on the server load balancer. I.e. the MGM server distributes 2 server addresses and we will support WS, UDP, custom TCP connections then every peer will send 6 connection attempts.

relay/auth/hmac/token_test.go
lixmal: Fixed

relay/server/peer.go
```diff
@@ -162,22 +172,30 @@ func (p *Peer) healthcheck(ctx context.Context, hc *healthcheck.Sender) {
 }
 
 func (p *Peer) handleTransportMsg(msg []byte) {
-	peerID, err := messages.UnmarshalTransportID(msg)
+	version, peerID, err := messages.UnmarshalTransportID(msg)
```
pappz, line 176: Already checked the version in the read loop based on the same information. I think here enough to handle the peerID and payload.

lixmal: Fixed

relay/auth/hmac/token_test.go
lixmal: Fixed

client/testdata/management.json
```diff
@@ -20,6 +20,11 @@
     "Secret": "c29tZV9wYXNzd29yZA==",
     "TimeBasedCredentials": true
   },
+  "Relay": {
+    "Address": "localhost:0",
```
pappz: This should be a list

relay/messages/message.go
```diff
@@ -55,37 +59,53 @@ type HelloResponse struct {
...
-	case MsgTypeHealthCheck:
+	case
+		MsgTypeHello,
+		MsgTypeTransport,
+		MsgTypeClose,
+		MsgTypeHealthCheck:
 		return msgType, nil
 	default:
-		return 0, fmt.Errorf("invalid msg type, len: %d", len(msg))
+		return 0, fmt.Errorf("invalid msg type %d, len: %d", msgType, len(msg))
```
pappz: The len(msg) is invalid because this function get the buf[1:n]

relay/messages/message.go
```diff
@@ -55,37 +59,53 @@ type HelloResponse struct {
 	InstanceAddress string
 }
 
-// DetermineClientMsgType determines the message type from the first byte of the message
-func DetermineClientMsgType(msg []byte) (MsgType, error) {
+// ValidateVersion checks if the given version is supported by the protocol
+func ValidateVersion(msg []byte) (int, error) {
```
pappz, line 64: If I look right we never use the returned int.

lixmal: Correct, but might use it if we need to branch based on version

relay/messages/message.go
lixmal: fixed

client/testdata/management.json
lixmal: fixed

relay/messages/message.go
```diff
@@ -55,37 +59,53 @@ type HelloResponse struct {
...
-	case MsgTypeHealthCheck:
+	case
+		MsgTypeHelloResponse,
+		MsgTypeTransport,
+		MsgTypeClose,
+		MsgTypeHealthCheck:
 		return msgType, nil
 	default:
-		return 0, fmt.Errorf("invalid msg type (len: %d)", len(msg))
+		return 0, fmt.Errorf("invalid msg type %d, len: %d", msgType, len(msg))
```
pappz: Message len is invalid because we skip the first byte of the msg.

pappz (approved):

relay/client/manager.go
lixmal: I think we can make this improvement soon after release

PR #2518, opened 2024-09-02
Add version byte to wire protocol

lixmal:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/relay/client/client.go
+++ b/relay/client/client.go
@@ -258,14 +258,10 @@ func (c *Client) handShake() error {
 		return err
 	}
 
-	version, msgType, err := messages.DetermineMessageType(buf[:n])
+	msgType, err := messages.DetermineServerMsgType(buf[:n])
 	if err != nil {
 		log.Errorf("failed to determine message type: %s", err)
 		return err
-	}
-
-	if version != messages.CurrentProtocolVersion {
-		return fmt.Errorf("unsupported protocol version: %d", version)
 	}
 
 	if msgType != messages.MsgTypeHelloResponse {
@@ -273,7 +269,7 @@ func (c *Client) handShake() error {
 		return fmt.Errorf("unexpected message type")
 	}
 
-	_, ia, err := messages.UnmarshalHelloResponse(buf[:n])
+	ia, err := messages.UnmarshalHelloResponse(buf[:n])
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


relay/messages/message.go
```diff
@@ -55,38 +59,14 @@ type HelloResponse struct {
...
-	default:
-		return 0, fmt.Errorf("invalid msg type (len: %d)", len(msg))
+// DetermineMessageType determines the message type and version from the first two bytes of the message
+func DetermineMessageType(msg []byte) (byte, MsgType, error) {
+	if len(msg) < 2 {
+		return 0, 0, fmt.Errorf("invalid message length")
 	}
+	version := msg[0]
+	msgType := MsgType(msg[1])
+	return version, msgType, nil
```
mlsmaycon: can you share why we don't need the message type validation here?

relay/messages/message.go
```diff
@@ -17,19 +17,23 @@ const (
 	MsgTypeHealthCheck   MsgType = 4
 
 	sizeOfMsgType       = 1
+	sizeOfVersionByte   = 1
 	sizeOfMagicByte     = 4
-	headerSizeTransport = sizeOfMsgType + IDSize                   // 1 byte for msg type, IDSize for peerID
-	headerSizeHello     = sizeOfMsgType + sizeOfMagicByte + IDSize // 1 byte for msg type, 4 byte for magic header, IDSize for peerID
+	headerSizeTransport = sizeOfVersionByte + sizeOfMsgType + IDSize                   // 1 byte for version, 1 byte for msg type, IDSize for peerID
```
pappz, line 22: The version information is enough in the initial message (hello)

relay/client/client.go
```diff
@@ -258,18 +258,22 @@ func (c *Client) handShake() error {
...
 	}
 
-	msgType, err := messages.DetermineServerMsgType(buf[:n])
+	version, msgType, err := messages.DetermineMessageType(buf[:n])
 	if err != nil {
 		log.Errorf("failed to determine message type: %s", err)
 		return err
 	}
 
+	if version != messages.CurrentProtocolVersion {
```
pappz, line 269: If on the server side we update the version then it won't be backward compatible. For example, if we extend the protocol with new message types that are useful for the new client, the old clients could also work because, in the current implementation, we drop the unknown message types.

pappz: I think makes no sense to send the version info by the server. If the server knows about a client that it is not compatible then just close the communication by the server.

relay/messages/message.go
lixmal: That was a mistake, fixed

relay/messages/message.go
lixmal: Correct but we won't be able to swap the whole proto for something else in that case

relay/client/client.go
lixmal: >If on the server side we update the version then it won't be backward compatible. For example, if we extend the protocol with new message types that are useful for the new client, the old clients could also work because, in the current implementation, we drop the unknown message types.

We don't need to update the proto version if the memory layout and the new message types are still compatible

>I think makes no sense to send the version info by the server. If the server knows about a client that it is not compatible then just close the communication by the server.

That's the behavior with only one version (!= expected version -> return error/close). If you have multiple the client can choose to keep supporting the older server version. Same thing on the server side

PR #2528, opened 2024-09-03
[client] Avoid deadlock when auto connect and early exit

### LINKED ISSUE #2527: Deadlock in latest master build when starting service
**Describe the problem**

Following PR created a deadlock
https://github.com/netbirdio/netbird/commit/13e7198046a0d73a9cd91bf8e063fafb3d41885c

Specifically this [line](https://github.com/netbirdio/netbird/commit/13e7198046a0d73a9cd91bf8e063fafb3d41885c#diff-b9a524a16fa54801148c3bf6f07149721310a5d097f234e839466a751ca900d8R151) and so this never gets executed https://github.com/netbirdio/netbird/blob/main/client/cmd/service_controller.go#L62 because of of the deadlock

**To Reproduce**

Not sure because it happens at various times, and independent of the flag

**Expected behavior**
GRPC Socket to be responsive and status cmd to work

no 

Please specify whether you use NetBird Cloud or self-host NetBird's control plane.

**NetBird version**

master

**NetBird status -dA output:**

NA

I did builds where I remove the wait, and sure enough `netbird status` would be functional again. On non working instances the log file was missing 
```
started daemon server
```

mlsmaycon:
resolves #2527 
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/server/server.go
+++ b/client/server/server.go
@@ -142,11 +142,13 @@ func (s *Server) Start() error {
 		s.sessionWatcher.SetOnExpireListener(s.onSessionExpire)
 	}
 
+	if config.DisableAutoConnect {
+		return nil
+	}
+
 	runningWg := sync.WaitGroup{}
 	runningWg.Add(1)
-	if !config.DisableAutoConnect {
-		go s.connectWithRetryRuns(ctx, config, s.statusRecorder, &runningWg)
-	}
+	go s.connectWithRetryRuns(ctx, config, s.statusRecorder, &runningWg)
 
 	runningWg.Wait()
 
--- a/client/server/server.go
+++ b/client/server/server.go
@@ -146,17 +146,9 @@ func (s *Server) Start() error {
 		return nil
 	}
 
-	runningChan := make(chan error)
-	go s.connectWithRetryRuns(ctx, config, s.statusRecorder, runningChan)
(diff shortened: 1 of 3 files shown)


hurricanehrndz:
Going to test this patch out

hurricanehrndz:
Nice fix!

client/server/server.go
```diff
@@ -587,13 +585,22 @@ func (s *Server) Up(callerCtx context.Context, _ *proto.UpRequest) (*proto.UpRes
...
-	runningWg.Wait()
+	runningChan := make(chan error)
+	go s.connectWithRetryRuns(ctx, s.config, s.statusRecorder, runningChan)
 
-	return &proto.UpResponse{}, nil
+	for {
+		select {
+		case err := <-runningChan:
+			if err != nil {
+				log.Debugf("waiting for engine to become ready failed: %s", err)
```
lixmal, line 595: Shouldn't this return the error?

pascal-fischer: No we do have the silent retry. If I would return this would break. I am just gonna log here. This is just a channel for us to get the errors outside.

lixmal: Got it

bcmmbaga (approved):

PR #2531, opened 2024-09-04
Use hello msg struct

lixmal:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/relay/messages/auth/auth.go
+++ b/relay/messages/auth/auth.go
@@ -1,0 +1,51 @@
+package auth
+
+import (
+	"bytes"
+	"encoding/gob"
+	"fmt"
+)
+
+type Algorithm int
+
+const (
+	AlgoUnknown Algorithm = iota
+	AlgoHMACSHA256
+	AlgoHMACSHA512
+)
+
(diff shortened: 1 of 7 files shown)


relay/server/peer.go
```diff
@@ -175,25 +175,20 @@ func (p *Peer) handleHealthcheckEvents(ctx context.Context, hc *healthcheck.Send
 }
 
 func (p *Peer) handleTransportMsg(msg []byte) {
-	version, peerID, err := messages.UnmarshalTransportID(msg)
+	peerID, err := messages.UnmarshalTransportID(msg[messages.SizeOfProtoHeader:])
```
pappz, line 178: The SizeOf* variables should not be exported. Just pass the full msg []byte to the UnmarshalTransportID and handle inside the offsets in the Unmarshal functions

relay/messages/auth/auth.go
```diff
@@ -0,0 +1,51 @@
+package auth
```
pappz, line 1: This package should be part of the relay/auth package

PR #2541, opened 2024-09-05
[client] Avoid panic when there is no conn client

mlsmaycon:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/connect.go
+++ b/client/internal/connect.go
@@ -297,7 +297,7 @@ func (c *ConnectClient) run(
 }
 
 func (c *ConnectClient) Engine() *Engine {
-	if c.engine == nil {
+	if c == nil {
 		return nil
 	}
 	var e *Engine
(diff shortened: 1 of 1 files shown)


client/internal/connect.go
```diff
@@ -297,6 +297,9 @@ func (c *ConnectClient) run(
 }
 
 func (c *ConnectClient) Engine() *Engine {
+	if c.engine == nil {
```
lixmal: This needs to be inside the lock. But I don't see the sense of this, if `c.engine` is nil then we already return nil.

client/internal/connect.go
```diff
@@ -305,6 +308,9 @@ func (c *ConnectClient) Engine() *Engine {
 }
 
 func (c *ConnectClient) Stop() error {
+	if c.engine == nil {
```
lixmal: This needs to be inside the lock

mlsmaycon: checked the wrong obj. fixed now

client/internal/connect.go
mlsmaycon: checked the wrong obj. now its fixed

PR #2546, opened 2024-09-06
Fix lock on down

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -244,7 +244,7 @@ func (e *Engine) Stop() error {
 
 	err := e.removeAllPeers()
 	if err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


client/internal/engine.go
```diff
@@ -244,7 +244,7 @@ func (e *Engine) Stop() error {
 
 	err := e.removeAllPeers()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to remove all peers: %s", err)
```
lixmal, line 247:
```suggestion
		return fmt.Errorf("remove all peers: %w", err)
```

Let's wrap the error, also "failed to" doesn't add any information, it's already an error.

PR #2568, opened 2024-09-09
Enforce permissions on Win

pappz:
## Describe your changes

Enforce folder permission on Windows, giving only administrators and system access to the NetBird folder.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


mlsmaycon (approved):

florian-obradovic:
I think it would be better to replace the permissions on every install / update.

I just checked two installations (MSI & EXE) and it won't fix already too open permissions
CleanShot 2024-09-17 at 18 28 48@2x

PR #2579, opened 2024-09-11
[management] Add accessible peers endpoint

bcmmbaga:
## Describe your changes

Refactored the accessible peers from the` /api/peers/{peerId}` response to a dedicated endpoint `/api/peers/{peerId}/accessible-peers`.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/management/server/http/api/openapi.yml
+++ b/management/server/http/api/openapi.yml
@@ -1825,7 +1825,7 @@ paths:
     get:
       summary: List accessible Peers
       description: Returns a list of peers that the specified peer can connect to within the network.
-      tags: [ Accessible Peers ]
+      tags: [ Peers ]
       security:
         - BearerAuth: [ ]
         - TokenAuth: [ ]
--- a/management/server/http/api/openapi.yml
+++ b/management/server/http/api/openapi.yml
@@ -251,7 +251,7 @@ components:
         - name
         - ssh_enabled
         - login_expiration_enabled
-    PeerBase:
+    Peer:
       allOf:
         - $ref: '#/components/schemas/PeerMinimum'
         - type: object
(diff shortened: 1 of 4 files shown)


management/server/http/api/openapi.yml
```diff
@@ -1806,6 +1821,38 @@ paths:
           "$ref": "#/components/responses/forbidden"
         '500':
           "$ref": "#/components/responses/internal_error"
+  /api/peers/{peerId}/accessible-peers:
+    get:
+      summary: List of accessible peers
```
pascal-fischer:
```suggestion
      summary: List accessible Peers
```
This is more is sync with the title we have for the other endpoint. This will go into the left side navigation on the API docs

management/server/http/api/openapi.yml
```diff
@@ -1806,6 +1821,38 @@ paths:
...
+      description: Returns a list of peers that the specified peer can connect to within the network.
+      tags: [ Accessible Peers ]
```
pascal-fischer:
```suggestion
      tags: [ Peers ]
```
As we decided to keep the enpoint together with the peers tab I would like to have the same for the docs. This tag will decide wether we add it to the Peers dropdown in doc or create a new page.

management/server/http/peers_handler.go
```diff
@@ -220,32 +212,66 @@ func (h *PeersHandler) setApprovalRequiredFlag(respBody []*api.PeerBatch, approv
...
+	if len(peerID) == 0 {
+		util.WriteError(r.Context(), status.Errorf(status.InvalidArgument, "invalid peer ID"), w)
+		return
+	}
+
+	dnsDomain := h.accountManager.GetDNSDomain()
+
+	validPeers, err := h.accountManager.GetValidatedPeers(account)
+	if err != nil {
+		log.WithContext(r.Context()).Errorf("failed to list appreoved peers: %v", err)
```
pascal-fischer:
```suggestion
		log.WithContext(r.Context()).Errorf("failed to list approved peers: %v", err)
```

pascal-fischer (changes requested):

bcmmbaga:
> We have a method `toPeerListItemResponse()` which is setting `PeerBatch.AccessiblePeersCount` but is always 0. Do we need it? It gives false value and if we split to the seperate endpoint this is probably not used in the future.

This will be removed in follow up PR

pascal-fischer (approved):

PR #2596, opened 2024-09-13
Fix leaked server connections

pappz:
## Describe your changes
Fix leaked server connections

- close unused connections in the client lib
- close deprecated connection in the server lib

The Server Picker is reusable in the guard if we want in the future. So we can support the server address changes.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/relay/client/picker.go
+++ b/relay/client/picker.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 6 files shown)


relay/client/picker.go
```diff
@@ -0,0 +1,93 @@
...
+		Url:         url,
+		Err:         err,
+	}
+}
+
+func (sp *ServerPicker) processConnResults(resultChan chan connResult, successChan chan connResult) {
+	var hasSuccess bool
+	for cr := range resultChan {
+		if cr.Err != nil {
+			continue
```
lixmal, line 78: We should keep logging these with debug or trace

lixmal (approved):

PR #2607, opened 2024-09-16
[client] Fix race condition while read/write conn status in peer conn

pappz:
## Describe your changes

The ConnStatus variable has been read in internal/peer/conn.go:330 and has been written in internal/peer/conn.go:422. It was not thread safe.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/peer/conn_status.go
+++ b/client/internal/peer/conn_status.go
@@ -1,6 +1,10 @@
 package peer
 
-import log "github.com/sirupsen/logrus"
+import (
+	"sync/atomic"
+
+	log "github.com/sirupsen/logrus"
+)
 
 const (
 	// StatusConnected indicate the peer is in connected state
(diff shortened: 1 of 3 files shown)


client/internal/peer/conn_status.go
```diff
@@ -12,7 +16,34 @@ const (
 )
 
 // ConnStatus describe the status of a peer's connection
-type ConnStatus int
+type ConnStatus int32
+
+// AtomicConnStatus is a thread-safe wrapper for ConnStatus
+type AtomicConnStatus struct {
+	status atomic.Int32
```
lixmal, line 23: I think we could remove the wrapper as it's only a single field, but that's a minor thing

lixmal (approved):

client/internal/peer/conn_status.go
pappz: I did the wrapper because of the atom.Int32 stores Int32 and the ConnStatus is a custom type (because of the String method). So I tried to avoid the cast steps.

lixmal: I meant doing `type AtomicConnStatus atomic.Int32` instead of the struct. But it's all right

PR #2615, opened 2024-09-18
[management] Remove redundant get account calls in GetAccountFromToken

bcmmbaga:
### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


management/server/account.go
```diff
@@ -1613,13 +1621,21 @@ func (am *DefaultAccountManager) handleNewUserAccount(ctx context.Context, domai
...
 	// only possible with the enabled IdP manager
 	if am.idpManager == nil {
 		log.WithContext(ctx).Warnf("invites only work with enabled IdP manager")
 		return nil
 	}
 
+	unlock := am.Store.AcquireReadLockByUID(ctx, accountID)
+
+	account, err := am.Store.GetAccount(ctx, accountID)
+	unlock()
```
pascal-fischer: This Lock makes no sense

management/server/account.go
```diff
@@ -1726,10 +1742,27 @@ func (am *DefaultAccountManager) GetAccountFromPAT(ctx context.Context, token st
...
+	}
+
+	if user.AccountID != accountID || (!user.HasAdminPower() && !user.IsServiceUser) {
+		return nil, status.Errorf(status.PermissionDenied, "the user has no permission to access account data")
+	}
+
+	unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
+	defer unlock()
+
+	return am.Store.GetAccount(ctx, accountID)
```
pascal-fischer: This lock also makes no sense

management/server/account.go
```diff
@@ -1739,110 +1772,110 @@ func (am *DefaultAccountManager) GetAccountFromToken(ctx context.Context, claims
...
-							log.WithContext(ctx).Errorf("failed to save account: %v", err)
-						}
-					}
-				}
-			} else {
-				log.WithContext(ctx).Debugf("JWT claim %q is not a string array", account.Settings.JWTGroupsClaimName)
+		if err := am.Store.SaveAccount(ctx, account); err != nil {
+			log.WithContext(ctx).Errorf("failed to save account: %v", err)
+			return nil
+		}
```
pascal-fischer, line 1856: Idealy this would get wrapped in a transaction and we completly get rid of the lock and the getAccount

management/server/account.go
```diff
@@ -1888,48 +1929,53 @@ func (am *DefaultAccountManager) getAccountWithAuthorizationClaims(ctx context.C
...
-		err = am.handleExistingUserAccount(ctx, account, primaryDomain, claims)
-		if err != nil {
-			return nil, err
+		primaryDomain := domainAccountID == "" || account.Id == domainAccountID
+		if err = am.handleExistingUserAccount(ctx, account, primaryDomain, claims); err != nil {
+			return "", err
 		}
-		return account, nil
+
+		return account.Id, nil
```
pascal-fischer, line 1971: We should not do a getAccount and pass it down. We should do the updates in `handleExistingUserAccount` within a transaction inside that method

management/server/account.go
```diff
@@ -1888,48 +1929,53 @@ func (am *DefaultAccountManager) getAccountWithAuthorizationClaims(ctx context.C
...
+				return "", err
 			}
 		}
-		return am.handleNewUserAccount(ctx, domainAccount, claims)
+
+		account, err := am.handleNewUserAccount(ctx, domainAccount, claims)
+		if err != nil {
+			return "", err
+		}
+		return account.Id, nil
```
pascal-fischer, line 1987: Same. We should not get the account here and do the update with transactions inside the `handleNewUserAccount`

management/server/group.go
```diff
@@ -27,70 +27,48 @@ func (e *GroupLinkError) Error() string {
...
+	return am.Store.GetAccountGroups(ctx, accountID)
 }
 
 // GetGroupByName filters all groups in an account by name and returns the one with the most peers
 func (am *DefaultAccountManager) GetGroupByName(ctx context.Context, groupName, accountID string) (*nbgroup.Group, error) {
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
-	defer unlock()
-
-	account, err := am.Store.GetAccount(ctx, accountID)
+	groups, err := am.Store.GetAccountGroups(ctx, accountID)
```
pascal-fischer: This should be its own database call `Store.GetGroupsByName`

management/server/http/groups_handler.go
```diff
@@ -59,7 +66,7 @@ func (h *GroupsHandler) GetAllGroups(w http.ResponseWriter, r *http.Request) {
 // UpdateGroup handles update to a group identified by a given ID
 func (h *GroupsHandler) UpdateGroup(w http.ResponseWriter, r *http.Request) {
 	claims := h.claimsExtractor.FromRequestContext(r)
-	account, user, err := h.accountManager.GetAccountFromToken(r.Context(), claims)
+	accountID, userID, err := h.accountManager.GetAccountIDFromToken(r.Context(), claims)
```
pascal-fischer, line 69: The `func (h *GroupsHandler) UpdateGroup()` is not thredsafe (also old implementation). We have no resource lock but do get and save operations. This whole method should be wrapped into an `Store.UpdateGroup method` With proper DB level locks (requires `GetAccount` and `SaveAccount` to also that DB level locks)

management/server/http/groups_handler.go
```diff
@@ -160,43 +174,37 @@ func (h *GroupsHandler) CreateGroup(w http.ResponseWriter, r *http.Request) {
 		Issued: nbgroup.GroupIssuedAPI,
 	}
 
-	err = h.accountManager.SaveGroup(r.Context(), account.Id, user.Id, &group)
+	err = h.accountManager.SaveGroup(r.Context(), accountID, userID, &group)
```
pascal-fischer, line 177: The `CreateGroup` method is not threadsave either (also existing issue)

management/server/http/groups_handler.go
```diff
@@ -114,23 +122,29 @@ func (h *GroupsHandler) UpdateGroup(w http.ResponseWriter, r *http.Request) {
 		ID:                   groupID,
 		Name:                 req.Name,
 		Peers:                peers,
-		Issued:               eg.Issued,
-		IntegrationReference: eg.IntegrationReference,
+		Issued:               existingGroup.Issued,
+		IntegrationReference: existingGroup.IntegrationReference,
+	}
+
+	if err := h.accountManager.SaveGroup(r.Context(), accountID, userID, &group); err != nil {
```
pascal-fischer, line 129: This is also not thread save

management/server/sql_store.go
```diff
@@ -399,20 +400,29 @@ func (s *SqlStore) DeleteTokenID2UserIDIndex(tokenID string) error {
...
+			return "", status.Errorf(status.NotFound, "account not found: provided domain is not registered or is not private")
 		}
 		log.WithContext(ctx).Errorf("error when getting account from the store: %s", result.Error)
-		return nil, status.Errorf(status.Internal, "issue getting account from store")
+		return "", status.Errorf(status.Internal, "issue getting account from store")
 	}
 
-	// TODO:  rework to not call GetAccount
-	return s.GetAccount(ctx, account.Id)
+	return account.Id, nil
```
pascal-fischer: The point of this is to not read the whole account but only one column. Check `GetAccountIDBySetupKey` for an example of how to read the string instead of the whole object.

management/server/sql_store.go
```diff
@@ -1024,3 +1034,18 @@ func (s *SqlStore) withTx(tx *gorm.DB) Store {
...
+		Where(idQueryCondition, accountID).First(&account)
+	if result.Error != nil {
+		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+			return "", "", status.Errorf(status.NotFound, "account not found")
+		}
+		return "", "", status.Errorf(status.Internal, "failed to retrieve account fields")
+	}
+
+	return account.Domain, account.DomainCategory, nil
+}
```
pascal-fischer: Same here, only read the two strings, not the whole account object

management/server/account.go
```diff
@@ -1252,25 +1255,30 @@ func (am *DefaultAccountManager) DeleteAccount(ctx context.Context, accountID, u
...
 // domain is used to create a new account if no account is found
-func (am *DefaultAccountManager) GetAccountByUserOrAccountID(ctx context.Context, userID, accountID, domain string) (*Account, error) {
+func (am *DefaultAccountManager) GetAccountIDByUserOrAccountID(ctx context.Context, userID, accountID, domain string) (string, error) {
 	if accountID != "" {
-		return am.Store.GetAccount(ctx, accountID)
+		_, _, err := am.Store.GetAccountDomainAndCategory(ctx, accountID)
+		if err != nil {
+			return "", err
+		}
+		return accountID, nil
```
pascal-fischer: Get account ID by * or account ID already shows an issue with this whole method. We should make the code clear here.

management/server/group.go
```diff
@@ -27,70 +27,48 @@ func (e *GroupLinkError) Error() string {
 
 // GetGroup object of the peers
 func (am *DefaultAccountManager) GetGroup(ctx context.Context, accountID, groupID, userID string) (*nbgroup.Group, error) {
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
-	defer unlock()
-
-	account, err := am.Store.GetAccount(ctx, accountID)
+	groups, err := am.GetAllGroups(ctx, accountID, userID)
```
pascal-fischer: We should not get all groups just to get a single one. This should be its own DB call. We can extract the getUser and check permissions into its own method to reduce duplication

management/server/http/peers_handler.go
```diff
@@ -144,13 +144,20 @@ func (h *PeersHandler) HandlePeer(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			util.WriteError(r.Context(), err, w)
+			return
+		}
+
+		if r.Method == http.MethodGet {
+			h.getPeer(r.Context(), account, peerID, userID, w)
+		} else {
+			h.updatePeer(r.Context(), account, userID, peerID, w, r)
+		}
```
pascal-fischer, line 160: We should not need to get the whole account to update or get a single peer

pascal-fischer: also make the switch nicely split between http.MethodGet, http.MethodPut without the additional if

management/server/http/peers_handler.go
```diff
@@ -159,28 +166,23 @@ func (h *PeersHandler) HandlePeer(w http.ResponseWriter, r *http.Request) {
...
 	claims := h.claimsExtractor.FromRequestContext(r)
-	account, user, err := h.accountManager.GetAccountFromToken(r.Context(), claims)
+	accountID, userID, err := h.accountManager.GetAccountIDFromToken(r.Context(), claims)
 	if err != nil {
 		util.WriteError(r.Context(), err, w)
 		return
 	}
 
-	peers, err := h.accountManager.GetPeers(r.Context(), account.Id, user.Id)
+	account, err := h.accountManager.GetAccountByID(r.Context(), accountID, userID)
```
pascal-fischer, line 176: we should not get account here, we should only get peer and pass accountID into the `GetValidatedPeers` and do seperate db calls for the tables there (probably also in a transaction)

management/server/http/policies_handler.go
```diff
@@ -76,41 +84,35 @@ func (h *Policies) UpdatePolicy(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	policyIdx := -1
-	for i, policy := range account.Policies {
-		if policy.ID == policyID {
-			policyIdx = i
-			break
-		}
+	account, err := h.accountManager.GetAccountByID(r.Context(), accountID, userID)
```
pascal-fischer: we do not need the account here, only load the policies

management/server/policy.go
```diff
@@ -350,15 +351,17 @@ func (am *DefaultAccountManager) SavePolicy(ctx context.Context, accountID, user
 		return err
 	}
 
-	exists := am.savePolicy(account, policy)
+	if err = am.savePolicy(account, policy, isUpdate); err != nil {
```
pascal-fischer, line 340: this should not load the whole account. Instead have a transaction and update only the fields

management/server/account.go
pascal-fischer: Can go into another PR

management/server/account.go
pascal-fischer: Can go into another PR

management/server/sql_store.go
pascal-fischer: And add locking clause here

bcmmbaga: This only loads the two columns `domain` and `domain_category`  as specified by the Select statement

pascal-fischer: Yes sorry, my bad, you are right.

management/server/sql_store.go
```diff
@@ -1035,10 +1035,53 @@ func (s *SqlStore) withTx(tx *gorm.DB) Store {
...
+
+// AccountExists checks whether an account exists by the given ID.
+func (s *SqlStore) AccountExists(ctx context.Context, id string) (bool, error) {
+	var count int64
+	result := s.db.WithContext(ctx).Model(&Account{}).Where(idQueryCondition, id).Count(&count)
+	if result.Error != nil {
+		return false, result.Error
+	}
+	return count > 0, nil
+}
```
pascal-fischer: this one should also have lockStrength (basically any Store operation will need this in the future)

management/server/sql_store.go
```diff
@@ -1087,12 +1087,24 @@ func (s *SqlStore) GetAccountDomainAndCategory(ctx context.Context, lockStrength
...
+	var group nbgroup.Group
+	result := s.db.WithContext(ctx).Model(&nbgroup.Group{}).Where(accountAndIDQueryCondition, accountID, groupID).First(&group)
+	if result.Error != nil {
+		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+			return nil, status.Errorf(status.NotFound, "group not found")
+		}
+		return nil, status.Errorf(status.Internal, "failed to get group from store: %s", result.Error)
+	}
+	return &group, nil
+}
```
pascal-fischer: this one should also have `lockStrength`

management/server/sql_store.go
```diff
@@ -1118,3 +1119,48 @@ func (s *SqlStore) GetGroupByName(ctx context.Context, lockStrength LockingStren
 	}
 	return &group, nil
 }
+
+func (s *SqlStore) GetAccountPolicies(ctx context.Context, accountID string) ([]*Policy, error) {
```
pascal-fischer: with `lockStrength` please

management/server/sql_store.go
```diff
@@ -1118,3 +1119,48 @@ func (s *SqlStore) GetGroupByName(ctx context.Context, lockStrength LockingStren
...
+	if err := result.Error; err != nil {
+		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+			return nil, status.Errorf(status.NotFound, "posture checks not found")
+		}
+		return nil, status.Errorf(status.Internal, "failed to get posture checks from store: %s", result.Error)
+	}
+	return policy, nil
+}
+
+func (s *SqlStore) GetAccountPostureChecks(ctx context.Context, accountID string) ([]*posture.Checks, error) {
```
pascal-fischer: with `lockStrength` please

pascal-fischer (dismissed):

pascal-fischer (approved):

PR #2618, opened 2024-09-18
[management] Restrict accessible peers to user-owned peers for non-admins

bcmmbaga:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/http/peers_handler.go
+++ b/management/server/http/peers_handler.go
@@ -215,7 +215,7 @@ func (h *PeersHandler) setApprovalRequiredFlag(respBody []*api.PeerBatch, approv
 // GetAccessiblePeers returns a list of all peers that the specified peer can connect to within the network.
 func (h *PeersHandler) GetAccessiblePeers(w http.ResponseWriter, r *http.Request) {
 	claims := h.claimsExtractor.FromRequestContext(r)
-	account, _, err := h.accountManager.GetAccountFromToken(r.Context(), claims)
+	account, user, err := h.accountManager.GetAccountFromToken(r.Context(), claims)
 	if err != nil {
 		util.WriteError(r.Context(), err, w)
 		return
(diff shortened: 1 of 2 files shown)


management/server/http/peers_handler_test.go
```diff
@@ -251,3 +298,112 @@ func TestGetPeers(t *testing.T) {
...
+	}
+
+	peer3 := &nbpeer.Peer{
+		ID:                     "peer3",
+		Key:                    "key3",
+		IP:                     net.ParseIP("100.64.0.3"),
+		Status:                 &nbpeer.PeerStatus{Connected: true},
+		Name:                   "peer3",
+		LoginExpirationEnabled: false,
+		UserID:                 regularUser,
```
mlsmaycon, line 335: please add a service user test too

management/server/http/peers_handler.go
```diff
@@ -228,6 +228,20 @@ func (h *PeersHandler) GetAccessiblePeers(w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	if !user.HasAdminPower() && !user.IsServiceUser {
+		peer, err := h.accountManager.GetPeer(r.Context(), account.Id, peerID, user.Id)
```
mlsmaycon: We are introducing an unnecessary get account here. I suggest either optimizing the GetPeer or reusing the account in line 218, but consider PR #2615

mlsmaycon (changes requested):

management/server/http/peers_handler.go
```diff
@@ -228,6 +228,16 @@ func (h *PeersHandler) GetAccessiblePeers(w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	// If the user is regular user and does not own the peer
+	// with the given peerID return an empty list
+	if !user.HasAdminPower() && !user.IsServiceUser {
+		peer, ok := account.Peers[peerID]
+		if !ok || peer.UserID != user.Id {
```
mlsmaycon: if !ok we should return an error as it should never happen

mlsmaycon (approved):

Issue #2616, opened 2024-09-18
Netbird can't query users when using newer versions than Zitadel 2.61.0


Kidswiss:
**Describe the problem**

When updating Zitadel to 2.61.2 or anything newer, then Netbird can't query the Zitadel user endpoint anymore.

**To Reproduce**

1. Install Netbird v0.29.3
2. Install Zitdadel 2.61.1 or newer
3. Get 403 during login

**Expected behavior**

Zitadel integration should still work if it gets updated.

**Are you using NetBird Cloud?**

Selfhosted

**NetBird version**

`0.29.3`

**Additional context**

Netbird management logs
```
2024-09-18T14:31:43Z WARN [context: SYSTEM] management/server/account.go:1017: failed warming up cache due to error: unable to post https://idp.secret.ch/management/v1/users/_search, statusCode 403
```

Zitadel log entries:
```	
time="2024-09-18T14:36:52Z" level=warning msg="token verifier repo: decrypt access token" caller="/home/runner/work/zitadel/zitadel/internal/authz/repository/eventsourcing/eventstore/token_verifier.go:283" error="ID=APP-ASdgg Message=invalid token"
```

I've tried re-creating the service account secret, but the error persisted. Also, not sure if this is an issue on Zitadel's side or on Netbird. But given that Netbird is the only app I had issues with, I opened a bug here.

landmass-deftly-reptile-budget:
Can confirm the same issue with Zitadel 2.62.1 and Netbird 0.29.3
Additional logs from netbird-management container:
ERRO [requestID: 098374cd-f244-4be6-91f4-9b3e02fb292f, context: HTTP] management/server/http/util/util.go:81: got a handler error: token invalid
ERRO [context: HTTP, requestID: 098374cd-f244-4be6-91f4-9b3e02fb292f] management/server/http/middleware/auth_middleware.go:89: Error when validating JWT claims: unable to post https://bla.blabla.com/management/v1/users/_search, statusCode 403

The logs in the Zitadel container are identical like above.

bcmmbaga:
I see that Zitadel released [v2.62.1](https://github.com/zitadel/zitadel/releases/tag/v2.62.1) two days ago, but they have now marked [v2.59.3](https://github.com/zitadel/zitadel/releases/tag/v2.59.3) as the latest version. Could you try using v2.59.3 (latest) for now or  rollback to the previous version that was working for you? 

[...]

PR #2634, opened 2024-09-23
[management] improve zitadel idp error response detail by decoding errors

adasauce:
## Describe your changes

Adds a zitadelErrorResponse struct to decode errors returned from zitadel rather than just returning a status code, so more information can be learned about potential configuration issues. Helped to diagnose #2616

I have some other small changes and improvements to compatibility later, but this was a pretty clear cut and standalone improvement.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/idp/zitadel.go
+++ b/management/server/idp/zitadel.go
@@ -44,6 +44,11 @@ type ZitadelCredentials struct {
 	appMetrics   telemetry.AppMetrics
 }
 
+type zitadelErrorResponse struct {
+	Code    int    `json:"code"`
+	Message string `json:"message"`
+}
+
 // zitadelEmail specifies details of a user email.
 type zitadelEmail struct {
 	Email           string `json:"email"`
@@ -489,7 +494,10 @@ func (zm *ZitadelManager) post(ctx context.Context, resource string, body string
 			zm.appMetrics.IDPMetrics().CountRequestStatusError()
 		}
 
-		return nil, fmt.Errorf("unable to post %s, statusCode %d", reqURL, resp.StatusCode)
+		bodyBytes, _ := io.ReadAll(resp.Body)
+		zErr := zm.readZitadelError(bodyBytes)
+
+		return bodyBytes, fmt.Errorf("unable to post %s, statusCode %d, zitadel: %w", reqURL, resp.StatusCode, zErr)
 	}
 
 	return io.ReadAll(resp.Body)
@@ -561,10 +569,23 @@ func (zm *ZitadelManager) get(ctx context.Context, resource string, q url.Values
 			zm.appMetrics.IDPMetrics().CountRequestStatusError()
 		}
 
-		return nil, fmt.Errorf("unable to get %s, statusCode %d", reqURL, resp.StatusCode)
+		bodyBytes, _ := io.ReadAll(resp.Body)
+		zErr := zm.readZitadelError(bodyBytes)
+
+		return bodyBytes, fmt.Errorf("unable to get %s, statusCode %d, zitadel: %w", reqURL, resp.StatusCode, zErr)
 	}
 
 	return io.ReadAll(resp.Body)
+}
+
+func (zm *ZitadelManager) readZitadelError(errorBody []byte) error {
+	var zitadelErr zitadelErrorResponse
+	err := zm.helper.Unmarshal(errorBody, &zitadelErr)
+	if err != nil {
+		return fmt.Errorf("error unparsable body: %s", errorBody)
+	}
+
+	return fmt.Errorf("error code: %d message: %s", zitadelErr.Code, zitadelErr.Message)
 }
 
 // userData construct user data from zitadel profile.
--- a/management/server/idp/zitadel.go
+++ b/management/server/idp/zitadel.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -42,11 +43,6 @@ type ZitadelCredentials struct {
 	jwtToken     JWTToken
 	mux          sync.Mutex
 	appMetrics   telemetry.AppMetrics
-}
-
-type zitadelErrorResponse struct {
-	Code    int    `json:"code"`
-	Message string `json:"message"`
 }
 
 // zitadelEmail specifies details of a user email.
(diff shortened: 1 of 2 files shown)


management/server/idp/zitadel.go
```diff
@@ -561,12 +569,25 @@ func (zm *ZitadelManager) get(ctx context.Context, resource string, q url.Values
...
+func (zm *ZitadelManager) readZitadelError(errorBody []byte) error {
+	var zitadelErr zitadelErrorResponse
+	err := zm.helper.Unmarshal(errorBody, &zitadelErr)
+	if err != nil {
+		return fmt.Errorf("error unparsable body: %s", errorBody)
+	}
+
+	return fmt.Errorf("error code: %d message: %s", zitadelErr.Code, zitadelErr.Message)
+}
+
```
bcmmbaga: We can pass the response body `io.ReadCloser` instead of body bytes `[]byte` to simplify  and reduce the need to handle reading the body in multiple places.

```go
func (zm *ZitadelManager) readZitadelError(body io.ReadCloser) error {
	bodyBytes, err := io.ReadAll(body)
	if err != nil {
		return fmt.Errorf("failed to read response body: %w", err)
	}


	var zitadelErr zitadelErrorResponse
	err = zm.helper.Unmarshal(bodyBytes, &zitadelErr)
	if err != nil {
		return fmt.Errorf("error unparsable body: %s", string(bodyBytes))
	}

	return fmt.Errorf("error code: %d message: %s", zitadelErr.Code, zitadelErr.Message)
}
```

management/server/idp/zitadel.go
```diff
@@ -489,7 +494,10 @@ func (zm *ZitadelManager) post(ctx context.Context, resource string, body string
 			zm.appMetrics.IDPMetrics().CountRequestStatusError()
 		}
 
-		return nil, fmt.Errorf("unable to post %s, statusCode %d", reqURL, resp.StatusCode)
+		bodyBytes, _ := io.ReadAll(resp.Body)
+		zErr := zm.readZitadelError(bodyBytes)
+
+		return bodyBytes, fmt.Errorf("unable to post %s, statusCode %d, zitadel: %w", reqURL, resp.StatusCode, zErr)
```
bcmmbaga:
```suggestion
		return nil, zm.readZitadelError(resp.Body)
```

management/server/idp/zitadel.go
```diff
@@ -561,12 +569,25 @@ func (zm *ZitadelManager) get(ctx context.Context, resource string, q url.Values
 			zm.appMetrics.IDPMetrics().CountRequestStatusError()
 		}
 
-		return nil, fmt.Errorf("unable to get %s, statusCode %d", reqURL, resp.StatusCode)
+		bodyBytes, _ := io.ReadAll(resp.Body)
+		zErr := zm.readZitadelError(bodyBytes)
+
+		return bodyBytes, fmt.Errorf("unable to get %s, statusCode %d, zitadel: %w", reqURL, resp.StatusCode, zErr)
```
bcmmbaga:
```suggestion
		return nil, zm.readZitadelError(resp.Body)
```

bcmmbaga (changes requested):
Could you also include handling for the Zitadel error when requesting JWT token?

https://github.com/netbirdio/netbird/blob/765aba2c1cc3e394a04df0746f01343913837322/management/server/idp/zitadel.go#L178-L180

adasauce:
> Could you also include handling for the Zitadel error when requesting JWT token?
> 
> https://github.com/netbirdio/netbird/blob/765aba2c1cc3e394a04df0746f01343913837322/management/server/idp/zitadel.go#L178-L180

yeah, i think since the only thing the method references internally is the helper, and they're two different structs it makes more sense to pull the parser out as a standalone utility function within the package and pass the bound helper of zm/zc + the io.readcloser of resp.Body.

I have the changes laid out like that locally here, but lmk if you'd like to see it go in that direction vs. something else.

bcmmbaga:
> > Could you also include handling for the Zitadel error when requesting JWT token?
> > https://github.com/netbirdio/netbird/blob/765aba2c1cc3e394a04df0746f01343913837322/management/server/idp/zitadel.go#L178-L180
> 
> yeah, i think since the only thing the method references internally is the helper, and they're two different structs it makes more sense to pull the parser out as a standalone utility function within the package and pass the bound helper of zm/zc + the io.readcloser of resp.Body.
> 
> I have the changes laid out like that locally here, but lmk if you'd like to see it go in that direction vs. something else.

Yes, you can extract the parser as a standalone function and initialize the helper within that function, as it's an instance of `JsonParser`

adasauce:
ah, have some tests to clean up to expect a zitadel message in the output. sorry didnt check first.

bcmmbaga:
I ran the tests and found an issue when the JWT request fails. It seems they're using a different error response format

`2024-09-27T20:01:23+03:00 WARN [context: SYSTEM] management/server/account.go:1017: failed warming up cache due to error: unable to get zitadel token, statusCode 400, zitadel: error code: 0 message: `

Error format:`{"error":"invalid_request","error_description":"Errors.User.Machine.Secret.Invalid"}`

To make it generic lets return the `string(bodyBytes)` as the new error. This way, we don't need to worry about the error structure.

```go
// readZitadelError parses errors returned by the zitadel APIs from a response.
func readZitadelError(body io.ReadCloser) error {
	bodyBytes, err := io.ReadAll(body)
	if err != nil {
		return fmt.Errorf("failed to read response body: %w", err)
	}

	return errors.New(string(bodyBytes))
}
```

Also, the Zitadel tests are failing. Could you take a look, or would you like some help with that?

adasauce:
yeah, I was just going to follow up to mention that it's a different error format with `error` and `error_description` fields with an actual failure. we could parse it, or just dump it as is.

I'm on board with just stringifying the body and returning that. the only downside I think is that with the other error response types you get extra data that isn't really important to the error, like encoded type info about the message field being a string.

bcmmbaga:
> {"error":"invalid_request","error_description":"Errors.User.Machine.Secret.Invalid"}

If there's a possibility of extra fields in the error response (which I couldn't reproduce), we can extend the `zitadelErrorResponse` to include these new fields. When constructing the error message, we should ensure to omit any fields with empty values

bcmmbaga:
This looks good. Just a minor change, remove the brackets around the error message and we’ll be good to go.

bcmmbaga (approved):

PR #2646, opened 2024-09-25
[client] Refactor/iface pkg

pappz:
## Describe your changes

Refactor the flat code structure

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


client/iface/device/device.go
```diff
@@ -0,0 +1,19 @@
+//go:build !android
+// +build !android
```
lixmal: obsolete

client/iface/configurer/interface.go
```diff
@@ -0,0 +1,21 @@
...
+	"errors"
+	"net"
+	"time"
+
+	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
+)
+
+var ErrPeerNotFound = errors.New("peer not found")
+
+type WGConfigurer interface {
```
lixmal: Interfaces should be close to the consumer, not the producer. So this should go into the `device` package, maybe `device.go`.

To avoid conflicts and tight coupling `NewKernelConfigurer` and `NewUSPConfigurer` should return their type and not the interface

client/internal/wgproxy/proxy_test.go
```diff
@@ -23,23 +23,23 @@ func TestMain(m *testing.M) {
 }
 
 type mocConn struct {
-	closeChan chan struct{}
-	closed    bool
+	closChan chan struct{}
```
lixmal: typo introduced?

lixmal (approved):

PR #2657, opened 2024-09-27
[client] Limit P2P attempts and restart on specific events

lixmal:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/peer/conn_monitor.go
pappz, line 1: Does it make no sense to move this file into a separate structure? I do not see any dependency from the conn that is not exchangeable.

client/internal/peer/conn_monitor.go
```diff
@@ -0,0 +1,181 @@
...
+	conn.candidatesMu.Lock()
+	defer conn.candidatesMu.Unlock()
+
+	if len(conn.currentCandidates) != len(newCandidates) {
+		conn.currentCandidates = newCandidates
+		return true
+	}
+
+	for i, candidate := range conn.currentCandidates {
+		if candidate.String() != newCandidates[i].String() {
```
mlsmaycon: We should consider a few things:
1. relay candidates will always be different
2. srflx candidates might have different ports if behind hard NAT

Let's add skip relay candidates and handle srflx better?

lixmal: >1. relay candidates will always be different

good point. I'll remove it

>2. srflx candidates might have different ports if behind hard NAT

I was contemplating using only host candidates but thought a different srflx could open up the possibility for a P22 connection. For some clients, this will result in infinite retries as well. Do you have any suggestions on what to do with srflx?

mlsmaycon: I think that just checking for `candidate.Address()` is enough.

mlsmaycon: using `candidateTypesP2P()` might be better to avoid Relay

client/internal/peer/conn_monitor.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	defer func() {
+		if err := agent.Close(); err != nil {
+			conn.log.Warnf("Failed to close ICE agent: %v", err)
+		}
+	}()
+
+	gatherDone := make(chan struct{})
+	err = agent.OnCandidate(func(c ice.Candidate) {
+		log.Debugf("Got candidate: %v", c)
```
mlsmaycon:
```suggestion
		log.Tracef("Got candidate: %v", c)
```

client/internal/peer/conn_monitor.go
```diff
@@ -0,0 +1,181 @@
...
+	case <-ctx.Done():
+		return fmt.Errorf("wait for gathering: %w", ctx.Err())
+	case <-gatherDone:
+	}
+
+	candidates, err := agent.GetLocalCandidates()
+	if err != nil {
+		return fmt.Errorf("get local candidates: %w", err)
+	}
+	log.Debugf("Got candidates: %v", candidates)
```
mlsmaycon:
```suggestion
	log.Tracef("Got candidates: %v", candidates)
```

client/internal/peer/conn_monitor.go
lixmal: Sonar will complain :)

client/internal/peer/conn_monitor.go
lixmal: done

client/internal/peer/conn_monitor.go
pappz: for what?

lixmal: That the file is too large

pappz: I can not follow you. If you move everything from conn_monitor.go to monitor.go and in it create a 
```
type ConnMonitor struct {
}
```
Then we can get a better separated code structure that is better testable by unit tests.

lixmal: I have moved it to a separate struct. But not its own package yet because that requires more refactoring on the other components

pappz: This looks nice! I will review it again.

PR #2673, opened 2024-09-30
Specify goreleaser version and update to 2

M0Rf30:
## Describe your changes
This allows usage of goreleaser 2 version both locally and in CI


mlsmaycon (approved):
Thanks @M0Rf30

PR #2678, opened 2024-09-30
Install sh alpine fixes

simen64:
## Describe your changes
1. Made the script check for doas to work with alpine linux
2. Specified a set DNS server in the curl statement that downloads the binaries. I had issues where the curl request wouldnt resolve the URL, my theory is that since one of the redirect URL's are very long the standard DNS wont resolve it. This is what i experienced on alpine linux, and setting the google dns in the curl request fixed the issue.
3. Change "-C" to "-c" in the sort statement, the sort version in alpine linux doesnt have the "-C" function and it works just as well with "-c"

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/release_files/install.sh
+++ b/release_files/install.sh
@@ -21,6 +21,8 @@ SUDO=""
 
 if command -v sudo > /dev/null && [ "$(id -u)" -ne 0 ]; then
     SUDO="sudo"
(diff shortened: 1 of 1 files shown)


mlsmaycon (approved):
Thanks @simen64 for the contribution.

PR #2683, opened 2024-10-02
[client] Adjust relay worker log level and message

mlsmaycon:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/peer/worker_relay.go
+++ b/client/internal/peer/worker_relay.go
@@ -74,7 +74,7 @@ func (w *WorkerRelay) OnNewOffer(remoteOfferAnswer *OfferAnswer) {
 	relayedConn, err := w.relayManager.OpenConn(srv, w.config.Key)
 	if err != nil {
 		if errors.Is(err, relayClient.ErrConnAlreadyExists) {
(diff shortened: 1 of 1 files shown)


client/internal/peer/worker_relay.go
```diff
@@ -74,7 +74,7 @@ func (w *WorkerRelay) OnNewOffer(remoteOfferAnswer *OfferAnswer) {
 	relayedConn, err := w.relayManager.OpenConn(srv, w.config.Key)
 	if err != nil {
 		if errors.Is(err, relayClient.ErrConnAlreadyExists) {
-			w.log.Infof("do not need to reopen relay connection")
+			w.log.Debugf("reusing relay connection")
```
pappz: This msg is not fully correct because we ignored this offer from a relay perspective, not reusing the existing connection. So we will get an Offer because of ICE and if this offer is irrelevant for Relay (because we already opened the connection) then we will do nothing.

mlsmaycon: Updated

pappz (approved):

PR #2691, opened 2024-10-03
[relay,client] Relay/fix/wg roaming

pappz:
## Describe your changes

If a peer connection switches from Relayed to ICE P2P, the Relayed proxy still consumes the data the other peer sends. Because the proxy is operating, the WireGuard switches back to the Relayed proxy automatically, thanks to the roaming feature.

Extend the Proxy implementation with pause/resume functions. Before switching to the p2p connection, pause the WireGuard proxy operation to prevent unnecessary package sources.
Consider waiting some milliseconds after the pause to be sure the WireGuard engine already processed all UDP msg in from the pipe.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/internal/peer/conn.go
```diff
@@ -420,54 +418,53 @@ func (conn *Conn) iCEConnectionIsReady(priority ConnPriority, iceConnInfo ICECon
...
+	if iceConnInfo.RelayedOnLocal {
+		wgProxy, err = conn.newProxy(iceConnInfo.RemoteConn)
+		if err != nil {
+			conn.log.Errorf("failed to add turn net.Conn to local proxy: %v", err)
+			return
+		}
+		ep = wgProxy.EndpointAddr()
+		conn.wgProxyICE = wgProxy
+	} else {
+		ep, _ = net.ResolveUDPAddr("udp", iceConnInfo.RemoteConn.RemoteAddr().String())
```
lixmal: Why do we ignore the error here? Let's at least log something to help debugging edge cases

client/internal/wgproxy/ebpf/wrapper.go
```diff
@@ -42,3 +80,45 @@ func (e *ProxyWrapper) CloseConn() error {
...
 }
+
+func (p *ProxyWrapper) proxyToLocal(ctx context.Context) {
+	defer p.WgeBPFProxy.removeTurnConn(uint16(p.wgEndpointAddr.Port))
+
+	buf := make([]byte, 1500)
+	for {
+		n, err := p.readFromRemote(ctx, buf)
+		if err != nil {
+			return
```
lixmal, line 93: Let's log the error here as well

client/internal/wgproxy/ebpf/wrapper.go
```diff
@@ -42,3 +80,45 @@ func (e *ProxyWrapper) CloseConn() error {
...
+	}
+}
+
+func (p *ProxyWrapper) readFromRemote(ctx context.Context, buf []byte) (int, error) {
+	n, err := p.remoteConn.Read(buf)
+	if err != nil {
+		if ctx.Err() != nil {
+			return 0, ctx.Err()
+		}
+		if err != io.EOF {
```
lixmal:
```suggestion
		if !errors.Is(err, io.EOF) {
```

client/internal/wgproxy/ebpf/wrapper.go
```diff
@@ -5,28 +5,66 @@ package ebpf
...
+}
+
+func (p *ProxyWrapper) Pause() {
+	if p.remoteConn == nil {
+		return
+	}
+
+	p.pausedMu.Lock()
+	p.paused = true
+	p.pausedMu.Unlock()
```
lixmal, line 69: Let's add some trace log here to track pausing/unpausing

client/internal/wgproxy/ebpf/wrapper.go
pappz: We log it in the readFromRemote function. I moved there because I reached the limit of the code complexity.

lixmal (dismissed):

lixmal (approved):

PR #2699, opened 2024-10-06
[management] Remove admin check on getAccountByID

pascal-fischer:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -18,7 +18,13 @@ import (
 
 	"github.com/eko/gocache/v3/cache"
 	cacheStore "github.com/eko/gocache/v3/store"
+	"github.com/hashicorp/go-multierror"
 	"github.com/miekg/dns"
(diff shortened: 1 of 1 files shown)


mlsmaycon (approved):

glaeqen:
@mlsmaycon 

I'm not sure if you guys have any regression testing but while this fixed an issue where a regular user with an unrestricted dashboard was getting 403s, now a regular user can see *all* the peers on `/api/peers`  (without permissions) while /api/peers/<id>`  returns
```json
{
    "message": "user ... has no access to peer ... under account ...",
    "code": 500
}
```
after clicking on a peer. I think this should be also HTTP `403` and not `500`.

Rolling back to `0.28.3` (that somewhat worked) for now after getting somewhat burned by `0.30.0` and `0.30.2`.

PR #2705, opened 2024-10-07
[client] Improve route acl

lixmal:
## Describe your changes

- Updates nftables library to v0.2.0
- Marks traffic that was originally destined for local and applies the input rules in the forward chain if said traffic was redirected (e.g. by Docker)
- Adds nft rules to internal map only if flush was successful
- Improves error message if handle is 0 (= not found or hasn't been refreshed)
- Adds debug logging when route rules are added
- Replaces nftables userdata (rule ID) with rule hash

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/firewall/iptables/acl_linux.go
```diff
@@ -232,6 +233,23 @@ func (m *aclManager) cleanChains() error {
...
+		return fmt.Errorf("list chains: %w", err)
+	}
+	if ok {
+		for _, rule := range m.entries["PREROUTING"] {
+			err := m.iptablesClient.DeleteIfExists("mangle", "PREROUTING", rule...)
+			if err != nil {
+				log.Errorf("failed to delete rule: %v, %s", rule, err)
+			}
+		}
+		err = m.iptablesClient.ClearChain("mangle", "PREROUTING")
```
mlsmaycon: This is cleaning any rules in the chain.

client/firewall/iptables/acl_linux.go
```diff
@@ -232,6 +233,23 @@ func (m *aclManager) cleanChains() error {
 		}
 	}
 
+	ok, err = m.iptablesClient.ChainExists("mangle", "PREROUTING")
```
mlsmaycon, line 244: should it be mangle or nat?

client/firewall/iptables/acl_linux.go
```diff
@@ -292,7 +310,10 @@ func (m *aclManager) seedInitialEntries() {
 
 	m.appendToEntries("FORWARD", []string{"-i", m.wgIface.Name(), "-j", "DROP"})
 	m.appendToEntries("FORWARD", []string{"-i", m.wgIface.Name(), "-j", m.routingFwChainName})
+	m.appendToEntries("FORWARD", []string{"-m", "mark", "--mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmark), "-j", chainNameInputRules})
 	m.appendToEntries("FORWARD", append([]string{"-o", m.wgIface.Name()}, established...))
+
+	m.appendToEntries("PREROUTING", []string{"-t", "nat", "-i", m.wgIface.Name(), "-m", "addrtype", "--dst-type", "LOCAL", "-j", "MARK", "--set-mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmark)})
```
mlsmaycon: nat or mangle?

mlsmaycon (changes requested):

client/firewall/iptables/acl_linux.go
lixmal: Indeed, see the comment on top of the fn

```
// todo write less destructive cleanup mechanism
```

I've restored the previous code when we still had prerouting rules.

client/firewall/iptables/acl_linux.go
lixmal: mangle

client/firewall/iptables/acl_linux.go
lixmal: mangle, fixed

client/firewall/iptables/acl_linux.go
mlsmaycon: let's not keep an old bug

lixmal: fixed

PR #2757, opened 2024-10-18
[client] Cleanup dns and route states on startup

lixmal:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/internal/statemanager/manager.go
```diff
@@ -0,0 +1,298 @@
...
+
+func (m *Manager) setState(name string, state State) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if _, exists := m.states[name]; !exists {
+		return fmt.Errorf("state %s not registered", name)
+	}
+
+	m.states[name] = state
```
pappz, line 145: if state is nil then we should not delete it from m.state[name] instead of set to nil?

client/internal/statemanager/manager.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/hashicorp/go-multierror"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/exp/maps"
+
+	nberrors "github.com/netbirdio/netbird/client/errors"
+)
+
+// State interface defines the methods that all state types must implement
+type State interface {
+	Name() string
```
pappz, line 23: If accidentally implement multiple states with the same name then it can cause an issue. Maybe would be good to maintain the names in this pkg.

client/internal/statemanager/manager.go
lixmal: No, we still need the key present for the registration check ~~and to (over)write an empty state for `name` in the json.~~

I think we could also check `stateTypes` instead. It doesn't really make a difference

client/internal/statemanager/manager.go
lixmal: I could add a check in `RegisterState` but that  means we need to return an `error` which adds more refactoring/boilerplate. Is it worth it?

pappz: I am just thinking in exported const string what will be used by the implementation of the interface.

pappz (approved):

PR #2758, opened 2024-10-18
[client] fix controller re-connection

pappz:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/iface/wgproxy/factory_usp.go
```diff
@@ -10,6 +12,7 @@ type USPFactory struct {
 }
 
 func NewUSPFactory(iceBind *bind.ICEBind) *USPFactory {
+	log.Infof("WireGuard Proxy Factory will produce bind proxy")
```
lixmal, line 15: Should these new logs be `Debug` level?

client/iface/wgproxy/udp/proxy.go
```diff
@@ -160,18 +161,16 @@ func (p *WGUDPProxy) proxyToRemote(ctx context.Context) {
 func (p *WGUDPProxy) proxyToLocal(ctx context.Context) {
 	defer func() {
 		if err := p.close(); err != nil {
-			log.Warnf("error in proxy to local loop: %s", err)
+			if err != io.EOF {
```
lixmal: This should use `errors.Is`

client/iface/wgproxy/udp/proxy.go
```diff
@@ -160,18 +161,16 @@ func (p *WGUDPProxy) proxyToRemote(ctx context.Context) {
...
 
 	buf := make([]byte, 1500)
 	for {
-		n, err := p.remoteConn.Read(buf)
+		n, err := p.remoteConnRead(ctx, buf)
 		if err != nil {
-			if ctx.Err() != nil {
-				return
-			}
-			log.Errorf("failed to read from remote conn: %s, %s", p.remoteConn.RemoteAddr(), err)
```
lixmal, line 174: Can we leave this at least as debug?

client/iface/wgproxy/udp/proxy.go
```diff
@@ -193,3 +192,15 @@ func (p *WGUDPProxy) proxyToLocal(ctx context.Context) {
 		}
 	}
 }
+
+func (p *WGUDPProxy) remoteConnRead(ctx context.Context, buf []byte) (n int, err error) {
+	n, err = p.remoteConn.Read(buf)
+	if err != nil {
+		if ctx.Err() != nil {
+			return
```
lixmal, line 201: This should rather return `n, nil` and the logging in line 202 should be done in the upper layer. Mostly cosmetic though

client/iface/wgproxy/factory_usp.go
pappz: No, this log line is really important to know in any case. And it will appear only once time. We support 3 different implementation and good to know what is the active.

client/iface/wgproxy/udp/proxy.go
pappz: I made this function because of the Sonarqube restriction. If I handle the error in upper layer then come code complexity issue.

lixmal (approved):

lixmal (approved):

lixmal (approved):

PR #2768, opened 2024-10-22
[client] Cleanup firewall state on startup

lixmal:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/firewall/create.go
```diff
@@ -11,10 +11,11 @@ import (
 
 	firewall "github.com/netbirdio/netbird/client/firewall/manager"
 	"github.com/netbirdio/netbird/client/firewall/uspfilter"
+	"github.com/netbirdio/netbird/client/internal/statemanager"
 )
 
 // NewFirewall creates a firewall manager instance
-func NewFirewall(context context.Context, iface IFaceMapper) (firewall.Manager, error) {
+func NewFirewall(_ context.Context, iface IFaceMapper, _ *statemanager.Manager) (firewall.Manager, error) {
```
pappz: You can fully remove context.Context parameter from here and from the create_linux.go too. It is unused.

client/firewall/create_linux.go
```diff
@@ -58,6 +59,12 @@ func NewFirewall(context context.Context, iface IFaceMapper) (firewall.Manager,
 		log.Info("no firewall manager found, trying to use userspace packet filtering firewall")
 	}
 
+	if fm != nil {
+		if err := fm.Init(stateManager); err != nil {
```
pappz: What will be if the **iface.IsUserspaceBind()** scope happens an error? Should not we revert this initialization?

lixmal: Probably, but this is out of scope for the PR. I just split out the init logic in the `Create` functions so it is not run when cleaning up only.

client/firewall/create.go
lixmal: done

pappz (approved):

PR #2810, opened 2024-10-29
[client] Persist route selection

lixmal:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/firewall/iptables/rulestore_linux.go
```diff
@@ -37,6 +37,11 @@ func (s *ipList) UnmarshalJSON(data []byte) error {
 		return err
 	}
 	s.ips = temp.IPs
+
+	if temp.IPs == nil {
```
pappz, line 41: Instead of check simpler to create map by default:
```
	temp := struct {
		IPs map[string]struct{} `json:"ips"`
	}{
		IPs: make(map[string]struct{}),
	}
```

client/firewall/iptables/rulestore_linux.go
```diff
@@ -89,5 +94,10 @@ func (s *ipsetStore) UnmarshalJSON(data []byte) error {
 		return err
 	}
 	s.ipsets = temp.IPSets
+
```
pappz, line 97: Here too

client/firewall/iptables/rulestore_linux.go
lixmal: Consider this example: https://go.dev/play/p/h7pdJWXBhL3

A user could manipulate the state file json, change the field to `null` and the map would be overwritten with nil. It will panic if we try to write to it.

pappz: Good to know!

pappz (approved):

Issue #2655, opened 2024-09-26
MariaDB/MySQL Support


ismail0234:
**Is your feature request related to a problem? Please describe.**
- I would be very grateful if MySQL and MariaDB support is added.

**Additional context**
- MariaDB is a MySQL fork. So when one is added the other will work the same way. It would be very nice if adding both will not cause problems. If it is possible to add only 1 of them, I prefer MySQL. Most web servers use mysql.

mgarces:
Thank you for your interest and feedback!

Support for MySQL/MariaDB/Percona is definitely something we have on our radar, and we understand how valuable it would be for many users. Currently, we need to conduct some internal testing and evaluate the integration thoroughly before we can commit to implementing this feature.

That being said, there isn’t a specific timeline or priority for this at the moment. We appreciate your patience and encourage you to keep an eye on future updates. [Contributions](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTING.md) or suggestions from the community are always welcome as well!

ismail0234:
I added this feature myself and it works very well. https://github.com/netbirdio/netbird/pull/2837

PR #2837, opened 2024-11-04
MySQL Support

ismail0234:
## Describe your changes

MySQL support for Netbird. Still in draft stage.  I have never used the “GO” language before, so if I have any mistakes you can help me to correct them.

- https://github.com/netbirdio/netbird/issues/2655

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


mlsmaycon:
Thanks for the submission @ismail0234. Can you enable github action tests for MySQL as well? 

See the current ones here: https://github.com/netbirdio/netbird/blob/4e918e55ba0fe9cd87a3b3eccc658d1a7deeda0f/.github/workflows/golang-test-linux.yml#L13

mlsmaycon:
@bcmmbaga can you share how to add a testcontainer for MySQL?

bcmmbaga:
@ismail0234 To add a MySQL test container, you will need to add MySQL  support in the `NewTestStoreFromSQL`
https://github.com/netbirdio/netbird/blob/a9d06b883fe742c5dd03b822ba2385203e1b1682/management/server/store.go#L253 

To spin up a new MySQL test container, you can follow the same approach we used for PostgreSQL https://github.com/netbirdio/netbird/blob/a9d06b883fe742c5dd03b822ba2385203e1b1682/management/server/testutil/store.go#L17

ismail0234:
The “security/snyk” step seems to have failed but I can't see what caused it.

bcmmbaga:
> The “security/snyk” step seems to have failed but I can't see what caused it.

Go modules seem out of sync with the codebase. Could you run `go mod tidy` and push the changes?

ismail0234:
hmm, the result still seems to be the same.

ismail0234:
Is this part used for testing?  Do I need to add mysql values?

https://github.com/netbirdio/netbird/blob/a9d06b883fe742c5dd03b822ba2385203e1b1682/.github/workflows/test-infrastructure-files.yml#L23

bcmmbaga:
> It seems to give the error “remote error: upload-pack: not our ref”. Is it a temporary problem?
> 
> https://github.com/netbirdio/netbird/actions/runs/11686156210/job/32552758527?pr=2837 seems to be successful.

Seems related to  the pipeline. I will restart the tests again

bcmmbaga:
@ismail0234, could you please force-push the latest commit to trigger the workflow? The last tests are currently queued and don’t seem to be progressing

ismail0234:
I am not sure why. But there are warnings like “Table for peer.Peer does not exist, no migration needed”. It says that the table and the data in it is missing. I didn't see any error output either.

ismail0234:
There seems to be a problem causing the MySQL container not to start. Increasing the timeout does not solve the problem either.

ismail0234:
The Container is working, but it's causing problems on the MySQL side because the way datetime are used in the Netbird database architecture is incorrectly designed. I will see what I can do for this.

ismail0234:
I temporarily disabled the other tests until the MySQL test passes.

ismail0234:
Strange, in the last tests it gives a connection refused error.

mlsmaycon:
there was a panic in one of the tests. https://github.com/netbirdio/netbird/pull/2837/checks#step:9:827 We are checking what it is. 

Your current version is missing a change on one of the workflow files to avoid canceling all matrix tests. You might want update your branch with main.

ismail0234:
Performances;

[Sqlite]
github.com/netbirdio/netbird/client/cmd => 13.338s

[Postgres]
github.com/netbirdio/netbird/client/cmd => 22.657s

[MySQL]
github.com/netbirdio/netbird/client/cmd => 46.300s

The reason why the MySQL Test is slow is that the MySQL TestContainer starts extremely slowly.

ismail0234:
@bcmmbaga 

Are we sure that the code here works in postgres and sqlite tests? My guess is that the code here does not work in sqlite and postgres tests. This is because the scheduled task should run after 10 minutes. But because mysql is slow to start the testcontainer, the test time exceeds 10 minutes and this test is triggered. As an error it says that the database connection is rejected.

https://github.com/netbirdio/netbird/blob/4be826450be50f6e8d6a7693081a041b448888ca/management/server/account.go#L1274-L1279

pascal-fischer:
Hi @ismail0234, I started similar topic a while ago [https://github.com/netbirdio/netbird/compare/main...feature/add-mysql-support#diff-998d2ca8f48ecbf2d1a259f3851048dacc8f8ae1dfc13c0e69a34e74eaae817a](https://github.com/netbirdio/netbird/compare/main...feature/add-mysql-support#diff-998d2ca8f48ecbf2d1a259f3851048dacc8f8ae1dfc13c0e69a34e74eaae817a ) but didn't progress nearly as far yet.

When you check the last file `management/server/user.go` this is how I solved the date issue. 

I will have a look into the inactivity expiration topic.

ismail0234:
> Hi @ismail0234, I started similar topic a while ago [main...feature/add-mysql-support#diff-998d2ca8f48ecbf2d1a259f3851048dacc8f8ae1dfc13c0e69a34e74eaae817a](https://github.com/netbirdio/netbird/compare/main...feature/add-mysql-support#diff-998d2ca8f48ecbf2d1a259f3851048dacc8f8ae1dfc13c0e69a34e74eaae817a) but didn't progress nearly as far yet.
> 
> When you check the last file `management/server/user.go` this is how I solved the date issue.
> 
> I will have a look into the inactivity expiration topic.

@pascal-fischer 

Setting NULL by default for datetime values does not work. I tried this, it works at first, but in some queries it fails because the system throws the value 0001-01-01 instead of NULL. To solve this, there is a sql_mode setting in MySQL. I configured mysql to allow dates with 0000.

ismail0234:
I sent the first fix to solve this problem. If I didn't break anything, it seems to be fine :) https://github.com/netbirdio/netbird/pull/2865

ismail0234:
This fix fixes the failure of SQL tests that take longer than 10 minutes. https://github.com/netbirdio/netbird/pull/2866

pascal-fischer:
Can you share what tests exactly fail and which PR fixes which test? https://github.com/netbirdio/netbird/pull/2865 should be no problem, although it should not be needed. I wanna understand how this can have an impact. And https://github.com/netbirdio/netbird/pull/2866 should not have any impact at all as the default should already be false so I wanna debug why this can have an impact and why.

ismail0234:
> Can you share what tests exactly fail and which PR fixes which test? #2865 should be no problem, although it should not be needed. I wanna understand how this can have an impact. And #2866 should not have any impact at all as the default should already be false so I wanna debug why this can have an impact and why.

@pascal-fischer 

I'm not at home at the moment. Those values for #2866 are inactive by default. That's why the tests were running that code and crashing. By setting "PeerInactivityExpirationEnabled" to false, we ensure that the issue is never triggered. But since there is another bug, we apply #2865.

ismail0234:
> Can you share what tests exactly fail and which PR fixes which test? #2865 should be no problem, although it should not be needed. I wanna understand how this can have an impact. And #2866 should not have any impact at all as the default should already be false so I wanna debug why this can have an impact and why.

I shared the necessary information in the related pull request.

ismail0234:
We are ready for the next test. Let's see what errors we get this time. -_-

ismail0234:
TestContainers have been optimized. Instead of creating a new test container each time, we create 1 for mysql and postgres and cache it.

Test time for Postgres decreased from 350+ seconds to 140 seconds.
Test time for MySQL decreased from 1800+ seconds to 150 seconds.

ismail0234:
The MySQL test seems to have been successful. 🥳

https://github.com/ismail0234/netbird/actions/runs/11777556741

ismail0234:
Hi @mlsmaycon ,

How can I use this in a new server?

ismail0234:
I tested it in production environment. Mysql connection is established and working without any problem. 

But when I use 127.0.0.1 in the mysql connection string, the connection is not established. When I enter the ip address of the server, the connection can be established. I am not sure if this is a docker related issue. Mysql and netbird are running on the same server.

image

ismail0234:
Some tests failed. Is it a general problem?

ismail0234:
Everything is complete. I've tested it in a production environment and so far I haven't had any problems. The peers can connect to each other and we can use the panel features. It looks like there's nothing else I can do. @bcmmbaga

ismail0234:
Using MySQL database made the api calls very fast. When I use sqlite, api calls take 1.5-4 seconds when there are 500 peers on the network, while it takes 0.2 seconds when we use mysql.

ismail0234:
@mlsmaycon @bcmmbaga @pascal-fischer Can you check the Pull Request to merge?

ismail0234:
If you want to merge PR, I reopened it. But if you don't want to merge, you can close or delete PR. @mlsmaycon

bcmmbaga:
@ismail0234, could you please run the following commands and push the changes?
```sh
 go get -u github.com/containerd/containerd && go mod tidy 
```

bcmmbaga:
> The vulnerability appears to be in the go-jose module. The vulnerability has been fixed on version v4.0.1. Even if containerd is updated, the issue still seems to persist.

You can revert the changes since they broke the gRPC. However, the Snyk issue seems to be a false positive, as we don't import the go-jose module at all. You can confirm with:
```sh
go mod why -m github.com/go-jose/go-jose
```

I'll take a closer look into this as well

ismail0234:
> > The vulnerability appears to be in the go-jose module. The vulnerability has been fixed on version v4.0.1. Even if containerd is updated, the issue still seems to persist.
> 
> You can revert the changes since they broke the gRPC. However, the Snyk issue seems to be a false positive, as we don't import the go-jose module at all. You can confirm with:
> 
> ```shell
> go mod why -m github.com/go-jose/go-jose
> ```
> 
> I'll take a closer look into this as well

@bcmmbaga yes, it says the module is not in use.

PR #2861, opened 2024-11-07
[management] Refactor setup key to use store methods

bcmmbaga:
## Describe your changes

Remove calls to get and save whole account in setup key operations.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/management/server/setupkey.go
+++ b/management/server/setupkey.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/sha256"
 	b64 "encoding/base64"
-	"fmt"
 	"hash/fnv"
 	"strconv"
 	"strings"
@@ -12,9 +11,8 @@ import (
 	"unicode/utf8"
 
 	"github.com/google/uuid"
-	log "github.com/sirupsen/logrus"
-
 	"github.com/netbirdio/netbird/management/server/activity"
+	nbgroup "github.com/netbirdio/netbird/management/server/group"
 	"github.com/netbirdio/netbird/management/server/status"
 )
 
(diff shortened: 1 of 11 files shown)


management/server/setupkey.go
```diff
@@ -387,37 +398,50 @@ func (am *DefaultAccountManager) GetSetupKey(ctx context.Context, accountID, use
...
-	for _, group := range autoGroups {
-		g, ok := account.Groups[group]
-		if !ok {
-			return status.Errorf(status.NotFound, "group %s doesn't exist", group)
+func validateSetupKeyAutoGroups(ctx context.Context, transaction Store, accountID string, autoGroupIDs []string) (map[string]*nbgroup.Group, error) {
+	autoGroups := map[string]*nbgroup.Group{}
+
+	for _, groupID := range autoGroupIDs {
+		group, err := transaction.GetGroupByID(ctx, LockingStrengthShare, groupID, accountID)
+		if err != nil {
```
pascal-fischer: For this specific case, we could think about having one getAllGroups operation just to not have too many DB requests for users that add many groups. Maybe we can also introduce some kind of comment similar to a `todo:` so that we can later easily find the places in the code where we could optimize. Idk maybe something like `// db-optimize: use getAllGroups or getGroupsByIDs` to load only the required groups.

management/server/setupkey.go
```diff
@@ -268,74 +273,62 @@ func (am *DefaultAccountManager) CreateSetupKey(ctx context.Context, accountID s
...
+
+	user, err := am.Store.GetUserByUserID(ctx, LockingStrengthShare, userID)
 	if err != nil {
 		return nil, err
 	}
 
+	if user.AccountID != accountID {
+		return nil, status.NewUserNotPartOfAccountError()
+	}
+
```
pascal-fischer: Although it is handled by the API layer, for security reasons we should also add the user permission code here:
```
if user.IsRegularUser() {
 	return nil, status.NewAdminPermissionError()
 }
```

management/server/setupkey.go
```diff
@@ -229,32 +227,39 @@ func (am *DefaultAccountManager) CreateSetupKey(ctx context.Context, accountID s
...
-	account, err := am.Store.GetAccount(ctx, accountID)
+	user, err := am.Store.GetUserByUserID(ctx, LockingStrengthShare, userID)
 	if err != nil {
 		return nil, err
 	}
 
-	if err := validateSetupKeyAutoGroups(account, autoGroups); err != nil {
-		return nil, err
+	if user.AccountID != accountID {
+		return nil, status.NewUserNotPartOfAccountError()
```
pascal-fischer, line 237: Although it is handled by the API layer, for security reasons we should also add the user permission code here:
```
if user.IsRegularUser() {
 	return nil, status.NewAdminPermissionError()
 }
```

management/server/sql_store.go
```diff
@@ -568,15 +568,15 @@ func (s *SqlStore) GetAccountUsers(ctx context.Context, accountID string) ([]*Us
 	return users, nil
 }
 
-func (s *SqlStore) GetAccountGroups(ctx context.Context, accountID string) ([]*nbgroup.Group, error) {
+func (s *SqlStore) GetAccountGroups(ctx context.Context, lockStrength LockingStrength, accountID string) ([]*nbgroup.Group, error) {
 	var groups []*nbgroup.Group
-	result := s.db.Find(&groups, accountIDCondition, accountID)
+	result := s.db.WithContext(ctx).Clauses(clause.Locking{Strength: string(lockStrength)}).Find(&groups, accountIDCondition, accountID)
```
pascal-fischer: due to recent issues we need to remove the WithContext()

management/server/sql_store.go
```diff
@@ -1220,12 +1220,59 @@ func (s *SqlStore) GetRouteByID(ctx context.Context, lockStrength LockingStrengt
 
 // GetAccountSetupKeys retrieves setup keys for an account.
 func (s *SqlStore) GetAccountSetupKeys(ctx context.Context, lockStrength LockingStrength, accountID string) ([]*SetupKey, error) {
-	return getRecords[*SetupKey](s.db, lockStrength, accountID)
+	var setupKeys []*SetupKey
+	result := s.db.WithContext(ctx).Clauses(clause.Locking{Strength: string(lockStrength)}).
```
pascal-fischer: remove WithContext() (and all other calls in this file)

management/server/setupkey.go
```diff
@@ -387,37 +383,76 @@ func (am *DefaultAccountManager) GetSetupKey(ctx context.Context, accountID, use
...
+			log.WithContext(ctx).Debugf("skipped adding group: %s GroupAddedToSetupKey activity: %v", g, err)
+			continue
+		}
+
+		meta := map[string]any{"group": group.Name, "group_id": group.ID, "setupkey": key.Name}
+		am.StoreEvent(ctx, userID, key.Id, accountID, activity.GroupAddedToSetupKey, meta)
+	}
+
+	return eventsToStore
+}
```
pascal-fischer: `eventsToStore` is always empty and the events are created immediately. This is also causing the order of the events to be not ideal.

pascal-fischer (approved):

PR #2863, opened 2024-11-08
[management] Remove context from database calls

pascal-fischer:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/sql_store.go
+++ b/management/server/sql_store.go
@@ -300,8 +300,6 @@ func (s *SqlStore) GetInstallationID() string {
 }
 
 func (s *SqlStore) SavePeer(ctx context.Context, accountID string, peer *nbpeer.Peer) error {
-	startTime := time.Now()
-
 	// To maintain data integrity, we create a copy of the peer's to prevent unintended updates to other fields.
 	peerCopy := peer.Copy()
 	peerCopy.AccountID = accountID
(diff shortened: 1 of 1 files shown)


management/server/sql_store.go
```diff
@@ -450,7 +450,7 @@ func (s *SqlStore) SaveUsers(accountID string, users map[string]*User) error {
 func (s *SqlStore) SaveUser(ctx context.Context, lockStrength LockingStrength, user *User) error {
 	startTime := time.Now()
 
-	result := s.db.WithContext(ctx).Clauses(clause.Locking{Strength: string(lockStrength)}).Save(user)
+	result := s.db.Clauses(clause.Locking{Strength: string(lockStrength)}).Save(user)
 	if result.Error != nil {
 		if errors.Is(result.Error, context.Canceled) {
 			return status.NewStoreContextCanceledError(time.Since(startTime))
```
bcmmbaga: Since we're no longer passing context into db calls, the context cancellation check is not needed anymore. GORM won’t report context cancellation without the context, so it's better to remove these checks

bcmmbaga (changes requested):
There are still some store methods checking for

management/server/sql_store.go
pascal-fischer: You are right. I will remove it

bcmmbaga (approved):

PR #2866, opened 2024-11-08
Fix Inactivity SQL Tests

ismail0234:
## Describe your changes

This fix fixes the failure of SQL tests that take longer than 10 minutes. 

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


ismail0234:
I checked again, these values are already false. When we apply issue #2865, this issue is fixed. So there is no need for this pull request. We just need to implement #2865.

PR #2867, opened 2024-11-08
[management] Refactor group to use store methods

bcmmbaga:
## Describe your changes

Remove calls to get and save whole account in group operations.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/management/server/group.go
+++ b/management/server/group.go
@@ -59,7 +59,7 @@ func (am *DefaultAccountManager) GetAllGroups(ctx context.Context, accountID, us
 		return nil, err
 	}
 
-	return am.Store.GetAccountGroups(ctx, accountID)
+	return am.Store.GetAccountGroups(ctx, LockingStrengthShare, accountID)
 }
 
 // GetGroupByName filters all groups in an account by name and returns the one with the most peers
--- a/management/server/group.go
+++ b/management/server/group.go
@@ -37,8 +37,12 @@ func (am *DefaultAccountManager) CheckGroupPermissions(ctx context.Context, acco
 		return err
 	}
 
-	if (!user.IsAdminOrServiceUser() && settings.RegularUsersViewBlocked) || user.AccountID != accountID {
-		return status.Errorf(status.PermissionDenied, "groups are blocked for users")
+	if user.AccountID != accountID {
+		return status.NewUserNotPartOfAccountError()
+	}
+
+	if user.IsRegularUser() && settings.RegularUsersViewBlocked {
+		return status.NewAdminPermissionError()
 	}
 
 	return nil
@@ -49,8 +53,7 @@ func (am *DefaultAccountManager) GetGroup(ctx context.Context, accountID, groupI
 	if err := am.CheckGroupPermissions(ctx, accountID, userID); err != nil {
 		return nil, err
 	}
-
-	return am.Store.GetGroupByID(ctx, LockingStrengthShare, groupID, accountID)
+	return am.Store.GetGroupByID(ctx, LockingStrengthShare, accountID, groupID)
 }
 
 // GetAllGroups returns all groups in an account
@@ -58,13 +61,12 @@ func (am *DefaultAccountManager) GetAllGroups(ctx context.Context, accountID, us
 	if err := am.CheckGroupPermissions(ctx, accountID, userID); err != nil {
 		return nil, err
 	}
-
 	return am.Store.GetAccountGroups(ctx, LockingStrengthShare, accountID)
 }
 
 // GetGroupByName filters all groups in an account by name and returns the one with the most peers
 func (am *DefaultAccountManager) GetGroupByName(ctx context.Context, groupName, accountID string) (*nbgroup.Group, error) {
-	return am.Store.GetGroupByName(ctx, LockingStrengthShare, groupName, accountID)
+	return am.Store.GetGroupByName(ctx, LockingStrengthShare, accountID, groupName)
 }
 
 // SaveGroup object of the peers
@@ -78,12 +80,19 @@ func (am *DefaultAccountManager) SaveGroup(ctx context.Context, accountID, userI
 // Note: This function does not acquire the global lock.
 // It is the caller's responsibility to ensure proper locking is in place before invoking this method.
 func (am *DefaultAccountManager) SaveGroups(ctx context.Context, accountID, userID string, newGroups []*nbgroup.Group) error {
-	account, err := am.Store.GetAccount(ctx, accountID)
+	user, err := am.Store.GetUserByUserID(ctx, LockingStrengthShare, userID)
 	if err != nil {
 		return err
 	}
 
-	var eventsToStore []func()
+	if user.AccountID != accountID {
+		return status.NewUserNotPartOfAccountError()
+	}
+
+	var (
+		eventsToStore []func()
+		groupsToSave  []*nbgroup.Group
+	)
 
 	for _, newGroup := range newGroups {
 		if newGroup.ID == "" && newGroup.Issued != nbgroup.GroupIssuedAPI {
@@ -91,7 +100,7 @@ func (am *DefaultAccountManager) SaveGroups(ctx context.Context, accountID, user
 		}
 
 		if newGroup.ID == "" && newGroup.Issued == nbgroup.GroupIssuedAPI {
-			existingGroup, err := account.FindGroupByName(newGroup.Name)
+			existingGroup, err := am.Store.GetGroupByName(ctx, LockingStrengthShare, accountID, newGroup.Name)
 			if err != nil {
 				s, ok := status.FromError(err)
 				if !ok || s.ErrorType != status.NotFound {
@@ -109,15 +118,15 @@ func (am *DefaultAccountManager) SaveGroups(ctx context.Context, accountID, user
 		}
 
 		for _, peerID := range newGroup.Peers {
-			if account.Peers[peerID] == nil {
+			if _, err = am.Store.GetPeerByID(ctx, LockingStrengthShare, accountID, peerID); err != nil {
 				return status.Errorf(status.InvalidArgument, "peer with ID \"%s\" not found", peerID)
 			}
 		}
 
-		oldGroup := account.Groups[newGroup.ID]
-		account.Groups[newGroup.ID] = newGroup
+		newGroup.AccountID = accountID
+		groupsToSave = append(groupsToSave, newGroup)
 
-		events := am.prepareGroupEvents(ctx, userID, accountID, newGroup, oldGroup, account)
+		events := am.prepareGroupEvents(ctx, userID, accountID, newGroup)
 		eventsToStore = append(eventsToStore, events...)
 	}
 
@@ -126,30 +135,45 @@ func (am *DefaultAccountManager) SaveGroups(ctx context.Context, accountID, user
 		newGroupIDs = append(newGroupIDs, newGroup.ID)
 	}
 
-	account.Network.IncSerial()
-	if err = am.Store.SaveAccount(ctx, account); err != nil {
+	updateAccountPeers, err := am.areGroupChangesAffectPeers(ctx, accountID, newGroupIDs)
+	if err != nil {
 		return err
 	}
 
-	if areGroupChangesAffectPeers(account, newGroupIDs) {
-		am.updateAccountPeers(ctx, account)
+	err = am.Store.ExecuteInTransaction(ctx, func(transaction Store) error {
+		if err = transaction.IncrementNetworkSerial(ctx, LockingStrengthUpdate, accountID); err != nil {
+			return err
+		}
+
+		if err = transaction.SaveGroups(ctx, LockingStrengthUpdate, groupsToSave); err != nil {
+			return fmt.Errorf("failed to save groups: %w", err)
+		}
+		return nil
+	})
+	if err != nil {
+		return err
 	}
 
 	for _, storeEvent := range eventsToStore {
 		storeEvent()
 	}
 
+	if updateAccountPeers {
+		am.updateAccountPeers(ctx, accountID)
+	}
+
 	return nil
 }
 
 // prepareGroupEvents prepares a list of event functions to be stored.
-func (am *DefaultAccountManager) prepareGroupEvents(ctx context.Context, userID string, accountID string, newGroup, oldGroup *nbgroup.Group, account *Account) []func() {
+func (am *DefaultAccountManager) prepareGroupEvents(ctx context.Context, userID string, accountID string, newGroup *nbgroup.Group) []func() {
 	var eventsToStore []func()
 
 	addedPeers := make([]string, 0)
 	removedPeers := make([]string, 0)
 
-	if oldGroup != nil {
+	oldGroup, err := am.Store.GetGroupByID(ctx, LockingStrengthShare, accountID, newGroup.ID)
+	if err == nil && oldGroup != nil {
 		addedPeers = difference(newGroup.Peers, oldGroup.Peers)
 		removedPeers = difference(oldGroup.Peers, newGroup.Peers)
 	} else {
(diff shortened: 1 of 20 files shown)


ismail0234:
@bcmmbaga When you use ExecuteInTransaction, storeEngine always returns empty value.

bcmmbaga:
> @bcmmbaga When you use ExecuteInTransaction, storeEngine always returns empty value.

I didn’t quite understand this. Are you getting an empty value when calling `transaction.(methodName)`, or is it happening in a different case?

ismail0234:
> > @bcmmbaga When you use ExecuteInTransaction, storeEngine always returns empty value.
> 
> I didn’t quite understand this. Are you getting an empty value when calling `transaction.(methodName)`, or is it happening in a different case?

For example, if you call “s.storeEngine” in this method, it returns an empty value. Because this method is called from a transaction.

https://github.com/netbirdio/netbird/blob/669904cd060186d0ea760057699aa3f4076ac13b/management/server/sql_store.go#L983-L993

ismail0234:
This can cause many problems in the future. When you try to access this value in the methods called by the transaction, it will return empty.

bcmmbaga:
> This can cause many problems in the future. When you try to access this value in the methods called by the transaction, it will return empty.

You're right. However, the Store methods should be generic across all databases and shouldn't include unique implementations for each supported database. Therefore, we shouldn't reference the specific engine in the Store implementation. There's currently a TODO in `GetGroupByName` that will be removed in this PR

ismail0234:
> > This can cause many problems in the future. When you try to access this value in the methods called by the transaction, it will return empty.
> 
> You're right. However, the Store methods should be generic across all databases and shouldn't include unique implementations for each supported database. Therefore, we shouldn't reference the specific engine in the Store implementation. There's currently a TODO in `GetGroupByName` that will be removed in this PR

This is needed for MySQL Integration. Because **key** is a reserved word for MySQL. You cannot use it in queries without enclosing it in double quotes. But when you use it with double quotes, an error occurs because Sqlite and Postresql queries do not support it. In this case, we should check for Mysql and Postgres and use the correct one.

The “queryCondition” value must take the correct value according to the current engine.

```

 func (s *SqlStore) GetSetupKeyBySecret(ctx context.Context, lockStrength LockingStrength, key string) (*SetupKey, error) { 

 	var queryCondition := "key"

	if s.storeEngine == MysqlStoreEngine {
		queryCondition = "`key`";
	}

 	var setupKey SetupKey 
 	result := s.db.Clauses(clause.Locking{Strength: string(lockStrength)}). 
 		First(&setupKey, queryCondition, key) 
 	if result.Error != nil { 
 		if errors.Is(result.Error, gorm.ErrRecordNotFound) { 
 			return nil, status.Errorf(status.NotFound, "setup key not found") 
 		} 
 		return nil, status.NewSetupKeyNotFoundError(result.Error) 
 	} 
 	return &setupKey, nil 
}
```

bcmmbaga:
@ismail0234 In that case, you can pass the store engine to this initialization, allowing the transaction to be aware of the current engine in use

https://github.com/netbirdio/netbird/blob/d58cf501276ba4712406849f8e8795fe11aced05/management/server/sql_store.go#L1117-L1121

```go
func (s *SqlStore) withTx(tx *gorm.DB) Store {
	return &SqlStore{
		db:          tx,
		storeEngine: s.storeEngine,
	}
}
```

ismail0234:
> @ismail0234 In that case, you can pass the store engine to this initialization, allowing the transaction to be aware of the current engine in use
> 
> https://github.com/netbirdio/netbird/blob/d58cf501276ba4712406849f8e8795fe11aced05/management/server/sql_store.go#L1117-L1121
> 
> ```go
> func (s *SqlStore) withTx(tx *gorm.DB) Store {
> 	return &SqlStore{
> 		db:          tx,
> 		storeEngine: s.storeEngine,
> 	}
> }
> ```

Thanks. Will you add this? 

I am currently querying the database name to overcome this problem.

```
func GetKeyQueryCondition(s *SqlStore) string {

	if s.storeEngine == MysqlStoreEngine {
		return mysqlKeyQueryCondition
	}

	if s.storeEngine == "" && s.db.Name() == "mysql" {
		return mysqlKeyQueryCondition
	}

	return keyQueryCondition
}
```

bcmmbaga:
> Thanks. Will you add this?
> 
Yes.  I’ll add that.

management/server/group.go
```diff
@@ -210,40 +214,47 @@ func difference(a, b []string) []string {
 }
 
 // DeleteGroup object of the peers.
-func (am *DefaultAccountManager) DeleteGroup(ctx context.Context, accountId, userId, groupID string) error {
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountId)
```
pascal-fischer, line 210: we should not remove the log yet

management/server/group.go
```diff
@@ -210,40 +214,47 @@ func difference(a, b []string) []string {
...
-		return status.Errorf(status.InvalidArgument, "deleting group ALL is not allowed")
-	}
+	var group *nbgroup.Group
 
-	if err = validateDeleteGroup(account, group, userId); err != nil {
-		return err
-	}
-	delete(account.Groups, groupID)
+	err = am.Store.ExecuteInTransaction(ctx, func(transaction Store) error {
+		group, err = transaction.GetGroupByID(ctx, LockingStrengthShare, accountID, groupID)
```
pascal-fischer: This should be locking strength update (we are deleting it)

management/server/group.go
```diff
@@ -210,40 +214,47 @@ func difference(a, b []string) []string {
 }
 
 // DeleteGroup object of the peers.
-func (am *DefaultAccountManager) DeleteGroup(ctx context.Context, accountId, userId, groupID string) error {
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountId)
-	defer unlock()
-
-	account, err := am.Store.GetAccount(ctx, accountId)
+func (am *DefaultAccountManager) DeleteGroup(ctx context.Context, accountID, userID, groupID string) error {
```
pascal-fischer, line 213: For the create we have one save groups and reuse it for the save group. Can we do the same for delete?

management/server/group.go
```diff
@@ -254,187 +265,249 @@ func (am *DefaultAccountManager) DeleteGroup(ctx context.Context, accountId, use
...
-	}
+	var allErrors error
+	var groupIDsToDelete []string
+	var deletedGroups []*nbgroup.Group
 
-	return allErrors
-}
+	err = am.Store.ExecuteInTransaction(ctx, func(transaction Store) error {
+		for _, groupID := range groupIDs {
+			group, err := transaction.GetGroupByID(ctx, LockingStrengthShare, accountID, groupID)
```
pascal-fischer: Should be locking strength update

management/server/group.go
```diff
@@ -254,187 +265,249 @@ func (am *DefaultAccountManager) DeleteGroup(ctx context.Context, accountId, use
...
+		am.StoreEvent(ctx, userID, group.ID, accountID, activity.GroupDeleted, group.EventMeta())
 	}
 
-	return groups, nil
+	return allErrors
 }
 
 // GroupAddPeer appends peer to the group
 func (am *DefaultAccountManager) GroupAddPeer(ctx context.Context, accountID, groupID, peerID string) error {
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
```
pascal-fischer, line 309: should not remove the lock

management/server/group.go
```diff
@@ -254,187 +265,249 @@ func (am *DefaultAccountManager) DeleteGroup(ctx context.Context, accountId, use
...
-	defer unlock()
+	var group *nbgroup.Group
+	var updateAccountPeers bool
+	var err error
+
+	err = am.Store.ExecuteInTransaction(ctx, func(transaction Store) error {
+		group, err = transaction.GetGroupByID(context.Background(), LockingStrengthShare, accountID, groupID)
```
pascal-fischer: should be locking strength update

management/server/group.go
```diff
@@ -254,187 +265,249 @@ func (am *DefaultAccountManager) DeleteGroup(ctx context.Context, accountId, use
...
+	if updateAccountPeers {
+		am.updateAccountPeers(ctx, accountID)
 	}
 
 	return nil
 }
 
 // GroupDeletePeer removes peer from the group
 func (am *DefaultAccountManager) GroupDeletePeer(ctx context.Context, accountID, groupID, peerID string) error {
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
```
pascal-fischer, line 347: should not remove the lock

management/server/group.go
```diff
@@ -254,187 +265,249 @@ func (am *DefaultAccountManager) DeleteGroup(ctx context.Context, accountId, use
...
-	defer unlock()
+	var group *nbgroup.Group
+	var updateAccountPeers bool
+	var err error
+
+	err = am.Store.ExecuteInTransaction(ctx, func(transaction Store) error {
+		group, err = transaction.GetGroupByID(context.Background(), LockingStrengthShare, accountID, groupID)
```
pascal-fischer: should be locking strength update

management/server/sql_store.go
```diff
@@ -1079,10 +1079,45 @@ func (s *SqlStore) AddPeerToAccount(ctx context.Context, peer *nbpeer.Peer) erro
...
+func (s *SqlStore) GetPeerByID(ctx context.Context, lockStrength LockingStrength, accountID, peerID string) (*nbpeer.Peer, error) {
+	var peer *nbpeer.Peer
+	result := s.db.Clauses(clause.Locking{Strength: string(lockStrength)}).
+		First(&peer, accountAndIDQueryCondition, accountID, peerID)
+	if result.Error != nil {
+		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+			return nil, status.Errorf(status.NotFound, "peer not found")
+		}
+		log.WithContext(ctx).Errorf("failed to get peer from store: %s", result.Error)
+		return nil, status.Errorf(status.Internal, "failed to get peer from store")
```
pascal-fischer, line 1092: we should not log if we return the same error. The logging should happen wherever the error is handled

management/server/sql_store.go
```diff
@@ -1079,10 +1079,45 @@ func (s *SqlStore) AddPeerToAccount(ctx context.Context, peer *nbpeer.Peer) erro
...
+}
+
+// GetPeersByIDs retrieves peers by their IDs and account ID.
+func (s *SqlStore) GetPeersByIDs(ctx context.Context, lockStrength LockingStrength, accountID string, peerIDs []string) (map[string]*nbpeer.Peer, error) {
+	var peers []*nbpeer.Peer
+	result := s.db.Clauses(clause.Locking{Strength: string(lockStrength)}).Find(&peers, accountAndIDsQueryCondition, accountID, peerIDs)
 	if result.Error != nil {
-		return status.Errorf(status.Internal, "issue incrementing network serial count: %s", result.Error)
+		log.WithContext(ctx).Errorf("failed to get peers by ID's from the store: %s", result.Error)
+		return nil, status.Errorf(status.Internal, "failed to get peers by ID's from the store")
```
pascal-fischer, line 1104: we should not log if we return the same error. The logging should happen wherever the error is handled

management/server/sql_store.go
```diff
@@ -1079,10 +1079,45 @@ func (s *SqlStore) AddPeerToAccount(ctx context.Context, peer *nbpeer.Peer) erro
...
+
+	return peersMap, nil
+}
+
+func (s *SqlStore) IncrementNetworkSerial(ctx context.Context, lockStrength LockingStrength, accountId string) error {
+	result := s.db.Clauses(clause.Locking{Strength: string(lockStrength)}).
+		Model(&Account{}).Where(idQueryCondition, accountId).Update("network_serial", gorm.Expr("network_serial + 1"))
+	if result.Error != nil {
+		log.WithContext(ctx).Errorf("failed to increment network serial count in store: %v", result.Error)
+		return status.Errorf(status.Internal, "failed to increment network serial count in store")
```
pascal-fischer, line 1120: we should not log if we return the same error. The logging should happen wherever the error is handled

management/server/sql_store.go
```diff
@@ -1155,12 +1191,22 @@ func (s *SqlStore) GetAccountDomainAndCategory(ctx context.Context, lockStrength
...
-	return getRecordByID[nbgroup.Group](s.db.Preload(clause.Associations), lockStrength, groupID, accountID)
+func (s *SqlStore) GetGroupByID(ctx context.Context, lockStrength LockingStrength, accountID, groupID string) (*nbgroup.Group, error) {
+	var group *nbgroup.Group
+	result := s.db.Clauses(clause.Locking{Strength: string(lockStrength)}).First(&group, accountAndIDQueryCondition, accountID, groupID)
+	if err := result.Error; err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return nil, status.NewGroupNotFoundError(groupID)
+		}
+		log.WithContext(ctx).Errorf("failed to get group from store: %s", err)
+		return nil, status.Errorf(status.Internal, "failed to get group from store")
```
pascal-fischer, line 1202: we should not log if we return the same error. The logging should happen wherever the error is handled

management/server/sql_store.go
```diff
@@ -1172,20 +1218,21 @@ func (s *SqlStore) GetGroupByName(ctx context.Context, lockStrength LockingStren
...
-	result := query.First(&group, "name = ? and account_id = ?", groupName, accountID)
+	result := query.First(&group, "account_id = ? AND name = ?", accountID, groupName)
 	if err := result.Error; err != nil {
 		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
-			return nil, status.Errorf(status.NotFound, "group not found")
+			return nil, status.NewGroupNotFoundError(groupName)
 		}
-		return nil, status.Errorf(status.Internal, "failed to get group from store: %s", result.Error)
+		log.WithContext(ctx).Errorf("failed to get group by name from store: %v", result.Error)
+		return nil, status.Errorf(status.Internal, "failed to get group by name from store")
```
pascal-fischer, line 1227: we should not log if we return the same error. The logging should happen wherever the error is handled

management/server/sql_store.go
```diff
@@ -1203,8 +1250,37 @@ func (s *SqlStore) GetGroupsByIDs(ctx context.Context, lockStrength LockingStren
 func (s *SqlStore) SaveGroup(ctx context.Context, lockStrength LockingStrength, group *nbgroup.Group) error {
 	result := s.db.Clauses(clause.Locking{Strength: string(lockStrength)}).Save(group)
 	if result.Error != nil {
-		return status.Errorf(status.Internal, "failed to save group to store: %v", result.Error)
+		log.WithContext(ctx).Errorf("failed to save group to store: %v", result.Error)
+		return status.Errorf(status.Internal, "failed to save group to store")
```
pascal-fischer, line 1254: we should not log if we return the same error. The logging should happen wherever the error is handled

management/server/sql_store.go
```diff
@@ -1203,8 +1250,37 @@ func (s *SqlStore) GetGroupsByIDs(ctx context.Context, lockStrength LockingStren
...
+	return nil
+}
+
+// DeleteGroup deletes a group from the database.
+func (s *SqlStore) DeleteGroup(ctx context.Context, lockStrength LockingStrength, accountID, groupID string) error {
+	result := s.db.Clauses(clause.Locking{Strength: string(lockStrength)}).
+		Delete(&nbgroup.Group{}, accountAndIDQueryCondition, accountID, groupID)
+	if err := result.Error; err != nil {
+		log.WithContext(ctx).Errorf("failed to delete group from store: %s", result.Error)
+		return status.Errorf(status.Internal, "failed to delete group from store")
```
pascal-fischer, line 1265: we should not log if we return the same error. The logging should happen wherever the error is handled

management/server/sql_store.go
```diff
@@ -1203,8 +1250,37 @@ func (s *SqlStore) GetGroupsByIDs(ctx context.Context, lockStrength LockingStren
...
+	return nil
+}
+
+// DeleteGroups deletes groups from the database.
+func (s *SqlStore) DeleteGroups(ctx context.Context, strength LockingStrength, accountID string, groupIDs []string) error {
+	result := s.db.Clauses(clause.Locking{Strength: string(strength)}).
+		Delete(&nbgroup.Group{}, accountAndIDsQueryCondition, accountID, groupIDs)
+	if result.Error != nil {
+		log.WithContext(ctx).Errorf("failed to delete groups from store: %v", result.Error)
+		return status.Errorf(status.Internal, "failed to delete groups from store: %v", result.Error)
```
pascal-fischer: we should not log if we return the same error. The logging should happen wherever the error is handled

pascal-fischer (changes requested):

pascal-fischer (dismissed):

PR #2887, opened 2024-11-13
Add disable option for event activities.

ismail0234:
## Describe your changes

Allows us to add an optional disable option for event activites. It is active by default. The reason why I need this is that I use api calls very intensively. 50-100k logs can accumulate monthly. The page where the logs are displayed is currently inaccessible to me and these logs are useless.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/event.go
+++ b/management/server/event.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"fmt"
+	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ismail0234:
Add the following line to the Docker compose file and logs will be disabled.

```
NB_EVENT_ACTIVITY_LOG_ENABLED=false
```

PR #2890, opened 2024-11-14
[client] Fix state manager race conditions

lixmal:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/internal/statemanager/manager.go
```diff
@@ -179,14 +179,18 @@ func (m *Manager) PersistState(ctx context.Context) error {
 		return nil
 	}
 
+	bs, err := json.MarshalIndent(m.states, "", "    ")
+	if err != nil {
+		return fmt.Errorf("marshal states: %w", err)
+	}
+
```
pappz, line 186: Why do not unlock the m.mu? It blocks the Updated function calls in the worst case for 5 sec. After this point, the code does not touch the m.state so make no sense to protect it.

client/internal/statemanager/manager.go
```diff
@@ -179,14 +179,18 @@ func (m *Manager) PersistState(ctx context.Context) error {
...
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
 	done := make(chan error, 1)
-
 	start := time.Now()
 	go func() {
-		done <- util.WriteJsonWithRestrictedPermission(ctx, m.filePath, m.states)
+		done <- util.WriteBytesWithRestrictedPermission(ctx, m.filePath, bs)
```
pappz, line 193: What will be if this function running more then 10 sec? The ticker will start a PersistState call and will be a conflict in the file.

client/internal/statemanager/manager.go
lixmal: We still need to clean up the `dirty` map further down. If we unlock here we might clear `dirty` with new entries that haven't been yet written.

client/internal/statemanager/manager.go
lixmal: That's what the ctx check and deadline is for in this fn

pappz: But the check and move is not an atomic operation. If the code runs parallel this lines with two different ctx the outcome is unpredictable.
```
	// Check context again
	if ctx.Err() != nil {
		return ctx.Err()
	}

	if err = os.Rename(tempFileName, file); err != nil {
		return fmt.Errorf("move %s to %s: %w", tempFileName, file, err)
	}
	```

client/internal/statemanager/manager.go
pappz: ok

pappz: The log running code (5 sec) really does not cause any issue outside of the statemanager?

lixmal: It can delay updating iptables/nftables/routes/dns on network map updates. Because of routes it might also delay p2p connections

client/internal/statemanager/manager.go
pappz: Lets do it in separated PR

pappz (approved):

PR #2893, opened 2024-11-14
[management] Refactor peers to use store methods

bcmmbaga:
## Describe your changes

Remove calls to get and save whole account in peers operations.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


management/server/account.go
```diff
@@ -1244,75 +1244,68 @@ func (am *DefaultAccountManager) handleInactivityExpirationSettings(ctx context.
...
+	if nextRun, ok := am.getNextPeerExpiration(ctx, accountID); ok {
+		go am.peerLoginExpiry.Schedule(ctx, nextRun, accountID, am.peerLoginExpirationJob(ctx, accountID))
 	}
 }
 
 // peerInactivityExpirationJob marks login expired for all inactive peers and returns the minimum duration in which the next peer of the account will expire by inactivity if found
 func (am *DefaultAccountManager) peerInactivityExpirationJob(ctx context.Context, accountID string) func() (time.Duration, bool) {
 	return func() (time.Duration, bool) {
-		unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
-		defer unlock()
```
pascal-fischer: don't remove the lock just yet

management/server/ephemeral.go
```diff
@@ -126,17 +120,21 @@ func (e *EphemeralManager) OnPeerDisconnected(ctx context.Context, peer *nbpeer.
...
-	for _, a := range accounts {
-		for id, p := range a.Peers {
-			if p.Ephemeral {
-				count++
-				e.addPeer(id, a, t)
-			}
+	for _, p := range peers {
+		if p.Ephemeral {
+			count++
+			e.addPeer(p.AccountID, p.ID, t)
```
pascal-fischer: the DB call already returns ephemeral peers. No need for the if and the counter, we can go Len(peers)

management/server/http/peers_handler.go
```diff
@@ -203,12 +194,18 @@ func (h *PeersHandler) GetAllPeers(w http.ResponseWriter, r *http.Request) {
...
 			return
 		}
-		groupMinimumInfo := toGroupsInfo(groupsMap, peer.ID)
+
+		peerGroups, err := h.accountManager.GetPeerGroups(r.Context(), accountID, peer.ID)
+		if err != nil {
+			util.WriteError(r.Context(), err, w)
+			return
+		}
+		groupMinimumInfo := toGroupsInfo(peerGroups)
```
pascal-fischer: I wonder if it makes sense here to keep get getAllGroups approach. As the method is getAllPeers, we will probably also need close to all groups. So I would keep it as one call

management/server/integrated_validator.go
```diff
@@ -73,6 +75,39 @@ func (am *DefaultAccountManager) GroupValidation(ctx context.Context, accountID
...
+		}
+
+		peers, err = transaction.GetAccountPeers(ctx, LockingStrengthShare, accountID)
+		if err != nil {
+			return err
+		}
+
+		settings, err = transaction.GetAccountSettings(ctx, LockingStrengthShare, accountID)
+		return err
+	})
```
pascal-fischer: Not sure how much effect this transaction will have but should not hurt either

management/server/management_proto_test.go
```diff
@@ -246,7 +246,7 @@ func Test_SyncProtocol(t *testing.T) {
 		t.Fatal("expecting SyncResponse to have non-nil NetworkMap")
 	}
 
-	if len(networkMap.GetRemotePeers()) != 3 {
+	if len(networkMap.GetRemotePeers()) != 4 {
 		t.Fatalf("expecting SyncResponse to have NetworkMap with 3 remote peers, got %d", len(networkMap.GetRemotePeers()))
```
pascal-fischer, line 253: Why do we need to update this? what changed? If so we also need to update the comment

pascal-fischer (changes requested):

management/server/integrated_validator.go
bcmmbaga: I was waiting for https://github.com/netbirdio/netbird/pull/2957 to measure the performance, but I am already planning to continue passing the account object instead of db calls, since all the methods calling this have access to the account object

management/server/management_proto_test.go
bcmmbaga: I have added the new peer in `store_with_expired_peers.sql` for the ephemeral peers tests, and this adds the counts in the network map. I will update the comment to match the number

pascal-fischer: Ok got it, thanks

management/server/account.go
```diff
@@ -466,33 +467,31 @@ func (am *DefaultAccountManager) peerLoginExpirationJob(ctx context.Context, acc
 		unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
 		defer unlock()
 
-		account, err := am.Store.GetAccount(ctx, accountID)
+		expiredPeers, err := am.getExpiredPeers(ctx, accountID)
 		if err != nil {
-			log.WithContext(ctx).Errorf("failed getting account %s expiring peers", accountID)
-			return account.GetNextPeerExpiration()
+			return 0, false
```
pascal-fischer: I am not sure we should return `0, false` here. If we have an issue connecting to the DB temporarily we would cancel the scheduler and not run any more expirations even though there might be more peers to expire. 

I would rather go with retry every 3 seconds (defined as some default retry const) and return true so the scheduler keeps running until it is absolutely sure there are no more peers to expire

management/server/account.go
```diff
@@ -466,33 +467,31 @@ func (am *DefaultAccountManager) peerLoginExpirationJob(ctx context.Context, acc
...
 
-		log.WithContext(ctx).Debugf("discovered %d peers to expire for account %s", len(peerIDs), account.Id)
+		log.WithContext(ctx).Debugf("discovered %d peers to expire for account %s", len(peerIDs), accountID)
 
-		if err := am.expireAndUpdatePeers(ctx, account, expiredPeers); err != nil {
-			log.WithContext(ctx).Errorf("failed updating account peers while expiring peers for account %s", account.Id)
-			return account.GetNextPeerExpiration()
+		if err := am.expireAndUpdatePeers(ctx, accountID, expiredPeers); err != nil {
+			log.WithContext(ctx).Errorf("failed updating account peers while expiring peers for account %s", accountID)
+			return 0, false
```
pascal-fischer: same as above

management/server/account.go
```diff
@@ -502,34 +501,33 @@ func (am *DefaultAccountManager) peerInactivityExpirationJob(ctx context.Context
 		unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
 		defer unlock()
 
-		account, err := am.Store.GetAccount(ctx, accountID)
+		inactivePeers, err := am.getInactivePeers(ctx, accountID)
 		if err != nil {
-			log.Errorf("failed getting account %s expiring peers", accountID)
-			return account.GetNextInactivePeerExpiration()
+			log.WithContext(ctx).Errorf("failed getting inactive peers for account %s", accountID)
+			return 0, false
```
pascal-fischer: same as above

management/server/account.go
```diff
@@ -502,34 +501,33 @@ func (am *DefaultAccountManager) peerInactivityExpirationJob(ctx context.Context
...
 
-		log.Debugf("discovered %d peers to expire for account %s", len(peerIDs), account.Id)
+		log.Debugf("discovered %d peers to expire for account %s", len(peerIDs), accountID)
 
-		if err := am.expireAndUpdatePeers(ctx, account, expiredPeers); err != nil {
-			log.Errorf("failed updating account peers while expiring peers for account %s", account.Id)
-			return account.GetNextInactivePeerExpiration()
+		if err := am.expireAndUpdatePeers(ctx, accountID, inactivePeers); err != nil {
+			log.Errorf("failed updating account peers while expiring peers for account %s", accountID)
+			return 0, false
```
pascal-fischer: same as above

management/server/ephemeral.go
```diff
@@ -22,10 +21,10 @@ var (
 )
 
 type ephemeralPeer struct {
-	id       string
-	account  *types.Account
```
pascal-fischer, line 26: This one is a great catch!

management/server/http/handlers/peers/peers_handler.go
```diff
@@ -72,11 +72,17 @@ func (h *Handler) getPeer(ctx context.Context, account *types.Account, peerID, u
...
 	dnsDomain := h.accountManager.GetDNSDomain()
 
-	groupsInfo := groups.ToGroupsInfo(account.Groups, peer.ID)
+	groupsMap := map[string]*types.Group{}
+	grps, _ := h.accountManager.GetAllGroups(ctx, accountID, userID)
+	for _, group := range grps {
+		groupsMap[group.ID] = group
+	}
+
+	groupsInfo := groups.ToGroupsInfo(groupsMap, peerID)
```
pascal-fischer: Inside `groups.ToGroupsInfo` we loop through the map an have no map access. We can save all the mappings from array to map for all the methods calling ToGroupsInfo if we just pass around arrays and only convert to map when actually needed

management/server/integrated_validator.go
```diff
@@ -76,8 +76,41 @@ func (am *DefaultAccountManager) GroupValidation(ctx context.Context, accountID
...
+	for _, group := range groups {
+		groupsMap[group.ID] = group
+	}
+
+	peersMap := make(map[string]*nbpeer.Peer, len(peers))
+	for _, peer := range peers {
+		peersMap[peer.ID] = peer
+	}
+
+	return am.integratedPeerValidator.GetValidatedPeers(accountID, groupsMap, peersMap, settings.Extra)
```
pascal-fischer: The GetValidatedPeers also run only in a loop through the map. We could avoid this whole re-mapping. For all the places that still operate on account.Peers as a map we could use account.PeersG as the account object contains map and array.

Decide if we should do this as part of this PR or only leave a comment. I would like to include it just because we already touched this whole part already. If you notice this is to much change, leave it for another PR and just leave a TODO

management/server/peer.go
```diff
@@ -153,29 +172,27 @@ func (am *DefaultAccountManager) updatePeerStatusAndLocation(ctx context.Context
...
 	}
 
-	account.UpdatePeer(peer)
-
 	log.WithContext(ctx).Tracef("saving peer status for peer %s is connected: %t", peer.ID, connected)
 
-	err := am.Store.SavePeerStatus(account.Id, peer.ID, *newStatus)
+	err := transaction.SavePeerStatus(ctx, store.LockingStrengthUpdate, accountID, peer.ID, *newStatus)
 	if err != nil {
-		return false, fmt.Errorf("failed to save peer status: %w", err)
```
pascal-fischer, line 184: Why do we remove the error chain?

management/server/peer.go
```diff
@@ -186,174 +203,172 @@ func (am *DefaultAccountManager) UpdatePeer(ctx context.Context, accountID, user
...
+	peerAccountID, err := am.Store.GetAccountIDByPeerID(ctx, store.LockingStrengthShare, peerID)
 	if err != nil {
 		return err
 	}
 
-	updateAccountPeers, err := am.isPeerInActiveGroup(ctx, account, peerID)
-	if err != nil {
-		return err
+	if peerAccountID != accountID {
+		return status.NewPeerNotPartOfAccountError()
```
pascal-fischer, line 361: We should also check the user.Account
```
	user, err := am.Store.GetUserByUserID(ctx, store.LockingStrengthShare, userID)
	if err != nil {
		return nil, err
	}

	if user.AccountID != accountID {
		return nil, status.NewUserNotPartOfAccountError()
	}
```

management/server/peer.go
```diff
@@ -1069,15 +1138,27 @@ func (am *DefaultAccountManager) UpdateAccountPeers(ctx context.Context, account
...
+	peer := account.GetPeer(peerId)
+	if peer == nil {
+		log.WithContext(ctx).Tracef("peer %s  doesn't exists in account %s", peerId, accountId)
+		return
+	}
+
+	if !am.peersUpdateManager.HasChannel(peerId) {
+		log.WithContext(ctx).Tracef("peer %s doesn't have a channel, skipping network map update", peerId)
+		return
+	}
```
pascal-fischer: lets do this check on the very top before loading the account

management/server/peer.go
```diff
@@ -1086,33 +1167,239 @@ func (am *DefaultAccountManager) UpdateAccountPeer(ctx context.Context, account
...
-	for _, label := range existingLabels {
-		labelMap[label] = struct{}{}
+// getNextPeerExpiration returns the minimum duration in which the next peer of the account will expire if it was found.
+// If there is no peer that expires this function returns false and a duration of 0.
+// This function only considers peers that haven't been expired yet and that are connected.
+func (am *DefaultAccountManager) getNextPeerExpiration(ctx context.Context, accountID string) (time.Duration, bool) {
+	peersWithExpiry, err := am.Store.GetAccountPeersWithExpiration(ctx, store.LockingStrengthShare, accountID)
+	if err != nil {
+		log.WithContext(ctx).Errorf("failed to get peers with expiration: %v", err)
+		return 0, false
```
pascal-fischer: same as above

management/server/peer.go
```diff
@@ -1086,33 +1167,239 @@ func (am *DefaultAccountManager) UpdateAccountPeer(ctx context.Context, account
...
-	return labelMap
+
+	if len(peersWithExpiry) == 0 {
+		return 0, false
+	}
+
+	settings, err := am.Store.GetAccountSettings(ctx, store.LockingStrengthShare, accountID)
+	if err != nil {
+		log.WithContext(ctx).Errorf("failed to get account settings: %v", err)
+		return 0, false
```
pascal-fischer: same as above

management/server/peer.go
```diff
@@ -1086,33 +1167,239 @@ func (am *DefaultAccountManager) UpdateAccountPeer(ctx context.Context, account
...
+}
+
+// GetNextInactivePeerExpiration returns the minimum duration in which the next peer of the account will expire if it was found.
+// If there is no peer that expires this function returns false and a duration of 0.
+// This function only considers peers that haven't been expired yet and that are not connected.
+func (am *DefaultAccountManager) getNextInactivePeerExpiration(ctx context.Context, accountID string) (time.Duration, bool) {
+	peersWithInactivity, err := am.Store.GetAccountPeersWithInactivity(ctx, store.LockingStrengthShare, accountID)
+	if err != nil {
+		log.WithContext(ctx).Errorf("failed to get peers with inactivity: %v", err)
+		return 0, false
```
pascal-fischer: same as above

management/server/peer.go
```diff
@@ -1086,33 +1167,239 @@ func (am *DefaultAccountManager) UpdateAccountPeer(ctx context.Context, account
...
+	}
+
+	if len(peersWithInactivity) == 0 {
+		return 0, false
+	}
+
+	settings, err := am.Store.GetAccountSettings(ctx, store.LockingStrengthShare, accountID)
+	if err != nil {
+		log.WithContext(ctx).Errorf("failed to get account settings: %v", err)
+		return 0, false
```
pascal-fischer: same as above

management/server/http/handlers/peers/peers_handler.go
bcmmbaga: `groups.ToGroupsInfo ` is currently used by network resources, which also pass a map from the groups manager. Changing this now would require significant modifications to the network resources.

pascal-fischer (approved):

Issue #2530, opened 2024-09-04
Teltonika RutOS (OpenWRT derivative) having trouble connecting to management service


nazarewk:
# Describe the problem

This is an issue dedicated to my problems with stock software running on Teltonika RUTX50 router (runs RutOS - a derivative of older version of OpenWRT):

```
root@yelk:/tmp/log/netbird# cat client.log 
2024-09-02T20:49:48Z INFO client/cmd/service_controller.go:24: starting Netbird service
2024-09-02T20:49:48Z INFO client/cmd/service_controller.go:64: started daemon server: /var/run/netbird.sock
2024-09-02T20:49:48Z INFO client/internal/connect.go:119: starting NetBird client version 0.28.9 on linux/arm
2024-09-02T20:49:48Z ERRO util/grpc/dialer.go:38: Failed to dial: dial: dial tcp 35.186.199.111:443: connect: invalid argument
2024-09-02T20:49:49Z ERRO util/grpc/dialer.go:38: Failed to dial: dial: dial tcp 35.186.199.111:443: connect: invalid argument
2024-09-02T20:49:51Z ERRO util/grpc/dialer.go:38: Failed to dial: dial: dial tcp 35.186.199.111:443: connect: invalid argument
2024-09-02T20:49:54Z ERRO util/grpc/dialer.go:38: Failed to dial: dial: dial tcp 35.186.199.111:443: connect: invalid argument
2024-09-02T20:49:58Z ERRO util/grpc/dialer.go:38: Failed to dial: dial: dial tcp 35.186.199.111:443: connect: invalid argument
2024-09-02T20:50:05Z ERRO util/grpc/dialer.go:38: Failed to dial: dial: dial tcp 35.186.199.111:443: connect: invalid argument
2024-09-02T20:50:14Z ERRO util/grpc/dialer.go:38: Failed to dial: dial: dial tcp 35.186.199.111:443: connect: invalid argument
2024-09-02T20:50:18Z INFO util/grpc/dialer.go:75: DialContext error: context deadline exceeded
2024-09-02T20:50:18Z INFO management/client/grpc.go:56: createConnection error: context deadline exceeded
2024-09-02T20:50:18Z ERRO management/client/grpc.go:64: failed creating connection to Management Service: context deadline exceeded
2024-09-02T20:50:19Z ERRO util/grpc/dialer.go:38: Failed to dial: dial: dial tcp 35.186.199.111:443: connect: invalid argument
...
```

[...]

PR #2899, opened 2024-11-18
add NB_SKIP_SOCKET_MARK & fix crash instead of returing an error

nazarewk:
## Describe your changes

I found some changes laying around on my computer when asked about #2530 (debugging my mobile router), can as well upstream those:
- allows disabling `SO_MARK` without disabling other custom routing features
- makes dialer return an error instead of crashing the process with `log.Fatalf`

https://github.com/netbirdio/netbird/issues/2530

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/util/grpc/dialer.go
+++ b/util/grpc/dialer.go
@@ -3,6 +3,7 @@ package grpc
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
 	"net"
 	"os/user"
 	"runtime"
@@ -20,24 +21,27 @@ import (
 
 func WithCustomDialer() grpc.DialOption {
 	return grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
+		log.Infof("Custom dialer called for address: %s", addr)
 		if runtime.GOOS == "linux" {
 			currentUser, err := user.Current()
 			if err != nil {
-				log.Fatalf("failed to get current user: %v", err)
+				return nil, fmt.Errorf("failed to get current user: %w", err)
 			}
 
-			// the custom dialer requires root permissions which are not required for use cases run as non-root
+			log.Infof("Current user UID: %s", currentUser.Uid)
 			if currentUser.Uid != "0" {
+				log.Debug("Not running as root, using standard dialer")
 				dialer := &net.Dialer{}
 				return dialer.DialContext(ctx, "tcp", addr)
 			}
 		}
 
+		log.Debug("Using nbnet.NewDialer()")
 		conn, err := nbnet.NewDialer().DialContext(ctx, "tcp", addr)
 		if err != nil {
-			log.Errorf("Failed to dial: %s", err)
-			return nil, err
+			return nil, fmt.Errorf("nbnet.NewDialer().DialContext failed: %v", err)
 		}
+		log.Debug("nbnet.NewDialer().DialContext succeeded")
 		return conn, nil
 	})
 }
--- a/util/net/dialer_nonios.go
+++ b/util/net/dialer_nonios.go
@@ -67,10 +67,12 @@ func (d *Dialer) DialContext(ctx context.Context, network, address string) (net.
 		}
 	}
 
+	log.Infof("Dialing %s %s", network, address)
 	conn, err := d.Dialer.DialContext(ctx, network, address)
 	if err != nil {
-		return nil, fmt.Errorf("dial: %w", err)
+		return nil, fmt.Errorf("d.Dialer.DialContext failed: %w", err)
 	}
+	log.Debug("d.Dialer.DialContext succeeded")
 
 	// Wrap the connection in Conn to handle Close with hooks
 	return &Conn{Conn: conn, ID: connID}, nil
--- a/util/net/net_linux.go
+++ b/util/net/net_linux.go
@@ -4,8 +4,13 @@ package net
 
 import (
 	"fmt"
+	"os"
 	"syscall"
+
+	log "github.com/sirupsen/logrus"
 )
+
+const envSkipSocketMark = "NB_SKIP_SOCKET_MARK"
 
 // SetSocketMark sets the SO_MARK option on the given socket connection
 func SetSocketMark(conn syscall.Conn) error {
@@ -28,16 +33,28 @@ func SetRawSocketMark(conn syscall.RawConn) error {
 	}
 
 	if setErr != nil {
-		return fmt.Errorf("set SO_MARK: %w", setErr)
+		return fmt.Errorf("set SO_MARK failed: %w", setErr)
 	}
 
 	return nil
 }
 
 func SetSocketOpt(fd int) error {
+	log.Debugf("Attempting to set SO_MARK on fd %d", fd)
 	if CustomRoutingDisabled() {
+		log.Infof("Custom routing is disabled, skipping SO_MARK")
 		return nil
 	}
 
-	return syscall.SetsockoptInt(fd, syscall.SOL_SOCKET, syscall.SO_MARK, NetbirdFwmark)
+	// Check for the new environment variable
+	if skipSocketMark := os.Getenv(envSkipSocketMark); skipSocketMark == "true" {
+		log.Info("NB_SKIP_SOCKET_MARK is set to true, skipping SO_MARK")
+		return nil
+	}
+
+	err := syscall.SetsockoptInt(fd, syscall.SOL_SOCKET, syscall.SO_MARK, NetbirdFwmark)
+	if err == nil {
+		log.Debugf("Successfully set SO_MARK to %d", NetbirdFwmark)
+	}
+	return err
 }


util/grpc/dialer.go
```diff
@@ -20,24 +21,27 @@ import (
...
+		log.Infof("Custom dialer called for address: %s", addr)
 		if runtime.GOOS == "linux" {
 			currentUser, err := user.Current()
 			if err != nil {
-				log.Fatalf("failed to get current user: %v", err)
+				return nil, fmt.Errorf("failed to get current user: %w", err)
 			}
 
-			// the custom dialer requires root permissions which are not required for use cases run as non-root
+			log.Infof("Current user UID: %s", currentUser.Uid)
```
lixmal: Let's keep the comment instead

util/grpc/dialer.go
```diff
@@ -20,24 +21,27 @@ import (
 
 func WithCustomDialer() grpc.DialOption {
 	return grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
+		log.Infof("Custom dialer called for address: %s", addr)
 		if runtime.GOOS == "linux" {
 			currentUser, err := user.Current()
 			if err != nil {
-				log.Fatalf("failed to get current user: %v", err)
+				return nil, fmt.Errorf("failed to get current user: %w", err)
```
lixmal: We should probably return a permanent error here to avoid retries.

```suggestion
				return nil, status.Errorf(codes.FailedPrecondition, "get current user: %v", err)
```

Needs

```
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
```

util/grpc/dialer.go
```diff
@@ -20,24 +21,27 @@ import (
 
 func WithCustomDialer() grpc.DialOption {
 	return grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
+		log.Infof("Custom dialer called for address: %s", addr)
```
lixmal:
```suggestion
```

There's a log within the dialer, we can drop this one

util/grpc/dialer.go
```diff
@@ -20,24 +21,27 @@ import (
...
 		}
 
+		log.Debug("Using nbnet.NewDialer()")
 		conn, err := nbnet.NewDialer().DialContext(ctx, "tcp", addr)
 		if err != nil {
-			log.Errorf("Failed to dial: %s", err)
-			return nil, err
+			return nil, fmt.Errorf("nbnet.NewDialer().DialContext failed: %v", err)
 		}
+		log.Debug("nbnet.NewDialer().DialContext succeeded")
```
lixmal: Let's only keep the logs for the exceptional path

```suggestion
```

util/net/dialer_nonios.go
```diff
@@ -67,10 +67,12 @@ func (d *Dialer) DialContext(ctx context.Context, network, address string) (net.
 		}
 	}
 
+	log.Infof("Dialing %s %s", network, address)
```
lixmal: Let's only keep the logs for the exceptional path

```suggestion
```

util/net/dialer_nonios.go
```diff
@@ -67,10 +67,12 @@ func (d *Dialer) DialContext(ctx context.Context, network, address string) (net.
...
 	conn, err := d.Dialer.DialContext(ctx, network, address)
 	if err != nil {
-		return nil, fmt.Errorf("dial: %w", err)
+		return nil, fmt.Errorf("d.Dialer.DialContext failed: %w", err)
```
lixmal:
```suggestion
		return nil, fmt.Errorf("d.Dialer.DialContext: %w", err)
```

util/grpc/dialer.go
```diff
@@ -20,24 +21,27 @@ import (
...
 				return dialer.DialContext(ctx, "tcp", addr)
 			}
 		}
 
+		log.Debug("Using nbnet.NewDialer()")
 		conn, err := nbnet.NewDialer().DialContext(ctx, "tcp", addr)
 		if err != nil {
-			log.Errorf("Failed to dial: %s", err)
-			return nil, err
+			return nil, fmt.Errorf("nbnet.NewDialer().DialContext failed: %v", err)
```
lixmal:
```suggestion
			log.Errorf("Failed to dial: %s", err)
			return nil, fmt.Errorf("nbnet.NewDialer().DialContext: %w", err)
```

- `failed` in a returned error doesn't add any information
- we should wrap the error
- we need to keep the extra log because grpc will only emit the error to the log if the grpc debugging env is enabled or it has exhausted all retries

util/net/dialer_nonios.go
```diff
@@ -67,10 +67,12 @@ func (d *Dialer) DialContext(ctx context.Context, network, address string) (net.
...
 	}
+	log.Debug("d.Dialer.DialContext succeeded")
```
lixmal: Let's only keep the logs for the exceptional path

```suggestion
```

util/net/net_linux.go
```diff
@@ -28,16 +33,28 @@ func SetRawSocketMark(conn syscall.RawConn) error {
 	}
 
 	if setErr != nil {
-		return fmt.Errorf("set SO_MARK: %w", setErr)
+		return fmt.Errorf("set SO_MARK failed: %w", setErr)
```
lixmal:
```suggestion
		return fmt.Errorf("set SO_MARK: %w", setErr)
```

util/net/net_linux.go
```diff
@@ -28,16 +33,28 @@ func SetRawSocketMark(conn syscall.RawConn) error {
...
 	}
 
 	return nil
 }
 
 func SetSocketOpt(fd int) error {
+	log.Debugf("Attempting to set SO_MARK on fd %d", fd)
```
lixmal:
```suggestion
```

util/net/net_linux.go
```diff
@@ -28,16 +33,28 @@ func SetRawSocketMark(conn syscall.RawConn) error {
...
+	// Check for the new environment variable
+	if skipSocketMark := os.Getenv(envSkipSocketMark); skipSocketMark == "true" {
+		log.Info("NB_SKIP_SOCKET_MARK is set to true, skipping SO_MARK")
 		return nil
 	}
 
-	return syscall.SetsockoptInt(fd, syscall.SOL_SOCKET, syscall.SO_MARK, NetbirdFwmark)
+	err := syscall.SetsockoptInt(fd, syscall.SOL_SOCKET, syscall.SO_MARK, NetbirdFwmark)
+	if err == nil {
+		log.Debugf("Successfully set SO_MARK to %d", NetbirdFwmark)
```
lixmal: Don't think this is necessary, it might pollute the log too much, let's revert this part

nazarewk:
I am not sure what is it with failing tests, doesn't seem like failures are related to my changes.

util/net/net_linux.go
```diff
@@ -28,16 +33,28 @@ func SetRawSocketMark(conn syscall.RawConn) error {
...
 
 func SetSocketOpt(fd int) error {
+	log.Debugf("Attempting to set SO_MARK on fd %d", fd)
 	if CustomRoutingDisabled() {
+		log.Infof("Custom routing is disabled, skipping SO_MARK")
+		return nil
+	}
+
+	// Check for the new environment variable
+	if skipSocketMark := os.Getenv(envSkipSocketMark); skipSocketMark == "true" {
```
lixmal: I suggest we add the env check here as well https://github.com/netbirdio/netbird/blob/52ea2e84e9aa3c03fe43c5098ab50c94ff2e0818/client/internal/routemanager/systemops/systemops_linux.go#L58

 This way we will fallback to exclusion routes automatically (instead of using fwmarks), to avoid routing loops.

lixmal (changes requested):

util/net/net_linux.go
nazarewk: do you mean `NB_SKIP_SOCKET_MARK=true` should automatically imply `NB_USE_LEGACY_ROUTING=true`?

lixmal:
> I am not sure what is it with failing tests, doesn't seem like failures are related to my changes.

Can you merge main? There's been an issue with a domain for the tests

util/grpc/dialer.go
```diff
@@ -23,20 +26,22 @@ func WithCustomDialer() grpc.DialOption {
 		if runtime.GOOS == "linux" {
 			currentUser, err := user.Current()
 			if err != nil {
-				log.Fatalf("failed to get current user: %v", err)
+				return nil, status.Errorf(codes.FailedPrecondition, "failed to get current user: %w", err)
```
lixmal:
```suggestion
				return nil, status.Errorf(codes.FailedPrecondition, "get current user: %v", err)
```

lixmal:
Thanks a lot

lixmal (approved):

PR #2922, opened 2024-11-21
[client] Add interface index in windows route update log

lixmal:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/internal/routemanager/systemops/systemops_windows.go
```diff
@@ -230,10 +230,13 @@ func (rm *RouteMonitor) parseUpdate(row *MIB_IPFORWARD_ROW2, notificationType MI
 	if idx != 0 {
 		intf, err := net.InterfaceByIndex(idx)
 		if err != nil {
-			return update, fmt.Errorf("get interface name: %w", err)
```
pappz, line 233: Based on the description of the PR, it was to help debug things only. But if you remove the return the logic will be different. Or not?

lixmal: The logic is slightly different, we won't error out here, instead, we will return the interface info we have (index). Index is sufficient for the current checks we have in the network monitor for Windows.

This case where we cannot lookup the interface only happens if we get a route update notification from the OS at the same time as the interface is removed (e.g. the interface is removed and all routes that point to it are removed as well, like in the linked issue).

pappz (approved):

PR #2936, opened 2024-11-22
[management] Add integration test for the setup-keys API endpoints

pascal-fischer:
## Describe your changes
This PR adds integration tests to test the full flow from API endpoint down to the store layer back up to the API response and the peers update message.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/setupkey.go
```diff
@@ -248,7 +248,7 @@ func (am *DefaultAccountManager) CreateSetupKey(ctx context.Context, accountID s
 
 	err = am.Store.ExecuteInTransaction(ctx, func(transaction Store) error {
 		if err = validateSetupKeyAutoGroups(ctx, transaction, accountID, autoGroups); err != nil {
-			return err
+			return status.Errorf(status.InvalidArgument, "invalid auto groups: %v", err)
```
lixmal, line 78: doesn't validateSetupKeyAutoGroups already return a `status.Errorf`? And the error is not wrapped, so the internal status code gets lost

management/server/setupkey.go
```diff
@@ -305,7 +305,7 @@ func (am *DefaultAccountManager) SaveSetupKey(ctx context.Context, accountID str
 
 	err = am.Store.ExecuteInTransaction(ctx, func(transaction Store) error {
 		if err = validateSetupKeyAutoGroups(ctx, transaction, accountID, keyToSave.AutoGroups); err != nil {
-			return err
+			return status.Errorf(status.InvalidArgument, "invalid auto groups: %v", err)
```
lixmal, line 135: here too

management/server/setupkey.go
pascal-fischer: Thats done on purpose. The internal status is not found in the database, but at this level I want to return invalid argument so the user gets a proper http status code. Not 404

bcmmbaga (approved):

PR #2937, opened 2024-11-22
[client] Android exit node via P2P

pappz:
## Describe your changes
Support exit node feature on Android via P2P communication

Protect the WireGuard UDP listeners with marks.
The implementation can support the VPN permission revocation events in thread safe way. It will be important if we start to support the running time route and DNS update features.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/iface/bind/control_android.go
+++ b/client/iface/bind/control_android.go
@@ -1,0 +1,12 @@
+package bind
+
+import (
+	wireguard "golang.zx2c4.com/wireguard/conn"
+
+	nbnet "github.com/netbirdio/netbird/util/net"
+)
+
+func init() {
+	// ControlFns is not thread safe and should only be modified during init.
+	*wireguard.ControlFns = append(*wireguard.ControlFns, nbnet.ControlProtectSocket)
+}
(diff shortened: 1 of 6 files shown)


client/iface/bind/control_android.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	wireguard "golang.zx2c4.com/wireguard/conn"
+
+	nbnet "github.com/netbirdio/netbird/util/net"
+)
+
+func init() {
+	// ControlFns is not thread safe and should only be modified during init.
+	*wireguard.ControlFns = append(*wireguard.ControlFns, nbnet.ControlProtectSocket)
```
lixmal, line 11: Should we check for nil here first? Just in case lib code changes

lixmal (approved):

client/iface/bind/control_android.go
pappz: hmm, hard question what is the best error handling here. If it is nil what can we do? We are in the init() function. Throw a fatal error? Actually we do it right now just without error check.

lixmal: I don't have a good idea either. Let's create a ticket to handle this later

PR #2939, opened 2024-11-22
[client] Account different policiy rules for routes firewall rules

mlsmaycon:
## Describe your changes
This fixes the behavior where multiple policies with different access levels were applied to all distribution group peers. 

Now, we ensure that route firewall rules generation will consider source group peers of access control policies and apply rules to peers from these groups. Peers from the distribution group that don't belong to any source group will have their traffic dropped.

This will change behavior for existing and valid policies benefiting from this flaw. 

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/route.go
+++ b/management/server/route.go
@@ -421,25 +421,32 @@ func (a *Account) getPeerRoutesFirewallRules(ctx context.Context, peerID string,
 
 		for _, accessGroup := range route.AccessControlGroups {
 			policies := getAllRoutePoliciesFromGroups(a, []string{accessGroup})
-			for _, policy := range policies {
-				if !policy.Enabled {
-					continue
-				}
-
-				for _, rule := range policy.Rules {
-					if !rule.Enabled {
-						continue
-					}
-
-					rulePeers := a.getRulePeers(rule, peerID, distributionPeers, validatedPeersMap)
-					rules := generateRouteFirewallRules(ctx, route, rule, rulePeers, firewallRuleDirectionIN)
-					routesFirewallRules = append(routesFirewallRules, rules...)
-				}
-			}
+			rules := a.getRouteFirewallRules(ctx, peerID, policies, route, validatedPeersMap, distributionPeers)
+			routesFirewallRules = append(routesFirewallRules, rules...)
 		}
 	}
 
 	return routesFirewallRules
+}
+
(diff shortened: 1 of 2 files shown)


management/server/route_test.go
```diff
@@ -1770,7 +1886,7 @@ func TestAccount_getPeersRoutesFirewall(t *testing.T) {
 				IsDynamic:    true,
 			},
 		}
-		assert.ElementsMatch(t, routesFirewallRules, expectedRoutesFirewallRules)
+		assert.ElementsMatch(t, orderRuleSourceRanges(routesFirewallRules), orderRuleSourceRanges(expectedRoutesFirewallRules))
```
lixmal, line 1889: >func ElementsMatch(t TestingT, listA interface{}, listB interface{}, msgAndArgs ...interface{}) (ok bool)
 
>ElementsMatch asserts that the specified listA(array, slice...) is equal to specified listB(array, slice...) ignoring the order of the elements. If there are duplicate elements, the number of appearances of each of them in both lists should match.

This works, without the change in this line, doesn't it?

There's another a few lines up

mlsmaycon: It does, I am just keeping the same style to avoid future issues and because this is a testing func.

management/server/route.go
```diff
@@ -417,25 +417,82 @@ func (a *Account) getPeerRoutesFirewallRules(ctx context.Context, peerID string,
...
+			rules := a.getRouteFirewallRules(ctx, peerID, policies, route, validatedPeersMap, distributionPeers)
+			routesFirewallRules = append(routesFirewallRules, rules...)
+		}
+	}
+
+	return routesFirewallRules
+}
+
+func (a *Account) getRouteFirewallRules(ctx context.Context, peerID string, policies []*Policy, route *route.Route, validatedPeersMap map[string]struct{}, distributionPeers map[string]struct{}) []*RouteFirewallRule {
+	fwRules := make([]*RouteFirewallRule, 0)
```
lixmal:
```suggestion
	var fwRules []*RouteFirewallRule
```

Saves an allocation if there will be no rules

lixmal (approved):
We should add a few more test cases, maybe in a follow-up PR

- peers in multiple source groups of the same policy
- peer in both source and access control group
- empty groups

lixmal (approved):

PR #2952, opened 2024-11-25
[relay client] choose nearest relay based on latency

### LINKED ISSUE #2950: [Relay] chosen home relay server is random
**Describe the problem**

chosen home Relay server is random and not based on latency.

**To Reproduce**

1. Deploy multiple relay servers (new relay) to different geographic regions.
2. Run client in same region as one of relay servers.
3. Observe chosen home relay server.

In my case specifically:
1. Deploy relay servers to AWS regions:
    a. us-east-1
    b. eu-central-1
    c. ap-south-1
2. Run client in us-east-1

**Expected behavior**

Expected client to chose home relay server to be closest server in latency.

Self-hosted

**NetBird version**

0.33.0

**NetBird status -dA output:**

N/A

**Do you face any (non-mobile) client issues?**

N/A

**Screenshots**

N/A

**Additional context**

I noticed in the code in `relay/client/picker.go` in `processConnResults()` it takes the first item with a successful connection from the `resultChan chan`, this is highly dependent on Golang scheduling and channel ordering rather than latency.

Might be relevant: tested client has only 1 core (so only 1 thread can run at a time, which would highly affect the ordering to be only based on the Go scheduling), another client which has 8 cores chose the nearest relay server consistently.

mohamed-essam:
## Describe your changes

Change picker to choose nearest relay based on connection latency.

https://github.com/netbirdio/netbird/issues/2950

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/relay/client/client.go
+++ b/relay/client/client.go
@@ -123,11 +123,12 @@ func (cc *connContainer) close() {
 // the client can be reused by calling Connect again. When the client is closed, all connections are closed too.
 // While the Connect is in progress, the OpenConn function will block until the connection is established with relay server.
 type Client struct {
-	log            *log.Entry
-	parentCtx      context.Context
-	connectionURL  string
-	authTokenStore *auth.TokenStore
-	hashedID       []byte
+	log                   *log.Entry
+	parentCtx             context.Context
+	connectionURL         string
+	authTokenStore        *auth.TokenStore
+	hashedID              []byte
+	InitialConnectionTime time.Duration
 
 	bufPool *sync.Pool
 
@@ -264,11 +265,12 @@ func (c *Client) Close() error {
 }
 
 func (c *Client) connect() error {
-	conn, err := ws.Dial(c.connectionURL)
+	conn, latency, err := ws.Dial(c.connectionURL)
 	if err != nil {
 		return err
 	}
 	c.relayConn = conn
+	c.InitialConnectionTime = latency
 
 	err = c.handShake()
 	if err != nil {
--- a/relay/client/dialer/ws/ws.go
+++ b/relay/client/dialer/ws/ws.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/http/httptrace"
 	"net/url"
 	"strings"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 	"nhooyr.io/websocket"
@@ -15,10 +17,10 @@ import (
 	nbnet "github.com/netbirdio/netbird/util/net"
 )
 
-func Dial(address string) (net.Conn, error) {
+func Dial(address string) (net.Conn, time.Duration, error) {
 	wsURL, err := prepareURL(address)
 	if err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 
 	opts := &websocket.DialOptions{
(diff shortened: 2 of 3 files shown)


pappz:
Hi @mohamed-essam,
First of all, thank you for your PR. Your observation regarding the scheduler is a good point.

When we designed this feature we focused on the fastest connection time as possible. In a corner case, your logic will flip this requirement and block the connection more then the defined timeout time. Imagine that situation if the length of the list of servers is 8.  And you have two really slow servers. Actually one of them is between 1-7 and 8. In this case, the code will block until 2*30 sec.

For your scenario, we suggest using only one domain with the Geo DNS service.

Hard to write good code for the scheduler but we can take try. I think your solution could works well because the 
client.Connect() does not start new go routines so the scheduler does not trigger context switches, luckily. But if we develop something inside the Client struct that break this rule then your solution will be provide invalid latency numbers again. So in a nutshell you measure the running time of the client.Connect() and not just the network latency.  And by the way I am working on a task that will start new parallel connection inside the Client code :/ . In the ideal case we should extend our network protocol, with elapsed time in the response (somehow). 

But back to your PR. What do you think if after [this](https://github.com/netbirdio/netbird/pull/2952/files#diff-8557a53ac9be99f44ecce48a45ec3a255e4bede53ed9b2283ae337263a2e3169R90) line, you check the length of the channel and if it contains more then one items, then drain the channel and compare the elapsed times on that? And do not wait for any new connections.

mohamed-essam:
Hello @pappz,

Thanks for the clarification of the requirement, I haven't really considered this case in my change, I think it might be possible to do something that is a mix of both, as in most cases delaying initial connection by some amount (think for example 200ms), _could_ result in a much faster relay off the get go.

As for the Geo DNS, my problem with that solution is that it would rely _only_ on geographical distance latency, and not take into account (for example) relay server load.

I'm working on a commit now that should address both the corner case of slow servers (by having some sort of upper limit of waiting after the first successful connection, which I'm thinking a value between 200-500ms should be sufficient to both allow cases of Go scheduling delays and weird networking to resolve), and the measurement of latency (by using [httptrace](https://pkg.go.dev/net/http/httptrace) inside `ws.go Dial()`)

Please let me know what you think of that previous approach even if I haven't committed it yet :pray:

pappz:
Measuring the time in deeper layers looks promising. Good idea. I will take a look and we will see.

mohamed-essam:
I don't have access to a single core machine today, so as a workaround I executed netbird -F with cpuaffinity to force it into a single CPU.
I added one more debug line before receving from the channel (to see if there is sometimes a connection already ready without the processConnResults goroutine halting), and that seems to be the case:

```
2024-11-29T13:38:58+02:00 WARN client/picker.go:99: beforeSelect:len(resultChan)=0
2024-11-29T13:38:59+02:00 WARN client/picker.go:110: afterSelect:len(resultChan)=0
2024-11-29T13:38:59+02:00 WARN client/picker.go:99: beforeSelect:len(resultChan)=1
2024-11-29T13:38:59+02:00 WARN client/picker.go:110: afterSelect:len(resultChan)=0
2024-11-29T13:38:59+02:00 WARN client/picker.go:99: beforeSelect:len(resultChan)=1
2024-11-29T13:38:59+02:00 WARN client/picker.go:110: afterSelect:len(resultChan)=0
```

In this case (3 relays), there were 2 times where a connection was ready immediately after the last connection was processed.

This also seemed to be dependent on whether the CPU was loaded or not, when loaded it was more likely to be always empty (before and after select statement).

One more thing I noticed, is that when CPU was free, the order of the connections was always (somewhat) already sorted by connection, but when CPU was loaded, the order of connections was more likely to be the order in which they were configured in management. I'm assuming this is because the first item in the array is likely to be connected to first if the CPU load is high enough for the process to wait a bit until connecting to the next relay.

All in all, the slight time delay seems to be always choosing the lowest latency in both cases.

pappz:
Thank you for sharing your experience.
I would like to draw your attention to [this](https://github.com/netbirdio/netbird/pull/2952#issuecomment-2507656874) work. What do you think is the `net/http/httptrace` usable in QUIC too? If so, I don't see any other blocking issues in this topic.

mlsmaycon:
@mohamed-essam with the addition to QUIC proto support, we caused a merge conflict. can you check them?

mohamed-essam:
Will try to do so over the weekend🙏

SilentEntity:
Any update on this ?

JimmyKmi:
Any update on this? This will be very useful, thanks to the contributor.

Kunlun-Hub:
Any update on this? This will be very useful, thanks to the contributor.

joachimtingvold:
Was this implemented somehow, or is it still selecting random relays? I would say latency-based relay-selection is very relevant (unless all your relays are in the same geographical area).

Also, relying on Geo-DNS is much more complex compared to using latency.

PR #2957, opened 2024-11-27
[management] refactor to use account object instead of separate db calls for peer update

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/posture_checks.go
+++ b/management/server/posture_checks.go
@@ -159,7 +159,7 @@ func (am *DefaultAccountManager) getPeerPostureChecks(account *Account, peerID s
 	}
 
 	for _, policy := range account.Policies {
-		if !policy.Enabled {
+		if !policy.Enabled || len(policy.SourcePostureChecks) == 0 {
 			continue
 		}
 
--- a/management/server/posture_checks.go
+++ b/management/server/posture_checks.go
@@ -151,28 +151,21 @@ func (am *DefaultAccountManager) ListPostureChecks(ctx context.Context, accountI
 }
 
 // getPeerPostureChecks returns the posture checks applied for a given peer.
-func (am *DefaultAccountManager) getPeerPostureChecks(ctx context.Context, account *Account, peerID string) ([]*posture.Checks, error) {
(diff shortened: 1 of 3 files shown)


management/server/posture_checks.go
```diff
@@ -149,30 +151,20 @@ func (am *DefaultAccountManager) ListPostureChecks(ctx context.Context, accountI
 }
 
 // getPeerPostureChecks returns the posture checks applied for a given peer.
-func (am *DefaultAccountManager) getPeerPostureChecks(ctx context.Context, accountID string, peerID string) ([]*posture.Checks, error) {
+func (am *DefaultAccountManager) getPeerPostureChecks(ctx context.Context, account *Account, peerID string) ([]*posture.Checks, error) {
 	peerPostureChecks := make(map[string]*posture.Checks)
 
 	err := am.Store.ExecuteInTransaction(ctx, func(transaction Store) error {
```
bcmmbaga: We can remove the store transaction here because we no longer interact with it

management/server/posture_checks.go
```diff
@@ -252,9 +244,10 @@ func addPolicyPostureChecks(ctx context.Context, transaction Store, accountID, p
 	}
 
 	for _, sourcePostureCheckID := range policy.SourcePostureChecks {
-		postureCheck, err := transaction.GetPostureChecksByID(ctx, LockingStrengthShare, accountID, sourcePostureCheckID)
-		if err != nil {
-			return err
+		account.getPostureChecks(sourcePostureCheckID)
```
bcmmbaga: duplicate call to get posture checks

bcmmbaga (changes requested):

management/server/posture_checks.go
pascal-fischer: yeah this I am removing

management/server/posture_checks.go
pascal-fischer: removed

bcmmbaga (approved):

PR #2977, opened 2024-12-03
Add missing parentheses on iphone hostname generation condition

jnohlgard:
## Describe your changes

Add missing parentheses around the comparison to fill the hostname field from the IdP user data when using iphone.

No ticket. Found this apparent bug when browsing the code for something completely unrelated, I don't believe this will affect anything in real world usage.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/peer.go
+++ b/management/server/peer.go
@@ -477,7 +477,7 @@ func (am *DefaultAccountManager) AddPeer(ctx context.Context, setupKey, userID s
 			setupKeyName = sk.Name
 		}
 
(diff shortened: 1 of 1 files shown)


pascal-fischer:
Hi @jnohlgard,

you are right, this will have no effect because when a peer doesn't have a userID the IdP lookup will fail and it will fallback to the default name anyway.

Anyway, the code should reflect the proper intended behavior and not rely on the surrounding code to handle this.

That was a very good catch! Thanks a lot for the fix.

pascal-fischer (approved):

PR #2989, opened 2024-12-05
[client] Init route selector early

lixmal:
## Describe your changes

Makes sure the route selector is initialized even if custom routing is disabled or we error out early.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/routemanager/manager.go
+++ b/client/internal/routemanager/manager.go
@@ -124,6 +124,8 @@ func NewManager(
 
 // Init sets up the routing
 func (m *DefaultManager) Init() (nbnet.AddHookFunc, nbnet.RemoveHookFunc, error) {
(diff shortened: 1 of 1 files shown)


client/internal/routemanager/manager.go
```diff
@@ -144,8 +146,6 @@ func (m *DefaultManager) Init() (nbnet.AddHookFunc, nbnet.RemoveHookFunc, error)
 		return nil, nil, fmt.Errorf("setup routing: %w", err)
```
pappz, line 146: What will happen if an issue occurs? Does the loaded state cause future issues?

lixmal: If we don't init the route selector before returning here it will be nil and we'll panic later

PR #2992, opened 2024-12-05
[client] Restart netbird-ui post-install in linux deb&rpm

mohamed-essam:
## Describe your changes

Add post-install script to netbird-ui deb and rpm packages to restart process if running.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/.goreleaser_ui.yaml
+++ b/.goreleaser_ui.yaml
@@ -50,6 +50,8 @@ nfpms:
       - netbird-ui
     formats:
       - deb
+    scripts:
+      postinstall: "release_files/ui-post-install.sh"
     contents:
       - src: client/ui/netbird.desktop
         dst: /usr/share/applications/netbird.desktop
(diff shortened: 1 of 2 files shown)


mgarces:
Hi @mohamed-essam , thank you for this contribution. 
Can you share how you tested this changes? I've tested both `rpm` and `deb` assets, both on an `Ubuntu Desktop 22.04` and on `Fedora Desktop 41`, and this didn't work at all; just want to make sure I can replicate your steps. 
Also, on Ubuntu, I got the following output:
```
/var/lib/dpkg/info/netbird-ui.postinst: 4 [[: not found 
```
I think this is due the script using `/bin/sh` as a shebang, which typically points to a more minimal shell. However, it uses `[[ ... ]]`, which is a feature of bash and not POSIX-compliant. This can cause the script to fail on systems where /bin/sh is not linked to bash. To fix this, you should either change the shebang to #!/bin/bash or use POSIX-compliant `[ ... ]`.

mohamed-essam:
Ah yes thanks, sorry I missed this and forgot that I had /bin/sh linked to /bin/bash on my system, I changed it now, and will be testing it on some other systems as well :pray:

mohamed-essam:
@mgarces Tested latest script on Ubuntu 22.04 (with and without /bin/sh linked to /bin/bash), and Fedora Desktop 40 and both work now perfectly :heavy_check_mark:

lixmal (approved):

PR #2994, opened 2024-12-06
[management] Feature - Migration to Zitadel v2 API

remipcomaite:
## Describe your changes

Hello,

I propose a PR to migrate to Zitadel API v2.

This allows to authenticate users from different organizations.

The endpoint for the management API changes from "/management/v1" to "/v2".

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


braginini:
Thank you for the contribution @remipcomaite 
Is this backwards compatible with the previous version?

remipcomaite:
Hello @braginini ,

The code is not backwards compatible because V1 of the API will be deprecated in the long term.
I changed the endpoints and some properties to match the API calls with V2.

mlsmaycon:
@remipcomaite when was v1 support removed from Zitadel?

remipcomaite:
> @remipcomaite when was v1 support removed from Zitadel?

I don't know but v1 and v2 are GA.
The v3 version of the API is in "Preview".

https://zitadel.com/docs/apis/introduction

mlsmaycon:
> > @remipcomaite when was v1 support removed from Zitadel?
> 
> @mlsmaycon I don't know but v1 (service-based) and v2 (resource-based) are GA. The v3 version of the API is in "Preview".
> 
> https://zitadel.com/docs/apis/introduction
> 

Thanks for checking @remipcomaite. I've asked in Zitadel's community discord, and it appears that they don't have a plan to deprecate the API yet. But the notification gives us an idea of the overall intentions. 

With that said, the problem with dropping V1 support on our end is that self-hosted users would have to upgrade their Zitadel's servers before updating NetBird's management service, and that will generate Github and Slack requests for assistance, which we can't provide. 

An alternative to that is to make v2 support optional on our side and keep v1 support for those users. 

That could be an IDP extra configuration, which, by default, uses V1, but we set our deployment scripts and guides to use V2. What do you think?

remipcomaite:
> That could be an IDP extra configuration, which, by default, uses V1, but we set our deployment scripts and guides to use V2. What do you think?

Yes, I think it's the best option.

mlsmaycon:
@remipcomaite do you want to work on this support? we can close this PR and open a new one when it is ready

remipcomaite:
@mlsmaycon Ok, I will try to make a new proposal to make the use of Zitadel API V2 optional.

Kind regards,

remipcomaite:
@mlsmaycon I'm sorry.
I didn't have time to work on it.
I'm not a developer so I don't really know how to make this easily optional.
Could you give me some pointers so I can look at how to adapt the code please?

remipcomaite:
@mlsmaycon @adsehgal I propose a new pull request to address this issue.

https://github.com/netbirdio/netbird/pull/4733

PR #3001, opened 2024-12-07
[client] Add peer conn init limit

mlsmaycon:
## Describe your changes
Limit the peer connection initialization at 200 peers at the same time

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/util/semaphore-group/semaphore_group.go
+++ b/util/semaphore-group/semaphore_group.go
@@ -1,0 +1,39 @@
+package semaphoregroup
+
+import (
+	"sync"
+)
+
+// SemaphoreGroup is a custom type that combines sync.WaitGroup and a semaphore.
+type SemaphoreGroup struct {
+	waitGroup sync.WaitGroup
+	semaphore chan struct{}
+}
+
+// NewSemaphoreGroup creates a new SemaphoreGroup with the specified semaphore limit.
+func NewSemaphoreGroup(limit int) *SemaphoreGroup {
+	return &SemaphoreGroup{
+		semaphore: make(chan struct{}, limit),
+	}
+}
+
+// Add increments the internal WaitGroup counter and acquires a semaphore slot.
+func (sg *SemaphoreGroup) Add() {
+	sg.waitGroup.Add(1)
+
+	// Acquire semaphore slot
+	sg.semaphore <- struct{}{}
+}
+
+// Done decrements the internal WaitGroup counter and releases a semaphore slot.
+func (sg *SemaphoreGroup) Done() {
+	sg.waitGroup.Done()
+
+	// Release semaphore slot
+	<-sg.semaphore
+}
+
+// Wait waits until the internal WaitGroup counter is zero.
+func (sg *SemaphoreGroup) Wait() {
+	sg.waitGroup.Wait()
+}
--- a/util/semaphore-group/semaphore_group.go
+++ b/util/semaphore-group/semaphore_group.go
@@ -1,6 +1,7 @@
 package semaphoregroup
 
 import (
+	"context"
 	"sync"
 )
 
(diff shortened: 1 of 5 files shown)


util/semaphore-group/semaphore_group.go
```diff
@@ -0,0 +1,39 @@
...
+// Done decrements the internal WaitGroup counter and releases a semaphore slot.
+func (sg *SemaphoreGroup) Done() {
+	sg.waitGroup.Done()
+
+	// Release semaphore slot
+	<-sg.semaphore
+}
+
+// Wait waits until the internal WaitGroup counter is zero.
+func (sg *SemaphoreGroup) Wait() {
```
pappz, line 46: Maybe I missed something but I do not see where you use the exported `Wait()`. I think the wg is unused, the channel is a good and eought limiter.

util/semaphore-group/semaphore_group.go
```diff
@@ -0,0 +1,39 @@
+package semaphoregroup
```
pappz, line 1: Better place inside the peer package.

util/semaphore-group/semaphore_group.go
```diff
@@ -0,0 +1,39 @@
...
+		semaphore: make(chan struct{}, limit),
+	}
+}
+
+// Add increments the internal WaitGroup counter and acquires a semaphore slot.
+func (sg *SemaphoreGroup) Add() {
+	sg.waitGroup.Add(1)
+
+	// Acquire semaphore slot
+	sg.semaphore <- struct{}{}
```
pappz: Maybe this line can cause an infinitie block (depending on how do we use this struct). Maybe it would be safer if we pass a context to the Add(ctx context.Context) and handle the cancel event.

util/semaphore-group/semaphore_group.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// Done decrements the internal WaitGroup counter and releases a semaphore slot.
+func (sg *SemaphoreGroup) Done() {
+	sg.waitGroup.Done()
```
pappz, line 35: If we run once time the `Add()` and run multiple times the `Done()` then it will cause an exception. If we use exported functions of this struct from multiple go routines then this part is dangerous.

util/semaphore-group/semaphore_group.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// Release semaphore slot
+	<-sg.semaphore
```
pappz: Same thing with the context cancel.

util/semaphore-group/semaphore_group.go
mlsmaycon: I am moving this package from another project and the util seems the most appropriated

util/semaphore-group/semaphore_group.go
mlsmaycon: any suggestions?

mohamed-essam: It carries the exact same risk as using a plain sync.WaitGroup, as long as the usage guarantees 1:1 Add and Done usage it should be fine, no?

pappz: I think the solution is simple. Just fully delete the  sg.waitGroup. The exported Wait has been used by the unit test only, but we can do it in different way. After all Go routine exited, check the len of the channel. It is not nice way to check internal variables in unit test but the code will be simpler and less risky to use.

pappz: @mohamed-essam I think this is Go philosophy question. The panic can force the developer to write, thread-safe code. I think returning with an error and handling it in the upper layer is a more defensive approach in a production-ready code. If this code snipped would be placed in internal package or it is not exported then the panic is okay from my point of view.

pappz (approved):

PR #3018, opened 2024-12-10
[management] Extract internal types into dedicated package

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/management/server/util/util.go
+++ b/management/server/util/util.go
@@ -1,6 +1,6 @@
 package util
 
-// difference returns the elements in `a` that aren't in `b`.
+// Difference returns the elements in `a` that aren't in `b`.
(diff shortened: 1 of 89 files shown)


management/server/util/util.go
```diff
@@ -0,0 +1,16 @@
+package util
+
+// difference returns the elements in `a` that aren't in `b`.
```
bcmmbaga:
```suggestion
// Difference returns the elements in `a` that aren't in `b`.
```

bcmmbaga (changes requested):

bcmmbaga (approved):

PR #3025, opened 2024-12-11
[management] Extends policy with source and destination resources

bcmmbaga:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/http/api/openapi.yml
+++ b/management/server/http/api/openapi.yml
@@ -782,15 +782,18 @@ components:
               items:
                 type: string
                 example: "ch8i4ug6lnn4g9hqv797"
+            sourceResource:
+              description: Policy rule source resource that the rule is applied to
+              $ref: '#/components/schemas/Resource'
             destinations:
               description: Policy rule destination group IDs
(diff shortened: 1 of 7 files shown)


management/server/http/api/openapi.yml
```diff
@@ -1176,6 +1182,24 @@ components:
             - id
             - network_type
         - $ref: '#/components/schemas/RouteRequest'
+    Resource:
+      type: object
+      properties:
+        id:
+          description: Resource ID
```
pascal-fischer: We should be more clear here what the ID is. In this case it is not the object ID but the ID of the actual resource we are targeting

management/server/http/api/openapi.yml
```diff
@@ -1176,6 +1182,24 @@ components:
...
+          type: string
+          example: chacdk86lnnboviihd7g
+        type:
+          description: Resource type
```
pascal-fischer: Same for the description

pascal-fischer (approved):

PR #3028, opened 2024-12-11
Create thread safe peer store

pappz:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -34,6 +34,7 @@ import (
 	"github.com/netbirdio/netbird/client/internal/peer"
 	"github.com/netbirdio/netbird/client/internal/peer/guard"
 	icemaker "github.com/netbirdio/netbird/client/internal/peer/ice"
+	"github.com/netbirdio/netbird/client/internal/peerstore"
 	"github.com/netbirdio/netbird/client/internal/relay"
 	"github.com/netbirdio/netbird/client/internal/rosenpass"
 	"github.com/netbirdio/netbird/client/internal/routemanager"
@@ -117,7 +118,7 @@ type Engine struct {
 	// mgmClient is a Management Service client
 	mgmClient mgm.Client
 	// peerConns is a map that holds all the peers that are known to this peer
-	peerConns map[string]*peer.Conn
+	peerStore *peerstore.Store
 
 	beforePeerHook nbnet.AddHookFunc
 	afterPeerHook  nbnet.RemoveHookFunc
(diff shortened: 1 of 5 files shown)


client/internal/peerstore/store.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	return p, true
+}
+
+func (s *Store) PeersPubKey() []string {
+	s.peerConnsMu.RLock()
+	defer s.peerConnsMu.RUnlock()
+
+	keys := make([]string, 0, len(s.peerConns))
+	for k, _ := range s.peerConns {
```
lixmal: You can use maps.Keys here, might be a bit prettier

client/internal/routemanager/manager.go
```diff
@@ -67,21 +68,10 @@ type DefaultManager struct {
...
-	dnsRouteInterval time.Duration,
-	wgInterface iface.IWGIface,
-	statusRecorder *peer.Status,
-	relayMgr *relayClient.Manager,
-	initialRoutes []*route.Route,
-	stateManager *statemanager.Manager,
-	dnsServer dns.Server,
-	peerConns map[string]*peer.Conn,
-) *DefaultManager {
+func NewManager(ctx context.Context, pubKey string, dnsRouteInterval time.Duration, wgInterface iface.IWGIface, statusRecorder *peer.Status, relayMgr *relayClient.Manager, initialRoutes []*route.Route, stateManager *statemanager.Manager, dnsServer dns.Server, peerStore *peerstore.Store) *DefaultManager {
```
lixmal: can you change this back to multiline?

client/internal/engine.go
```diff
@@ -462,8 +463,8 @@ func (e *Engine) modifyPeers(peersUpdate []*mgmProto.RemotePeerConfig) error {
 	var modified []*mgmProto.RemotePeerConfig
 	for _, p := range peersUpdate {
 		peerPubKey := p.GetWgPubKey()
-		if peerConn, ok := e.peerConns[peerPubKey]; ok {
-			if peerConn.WgConfig().AllowedIps != strings.Join(p.AllowedIps, ",") {
+		if allowedIP, ok := e.peerStore.AllowedIP(peerPubKey); ok {
+			if allowedIP.String() != strings.Join(p.AllowedIps, ",") {
```
lixmal: I think this is different than before. `allowedIP` is a `net.IP`, so always a single one, while the one from wgconfig could be multiple

PR #3044, opened 2024-12-13
[management] Generate network map for new network concept

bcmmbaga:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/types/account.go
```diff
@@ -1198,3 +1202,176 @@ func GetAllRoutePoliciesFromGroups(account *Account, accessControlGroups []strin
...
+
+				resources := a.getNetworkResources(router.NetworkID)
+				routes = append(routes, a.getNetworkResourcesRoutes(resources, router, peer)...)
+			}
+		}
+
+		if router.Peer == peerID {
+			resources := a.getNetworkResources(router.NetworkID)
+			routes = append(routes, a.getNetworkResourcesRoutes(resources, router, peer)...)
+		}
```
mlsmaycon, line 1242:
```suggestion
	for _, router := range a.NetworkRouters {
	        isMember := false
		for _, groupID := range router.PeerGroups {
			group := a.GetGroup(groupID)
			if group == nil {
				log.WithContext(ctx).Warnf("router %s has peers group %s that doesn't exist under account %s, will continue map generation", router.ID, groupID, a.Id)
				continue
			}

			for _, id := range group.Peers {
				if id == peerID {
					isMember = true
					break
				}

			}
		}

		if isMember || router.Peer == peerID {
			resources := a.getNetworkResources(router.NetworkID)
			routes = append(routes, a.getNetworkResourcesRoutes(resources, router, peer)...)
		}
```

mlsmaycon: this can be done in a new PR

PR #3054, opened 2024-12-15
[client] Restrict Unix socket permissions to 0600 for enhanced security

zakkuuno:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/cmd/service_controller.go
+++ b/client/cmd/service_controller.go
@@ -48,7 +48,7 @@ func (p *program) Start(svc service.Service) error {
 		defer listen.Close()
 
 		if split[0] == "unix" {
-			err = os.Chmod(split[1], 0666)
+			err = os.Chmod(split[1], 0600)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


zakkuuno:
@mlsmaycon That sounds great! What do you think about creating a dedicated "netbird" group during installation, similar to how Docker manages access? The group would have read/write permissions to the Unix socket, and the first user can be added automatically to the group. This gives users access to the Unix socket without administrators permission.

mlsmaycon:
> @mlsmaycon That sounds great! What do you think about creating a dedicated "netbird" group during installation, similar to how Docker manages access? The group would have read/write permissions to the Unix socket, and the first user can be added automatically to the group. This gives users access to the Unix socket without administrators permission.

That's an option, but we want to avoid creating resources for that. This would also imply handling group/user membership on MacOS, FreeBSD and cases with LDAP authentication enabled systems

mlsmaycon:
will close this one to handle in a different PR

PR #3059, opened 2024-12-16
[management] Remove resource from groups when resource deleted

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/networks/manager.go
+++ b/management/server/networks/manager.go
@@ -2,6 +2,7 @@ package networks
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/rs/xid"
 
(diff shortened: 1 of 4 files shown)


management/server/networks/manager.go
```diff
@@ -98,7 +99,34 @@ func (m *managerImpl) DeleteNetwork(ctx context.Context, accountID, userID, netw
...
+		if err != nil {
+			return fmt.Errorf("failed to get routers in network: %w", err)
+		}
+
+		for _, router := range routers {
+			err = m.routersManager.DeleteRouter(ctx, accountID, userID, networkID, router.ID)
+			if err != nil {
+				return fmt.Errorf("failed to delete router: %w", err)
+			}
+		}
```
bcmmbaga: Deleting resources and routers using managers will be blocked because the store is already locked with a transaction. Consider calling the method directly using `transaction.Delete...` instead

management/server/networks/manager.go
```diff
@@ -98,7 +99,34 @@ func (m *managerImpl) DeleteNetwork(ctx context.Context, accountID, userID, netw
...
+
+		return m.store.DeleteNetwork(ctx, store.LockingStrengthUpdate, accountID, networkID)
```
bcmmbaga:
```suggestion
		return transaction.DeleteNetwork(ctx, store.LockingStrengthUpdate, accountID, networkID)
```

bcmmbaga (changes requested):

bcmmbaga (approved):

PR #3093, opened 2024-12-20
[client] Add stateful userspace firewall and remove egress filters

lixmal:
## Describe your changes

- Add stateful firewall functionality for UDP/TCP/ICMP in userspace firewalll
- Removes all egress drop rules/filters, still needs refactoring so we don't add output rules to any chains/filters.
- on Linux, if the OUTPUT policy is DROP  then we don't do anything about it (no extra allow rules). This is up to the user, if they don't want anything leaving their machine they'll have to manage these rules explicitly.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/firewall/uspfilter/conntrack/tcp.go
```diff
@@ -0,0 +1,312 @@
...
+		timeout:       timeout,
+	}
+
+	go tracker.cleanupRoutine()
+	return tracker
+}
+
+// TrackOutbound processes an outbound TCP packet and updates connection state
+func (t *TCPTracker) TrackOutbound(srcIP net.IP, dstIP net.IP, srcPort uint16, dstPort uint16, flags uint8) {
+	t.mutex.Lock()
```
pappz: We can apply the locks later, right before touch the map.

client/firewall/uspfilter/conntrack/tcp.go
```diff
@@ -0,0 +1,312 @@
...
+	DestPort    uint16
+	State       TCPState
+	LastSeen    time.Time
+	established bool
+}
+
+// TCPTracker manages TCP connection states
+type TCPTracker struct {
+	connections   map[TCPConnKey]*TCPConnTrack
+	mutex         sync.RWMutex
```
pappz, line 70: It is RWMutex but it has been used without RW feature

lixmal: True. Looking at the locks we also modify conn under read locks

client/firewall/uspfilter/conntrack/tcp.go
lixmal: fixed

client/firewall/uspfilter/conntrack/tcp.go
lixmal: I've addressed the locking issues

mlsmaycon (approved):

PR #3108, opened 2024-12-24
[management] Add MySQL Support

bcmmbaga:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


ismail0234:
All of these commits are unnecessary. In the PR I have already opened, mysql is running without any problems. I don't understand why you send so many useless commits.

bcmmbaga:
> All of these commits are unnecessary. In the PR I have already opened, mysql is running without any problems. I don't understand why you send so many useless commits.

The current setup with MySQL works only when the SQL mode is not set, but with the default strict mode, it encounters issues. Some of the broken functionality includes adding new users, creating user PATs, creating new setup keys, and adding new peers. These additional commits aim to address these problems and ensure the MySQL test container runs in strict mode to verify everything works correctly and to catch any potential issues that could break MySQL support

ismail0234:
> > All of these commits are unnecessary. In the PR I have already opened, mysql is running without any problems. I don't understand why you send so many useless commits.
> 
> The current setup with MySQL works only when the SQL mode is not set, but with the default strict mode, it encounters issues. Some of the broken functionality includes adding new users, creating user PATs, creating new setup keys, and adding new peers. These additional commits aim to address these problems and ensure the MySQL test container runs in strict mode to verify everything works correctly and to catch any potential issues that could break MySQL support

It should not be difficult for users to set sql mode. I have been using the mysql engine in a production environment for 1 month. I have not encountered any problems so far.

250+ onlines
600+ peers

Performance is quite good.

mlsmaycon (approved):

PR #3134, opened 2024-12-31
[client] Add experimental support for userspace routing

lixmal:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/firewall/uspfilter/forwarder/endpoint.go
+++ b/client/firewall/uspfilter/forwarder/endpoint.go
@@ -1,15 +1,17 @@
 package forwarder
 
 import (
-	log "github.com/sirupsen/logrus"
 	wgdevice "golang.zx2c4.com/wireguard/device"
 	"gvisor.dev/gvisor/pkg/tcpip"
 	"gvisor.dev/gvisor/pkg/tcpip/header"
 	"gvisor.dev/gvisor/pkg/tcpip/stack"
+
+	nblog "github.com/netbirdio/netbird/client/firewall/uspfilter/log"
 )
 
 // endpoint implements stack.LinkEndpoint and handles integration with the wireguard device
 type endpoint struct {
+	logger     *nblog.Logger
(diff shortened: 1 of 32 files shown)


Silex:
@lixmal: wonderful, tell me if I can help with testing. I have cameras that only allow user space and thus tailscale was the only one working, I can quickly tell you if what you did works for netbird.

Are there binaries of this branch built somewhere? If not, what's the simplest way to build them?

I'll need to crosscompile for aarch64 and armv7hf (arm32) to test, I also have routers that use mipsel arch but there userspace is not needed so the normal binaries work.

client/firewall/uspfilter/forwarder/endpoint.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (e *endpoint) MaxHeaderLength() uint16 {
+	return 0
+}
+
+func (e *endpoint) LinkAddress() tcpip.LinkAddress {
+	return ""
+}
+
+func (e *endpoint) WritePackets(pkts stack.PacketBufferList) (int, tcpip.Error) {
```
pappz, line 44: This never return with error

lixmal: True, but we have to implement this interface:

```
// LinkWriter is an interface that supports sending packets via a data-link
// layer endpoint. It is used with QueueingDiscipline to batch writes from
// upper layer endpoints.
type LinkWriter interface {
	// WritePackets writes packets. Must not be called with an empty list of
	// packet buffers.
	//
	// Each packet must have the link-layer header set, if the link requires
	// one.
	//
	// WritePackets may modify the packet buffers, and takes ownership of the PacketBufferList.
	// it is not safe to use the PacketBufferList after a call to WritePackets.
	WritePackets(PacketBufferList) (int, tcpip.Error)
}
```

pappz (approved):

pappz (approved):

PR #3142, opened 2025-01-02
[management] exclude self from network map if self is routing peer

pascal-fischer:
## Describe your changes
If the peer is a router and it would be added to the accessible peers part of the network map. This PR excludes a peer so it is not shown as connecting to itself.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/types/account.go
+++ b/management/server/types/account.go
@@ -335,6 +335,9 @@ func (a *Account) addNetworksRoutingPeers(
 		}
 	}
 	for p := range networkRoutesPeers {
+		if p == peer.ID {
+			continue
+		}
 		missingPeers[p] = struct{}{}
 	}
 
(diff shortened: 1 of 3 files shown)


management/server/types/account.go
```diff
@@ -317,8 +317,6 @@ func (a *Account) addNetworksRoutingPeers(
 		networkRoutesPeers[r.PeerID] = struct{}{}
 	}
 
-	delete(sourcePeers, peer.ID)
```
mlsmaycon, line 320: This should be enough, no?

pascal-fischer: no because it would need to be `delete(networkRoutesPeers, peer.ID)`

mlsmaycon: then the following is more clear:
```
delete(sourcePeers, peer.ID)
delete(networkRoutesPeers, peer.ID)
```

mlsmaycon: or when generating the `networkResourcesRoutes` map

pascal-fischer: ok

mlsmaycon (approved):

PR #3153, opened 2025-01-06
[client] Add disable system flags

lixmal:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/cmd/up.go
```diff
@@ -264,6 +277,19 @@ func runInDaemonMode(ctx context.Context, cmd *cobra.Command) error {
 		loginRequest.DnsRouteInterval = durationpb.New(dnsRouteInterval)
 	}
 
+	if cmd.Flag(disableClientRoutesFlag).Changed {
```
mlsmaycon, line 280: we are missing support to the foreground mode as well

client/internal/routemanager/manager.go
```diff
@@ -202,6 +215,11 @@ func (m *DefaultManager) initSelector() *routeselector.RouteSelector {
 }
 
 func (m *DefaultManager) EnableServerRouter(firewall firewall.Manager) error {
+	if m.disableServerRoutes {
```
mlsmaycon, line 218: should we handle nil firewall manager?

client/cmd/up.go
lixmal: It's here https://github.com/netbirdio/netbird/pull/3153/files#diff-0db67eedcdec93665cc9dea08a99e38d7ed16f94db2e98f842af921340815542R150-R162

client/internal/routemanager/manager.go
lixmal: There's a nil check further up in the chain, but I'll add it here explicitly as well

client/cmd/up.go
mlsmaycon: Totally missed that. Thanks.

client/internal/routemanager/manager.go
lixmal: done

mlsmaycon (approved):

PR #3217, opened 2025-01-19
[client] Add reverse dns zone

### LINKED ISSUE #3214: Question: Any possible way to get rDNS / PTR records for netbird clients?
As far as I know, there is currently no way to get rDNS resolved for netbird IPs. 
If there is, please point me in the right direction.

Use Case:
When playing around with and especially troubleshooting HA routes, I often use mtr / traceroute to get the current active route. While I can use the web interface to decipher which IP corresponds to which host, it would be way easier if the node name shows up right away. 
I'm sure there are many more use cases where this could be helpful.

I would love to hear your opinion on this.

lixmal:

Closes #3214
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


mlsmaycon:
are we adding the PTR zone as search domain?

lixmal:
> are we adding the PTR zone as search domain?

I've excluded it but seems it's not marked by OS tools anyway

hakansa (changes requested):
we can split the logic to make it more readable
```go
func addReverseZone(config *nbdns.Config, ipNet *net.IPNet) {
	zoneName, err := reverseZoneName(ipNet)
	if err != nil {
		log.Warn(err)
		return
	}

	if zoneExists(config, zoneName) {
		log.Debugf("reverse DNS zone %s already exists", zoneName)
		return
	}

	records := collectPTRRecords(config.CustomZones, ipNet)
	config.CustomZones = append(config.CustomZones, nbdns.CustomZone{
		Domain:  zoneName,
		Records: records,
	})

	log.Debugf("added reverse DNS zone: %s with %d records", zoneName, len(records))
}

// reverseZoneName computes the reverse DNS zone name
func reverseZoneName(ipNet *net.IPNet) (string, error) {
	networkIP := ipNet.IP.Mask(ipNet.Mask)
	maskOnes, _ := ipNet.Mask.Size()
	// round up to nearest byte
	octetsToUse := (maskOnes + 7) / 8

	octets := strings.Split(ip4.String(), ".")
	if octetsToUse > len(octets) {
		return "", fmt.Errorf("invalid network mask size for reverse DNS: %d", maskOnes)
	}

	// reverse the first octetsToUse octets.
	reversed := make([]string, octetsToUse)
	for i := 0; i < octetsToUse; i++ {
		reversed[octetsToUse-1-i] = octets[i]
	}

	zone := fmt.Sprintf("%s.in-addr.arpa", strings.Join(reversed, "."))
	return dns.Fqdn(zone), nil
}

// zoneExists returns true if the given zone name already exists in the configuration.
func zoneExists(config *nbdns.Config, zoneName string) bool {
	for _, zone := range config.CustomZones {
		if zone.Domain == zoneName {
			return true
		}
	}
	return false
}

// collectPTRRecords iterates over all A records in the existing zones,
// creating PTR records where applicable.
func collectPTRRecords(zones []nbdns.CustomZone, ipNet *net.IPNet) []nbdns.SimpleRecord {
	var records []nbdns.SimpleRecord
	for _, zone := range zones {
		for _, record := range zone.Records {
			if record.Type != int(dns.TypeA) {
				continue
			}
			if ptrRecord, ok := createPTRRecord(record, ipNet); ok {
				records = append(records, ptrRecord)
			}
		}
	}
	return records
}
```

hakansa (approved):

PR #3232, opened 2025-01-23
[client] Support port ranges in peer ACLs

lixmal:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/acl/manager.go
```diff
@@ -268,13 +268,16 @@ func (d *DefaultManager) protoRuleToFirewallRule(
 	}
 
 	var port *firewall.Port
-	if r.Port != "" {
+	if r.PortInfo != nil {
+		port = convertPortInfo(r.PortInfo)
```
pappz, line 272: Because the port type has been changed from int to uint16, should write validation in convertPortInfo. In the worst case, it can cause overflow.

lixmal: Let's add this in a follow-up PR with refactoring and more validation

PR #3240, opened 2025-01-27
Fix broken dependency locations

jameshilliard:
Should fix dependency downloads with `GOPROXY=direct`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -252,3 +252,7 @@ replace github.com/cloudflare/circl => github.com/cunicu/circl v0.0.0-2023080111
 replace github.com/pion/ice/v3 => github.com/netbirdio/ice/v3 v3.0.0-20240315174635-e72a50fcb64e
 
 replace github.com/libp2p/go-netroute => github.com/netbirdio/go-netroute v0.0.0-20240611143515-f59b0e1d3944
+
+replace go.opencensus.io => github.com/census-instrumentation/opencensus-go v0.24.0
+
+replace nhooyr.io/websocket => github.com/coder/websocket v1.8.11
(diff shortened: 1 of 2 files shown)


paralin:
Hi @lixmal , all, just pinging with a +1 as this is a blocker for us merging a Netbird package into the Buildroot distribution: https://buildroot.org

When we add it update Go packages we expect them to build with GOPROXY=direct

While this does not prevent the vulnerability discussed in the following link, it does help to indicate something has gone wrong somewhere when the code in the go proxy does not match the Git source. We also prefer to be able to build fetching directly from the git source instead of using the proxy on default. See; https://socket.dev/blog/malicious-package-exploits-go-module-proxy-caching-for-persistence and https://lore.kernel.org/all/CA+h8R2oJWZw_S08XXhjE--_So0UJohVLWSoj30PpR9z66x8U1A@mail.gmail.com/

Thanks for having a look at this!

lixmal:
Hi @jameshilliard and @paralin,
Thank you for submitting the PR and providing the reference links.
We have an ongoing effort to reduce our usage of replace directives, so adding new ones would go against our current goals.

Regarding the dependencies:

For `go.opencensus.io`: We don't have a direct import of this package - it's coming in as a transitive dependency through `google.golang.org/api/admin/directory/v1`. 
From what I've seen newer versions don't use that import anymore, so bumping `google.golang.org/api/admin/directory/v1` instead should be fine.

For nhooyr.io/websocket: Could you update the import statement directly in relay/client/dialer/ws/ws.go instead of using a replace directive?

Thanks for bringing this to our attention.

paralin:
@lixmal No problem, I will send a PR shortly!

lixmal:
Both PRs are merged, I'm going to clsoe this one

PR #3242, opened 2025-01-27
Feature/port forwarding client ingress

pappz:
## Describe your changes

Handle new forwarding rules message type and forward the DNAT rules to the underlying firewall.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/ingressgw/manager.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, fwdRule := range forwardRules {
+		id := fwdRule.ID()
+		if _, ok := h.rules[id]; ok {
+			delete(toDelete, id)
+			continue
+		}
+
+		rule, err := h.dnatFirewall.AddDNATRule(fwdRule)
+		if err != nil {
+			mErr = multierror.Append(mErr, fmt.Errorf("failed to add forward rule '%s': %v", fwdRule.String(), err))
```
lixmal: I think we can drop the `failed to ` prefix in wrapped errors, it just makes the message longer without adding any information

client/internal/ingressgw/manager_test.go
```diff
@@ -0,0 +1,244 @@
...
+import (
+	"fmt"
+	"net/netip"
+	"testing"
+
+	firewall "github.com/netbirdio/netbird/client/firewall/manager"
+)
+
+var (
+	_ firewall.Rule = &MocFwRule{}
```
lixmal:
```suggestion
	_ firewall.Rule = (*MocFwRule)(nil)
```
might be a bit better

client/internal/ingressgw/manager.go
pappz: Yes, but I am not sure in case of an error how can we figure out the original fwRule

client/internal/ingressgw/manager.go
```diff
@@ -0,0 +1,106 @@
...
+		log.Infof("forward rule has been added '%s'", fwdRule)
+		h.rules[id] = RulePair{
+			ForwardRule: fwdRule,
+			Rule:        rule,
+		}
+	}
+
+	// Remove deleted rules
+	for id, rulePair := range toDelete {
+		if err := h.dnatFirewall.DeleteDNATRule(rulePair.Rule); err != nil {
```
lixmal, line 71: We have a slight issue here. If the `rule, err := h.dnatFirewall.AddDNATRule(fwdRule)` above fails, we'll end up trying to remove that failed-to-add-rule here which will result in another error

pappz: I think this algorithm is correct, or I misunderstand how the underlying system works.

1. When h.dnatFirewall.AddDNATRule returns an error, the ingress gateway manager doesn't store the rule locally

2. As a consequence, when deletion is requested later, DeleteDNATRule isn't called (since there's no local record of the rule)

The logic makes sense from a data consistency perspective, if a rule failed to be added, we shouldn't track it locally since it doesn't exist in the actual firewall.

I made a unit test for these scenario. Could you double check it?

client/internal/ingressgw/manager.go
lixmal: Sorry, I mean just the `failed to` part

Basically:
```suggestion
			mErr = multierror.Append(mErr, fmt.Errorf("add forward rule '%s': %v", fwdRule.String(), err))
```

client/internal/ingressgw/manager.go
lixmal: All right, thank you

lixmal (approved):

PR #3252, opened 2025-01-30
[client, management] Support DNS Labels for Peer Addressing

hakansa:
## Describe your changes
Adding support for DNS labels that allow peers to be addressed using user-defined names.
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/dns/local.go
```diff
@@ -53,37 +55,55 @@ func (d *localResolver) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
...
+
+	value, found := d.records.Load(key)
 	if !found {
 		return nil
 	}
 
-	return record.(dns.RR)
+	records, ok := value.([]dns.RR)
+	if !ok {
+		return nil
```
lixmal, line 80: Let's add some log here. Something is really wrong if this fails.

management/domain/validate.go
```diff
@@ -0,0 +1,40 @@
...
+
+	var domainList List
+
+	for _, d := range domains {
+		d := strings.ToLower(d)
+
+		// handles length and idna conversion
+		punycode, err := FromString(d)
+		if err != nil {
+			return domainList, fmt.Errorf("failed to convert domain to punycode: %s: %v", d, err)
```
lixmal: Old bug:

```suggestion
			return domainList, fmt.Errorf("convert domain to punycode: %s: %w", d, err)
```

management/proto/management.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.26.0
-// 	protoc        v4.24.3
+// 	protoc        v5.29.3
```
lixmal: You might want to use the same protoc version to avoid issues.

client/server/server.go
```diff
@@ -404,6 +405,16 @@ func (s *Server) Login(callerCtx context.Context, msg *proto.LoginRequest) (*pro
 		s.latestConfigInput.BlockLANAccess = msg.BlockLanAccess
 	}
 
+	if msg.CleanDNSLabels {
+		inputConfig.DNSLabels = domain.List{}
+		s.latestConfigInput.DNSLabels = nil
+	} else if msg.DnsLabels != nil {
+		// TODO(hakan): discuss with @Viktor
```
lixmal: Is there anything left to discuss here?

client/internal/config.go
```diff
@@ -489,6 +494,14 @@ func (config *Config) apply(input ConfigInput) (updated bool, err error) {
 		}
 	}
 
+	if input.DNSLabels != nil && !reflect.DeepEqual(config.DNSLabels, input.DNSLabels) {
```
lixmal: We could use [slices.Equal](https://pkg.go.dev/slices#Equal) here

hakansa: done

client/internal/dns/local.go
hakansa: done

management/proto/management.pb.go
hakansa: done

client/server/server.go
hakansa: is it ok to assume that ValidateDomains won't return err here?

lixmal: Ah! I think we can use `domain.FromPunycodeList` here as we pass it as punycode from the cli already

lixmal (dismissed):

lixmal (dismissed):

lixmal (dismissed):

lixmal (approved):

pktiuk:
@hakansa     
Could you provide some screens or instruction showing how to use this feature? (TBH I don't know where I should look for this option)

client/internal/dns/local.go
```diff
@@ -44,11 +44,12 @@ func (d *localResolver) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
...
-	response := d.lookupRecord(r)
-	if response != nil {
-		replyMessage.Answer = append(replyMessage.Answer, response)
+	// lookup all records matching the question
+	records := d.lookupRecords(r)
+	if len(records) > 0 {
+		replyMessage.Rcode = dns.RcodeSuccess
+		replyMessage.Answer = append(replyMessage.Answer, records...)
 	} else {
 		replyMessage.Rcode = dns.RcodeNameError
```
CertainLach, line 54: This change has broken DNS for me.

I don't have DNS server added to netbird, thus netbird can only resolve subdomains of `netbird.local`

However, netbird sets search domain to `netbird.local`, thus any search first comes through netbird embedded DNS

E.g, when I do `resolvectl query google.com`, resolved first tries to ask netbird for `google.com.netbird.local`.

And there is a bug: in my case netbird can't lookup google.com, and thus returns `NXDOMAIN` (`RcodeNameError`) error, making `systemd-resolved` skip all other DNS servers and return the answer directly:

`google.com: resolve call failed: Could not resolve 'google.com', server or network returned error: SERVFAIL`

If netbird DNS server is not authoritative (i.e it has no DNS server defined in dashboard), it should instead return `RcodeSuccess` with zero records, so that `resolved` can continue the search using system DNS server.

The previous code was working, because it was responding with zero records in that case.

CertainLach: ...Or I misunderstood the DNS specification?
For now, I have added DNS in netbird dashboard and it works, but it doesn't seem right that system DNS is skipped now

Edit: In systemd-resolved, FallbackDNS only works when there is no other DNS server defined
Adding 1.1.1.1 as global DNS server and removing DNS servers from netbird dashboard have fixed the issue for me, I'm not sure why did it work out of the box on previous netbird version, but it doesn't seem right that netbird tries to handle all DNS by itself when no DNS server is defined in dashboard.
Looks like I have bisected it to the wrong place, but it definitely broken after merging this PR.

PR #3267, opened 2025-02-01
[misc] Replace Wiretrustee with Netbird

lixmal:
### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


release_files/install.sh
```diff
@@ -263,16 +263,16 @@ install_netbird() {
         add_aur_repo
     ;;
     brew)
-        # Remove Wiretrustee if it had been installed using Homebrew before
```
lixmal, line 266: Is this block correct or should it have been `netbird`?

release_files/install.sh
```diff
@@ -263,16 +263,16 @@ install_netbird() {
...
-        if brew ls --versions wiretrustee >/dev/null 2>&1; then
```
lixmal, line 267: Is this block correct or should it have been `netbird`?

pappz: I think the wiretrustee has been deprecated in brew. So this change looks correct.

pappz (approved):

PR #3268, opened 2025-02-01
[manager] ingress ports manager support

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/types/network.go
```diff
@@ -33,6 +34,52 @@ type NetworkMap struct {
 	OfflinePeers        []*nbpeer.Peer
 	FirewallRules       []*FirewallRule
 	RoutesFirewallRules []*RouteFirewallRule
+	ForwardingRules     []*ForwardingRule
+}
+
+func (nm *NetworkMap) Merge(other *NetworkMap) {
```
mlsmaycon, line 40: we might endup with duplicates on some key fields. We should check for unique on:
Peers
Routes
OfflinePeers

mlsmaycon (changes requested):

PR #3272, opened 2025-02-02
[client] use embedded root CA if system certpool is empty

hakansa:
…oots

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/auth/device_flow.go
+++ b/client/internal/auth/device_flow.go
@@ -2,6 +2,8 @@ package auth
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -11,7 +13,10 @@ import (
 	"strings"
 	"time"
 
+	log "github.com/sirupsen/logrus"
+
 	"github.com/netbirdio/netbird/client/internal"
+	"github.com/netbirdio/netbird/util/embeddedroots"
 )
 
 // HostedGrantType grant type for device flow on Hosted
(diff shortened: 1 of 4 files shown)


client/internal/auth/device_flow.go
```diff
@@ -56,6 +61,18 @@ func NewDeviceAuthorizationFlow(config internal.DeviceAuthProviderConfig) (*Devi
 	httpTransport := http.DefaultTransport.(*http.Transport).Clone()
 	httpTransport.MaxIdleConns = 5
 
+	certPool, err := x509.SystemCertPool()
+	if err != nil || certPool == nil {
+		log.Debug("System cert pool not available; falling back to embedded cert.")
```
lixmal: Let's log the `err` here and in the other places as well

util/embeddedroots/embeddedroots.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (r *roots) load() {
+	r.once.Do(func() {
+		p := x509.NewCertPool()
+		p.AppendCertsFromPEM([]byte(isrgRootX1RootPEM))
+		r.p = p
+	})
+}
+
+const isrgRootX1RootPEM = `-----BEGIN CERTIFICATE-----
```
lixmal: I'd suggest we use [embed](https://pkg.go.dev/embed) instead, that makes updating the file easier.

This is the RSA root, we might also want to add the ECDSA one.

Let's also add some comments about when it expires

client/internal/auth/device_flow.go
hakansa: done

util/embeddedroots/embeddedroots.go
hakansa: done

lixmal (dismissed):

PR #3287, opened 2025-02-05
[relay] Use new upstream for nhooyr.io/websocket package

paralin:
The nhooyr.io/websocket package was renamed to github.com/coder/websocket when the project was transferred to "coder" as the new maintainer.

Use the new import path and update go.mod and go.sum accordingly.

See #3240 cc @lixmal

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	github.com/c-robinson/iplib v1.0.3
 	github.com/caddyserver/certmagic v0.21.3
 	github.com/cilium/ebpf v0.15.0
+	github.com/coder/websocket v1.8.12
 	github.com/coreos/go-iptables v0.7.0
 	github.com/creack/pty v1.1.18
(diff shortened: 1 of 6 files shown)


paralin:
Hey @lixmal - the test that failed here:

https://github.com/netbirdio/netbird/blob/aed3d3d13d4998c1b64e51b193802d0f1feb5809/management/server/account_test.go#L3076

It seems to have failed because it took `51.64 ms/op, maximum 35.00 ms/op` - 16 ms longer than usual?

Does that code even use websocket at all, though? As far as I can see websocket is just used in relay.

Could this just be a test flake? Thanks!

lixmal:
It's just a flake.

Thanks for the PR, appreciate it!

lixmal (approved):

paralin:
@lixmal No problem, thank you for addressing this so quickly.

PR #3288, opened 2025-02-05
Update google.golang.org/api to latest

paralin:
Updating google.golang.org/api library to fix indirect dependency issues with older versions of OpenTelemetry.
    
See: #3240 

This PR also includes required updates to dependencies of google.golang.org/api separated into one commit per dependency.


paralin:
I can't see what dependency made snyk fail - any guidance? @lixmal  Thanks!

lixmal:
This is what it's complaining about
image

lixmal:
@mlsmaycon do you know how to fix these? I think we had that before

lixmal (approved):

PR #3291, opened 2025-02-06
[management] Support Extra DNS Labels for Peer Addressing

hakansa:
## Describe your changes
Support Extra DNS Labels for Peer Addressing

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/types/account.go
```diff
@@ -459,8 +459,23 @@ func (a *Account) GetPeersCustomZone(ctx context.Context, dnsDomain string) nbdn
 			TTL:   defaultTTL,
 			RData: peer.IP.String(),
 		})
-
 		sb.Reset()
+
+		for _, extraLabel := range peer.ExtraDNSLabels {
+			sb.Grow(len(extraLabel) + len(domainSuffix))
+			sb.WriteString(extraLabel)
```
lixmal, line 466: Do we ensure this doesn't already end with a dot?

management/domain/validate.go
```diff
@@ -38,3 +38,28 @@ func ValidateDomains(domains []string) (List, error) {
...
+// ValidateDomainsStrSlice checks if each domain in the list is valid
+func ValidateDomainsStrSlice(domains []string) ([]string, error) {
+	if len(domains) == 0 {
+		return nil, nil
+	}
+	if len(domains) > maxDomains {
+		return nil, fmt.Errorf("domains list exceeds maximum allowed domains: %d", maxDomains)
+	}
+
+	domainRegex := regexp.MustCompile(`^(?:\*\.)?(?:(?:xn--)?[a-zA-Z0-9_](?:[a-zA-Z0-9-_]{0,61}[a-zA-Z0-9])?\.)*(?:xn--)?[a-zA-Z0-9](?:[a-zA-Z0-9-_]{0,61}[a-zA-Z0-9])?$`)
```
lixmal, line 51: In this particular case, we might not want to allow wildcards. Clients won't be able to resolve these.

~~We should probably do the same on the client side, meaning we'd have to split the validate functions further. Not a big fan of that, any ideas?~~
Let's leave client like it is for now.

management/server/types/account.go
hakansa: domain.ValidateDomains throw an error if a label ends with a dot.

management/server/peer.go
```diff
@@ -285,6 +289,8 @@ func (am *DefaultAccountManager) UpdatePeer(ctx context.Context, accountID, user
 			inactivityExpirationChanged = true
 		}
 
+		//TODO(hakan): do we need to update extra dns labels here?
```
bcmmbaga: Let's remove this

bcmmbaga (changes requested):

bcmmbaga (approved):

PR #3296, opened 2025-02-06
[management] refactor auth

pnmcosta:
## Describe your changes
New Auth Manager that isolates user authentication (JWT's and PAT's) from handlers by passing a UserAuth struct in the Request's Context instead.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


management/server/http/middleware/auth_middleware.go
```diff
@@ -84,108 +58,106 @@ func (m *AuthMiddleware) Handler(h http.Handler) http.Handler {
...
 		default:
 			util.WriteError(r.Context(), status.Errorf(status.Unauthorized, "no valid authentication provided"), w)
 			return
 		}
-		claims := m.claimsExtractor.FromRequestContext(r)
-		//nolint
-		ctx := context.WithValue(r.Context(), nbContext.UserIDKey, claims.UserId)
-		//nolint
-		ctx = context.WithValue(ctx, nbContext.AccountIDKey, claims.AccountId)
-		h.ServeHTTP(w, r.WithContext(ctx))
```
pascal-fischer, line 109: We should not remove the accountID and userID from the context here. Or we have to update our logging hooks [here](https://github.com/netbirdio/netbird/blob/765aba2c1cc3e394a04df0746f01343913837322/formatter/hook.go#L96)
 which give us the tags on our log messages
```
2025-02-19T23:22:38+01:00 DEBG [context: HTTP, requestID: 6a5d41ba-c2c5-45c6-b28c-e2fdff0714d9, accountID: ctgn35cjcko7lol31111, userID: google-oauth2|111111111111111111111] management/server/account.go:1209: overriding JWT Domain and DomainCategory claims since single account mode is enabled
```

Keep in mind we have the same done for the grpc API as well. If we replace them we have to update the grpc interceptors [here](https://github.com/netbirdio/netbird/blob/e4b2e10d9bc2d16db72a550244652e17cf4f3e24/management/cmd/management.go#L369) and [here](https://github.com/netbirdio/netbird/blob/e4b2e10d9bc2d16db72a550244652e17cf4f3e24/management/cmd/management.go#L383) as well. As well as some other places directly in the code maybe.

Those logging hooks should probably be unified by now as they are the same everywhere.

management/server/http/testing/testing_tools/tools.go
```diff
@@ -115,11 +117,20 @@ func BuildApiBlackBoxWithDBState(t TB, sqlFile string, expectedPeerUpdate *serve
 		t.Fatalf("Failed to create manager: %v", err)
 	}
 
+	// @note this is required so that PAT's validate from store, but JTW's are mocked
```
pascal-fischer:
```suggestion
	// @note this is required so that PAT's validate from store, but JWT's are mocked
```

management/server/http/testing/testing_tools/tools.go
```diff
@@ -115,11 +117,20 @@ func BuildApiBlackBoxWithDBState(t TB, sqlFile string, expectedPeerUpdate *serve
...
+	authManager := auth.NewManager(store, "", "", "", "", []string{}, false)
+	mockedManager := &auth.MockManager{
```
pascal-fischer: to stay consistent this should be `authManagerMock`

pascal-fischer (changes requested):

management/server/http/middleware/auth_middleware.go
pnmcosta: Done, when setting the userAuth to the context, it ensures those are added too https://github.com/netbirdio/netbird/blob/refactor/auth/management/server/context/auth.go#L56

pascal-fischer (dismissed):

PR #3322, opened 2025-02-14
[client] ui update systray icons

hakansa:
## Describe your changes
updating Client UI systray icons with @karsa-mistmere's amazing work.
Ekran Resmi 2025-02-14 11 20 33
Ekran Resmi 2025-02-14 11 20 57

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


karsa-mistmere:
These aren't the designated icons, using the error status for disconnected seems like bad UX.

Also, using the fallback colors... works, but has much less than desirable contrast, can we explore dark mode support?

hakansa:
Yes, we can use base icon for disconnect status, wdyt @mlsmaycon @braginini.

Dark Mode support not an option for now since we're supporting all Major OS'es and it's an effortful task.

braginini:
> Yes, we can use base icon for disconnect status, wdyt @mlsmaycon @braginini.
> 

Yes, we need to support 4 icons for now (fallback mode): 

Screenshot 2025-02-14 at 10 27 02

braginini:
> > Dark Mode support not an option for now since we're supporting all Major OS'es and it's an effortful task.
> 
> Then maybe it would be best to either postpone this until dark mode support is an option, or use some other style that has better contrast, e.g.
> 
> image

This looks suitabke too!

karsa-mistmere:
See #3335, which I believe supersedes this PR.

PR #3324, opened 2025-02-14
[client] Update local interface addresses when gathering candidates

lixmal:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/stdnet/stdnet.go
+++ b/client/internal/stdnet/stdnet.go
@@ -54,6 +54,9 @@ func (n *Net) UpdateInterfaces() (err error) {
 // Interfaces returns a slice of interfaces which are available on the
 // system
 func (n *Net) Interfaces() ([]*transport.Interface, error) {
+	if err := n.UpdateInterfaces(); err != nil {
+		return nil, fmt.Errorf("update interfaces: %w", err)
+	}
 	return n.interfaces, nil
 }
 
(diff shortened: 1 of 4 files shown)


client/internal/stdnet/stdnet.go
```diff
@@ -48,12 +52,22 @@ func (n *Net) UpdateInterfaces() (err error) {
...
 	}
 	n.interfaces = n.filterInterfaces(allIfaces)
+
 	return nil
 }
 
 // Interfaces returns a slice of interfaces which are available on the
 // system
 func (n *Net) Interfaces() ([]*transport.Interface, error) {
+	if time.Since(n.lastUpdate) < updateInterval {
```
pappz, line 80: I think this function should be thread-safe because we use the instance of this struct in several parts of the code. The engine distribute it.

lixmal: Fixed

client/internal/stdnet/stdnet.go
```diff
@@ -43,18 +52,40 @@ func NewNet(disallowList []string) (*Net, error) {
...
+
+	if time.Since(n.lastUpdate) < updateInterval {
+		return slices.Clone(n.interfaces), nil
+	}
+
+	if err := n.updateInterfaces(); err != nil {
+		return nil, fmt.Errorf("update interfaces: %w", err)
+	}
+
+	n.lastUpdate = time.Now()
```
pappz: This line should be inside the `updateInterfaces` function

lixmal: Fixed

pappz (approved):

PR #3338, opened 2025-02-17
[client] Fix/pkg loss

pappz:
## Describe your changes

The Relayed connection setup is optimistic. It does not have any confirmation of an established end-to-end connection. Peers start sending WireGuard handshake packets immediately after the successful offer-answer handshake.
Meanwhile, for successful P2P connection negotiation, we change the WireGuard endpoint address, but this change does not trigger new handshake initiation. Because the peer switched from Relayed connection to P2P, the packets from the Relay server are dropped and must wait for the next WireGuard handshake via P2P.

To avoid this scenario, the relayed WireGuard proxy no longer drops the packets. Instead, it rewrites the source address to the new P2P endpoint and continues forwarding the packets.

We still have one corner case: if the Relayed server negotiation chooses a server that has not been used before. In this case, one side of the peer connection will be slower to reach the Relay server, and the Relay server will drop the handshake packet.

If everything goes well we should see exactly 5 seconds improvements between the WireGuard configuration time and the handshake time.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:


client/iface/wgproxy/udp/proxy.go
```diff
@@ -175,14 +228,20 @@ func (p *WGUDPProxy) proxyToLocal(ctx context.Context) {
...
+			p.pausedCond.L.Lock()
+			if p.paused {
+				p.pausedCond.Wait()
+				if !p.paused {
+					break
+				}
+				p.pausedCond.L.Unlock()
+				continue
+			}
+			break
```
lixmal: Would this be the same but much simpler?
```suggestion
		p.pausedCond.L.Lock()
		for p.paused {
			p.pausedCond.Wait()
```

client/iface/wgproxy/udp/rawsocket.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	return ipH, udpH, nil
+}
+
+func prepareSenderRawSocket() (net.PacketConn, error) {
+	// Create a raw socket.
+	fd, err := syscall.Socket(syscall.AF_INET, syscall.SOCK_RAW, syscall.IPPROTO_RAW)
+	if err != nil {
+		return nil, fmt.Errorf("creating raw socket failed: %w", err)
```
lixmal: I would personally drop all `failed` parts from the errors. Since it's already an error it's clear something went wrong and upstream will clarify when printing the log.
But I'm aware this reused code.

pappz: I merged this code with the reused code.
I am not sure the `syscall.Socket` responded with proper verbose code and maybe will be hard to detect where come from the original error. But I will keep in mind if I see an error message here and it is too long.

lixmal (approved):

client/iface/wgproxy/rawsocket/rawsocket.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	// Bind the socket to the "lo" interface.
+	err = syscall.SetsockoptString(fd, syscall.SOL_SOCKET, syscall.SO_BINDTODEVICE, "lo")
+	if err != nil {
+		return nil, fmt.Errorf("binding to lo interface failed: %w", err)
+	}
+
+	// Set the fwmark on the socket.
+	err = nbnet.SetSocketOpt(fd)
```
lixmal, line 34: This is redundant if we use the socket with loopback addresses only, but it won't introduce issues either.

lixmal (approved):

PR #3362, opened 2025-02-21
[client] Replace string to netip.Prefix

pappz:
## Describe your changes

The allowedIPs is a string type and it causes inconsistency. It could be:
- "100.108.186.247"
- "100.108.186.247/32"
- "100.108.186.247/32,100.108.186.2/32"

In this PR convert allowed IP variable type to netip.Prefix

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -526,15 +526,18 @@ func (e *Engine) modifyPeers(peersUpdate []*mgmProto.RemotePeerConfig) error {
 	var modified []*mgmProto.RemotePeerConfig
 	for _, p := range peersUpdate {
 		peerPubKey := p.GetWgPubKey()
-		if allowedIPs, ok := e.peerStore.AllowedIPs(peerPubKey); ok {
-			if allowedIPs != strings.Join(p.AllowedIps, ",") {
-				modified = append(modified, p)
-				continue
-			}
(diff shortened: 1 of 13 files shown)


client/internal/engine.go
```diff
@@ -1812,3 +1826,37 @@ func getInterfacePrefixes() ([]netip.Prefix, error) {
...
+	})
+
+	// Compare sorted slices
+	for i := range list1 {
+		if list1[i] != list2Prefixes[i] {
+			return false
+		}
+	}
+	return true
+}
```
hakansa: i assume that ordering is not important when comparing the lists, so we can use a freq map.

```go
func compareNetIPLists(list1 []netip.Prefix, list2 []string) bool {
    if len(list1) != len(list2) {
        return false
    }

    freq := make(map[string]int, len(list1))
    for _, p := range list1 {
        freq[p.String()]++
    }

    for _, s := range list2 {
        p, err := netip.ParsePrefix(s)
        if err != nil {
            return false // invalid prefix in list2.
        }
        key := p.String()
        if freq[key] == 0 {
            return false
        }
        freq[key]--
    }

    // all counts should be zero if lists are equal.
    for _, count := range freq {
        if count != 0 {
            return false
        }
    }
    return true
}
```

hakansa (approved):

PR #3379, opened 2025-02-24
[client] Feature/lazy connection

pappz:
## Describe your changes
With the lazy connection feature, the peer will connect to target peers on-demand. The trigger can be any IP traffic.

This feature can be enabled with the NB_ENABLE_EXPERIMENTAL_LAZY_CONN environment variable.

When the engine receives a network map, it binds a free UDP port for every remote peer, and the system configures WireGuard endpoints for these ports. When traffic appears on a UDP socket, the system removes this listener and starts the peer connection procedure immediately.

### Key changes
- Fix slow `netbird status -d` command
- Move from engine.go file to conn_mgr.go the peer connection related code
- Refactor the iface interface usage and moved interface file next to the engine code
- Add new command line flag and UI option to enable feature
- The peer.Conn struct is reusable after it has been closed.
- Change connection states

### Connection states
Idle: The peer is not attempting to establish a connection. This typically means it's in a lazy state or the remote peer is expired.

Connecting: The peer is actively trying to establish a connection. This occurs when the peer has entered an active state and is continuously attempting to reach the remote peer.

Connected: A successful peer-to-peer connection has been established and communication is active.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


pappz:
> @pappz let me know if you need help testing. I have some IoT devices were I had to disable Netbird again, as it was consuming too much data, so I'm really interesting in helping this PR along :)

@Lauszus, thank you!
The feature is ready for review. In the meantime, I am working on bug fixes based on feedback. I would like to highlight that this feature also requires an update to the Management server; without it, we cannot establish a backward-compatible lazy connection. By default, this feature is disabled. You can enable it using an environment variable or command-line options.

Lauszus:
> > @pappz let me know if you need help testing. I have some IoT devices were I had to disable Netbird again, as it was consuming too much data, so I'm really interesting in helping this PR along :)
> 
> 
> 
> @Lauszus, thank you!
> 
> The feature is ready for review. In the meantime, I am working on bug fixes based on feedback. I would like to highlight that this feature also requires an update to the Management server; without it, we cannot establish a backward-compatible lazy connection. By default, this feature is disabled. You can enable it using an environment variable or command-line options.

No problem. I can enable it and test it on my side. I have ~50 devices where I can try it on.

client/cmd/root.go
```diff
@@ -179,6 +181,7 @@ func init() {
 	upCmd.PersistentFlags().BoolVar(&rosenpassPermissive, rosenpassPermissiveFlag, false, "[Experimental] Enable Rosenpass in permissive mode to allow this peer to accept WireGuard connections without requiring Rosenpass functionality from peers that do not have Rosenpass enabled.")
 	upCmd.PersistentFlags().BoolVar(&serverSSHAllowed, serverSSHAllowedFlag, false, "Allow SSH server on peer. If enabled, the SSH server will be permitted")
 	upCmd.PersistentFlags().BoolVar(&autoConnectDisabled, disableAutoConnectFlag, false, "Disables auto-connect feature. If enabled, then the client won't connect automatically when the service starts.")
+	upCmd.PersistentFlags().BoolVar(&lazyConnEnabled, enableLazyConnectionFlag, false, "[Experimental] Enable lazy connection feature. If enabled, the client will establish connections on-demand.")
```
lixmal:
```suggestion
	upCmd.PersistentFlags().BoolVar(&lazyConnEnabled, enableLazyConnectionFlag, false, "[Experimental] Enable the lazy connection feature. If enabled, the client will establish connections on-demand.")
```

client/internal/lazyconn/activity/listener.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (d *Listener) ReadPackets() {
+	for {
+		buffer := make([]byte, 10)
+		n, remoteAddr, err := d.conn.ReadFromUDP(buffer)
+		if err != nil {
+			if d.isClosed {
+				d.peerCfg.Log.Debugf("exit from activity listener")
+			} else {
+				d.peerCfg.Log.Errorf("failed to read from activity listener: %s", err)
```
lixmal, line 53: Should we not retry (continue) reading if there's an error != closed? Maybe with a ctx check in-between

client/internal/lazyconn/activity/listener.go
```diff
@@ -0,0 +1,81 @@
...
+	cfg.Log.Infof("created activity listener: %s", addr.String())
+	return d, nil
+}
+
+func (d *Listener) ReadPackets() {
+	for {
+		buffer := make([]byte, 10)
+		n, remoteAddr, err := d.conn.ReadFromUDP(buffer)
+		if err != nil {
+			if d.isClosed {
```
lixmal: This field seems to be not thread safe. We access it here without any lock here or in the caller

client/internal/lazyconn/activity/listener.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, err
+	}
+	d.done.Lock()
+	cfg.Log.Infof("created activity listener: %s", addr.String())
+	return d, nil
+}
+
+func (d *Listener) ReadPackets() {
+	for {
+		buffer := make([]byte, 10)
```
lixmal: Why allocate a buffer of 10 when we only need 4? Should we allocate the buffer outside the loop to reduce allocations?

client/internal/lazyconn/activity/manager.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	conn, addr, err := m.portGenerator.newConn()
+	if err != nil {
+		return fmt.Errorf("failed to bind activity listener: %v", err)
+	}
+
+	listener, err := NewListener(m.wgIface, peerCfg, conn, addr)
+	if err != nil {
+		return err
```
lixmal, line 44: Let's add some err context here as well

client/internal/lazyconn/activity/allocator.go
```diff
@@ -0,0 +1,55 @@
...
+func newPortAllocator() *portAllocator {
+	return &portAllocator{
+		nextFreePort: 65535,
+	}
+}
+
+func (p *portAllocator) newConn() (*net.UDPConn, *net.UDPAddr, error) {
+	for i := 0; i < retryLimit; i++ {
+		addr := &net.UDPAddr{
+			Port: p.nextPort(),
```
lixmal: Why not use port 0 and let the OS decide? That should be safer and more performant.

client/internal/engine.go
```diff
@@ -1878,7 +1912,33 @@ func (e *Engine) updateForwardRules(rules []*mgmProto.ForwardingRule) error {
...
+	return forwardingRules, nberrors.FormatErrorOrNil(merr)
+}
+
+func (e *Engine) toExcludedLazyPeers(routes []*route.Route, rules []firewallManager.ForwardRule, peers []*mgmProto.RemotePeerConfig) []string {
+	excludedPeers := make([]string, 0)
+	for _, r := range routes {
+		if r.Peer == "" {
+			continue
+		}
+		log.Infof("excluded peer from lazy connection: %s", r.Peer)
```
lixmal: Let's add a reason here e.g. r.NetString()

client/internal/lazyconn/activity/listen_ip.go
```diff
@@ -0,0 +1,9 @@
+//go:build !linux || android
+
+package activity
+
+import "net"
+
+var (
+	listenIP = net.ParseIP("127.0.0.1")
```
lixmal: nitpick: better use `net.MustParseIP` or hardcode as `net.IP{127, 0, 0, 1}`

client/internal/lazyconn/activity/manager.go
```diff
@@ -0,0 +1,103 @@
...
+	defer m.mu.Unlock()
+
+	if _, ok := m.peers[peerCfg.PeerConnID]; ok {
+		log.Warnf("activity listener already exists for: %s", peerCfg.PublicKey)
+		return nil
+	}
+
+	conn, addr, err := m.portGenerator.newConn()
+	if err != nil {
+		return fmt.Errorf("failed to bind activity listener: %v", err)
```
lixmal:
```suggestion
		return fmt.Errorf("bind activity listener: %v", err)
```

lixmal (commented):
PR description doesn't match code anymore (env vars)

client/internal/lazyconn/activity/listener.go
pappz: With it, we risk a high load loop. If it fails to read from UDP once, what makes us think it will succeed the second time?

lixmal: > With it, we risk a high load loop. 

True

> If it fails to read from UDP once, what makes us think it will succeed the second time?

It could be a temporary error, e.g. OS ran out of file descriptors.
But yeah, in general this should be rare. Maybe we can add a comment here explaining the implication for the time being?

lixmal (approved):

lixmal (dismissed):

lixmal (approved):

lixmal (approved):

PR #3435, opened 2025-03-04
[client] Fix engine restart

pappz:
## Describe your changes

- Refactor the network monitoring to handle one event and it after return
- In the engine restart cancel the upper layer context and the responsibility of the engine stop will be the upper layer
- Before triggering a restart, the engine checks whether the state is already down. This helps avoid unnecessary delayed network restart events.

Originally, it processed multiple events. Now, it is a one-time usage structure. Essentially, the effects remain the same but without extra goroutines or timers.

### Note
Keep in mind that the restart function only restarts the engine implicitly, as it is already handled by the connect backoff code.
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


lixmal (approved):
This line can go, otherwise it will be repeated without any use:
https://github.com/netbirdio/netbird/pull/3435/files#diff-f6717c1eaf9f930621027397547906b22e92e813e0cce5ca2dd676ef3fa26248R30

pappz:
- Refactor the network monitoring to handle one event and it after return
- In the engine restart cancel the upper layer context and the responsibility of the engine stop will be the upper layer
- Before triggering a restart, the engine checks whether the state is already down. This helps avoid unnecessary delayed network restart events.

PR #3444, opened 2025-03-05
[management] permission manager validate account access

pnmcosta:
### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


management/server/account.go
```diff
@@ -597,9 +602,10 @@ func (am *DefaultAccountManager) DeleteAccount(ctx context.Context, accountID, u
 		return err
 	}
 
-	if !user.HasAdminPower() {
-		return status.Errorf(status.PermissionDenied, "user is not allowed to delete account")
-	}
+	// @note not necessary, below it explicitly checks for Owner role
+	// if !user.HasAdminPower() {
+	// 	return status.Errorf(status.PermissionDenied, "user is not allowed to delete account")
+	// }
```
pascal-fischer: True, this can be removed

management/server/account.go
```diff
@@ -1144,6 +1150,7 @@ func (am *DefaultAccountManager) GetAccountIDFromUserAuth(ctx context.Context, u
 		return accountID, user.Id, nil
 	}
 
+	// @note, this can remain cause above we explicitly early return if auth id for a child account
```
pascal-fischer: Would it still make sense to replace this with the manager? Just to have a unified approach of checking permissions throughout the project? (also intending to move the owner role check and admin check to this manager). This way we will also keep the returned errors consistent throughout the project

management/server/user.go
```diff
@@ -338,15 +338,16 @@ func (am *DefaultAccountManager) CreatePAT(ctx context.Context, accountID string
...
+	if err := am.permissionsManager.ValidateAccountAccess(ctx, accountID, initiatorUser); err != nil {
+		return nil, err
 	}
 
 	targetUser, err := am.Store.GetUserByUserID(ctx, store.LockingStrengthShare, targetUserID)
 	if err != nil {
 		return nil, err
 	}
 
+	// @todo how to handle this case, PAT can only be created own user?
```
pascal-fischer: Wouldn't is suffice to keep checking if the user the token is created for is the same user that is creating the token? do you intend this logic to be moved to permissions manager as well?

management/server/user.go
```diff
@@ -967,6 +971,8 @@ func (am *DefaultAccountManager) DeleteRegularUsers(ctx context.Context, account
 		return err
 	}
 
+	// @todo maybe add ValidateAccountPermission?
```
pascal-fischer: yes

management/server/account.go
pnmcosta: yeah good call, can't see why not, since we are indeed returning early if it's a child account, there's no reason we couldn't use `ValidateAccountAccess` here too.

management/server/user.go
pnmcosta: Note is misleading, as I misread the check below. Users can currently create PAT's for each other as long as they have admin rights.

pascal-fischer: No this should not be. Users should only be able to (if they are admins) to create PATs for their own user, not for each other. `initiatorUserID != targetUserID` should take care of that

pnmcosta: Ok, so there's nothing to change here.

mlsmaycon:
@pnmcosta can you handle sonar findings?

pascal-fischer (approved):

PR #3445, opened 2025-03-05
[client] add resource id fields to netflow events

hakansa:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/peer/route.go
```diff
@@ -0,0 +1,100 @@
...
+	// check resolved ip's first
+	resId, ok := r.resolvedIPs.Load(src)
+	if ok {
+		srcResourceID = resId.(string)
+	} else {
+		resId, ok := r.resolvedIPs.Load(dst)
+		if ok {
+			dstResourceID = resId.(string)
+		}
+	}
```
lixmal: We can also include this part in the direction switch. Since routing peers don't currently know about resolved IPs, we can stick to the perspective of routing clients for now.

egress -> resolve dst IP
ingress -> resolve src IP

client/internal/peer/route.go
```diff
@@ -0,0 +1,100 @@
...
+				}
+
+				if srcResourceID != "" && dstResourceID != "" {
+					return false
+				}
+
+				return true
+			})
+		}
+	}
```
lixmal: I think this part is not quite correct. In both direction cases, we need to traverse both local/remote maps. The direction defines whether src or dst needs to be looked up in the respective map.

client/internal/peer/status.go
```diff
@@ -340,11 +351,28 @@ func (d *Status) RemovePeerStateRoute(peer string, route string) error {
...
 	return nil
 }
 
+// CheckRoutes checks if the source and destination addresses are within the same route
+// and returns the resource ID of the route that contains the addresses
+func (d *Status) CheckRoutes(src, dst netip.Addr, direction nftypes.Direction) (srcResId string, dstResId string) {
+	if d == nil {
+		return
+	}
+
```
lixmal: We should check here if any of the IPs are part of the netbird network, then we can skip the route lookup for that IP. For efficiency and avoiding wrong matches (e.g. Netbird IP matching to a default route)

lixmal (changes requested):

client/internal/netflow/logger/logger.go
```diff
@@ -21,15 +24,29 @@ type Logger struct {
...
 	Store          types.Store
 }
 
-func New(ctx context.Context) *Logger {
+func New(ctx context.Context, statusRecorder *peer.Status, wgIfaceIP wgaddr.Address) *Logger {
+
+	var addr netip.Addr
+	if wgIfaceIP.IP != nil {
+		var err error
+		addr, err = netip.ParseAddr(wgIfaceIP.IP.String())
```
lixmal: We should compare the event src/dst to the whole `wgIfaceIP.Network`, not just the IP, especially for Egress conns

lixmal (approved):

PR #3472, opened 2025-03-09
[client] Ignore candidates that are part of the the wireguard subnet

lixmal:
## Describe your changes

Candidates with an IP from the overlay network are never valid, hence we block it in the mux. Not blocking it in `OnCandidate` instead since this wouldn't catch `prflx` candidates.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/iface/bind/udp_mux_universal.go
+++ b/client/iface/bind/udp_mux_universal.go
@@ -165,8 +165,8 @@ func (u *udpConn) performFilterCheck(addr net.Addr) error {
 	}
 
 	if u.address.Network.Contains(a.AsSlice()) {
-		log.Warnf("Address %s is part of the wireguard network %s, refusing to write", addr, u.address)
-		return fmt.Errorf("address %s is part of the wireguard network %s, refusing to write", addr, u.address)
+		log.Warnf("Address %s is part of the NetBird network %s, refusing to write", addr, u.address)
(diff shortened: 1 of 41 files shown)


client/iface/bind/udp_mux_universal.go
```diff
@@ -159,6 +164,11 @@ func (u *udpConn) performFilterCheck(addr net.Addr) error {
 		return nil
 	}
 
+	if u.address.Network.Contains(a.AsSlice()) {
+		log.Warnf("Address %s is part of the wireguard network %s, refusing to write", addr, u.address)
```
mlsmaycon, line 168:
```suggestion
		log.Warnf("Address %s is part of the NetBird network %s, refusing to write", addr, u.address)
```

client/iface/bind/udp_mux_universal.go
```diff
@@ -159,6 +164,11 @@ func (u *udpConn) performFilterCheck(addr net.Addr) error {
...
+		return fmt.Errorf("address %s is part of the wireguard network %s, refusing to write", addr, u.address)
+	}
```
mlsmaycon:
```suggestion
	if u.address.Network.Contains(a.AsSlice()) {
		log.Warnf("Address %s is part of the NetBird network %s, refusing to write", addr, u.address)
		return fmt.Errorf("address %s is part of the NetBird network %s, refusing to write", addr, u.address)
	}
```

PR #3496, opened 2025-03-13
[client] Fix macos privacy warning when checking static info

hurricanehrndz:
## Describe your changes
Detect platform only when running as root or when not running on Darwin

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/system/static_info.go
+++ b/client/system/static_info.go
@@ -16,12 +16,6 @@ var (
 	once       sync.Once
 )
 
-func init() {
-	go func() {
-		_ = updateStaticInfo()
-	}()
-}
-
 func updateStaticInfo() StaticInfo {
 	once.Do(func() {
 		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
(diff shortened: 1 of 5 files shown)


hurricanehrndz:
I have tested this fix

https://github.com/user-attachments/assets/e1c6f7ed-c122-4816-b6d1-cec82c8be1b2

client/system/static_info.go
```diff
@@ -27,20 +29,28 @@ func updateStaticInfo() StaticInfo {
...
-		go func() {
-			staticInfo.Environment.Cloud = detect_cloud.Detect(ctx)
-			wg.Done()
-		}()
-		go func() {
-			staticInfo.Environment.Platform = detect_platform.Detect(ctx)
-			wg.Done()
-		}()
+		// run on all OSes, on Darwin run only when executing as root
+		if runtime.GOOS != "darwin" || isRoot() {
```
lixmal: There's one small caveat here. If we run in netstack mode it might not be under root. While this isn't very common it's a possibility. Any idea for a low effort fix?

hurricanehrndz: After discussing it Maycon, I have updated the PR

mlsmaycon (approved):
looks good @hurricanehrndz. I will move the login check to after the console check in another Pr

PR #3538, opened 2025-03-18
[client] fix netflow resource id parse & loggin

hakansa:
…and consistency

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/netflow/manager.go
+++ b/client/internal/netflow/manager.go
@@ -32,7 +32,7 @@ type Manager struct {
 	publicKey      []byte
 }
 
-// NewManager creates a new netflow manager
+// NewManager creates a new netnetflow manager
 func NewManager(ctx context.Context, iface nftypes.IFaceMapper, publicKey []byte, statusRecorder *peer.Status) *Manager {
 	var ipNet net.IPNet
 	if iface != nil {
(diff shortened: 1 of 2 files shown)


client/internal/netflow/manager.go
```diff
@@ -130,10 +134,24 @@ func (m *Manager) Update(update *nftypes.FlowConfig) error {
...
 
-	return m.disableFlow()
+	log.Infof("netflow manager disabled; stopping netflow manager")
+	err := m.disableFlow()
+	if err != nil {
+		log.Errorf("failed to disable netflow manager: %v", err)
+	} else {
+		log.Infof("netflow manager disabled successfully")
+	}
+	return err
```
pascal-fischer: Maybe it's just my personal preference but to not pollute logs I would drop the else cases. The info on what we are doing and, in case of an error, a second log should be enough. No need for an additional success log

client/internal/netflow/manager.go
```diff
@@ -130,10 +134,20 @@ func (m *Manager) Update(update *nftypes.FlowConfig) error {
 	m.logger.UpdateConfig(update.DNSCollection, update.ExitNodeCollection)
 
 	if update.Enabled {
-		return m.enableFlow(previous)
+		log.Infof("netflow manager enabled; starting netflow manager")
+		err := m.enableFlow(previous)
```
lixmal: I think we might want to keep bubbling up the error so engine restarts if we cannot make flows work. What do you think @mlsmaycon?

mlsmaycon: sounds good

hakansa: done

PR #3596, opened 2025-03-28
[client] Stop flow grpc receiver properly

lixmal:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/netflow/manager.go
+++ b/client/internal/netflow/manager.go
@@ -27,18 +27,18 @@ type Manager struct {
 	logger         nftypes.FlowLogger
 	flowConfig     *nftypes.FlowConfig
 	conntrack      nftypes.ConnTracker
-	ctx            context.Context
 	receiverClient *client.GRPCClient
 	publicKey      []byte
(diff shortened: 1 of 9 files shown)


client/internal/netflow/manager.go
```diff
@@ -96,17 +83,50 @@ func (m *Manager) enableFlow(previous *nftypes.FlowConfig) error {
...
+	}
+	log.Infof("flow client configured to connect to %s", m.flowConfig.URL)
+
+	m.receiverClient = flowClient
+
+	if m.cancel != nil {
+		m.cancel()
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
hakansa, line 105: with suggested change, cancel would be triggerred if m.ctx is done
```suggestion
	ctx, cancel := context.WithCancel(m.ctx)
```

lixmal: There's no m.ctx anymore

hakansa (approved):

PR #3604, opened 2025-03-29
[client] Force new user login on PKCE auth in CLI (behavior change)

mlsmaycon:
## Describe your changes
With this change, browser sessions won't be considered for CLI authentication, and credentials will be requested.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/auth/pkce_flow.go
+++ b/client/internal/auth/pkce_flow.go
@@ -99,6 +99,7 @@ func (p *PKCEAuthorizationFlow) RequestAuthInfo(ctx context.Context) (AuthFlowIn
 		oauth2.SetAuthURLParam("code_challenge_method", "S256"),
 		oauth2.SetAuthURLParam("code_challenge", codeChallenge),
 		oauth2.SetAuthURLParam("audience", p.providerConfig.Audience),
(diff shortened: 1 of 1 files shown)


avg-I:
This change introduced some annoying behavior for me.

I understand that requiring a new login (ignoring existing SSO login) may be a good feature.
But some details are annoying.

The VPN that I am using has login.microsoftonline.com as the token and authorization endpoint.
The SSO is configured to work with MS Authenticator mobile app.
So  my usual SSO login procedure is:
- I am presented with a login page where I see a list of my accounts (I have just one);
- I click the account;
- I am given a number to use with the Authenticator app;
- the Authenticator app activates on my phone;
- I enter the number and then my fingerprint;
- I am logged in.

But with netbird it's _almost_ the same except for one detail: the initial login page requires me to type in (or copy+paste, of course) my full email address. There is no way to just click my account. Afterwards, the rest of the process is the same.

It's having to always type in the email that gets annoying.

I am not sure if it is just something on the Microsoft side when it sees prompt=login or if there is something else (or something missing) that tells the server to prompt for an email instead of listing known accounts for my browser.

PR #3623, opened 2025-04-04
[client] Mark netbird data plane traffic to identify interface traffic correctly

lixmal:
## Describe your changes

This change enables site2site traffic flow support in kernel mode.

Packets in and out of the NetBird interface mark the associated connection with the respective fwmark (per direction).
This is then used to filter events from the conntrack socket and determine the conn direction.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/firewall/iptables/router_linux.go
+++ b/client/firewall/iptables/router_linux.go
@@ -438,6 +438,7 @@ func (r *router) setupDataPlaneMark() error {
 	var merr *multierror.Error
 	preRule := []string{
 		"-i", r.wgIface.Name(),
+		"-m", "conntrack", "--ctstate", "NEW",
 		"-j", "CONNMARK", "--set-mark", fmt.Sprintf("%#x", nbnet.DataPlaneMarkIn),
 	}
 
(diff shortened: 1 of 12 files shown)


client/firewall/iptables/router_linux.go
```diff
@@ -423,6 +433,55 @@ func (r *router) createContainers() error {
...
+// setupDataPlaneMark configures the fwmark for the data plane
+func (r *router) setupDataPlaneMark() error {
+	var merr *multierror.Error
+	preRule := []string{
+		"-i", r.wgIface.Name(),
+		"-j", "CONNMARK", "--set-mark", fmt.Sprintf("%#x", nbnet.DataPlaneMarkIn),
+	}
+
+	if err := r.iptablesClient.AppendUnique(tableMangle, chainPREROUTING, preRule...); err != nil {
+		merr = multierror.Append(merr, fmt.Errorf("add mangle prerouting rule: %w", err))
```
hakansa, line 446: why we dont return here if there's an error?

lixmal: I was thinking that partial marking would be better than nothing, but in practice it won't make a difference

hakansa (approved):

PR #3644, opened 2025-04-08
[management] Buffer updateAccountPeers calls

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/peer.go
```diff
@@ -1226,6 +1228,28 @@ func (am *DefaultAccountManager) UpdateAccountPeers(ctx context.Context, account
...
+
+	if !lock.TryLock() {
+		return
+	}
+
+	go func() {
+		intervalStr := os.Getenv("PEER_UPDATE_INTERVAL_MS")
+		interval, err := strconv.Atoi(intervalStr)
+		if err != nil {
+			interval = 1000
```
mlsmaycon: let's default to 1ms?

management/server/peer.go
```diff
@@ -1226,6 +1228,28 @@ func (am *DefaultAccountManager) UpdateAccountPeers(ctx context.Context, account
...
+func (am *DefaultAccountManager) BufferUpdateAccountPeers(ctx context.Context, accountID string) {
+	mu, _ := am.accountUpdateLocks.LoadOrStore(accountID, &sync.Mutex{})
+	lock := mu.(*sync.Mutex)
+
+	if !lock.TryLock() {
+		return
+	}
+
+	go func() {
+		intervalStr := os.Getenv("PEER_UPDATE_INTERVAL_MS")
```
mlsmaycon: should we go with a pre check for the env?

management/server/peer.go
```diff
@@ -1226,6 +1228,28 @@ func (am *DefaultAccountManager) UpdateAccountPeers(ctx context.Context, account
...
+		interval, err := strconv.Atoi(intervalStr)
+		if err != nil {
+			interval = 1000
+		}
+		time.Sleep(time.Duration(interval) * time.Millisecond)
+		am.UpdateAccountPeers(ctx, accountID)
+
+		am.accountUpdateLocks.Delete(accountID)
```
mlsmaycon: don't think we need to delete the sync here. maybe when deleting an account?

mlsmaycon (changes requested):

mlsmaycon (approved):

PR #3651, opened 2025-04-10
[client] Refactor DNSForwarder to improve  handle wildcard domain resource id matching

hakansa:
[client] Refactor DNSForwarder to improve  handle wildcard domain resource id matching

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/dnsfwd/forwarder.go
+++ b/client/internal/dnsfwd/forwarder.go
@@ -3,6 +3,7 @@ package dnsfwd
 import (
 	"context"
 	"errors"
+	"math"
 	"net"
 	"net/netip"
 	"strings"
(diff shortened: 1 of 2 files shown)


client/internal/dnsfwd/forwarder.go
```diff
@@ -204,6 +203,31 @@ func (f *DNSForwarder) addIPsToResponse(resp *dns.Msg, domain string, ips []neti
 	}
 }
 
+func (f *DNSForwarder) getResIdForDomain(domain string) string {
```
lixmal, line 207: Let's add a unit test for this one

hakansa: done

client/internal/dnsfwd/forwarder_test.go
```diff
@@ -0,0 +1,72 @@
...
+			storedMappings: map[string]string{"*.example.com": "res4"},
+			queryDomain:    "foo.notexample.com",
+			expectedResId:  "",
+		},
+		{
+			name:           "Non-wildcard pattern does not match subdomain",
+			storedMappings: map[string]string{"example.com": "res5"},
+			queryDomain:    "foo.example.com",
+			expectedResId:  "",
+		},
```
lixmal, line 50: Let's also add test cases with multiple stored mappings. That copilot suggestion with the most specific match also makes sense, we can add that case

lixmal (approved):

PR #3653, opened 2025-04-10
[client] add byte counters & ruleID for routed traffic on userspace

hakansa:
## Describe your changes
[client] add byte counters for routed traffic on userspace
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/firewall/uspfilter/forwarder/icmp.go
```diff
@@ -61,18 +61,19 @@ func (f *Forwarder) handleICMP(id stack.TransportEndpointID, pkt stack.PacketBuf
 
 	// For Echo Requests, send and handle response
 	if header.ICMPv4Type(icmpType) == header.ICMPv4Echo {
-		f.handleEchoResponse(icmpHdr, conn, id)
-		f.sendICMPEvent(nftypes.TypeEnd, flowID, id, icmpType, icmpCode)
+		rxBytes := len(payload)
```
lixmal: this should be the len of the whole `pkt`
```suggestion
		rxBytes := pkt.Size()
```

client/firewall/uspfilter/forwarder/udp.go
```diff
@@ -212,49 +211,53 @@ func (f *Forwarder) handleUDP(r *udp.ForwarderRequest) {
...
+	wg.Wait()
+
+	if outboundErr != nil && !isClosedError(outboundErr) {
+		f.logger.Error("proxyUDP: copy error (outbound->inbound): %v", outboundErr)
+	}
+	if inboundErr != nil && !isClosedError(inboundErr) {
+		f.logger.Error("proxyUDP: copy error (inbound->outbound): %v", inboundErr)
+	}
+
+	f.logger.Trace("proxyUDP: bytes transferred for connection %v: inbound bytes: %d, outbound bytes: %d", epID(id), rxBytes, txBytes)
```
lixmal: This kinda lost the message that this UDP conn is done, like here

```
f.logger.Trace("forwarder: tearing down UDP connection %v due to context done", epID(id))
```

Let's format it close to what peer 2 peer conns look like:

```
			t.logger.Trace("Removed UDP connection %s (timeout) [in: %d Pkts/%d B, out: %d Pkts/%d B]",
```

the bracket part

client/firewall/uspfilter/forwarder/tcp.go
```diff
@@ -65,48 +65,51 @@ func (f *Forwarder) handleTCP(r *tcp.ForwarderRequest) {
...
 
-	select {
-	case <-ctx.Done():
-		f.logger.Trace("forwarder: tearing down TCP connection %v due to context done", epID(id))
-		return
-	case err := <-errChan:
-		if err != nil && !isClosedError(err) {
-			f.logger.Error("proxyTCP: copy error: %v", err)
-		}
-		f.logger.Trace("forwarder: tearing down TCP connection %v", epID(id))
```
lixmal, line 104: This log is lost

client/firewall/uspfilter/forwarder/tcp.go
```diff
@@ -65,48 +65,51 @@ func (f *Forwarder) handleTCP(r *tcp.ForwarderRequest) {
...
-		_, err := io.Copy(inConn, outConn)
-		errChan <- err
+		defer wg.Done()
+		var n int64
+		n, errOutToIn = io.Copy(inConn, outConn)
+		bytesFromOutToIn = n
 	}()
 
-	select {
-	case <-ctx.Done():
```
lixmal, line 97: How do we handle context cancellation now?

hakansa: it was missing, added.

client/firewall/uspfilter/forwarder/tcp.go
```diff
@@ -65,48 +66,69 @@ func (f *Forwarder) handleTCP(r *tcp.ForwarderRequest) {
...
 	ctx, cancel := context.WithCancel(f.ctx)
 	defer cancel()
 
-	errChan := make(chan error, 2)
+	go func() {
+		<-ctx.Done()
+		inConn.Close()
+		outConn.Close()
+		ep.Close()
+	}()
```
lixmal: Are we closing everything twice now?

client/firewall/uspfilter/forwarder/udp.go
```diff
@@ -212,49 +211,60 @@ func (f *Forwarder) handleUDP(r *udp.ForwarderRequest) {
...
 
+	// inbound->outbound: copy from pConn.outConn to pConn.conn
 	go func() {
-		errChan <- pConn.copy(ctx, pConn.outConn, pConn.conn, &f.udpForwarder.bufPool, "inbound->outbound")
+		defer wg.Done()
+		rxBytes, inboundErr = pConn.copy(ctx, pConn.outConn, pConn.conn, &f.udpForwarder.bufPool, "inbound->outbound")
 	}()
 
-	select {
-	case <-ctx.Done():
```
lixmal, line 244: This one is also lost

client/firewall/uspfilter/forwarder/udp.go
```diff
@@ -212,49 +211,60 @@ func (f *Forwarder) handleUDP(r *udp.ForwarderRequest) {
...
+	}
+
+	var rxPackets, txPackets uint64
+	if tcpStats, ok := ep.Stats().(*tcpip.TransportEndpointStats); ok {
+		// fields are flipped since this is the in conn
+		rxPackets = tcpStats.PacketsSent.Value()
+		txPackets = tcpStats.PacketsReceived.Value()
+	}
+
+	f.logger.Trace("Removed UDP connection %s (timeout) [in: %d Pkts/%d B, out: %d Pkts/%d B]", epID(id), rxPackets, rxBytes, txPackets, txBytes)
```
lixmal: Not sure about the reason timeout here, it could also come from shutdown, right? Can you double check?

hakansa: yes, that's right!

client/firewall/uspfilter/forwarder/udp.go
hakansa: fixed.

client/firewall/uspfilter/forwarder/tcp.go
hakansa: fixed.

client/firewall/uspfilter/forwarder/tcp.go
```diff
@@ -128,5 +149,12 @@ func (f *Forwarder) sendTCPEvent(typ nftypes.Type, flowID uuid.UUID, id stack.Tr
 		}
 	}
 
+	remoteIp, _ := netip.ParseAddr(id.RemoteAddress.String())
+	localIp, _ := netip.ParseAddr(id.LocalAddress.String())
```
lixmal: This is bad for perf. We can (re)use

```
SourceIP: netip.AddrFrom4(id.RemoteAddress.As4()),
DestIP:  netip.AddrFrom4(id.LocalAddress.As4()),
```

Also src == remote and dest == local

client/firewall/uspfilter/forwarder/forwarder.go
```diff
@@ -167,3 +171,31 @@ func (f *Forwarder) determineDialAddr(addr tcpip.Address) net.IP {
 	}
 	return addr.AsSlice()
 }
+
+func (f *Forwarder) RegisterRuleID(srcIP, dstIP netip.Addr, srcPort, dstPort uint16, ruleID []byte) {
+	key := buildKey(srcIP, dstIP, srcPort, dstPort)
+	f.ruleIdMap.Store(key, ruleID)
```
lixmal: We're calling this method for every packet (currently no established check for routed traffic), can you bench if perf is better with `LoadOrStore` ?

`sync.Map` is optimized for the many reads, few writes scenario. Then again we might need to overwrite an entry if rules have changed. Maybe clear this map whenever rules were changed in the uspfilter?

client/firewall/uspfilter/forwarder/tcp.go
```diff
@@ -128,5 +149,12 @@ func (f *Forwarder) sendTCPEvent(typ nftypes.Type, flowID uuid.UUID, id stack.Tr
...
+
+	if ruleId, ok := f.getRuleID(typ, remoteIp, localIp, id.RemotePort, id.LocalPort); ok {
```
lixmal: Let's do this only on `Start` to align with Peer ACLs

client/firewall/uspfilter/forwarder/udp.go
```diff
@@ -222,39 +228,53 @@ func (f *Forwarder) proxyUDP(ctx context.Context, pConn *udpPacketConn, id stack
...
-	case <-ctx.Done():
-		f.logger.Trace("forwarder: tearing down UDP connection %v due to context done", epID(id))
-		return
-	case err := <-errChan:
-		if err != nil && !isClosedError(err) {
-			f.logger.Error("proxyUDP: copy error: %v", err)
-		}
-		f.logger.Trace("forwarder: tearing down UDP connection %v", epID(id))
-		return
+	wg.Wait()
```
lixmal, line 250: I think there's one edge case here. If one of the connections errors out, then we'll wait for the other one here (read from the other conn will keep looping). Until the udp timeout has been reached, which will clean up both.
Not a blocker, but needs some investigation.

client/firewall/uspfilter/forwarder/udp.go
```diff
@@ -212,7 +211,14 @@ func (f *Forwarder) handleUDP(r *udp.ForwarderRequest) {
...
+
+	ctx, cancel := context.WithCancel(f.ctx)
+	defer cancel()
+
+	go func() {
+		<-ctx.Done()
+		f.logger.Trace("forwarder: tearing down UDP connection %v due to context done", epID(id))
+
 		pConn.cancel()
 		if err := pConn.conn.Close(); err != nil {
```
lixmal: I think we can ignore isClosedError here too

client/firewall/uspfilter/forwarder/tcp.go
```diff
@@ -65,48 +66,66 @@ func (f *Forwarder) handleTCP(r *tcp.ForwarderRequest) {
...
 		}
-		f.logger.Trace("forwarder: tearing down TCP connection %v", epID(id))
-		return
 	}
+	if errOutToIn != nil {
+		if !isClosedError(errOutToIn) {
+			f.logger.Error("proxyTCP: copy error (out -> in): %v", errOutToIn)
+		}
+	}
+
```
lixmal, line 122: We should have a log somewhere here with bytes/pkts as well. Ideally only one, if possible (udp too, it has two currently)

client/firewall/uspfilter/forwarder/tcp.go
```diff
@@ -65,48 +66,66 @@ func (f *Forwarder) handleTCP(r *tcp.ForwarderRequest) {
...
+		errInToOut       error
+		errOutToIn       error
+	)
 
 	go func() {
-		_, err := io.Copy(outConn, inConn)
-		errChan <- err
+		defer wg.Done()
+		var n int64
+		n, errInToOut = io.Copy(outConn, inConn)
```
lixmal: We can skip the `n` var altogether

client/firewall/uspfilter/forwarder/tcp.go
```diff
@@ -65,48 +66,66 @@ func (f *Forwarder) handleTCP(r *tcp.ForwarderRequest) {
 }
 
 func (f *Forwarder) proxyTCP(id stack.TransportEndpointID, inConn *gonet.TCPConn, outConn net.Conn, ep tcpip.Endpoint, flowID uuid.UUID) {
```
lixmal, line 69: Something is wrong here. We get the established message

```
2025-04-24T18:30:50+02:00 TRAC forwarder: established TCP connection 100.64.106.139:51506 -> 192.168.20.2:8000
```

but it's never cleaned up (compare to `main` where it is cleaned up immediately after the conn is done).

client/firewall/uspfilter/forwarder/udp.go
```diff
@@ -212,7 +211,14 @@ func (f *Forwarder) handleUDP(r *udp.ForwarderRequest) {
 }
 
 func (f *Forwarder) proxyUDP(ctx context.Context, pConn *udpPacketConn, id stack.TransportEndpointID, ep tcpip.Endpoint) {
-	defer func() {
+
+	ctx, cancel := context.WithCancel(f.ctx)
+	defer cancel()
+
+	go func() {
```
lixmal, line 218: Very elegant! Either this goes first because the context is done or this goes last because of `defer cancel()`

lixmal (changes requested):

client/firewall/uspfilter/forwarder/udp.go
hakansa: Yeah i thought that case but couldn't figure a scenario that causes this. We can trigger `ctx.Cancel()` if one of the connections errors out so the other one would close immediately. Wdyt?

client/firewall/uspfilter/forwarder/tcp.go
hakansa: done

client/firewall/uspfilter/forwarder/tcp.go
hakansa: done

client/firewall/uspfilter/forwarder/tcp.go
hakansa: done

client/firewall/uspfilter/forwarder/udp.go
hakansa: 🙌

client/firewall/uspfilter/forwarder/forwarder.go
hakansa: yes you're right, `LoadOrStore` is slightly faster since we already have the entry most of the time. 

 Right now we only get rule id when the connection starts and remove the entry when we get the conn is done. So i dont think there's need to clear the map

client/firewall/uspfilter/forwarder/tcp.go
hakansa: Removed the log in the goroutine, we can log after both `io.Copy`s are done.

client/firewall/uspfilter/forwarder/forwarder.go
lixmal: True, forgot about the removal part.

client/firewall/uspfilter/uspfilter.go
```diff
@@ -792,6 +793,8 @@ func (m *Manager) handleRoutedTraffic(d *decoder, srcIP, dstIP netip.Addr, packe
 	} else {
 		if err := fwd.InjectIncomingPacket(packetData); err != nil {
 			m.logger.Error("Failed to inject routed packet: %v", err)
+		} else {
+			fwd.RegisterRuleID(srcIP, dstIP, srcPort, dstPort, ruleID)
```
lixmal: Is there a potential race here? We inject the packet before we register the rule, does the forwarder get the map entry in time?

client/firewall/uspfilter/forwarder/forwarder.go
```diff
@@ -167,3 +171,31 @@ func (f *Forwarder) determineDialAddr(addr tcpip.Address) net.IP {
...
+func (f *Forwarder) deleteRuleID(srcIP, dstIP netip.Addr, srcPort, dstPort uint16) {
+	if _, ok := f.ruleIdMap.LoadAndDelete(buildKey(srcIP, dstIP, srcPort, dstPort)); ok {
+		return
+	} else if _, ok := f.ruleIdMap.LoadAndDelete(buildKey(dstIP, srcIP, dstPort, srcPort)); ok {
+		return
+	}
+}
+
+func buildKey(srcIP, dstIP netip.Addr, srcPort, dstPort uint16) string {
+	return fmt.Sprintf("%s:%d-%s:%d", srcIP, srcPort, dstIP, dstPort)
```
lixmal: This isn't great, it adds an allocation, better would be to return a struct (but not a pointer to it)

hakansa: done

client/firewall/uspfilter/uspfilter.go
hakansa: never faced but you're right, it's better to register ruleId's first.

client/firewall/uspfilter/forwarder/tcp.go
hakansa: fixed.

client/firewall/uspfilter/forwarder/udp.go
hakansa: triggering `ctx.Cancel()` ensures the same behaviour, done.

client/firewall/uspfilter/forwarder/forwarder.go
```diff
@@ -167,3 +171,43 @@ func (f *Forwarder) determineDialAddr(addr tcpip.Address) net.IP {
...
+
+func (f *Forwarder) DeleteRuleID(srcIP, dstIP netip.Addr, srcPort, dstPort uint16) {
+	if _, ok := f.ruleIdMap.LoadAndDelete(buildKey(srcIP, dstIP, srcPort, dstPort)); ok {
+		return
+	} else if _, ok := f.ruleIdMap.LoadAndDelete(buildKey(dstIP, srcIP, dstPort, srcPort)); ok {
+		return
+	}
+}
+
+type ruleIdMapKey struct {
```
lixmal: We can reuse `conntrack.ConnKey` if there's no circular dependency issue

client/firewall/uspfilter/forwarder/icmp.go
```diff
@@ -122,6 +133,21 @@ func (f *Forwarder) sendICMPEvent(typ nftypes.Type, flowID uuid.UUID, id stack.T
...
 
-		// TODO: get packets/bytes
-	})
+		RxBytes:   rxBytes,
+		TxBytes:   txBytes,
+		RxPackets: rxPackets,
+		TxPackets: txPackets,
+	}
+
+	srcIp := netip.AddrFrom4(id.RemoteAddress.As4())
```
lixmal: Can we declare these further up and reuse them in `fields`?

lixmal (approved):

lixmal (approved):

Issue #1168, opened 2023-09-26
Install fails due to expired PGP key


risasoft:
**Describe the problem**
Attempt to install via cli:

``` curl -fsSL https://pkgs.netbird.io/install.sh | sh```

I am getting error about PGP key:

```
➜  ~ curl -fsSL https://pkgs.netbird.io/install.sh | sh
NetBird UI installation will be omitted as Linux does not run desktop environment
The installation will be performed using apt package manager
Hit:1 http://nova.clouds.archive.ubuntu.com/ubuntu jammy InRelease
Hit:2 http://nova.clouds.archive.ubuntu.com/ubuntu jammy-updates InRelease                                                               
Hit:3 http://nova.clouds.archive.ubuntu.com/ubuntu jammy-backports InRelease                                                                                     
Hit:4 https://swupdate.openvpn.net/community/openvpn3/repos jammy InRelease                                                                                      
Get:5 https://pkgs.tailscale.com/stable/ubuntu jammy InRelease                                                                                                   
Hit:6 http://security.ubuntu.com/ubuntu jammy-security InRelease                            
Get:7 https://pkgs.netbird.io/debian stable InRelease [5934 B]        
Err:7 https://pkgs.netbird.io/debian stable InRelease
  The following signatures were invalid: EXPKEYSIG 83F79AD029778355 Wiretrustee <dev@wiretrustee.com>
Reading package lists... Done
W: GPG error: https://pkgs.netbird.io/debian stable InRelease: The following signatures were invalid: EXPKEYSIG 83F79AD029778355 Wiretrustee <dev@wiretrustee.com>
E: The repository 'https://pkgs.netbird.io/debian stable InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

**Comment by gene1wood** [comment]
This key which expired today on 2023-09-25 is not only breaking new installs but `apt upgrade` also errors out now due to the key expiration

```
$ sudo bash -c "apt update && apt -y upgrade"
Hit:1 http://us.archive.ubuntu.com/ubuntu focal InRelease
...snip...
Err:6 https://pkgs.wiretrustee.com/debian stable InRelease                                                                                                     
  The following signatures were invalid: EXPKEYSIG 83F79AD029778355 Wiretrustee <dev@wiretrustee.com>
Hit:8 https://downloads.plex.tv/repo/deb public InRelease                                                                               
...snip...
[...]

PR #3663, opened 2025-04-11
[misc] update gpt file permissions in install.sh

Lamera:
Fix install.sh for some installations by explicitly setting the file permissions correctly.

https://github.com/netbirdio/netbird/issues/1168

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/release_files/install.sh
+++ b/release_files/install.sh
@@ -110,7 +110,7 @@ add_apt_repo() {
     | ${SUDO} gpg --dearmor -o /usr/share/keyrings/netbird-archive-keyring.gpg
 
     # Explicitly set the file permission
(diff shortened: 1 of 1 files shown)


mlsmaycon:
@Lamera the expired key status was linked to a file permission? In those cases, what was the resulting permission of the key file?

Lamera:
@mlsmaycon Yes, the file permission results from host to host when you configure umask. I have some older hosts here where this probably got changed. That's why I want to set the permissions explicitly.
Before changing manually the permissions looked like this:
-rw------- 1 root root 1,8K Apr 11 12:05 netbird-archive-keyring.gpg

After changing with chmod 644 to the file apt update worked just fine.

mlsmaycon (approved):

PR #3673, opened 2025-04-14
[client] set TLS ServerName for hostname-based QUIC connections

### LINKED ISSUE #3672: QUIC dialer fails with "doesn't contain any IP SANs" when connecting to relay hostname
**Describe the problem**

When connecting to a relay server by hostname, the QUIC dialer fails with certificate validation errors while the WebSocket dialer works correctly. The QUIC dialer validates the certificate against the resolved IP address instead of the hostname.

**To Reproduce**

1. Set up a self-hosted relay server with a Let's Encrypt certificate (or any certificate with DNS SANs only)
2. Configure NetBird to use this relay server by hostname (not IP address)
3. Attempt to connect to the relay
4. Error in logs:
`failed to dial to Relay server via QUIC 'example.com:33080': CRYPTO_ERROR 0x12a (local): tls: failed to verify certificate: x509: cannot validate certificate for 192.0.2.10 because it doesn't contain any IP SANs`

**Expected behavior**

The QUIC dialer should validate the certificate against the hostname used for connection, not the resolved IP address. This would allow successful connections with certificates that only have DNS SANs (like standard Let's Encrypt certificates).

Self-hosted NetBird control plane.

**NetBird version**

0.41.0

No.

**Additional context**

The issue is in `relay/client/dialer/quic/quic.go` where the TLS config doesn't set ServerName when connecting via hostname. The WebSocket dialer works properly because the HTTP client automatically sets the hostname for certificate validation.

alindt:
This sets ServerName in the TLS config when using hostnames to fix the validation.

Potentially fixes #3672

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/relay/client/dialer/quic/quic.go
+++ b/relay/client/dialer/quic/quic.go
@@ -28,6 +28,16 @@ func (d Dialer) Dial(ctx context.Context, address string) (net.Conn, error) {
 		return nil, err
 	}
 
+	// Get the base TLS config
+	tlsClientConfig := quictls.ClientQUICTLSConfig()
+
+	// Set ServerName to hostname if not an IP address
+	host, _, splitErr := net.SplitHostPort(quicURL)
+	if splitErr == nil && net.ParseIP(host) == nil {
+		// It's a hostname, not an IP - modify directly
+		tlsClientConfig.ServerName = host
+	}
+
 	quicConfig := &quic.Config{
 		KeepAlivePeriod:   30 * time.Second,
 		MaxIdleTimeout:    4 * time.Minute,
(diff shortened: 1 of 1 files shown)


pappz:
The URL may not have a port number in the URL string. Could you support this scenario too?

alindt:
@pappz Done.

pappz:
@alindt 
Modifying the prepareURL function is a little bit risky. We will lose the URL path.

I recommend that you create a function that figures out the serverName from the given URL string properly. I take a quick look to the WS package, the first step is the URL string will be converted to URL type and after that the net.SplitHostPort

alindt:
@pappz I don't understand what you mean by "losing the URL path".

The only way the QUIC dialer uses `quicURL` is as a parameter for `net.ResolveUDPAddr()` which expects a `host[:port]` string, which is exactly what my changed `prepareURL` (still) does.

The only change is in appending the default ports as per your request, meaning it will always include a `:port` part.

Sorry, am I missing something?

pappz:
You are right. I mixed the things a little bit. I take a look again.

B08Z:
> You are right. I mixed the things a little bit. I take a look again.

Any update on if this will be pulled in. 

I think it's stopping my clients connecting to relay using quic

PR #3690, opened 2025-04-16
[management] permissions manager use crud operations

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/permissions/roles/user.go
+++ b/management/server/permissions/roles/user.go
@@ -8,7 +8,9 @@ import (
 var User = RolePermissions{
 	Role: types.UserRoleUser,
 	AutoAllowNew: map[operations.Operation]bool{
-		operations.Read:  false,
-		operations.Write: false,
+		operations.Read:   false,
+		operations.Create: false,
+		operations.Update: false,
+		operations.Delete: false,
 	},
 }
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -283,7 +283,7 @@ func (am *DefaultAccountManager) UpdateAccountSettings(ctx context.Context, acco
 		return nil, err
 	}
 
-	allowed, err := am.permissionsManager.ValidateUserPermissions(ctx, accountID, userID, modules.Settings, operations.Write)
+	allowed, err := am.permissionsManager.ValidateUserPermissions(ctx, accountID, userID, modules.Settings, operations.Update)
 	if err != nil {
 		return nil, fmt.Errorf("failed to validate user permissions: %w", err)
 	}
@@ -533,7 +533,7 @@ func (am *DefaultAccountManager) DeleteAccount(ctx context.Context, accountID, u
 		return err
 	}
 
-	allowed, err := am.permissionsManager.ValidateUserPermissions(ctx, accountID, userID, modules.Accounts, operations.Write)
+	allowed, err := am.permissionsManager.ValidateUserPermissions(ctx, accountID, userID, modules.Accounts, operations.Delete)
 	if err != nil {
 		return fmt.Errorf("failed to validate user permissions: %w", err)
 	}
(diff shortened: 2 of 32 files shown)


management/server/permissions/roles/user.go
```diff
@@ -8,7 +8,9 @@ import (
 var User = RolePermissions{
```
pnmcosta, line 8: @pascal-fischer this still is very weird to me, and can create dashboard and custom role unnecessary complications/workarounds as I mentioned before.

IMO User Role should at least have module Peers permissions, manager/BL should filter accordingly to only it's own Peers.

pascal-fischer: I mean we can do it. Then we need to rename the settings (regular user view blocked) and have custom role handling. 

Right now I say if you have permission you get to see ALL peers. If you don't have permissions you see only your OWN peers. If additionally the settings `regular user view blocked` is set then you see NONE.

I just did it for simplicity. This way I don't care about the roles. If we set permissions peer to true the  `regular user view blocked`  is obsolete (which is not a bad thing) but then how do you define who can see all peers and who can see it's own? Then you need to set some new settings or special handling for each existing role. I am open for suggestions

pascal-fischer: Or we need to change Read permissions from bool to some multi-state approach and have, ALL, LIMITED, NONE. This could work and would probably be the cleanest approach but would require some better handling of roles, probably as we need to pass the states into each package, not only peers

pnmcosta: > Then we need to rename the settings (regular user view blocked) 

Why would that be? The regular user view blocked settings would simply flip all modules permissions to false, that's it's applicable feature.

> I just did it for simplicity. This way I don't care about the roles.

I don't think this is correct, you are caring, indirectly by having a special handling when the user is not allowed access to the module https://github.com/netbirdio/netbird/blob/main/management/server/peer.go#L64

Conversely, if a new role is created with Peers Read=false, it will always have access to it's peers and won't be covered by the `regular user view blocked` nor will have access to none. 😅

> Or we need to change Read permissions from bool to some multi-state approach and have, ALL, LIMITED, NONE. 

Perhaps, but not all packages would require this kind of handling, but we should be able to opt them in. Read/ReadMany comes to mind right now.

pascal-fischer: > Conversely, if a new role is created with Peers Read=false, it will always have access to it's peers and won't be covered by the regular user view blocked nor will have access to none. 😅

This statement is not true, all roles that have permission Peers false will see their own, and all roles that have permission false and `regular user view blocked` true will see none, regardless if it is a user, a billings admin or anything coming in the future. This is fully decoupled from roles now.

> Perhaps, but not all packages would require this kind of handling, but we should be able to opt them in. Read/ReadMany comes to mind right now.

I know but read one, read many will not work right now. unless you change 
```
allowed, err := am.permissionsManager.ValidateUserPermissions(ctx, accountID, userID, modules.Peers, operations.Read)
if err != nil {
	return nil, status.NewPermissionValidationError(err)
}
```
to return not allowed, but also allowed for one, allowed for many. Or some completely different return value. Otherwise how will you controll the downstream logic on `GetPeers`?

pnmcosta: > Otherwise how will you controll the downstream logic on GetPeers?

There, in the business logic, where the check to show own peers, all or none would be explicit. We wouldn't need any multi-state operation approaches, nor would the settings restriction have to be separate from the permissions for dashboard logic. 🤷

pascal-fischer: Just to clarify, I am in favor of having the permissions in sync, and even having Peers as true. I just don't wanna have any more `role ==` or `user.IsAdmin` in the code. And with the settings I had available this was the path of least resistance to maintain the old functionality without too much refactoring and migration. 

If we spend more time on this to develop a proper solution, lets go for it. I am even ok with the multi-state solution

PR #3702, opened 2025-04-17
[client] Set up firewall rules for dns routes dynamically based on dns response

lixmal:
## Describe your changes

This PR enhances NetBird's firewall implementation by replacing static wide firewall rules for DNS routes with dynamic rule generation.

- Firewall rules now start with empty prefix sets that are populated dynamically as DNS resolution occurs
- When the DNS forwarder resolves domain names to IP addresses, it automatically updates the corresponding prefix sets
- For the userspace firewall implementation, destination prefixes initialize as empty and update/merge dynamically as new IPs are resolved
- Debug bundle now collects IP sets for iptables
- Fixes uspfilter closing trackers on Windows properly
- Handles duplicate domains in the dns forwarder being part of different routes (adds resolved IPs to all of them)
- Fixes a bug where, in some cases, the userspace firewall would not block invalid routed subnets
- Aligns the order of expressions in one nftables rule
- Fixes a bug where netbird status would not display unicode fqdns properly
- Updates the nftables library to make use of the comment feature for sets
- Fixes case handling in the dns forwarder
- Fixes a potential panic in the dns forwarder if Update was called before creating the multiplexer

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/internal/dnsfwd/manager.go
```diff
@@ -19,6 +21,13 @@ const (
 	dnsTTL     = 60 //seconds
 )
 
+// ForwarderEntry is a mapping from a domain to a resource ID and a hash of the parent domain list.
+type ForwarderEntry struct {
+	Domain     domain.Domain
+	ResId      route.ID
```
hakansa: I'd prefer to use string type here. We store `<resId>:<peerId>` in `route.ID` but here we only store resId, we might forget and get confused in future 😄

lixmal: Well for network routes it's just a single ID, so it's not consistent.
Maybe a new type then? All those `string`s, especially in maps are really confusing and error-prone.

hakansa: yeah a new type sounds good, totally agree with the string thing

lixmal: See https://github.com/netbirdio/netbird/pull/3702/commits/cbf3a005d76989f0ef7143eed3fafe19deaae916

hakansa (approved):

PR #3718, opened 2025-04-23
[client] Feature/lazy connection backward

pappz:
## Describe your changes
Check the minimum version number of the remote peer, and if the remote peer does not support the lazy connection, then always use a permanent connection.

Keep in mind just because the user updated the agent, the management server will not push a new network map with the new version information.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/conn_mgr.go
```diff
@@ -41,7 +36,7 @@ func NewConnMgr(engineConfig *EngineConfig, statusRecorder *peer.Status, peerSto
 	e := &ConnMgr{
 		peerStore: peerStore,
 	}
-	if engineConfig.LazyConnectionEnabled || os.Getenv(envEnableLazyConn) == "true" {
+	if engineConfig.LazyConnectionEnabled || os.Getenv(lazyconn.EnvEnableLazyConn) == "true" {
```
lixmal: Let's use strconv here. Example:

```
	if val := os.Getenv(EnvDisableConntrack); val != "" {
		disableConntrack, err = strconv.ParseBool(val)
		if err != nil {
			log.Warnf("failed to parse %s: %v", EnvDisableConntrack, err)
		}
	}
```

Why is it a separate env var anyway, we already have the CLI flag that also provides us with an env var, right?

pappz: We do not process all CLI flags automatically with env var. It is true for root cmd but this flag is part of the up command, just like the rosenpass. Maybe it is a bug or unexpected behaviour.

lixmal (approved):

PR #3728, opened 2025-04-23
[management] user info with role permissions

pnmcosta:
## Describe your changes
Get current user response info with role permissions by module and operation

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/http/api/openapi.yml
```diff
@@ -216,11 +216,29 @@ components:
...
+              type: string
+              enum:
+                - read
+                - write
+          propertyNames:
+            type: string
+            enum:
+              - read
+              - write
+          example: {"networks": { "read": true, "write": false}, "peers": { "read": false, "write": false} }
```
pascal-fischer: we moved to CRUD. This should be represented here as well

pascal-fischer: also one is property names create, read, update, delete. The other should be module no? can we skip the enum here?

pascal-fischer (dismissed):

bcmmbaga (approved):

PR #3733, opened 2025-04-23
[management] Skip IdP cache warm-up on Redis if data exists

bcmmbaga:
## Describe your changes

Implements logic to conditionally skip IDP cache warm-up on startup when using Redis as cache store. If account data already present in the cache the warm-up is skipped. A delay flag is also supported via env `NB_IDP_CACHE_WARMUP_DELAY` to control warm-up timing.

The delay can be configured using s (seconds), m (minutes), or h (hours). eg `NB_IDP_CACHE_WARMUP_DELAY=2m`

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	cacheStore "github.com/eko/gocache/lib/v4/store"
+	"github.com/eko/gocache/store/redis/v4"
 	"github.com/rs/xid"
 	log "github.com/sirupsen/logrus"
 	"github.com/vmihailenco/msgpack/v5"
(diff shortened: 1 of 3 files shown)


management/server/store/sql_store.go
```diff
@@ -785,6 +785,19 @@ func (s *SqlStore) GetAccountByPeerPubKey(ctx context.Context, peerKey string) (
 	return s.GetAccount(ctx, peer.AccountID)
 }
 
+func (s *SqlStore) GetAnyAccountID(ctx context.Context) (string, error) {
```
mlsmaycon, line 803: let's add a test for this call

management/server/account.go
```diff
@@ -524,6 +543,32 @@ func (am *DefaultAccountManager) warmupIDPCache(ctx context.Context) error {
 	return nil
 }
 
+// isCacheCold checks if the cache needs warming up.
+func (am *DefaultAccountManager) isCacheCold(ctx context.Context, store cacheStore.StoreInterface) (bool, error) {
```
mlsmaycon, line 557: let's add a test too

mlsmaycon (changes requested):
Looks good. just missing some tests

mlsmaycon (approved):

PR #3798, opened 2025-05-07
[management] fix a bug with missed extra dns labels for a new peer

crn4:
## Describe your changes
during a registration of a new peer we miss to add extra dns labels to peer object
### Checklist
- [X] Is it a bug fix
- [ ] It is a refactor


management/server/peer.go
```diff
@@ -875,16 +875,6 @@ func (am *DefaultAccountManager) LoginPeer(ctx context.Context, login types.Peer
 			return status.Errorf(status.PreconditionFailed, "couldn't login peer: setup key doesn't allow extra DNS labels")
 		}
 
-		extraLabels, err := domain.ValidateDomainsStrSlice(login.ExtraDNSLabels)
-		if err != nil {
-			return status.Errorf(status.InvalidArgument, "invalid extra DNS labels: %v", err)
-		}
```
bcmmbaga, line 881: We should not always trust the extra labels at login. We should validate them

management/server/peer.go
```diff
@@ -875,16 +875,6 @@ func (am *DefaultAccountManager) LoginPeer(ctx context.Context, login types.Peer
...
-
-		if !slices.Equal(peer.ExtraDNSLabels, extraLabels) {
-			peer.ExtraDNSLabels = extraLabels
-			shouldStorePeer = true
-		}
```
bcmmbaga, line 886: Removing this will prevent the extra DNS labels from being updated. Eg. if you run` netbird down` and then `netbird up --extra-dns-labels {new_labels}`, the peer will ignore the new labels and keep the old ones

bcmmbaga (approved):

PR #3805, opened 2025-05-09
[client, android] Do not propagate empty routes

pappz:
## Describe your changes

If we get domain routes the Network prefix variable in route structure will be invalid (engine.go:1057). When we handower to Android the routes, we must to filter out the domain routes. If we do not do it the Android code will get "invalid prefix" string as a route.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/routemanager/notifier/notifier.go
+++ b/client/internal/routemanager/notifier/notifier.go
@@ -33,7 +33,7 @@ func (n *Notifier) SetInitialClientRoutes(clientRoutes []*route.Route) {
 	nets := make([]string, 0)
 	for _, r := range clientRoutes {
 		// filter out domain routes
-		if !r.Network.IsValid() {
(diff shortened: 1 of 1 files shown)


client/internal/routemanager/notifier/notifier.go
```diff
@@ -32,6 +32,10 @@ func (n *Notifier) SetListener(listener listener.NetworkChangeListener) {
 func (n *Notifier) SetInitialClientRoutes(clientRoutes []*route.Route) {
 	nets := make([]string, 0)
 	for _, r := range clientRoutes {
+		// filter out domain routes
+		if !r.Network.IsValid() {
```
lixmal:
```suggestion
		if r.IsDynamic()
```

lixmal (approved):

PR #3818, opened 2025-05-13
[client] Feature/lazy manageable feature by mgm server

pappz:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/conn_mgr.go
```diff
@@ -22,48 +22,79 @@ import (
 // - Managing lazy connections via the lazyConnManager
 // - Maintaining a list of excluded peers that should always have permanent connections
 // - Handling connection establishment based on peer signaling
+//
+// The implementation is not thread-safe; it is protected by engine.syncMsgMux.
```
lixmal, line 26: Is that a good idea? The next maintainer might be unaware of this comment when moving things around in the engine code. I'm in favor of encapsulation here

pappz: Me too, I’ve been thinking about it a lot. The full conn_mgr was originally part of the engine code, and logically, it still should be. But the engine is currently doing too much.

As a next step, we should move the logic for peer creation, removal, and modification out of the engine and into conn_mgr. At that point, it would make sense to revisit the locking strategy.

Right now, I think adding an extra lock would introduce unnecessary risk, since the current engine implementation is already safe.

lixmal (approved):

PR #3822, opened 2025-05-14
[client] Add FreeBSD desktop client support to OAuth flow

hakansa:
## Describe your changes
[client] Add FreeBSD desktop client support to OAuth flow
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


avg-I:
I am curious if the additional parameter for FreeBSD desktop is really needed.
It seems that desktop detection and handling is totally identical between Linux and FreeBSD.
So, could they just use the same "is(Unix)Desktop" flag?

hakansa:
> I am curious if the additional parameter for FreeBSD desktop is really needed. It seems that desktop detection and handling is totally identical between Linux and FreeBSD. So, could they just use the same "is(Unix)Desktop" flag?

@avg-I what about Wayland? Haven't used it before. So I'm not sure if it's being used widely.

avg-I:
> @avg-I what about Wayland? Haven't used it before. So I'm not sure if it's being used widely.

@hakansa I think that there should not be any difference between Linux and FreeBSD either.
I.e., the DESKTOP / XDG checks should be sufficient for most desktop environments.
If you want to double-check based on DISPLAY / WAYLAND_DISPLAY (e.g., for people running some minimalist graphic setup), then I do not see a reason to _not_ do that on Linux as well.

PR #3824, opened 2025-05-14
[management,client] PKCE add query parameter flag prompt=login or max…

crn4:
## Describe your changes
Introduces a new flag for the PKCE authentication flow. When active, this flag appends query parameters such as 'prompt=login' or 'max_age=0' to the authorization request.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/pkce_auth.go
```diff
@@ -99,7 +102,7 @@ func GetPKCEAuthorizationFlowInfo(ctx context.Context, privateKey string, mgmURL
 			RedirectURLs:          protoPKCEAuthorizationFlow.GetProviderConfig().GetRedirectURLs(),
 			UseIDToken:            protoPKCEAuthorizationFlow.GetProviderConfig().GetUseIDToken(),
 			ClientCertPair:        clientCert,
-			DisablePromptLogin:    protoPKCEAuthorizationFlow.GetProviderConfig().GetDisablePromptLogin(),
```
mlsmaycon, line 102: we should keep this one

management/client/common/types.go
```diff
@@ -0,0 +1,21 @@
...
+// LoginFlag introduces additional login flags to the PKCE authorization request
+type LoginFlag uint8
+
+const (
+	// LoginFlagDisabled disables additional login flags
+	LoginFlagDisabled LoginFlag = iota
+	// LoginFlagPrompt adds prompt=login to the authorization request
+	LoginFlagPrompt
+	// LoginFlagMaxAge0 adds max_age=0 to the authorization request
+	LoginFlagMaxAge0
```
mlsmaycon:
```suggestion
	// LoginFlagPrompt adds prompt=login to the authorization request
	LoginFlagPrompt LoginFlag = iota
	// LoginFlagMaxAge0 adds max_age=0 to the authorization request
	LoginFlagMaxAge0
```

We don't need a disabled flag in this case since we have a dedicated field for it with the `DisablePromptLogin ` bool.

mlsmaycon (changes requested):

management/client/common/types.go
crn4: yes, sure

mlsmaycon (approved):

management/client/common/types.go
```diff
@@ -0,0 +1,19 @@
+package common
+
+// LoginFlag introduces additional login flags to the PKCE authorization request
+type LoginFlag uint8
+
+const (
+	// LoginFlagPrompt adds prompt=login to the authorization request
+	LoginFlagPrompt LoginFlag = iota
```
avg-I, line 8: I don't fully understand this part of the change.
The code means that `LoginFlagPrompt` has a numeric value of zero.
Which means that if the Management Server does not send any login flag then that would translate to `LoginFlagPrompt` .
Is that the intended behavior?  To have `LoginFlagPrompt` as a default?

Another question, then, is how to unset this flag?
Is setting `LoginFlagMaxAge0` the only way to not set `LoginFlagPrompt`?

PR #3855, opened 2025-05-21
[management] lazy connection account setting

pnmcosta:
## Describe your changes
Introduces an account setting for the experimental [lazy connections concept](https://docs.netbird.io/how-to/lazy-connection) feature

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/http/api/openapi.yml
```diff
@@ -118,6 +118,10 @@ components:
           example: my-organization.org
         extra:
           $ref: '#/components/schemas/AccountExtraSettings'
+        lazy_connection_enabled:
+          description: Enables or disables lazy connection
```
mlsmaycon: let's add an experimental note here.

management/server/account.go
```diff
@@ -342,6 +342,16 @@ func (am *DefaultAccountManager) UpdateAccountSettings(ctx context.Context, acco
...
 	}
 
+	if oldSettings.LazyConnectionEnabled != newSettings.LazyConnectionEnabled {
+		if newSettings.LazyConnectionEnabled {
+			am.StoreEvent(ctx, userID, accountID, accountID, activity.AccountLazyConnectionEnabled, nil)
+		} else {
+			am.StoreEvent(ctx, userID, accountID, accountID, activity.AccountLazyConnectionDisabled, nil)
+		}
+		updateAccountPeers = true
+		account.Network.Serial++
```
pascal-fischer: Not a new issue from this PR but just noticed, maybe we should move the `account.Network.Serial++` to the bottom just before the `SaveAccount` so we only increase network serial once even though multiple setttings got chnaged. This way we do not grow the serial unneccessarily.

pnmcosta: yup makes sense

management/server/http/api/openapi.yml
pascal-fischer: can you add `x-experimental: true`, I will try to update the parser to consider this flag not only on endpoints but on parameters as well

pnmcosta: sure, that's done

PR #3856, opened 2025-05-21
[client] Do not reconnect to mgm server in case of handler error

pappz:
## Describe your changes

- Do not reconnect to mgm server in case of handler error
- Set to nil the flow grpc client to prevent double close call

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/netflow/manager.go
+++ b/client/internal/netflow/manager.go
@@ -123,10 +123,14 @@ func (m *Manager) disableFlow() error {
 
 	m.logger.Close()
 
-	if m.receiverClient != nil {
-		clientErr := m.receiverClient.Close()
-		m.receiverClient = nil
-		return clientErr
+	if m.receiverClient == nil {
+		return nil
+	}
+
+	err := m.receiverClient.Close()
+	m.receiverClient = nil
+	if err != nil {
+		return fmt.Errorf("close: %w", err)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


lixmal (approved):

client/internal/netflow/manager.go
```diff
@@ -124,7 +124,9 @@ func (m *Manager) disableFlow() error {
 	m.logger.Close()
 
 	if m.receiverClient != nil {
-		return m.receiverClient.Close()
+		clientErr := m.receiverClient.Close()
+		m.receiverClient = nil
+		return clientErr
```
lixmal, line 129: Hold on, we need to check if this is nil, otherwise we might return a non-nil interface containing nil

lixmal: Hold on, we need to check if this is nil, otherwise we might return a non-nil interface containing nil

```suggestion
		client := m.receiverClient
		m.receiverClient = nil
		if err := client.Close(); err != nil {
				return fmt.Errorf("close: %w", err)
		}
		return nil
```

lixmal (approved):

matanbaruch:
This PR is breaking netbird in version 0.45.1. netbird peers are dropping and not recovering

mlsmaycon:
@matanbaruch can you please share the client logs that didn't recover?

PR #3907, opened 2025-06-01
[client] Apply routes right away instead of on peer connection

lixmal:
### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/client/internal/routemanager/client/client.go
+++ b/client/internal/routemanager/client/client.go
@@ -1,4 +1,4 @@
-package routemanager
+package client
 
 import (
 	"context"
@@ -43,9 +43,9 @@ type routerPeerStatus struct {
 	latency   time.Duration
 }
 
-type routesUpdate struct {
-	updateSerial uint64
-	routes       []*route.Route
+type RoutesUpdate struct {
+	UpdateSerial uint64
+	Routes       []*route.Route
 }
 
 // RouteHandler defines the interface for handling routes
@@ -57,7 +57,7 @@ type RouteHandler interface {
 	RemoveAllowedIPs() error
 }
 
-type ClientNetworkConfig struct {
+type WatcherConfig struct {
 	Context          context.Context
 	DNSRouteInterval time.Duration
 	WGInterface      iface.WGIface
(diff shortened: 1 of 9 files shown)


client/ui/debug.go
```diff
@@ -686,7 +686,7 @@ func showBundleCreatedDialog(w fyne.Window, localPath string) {
 			"Administrator privileges may be required to access the file.", localPath)),
 	)
 
-	customDialog := dialog.NewCustom("Debug Bundle Created", "Cancel", content, w)
+	customDialog := dialog.NewCustom("Debug Bundle Created", "cancel", content, w)
```
lixmal:
```suggestion
	customDialog := dialog.NewCustom("Debug Bundle Created", "Cancel", content, w)
```

client/internal/routemanager/client/client.go
```diff
@@ -416,86 +406,95 @@ func (c *clientNetwork) disconnectEvent(rsn reason) {
...
+	w.startNewPeerStatusWatchers()
+}
+
+// Stop stops the watcher and cleans up resources.
+func (w *Watcher) Stop() {
+	log.Debugf("Stopping watcher for network [%v]", w.handler)
+
+	w.cancel()
+
+	if err := w.removeAllowedIPs(w.currentChosen, reasonShutdown); err != nil {
```
hakansa, line 497: we might need a nil check for `w.currentChosen`

```suggestion
        if w.currentChosen == nil {
            return
        }
	if err := w.removeAllowedIPs(w.currentChosen, reasonShutdown); err != nil {
```

client/internal/routemanager/client/client.go
```diff
@@ -416,86 +406,95 @@ func (c *clientNetwork) disconnectEvent(rsn reason) {
 	)
 }
 
-func (c *clientNetwork) sendUpdateToClientNetworkWatcher(update routesUpdate) {
+func (w *Watcher) SendUpdate(update RoutesUpdate) {
 	go func() {
-		c.routeUpdate <- update
+		w.routeUpdate <- update
```
hakansa: should we check Watcher ctx here to avoid possible blocking of this goroutine?

lixmal: Yep, definitely

PR #3908, opened 2025-06-01
[client] Add lazy connections to routed networks

lixmal:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/dns/upstream.go
+++ b/client/internal/dns/upstream.go
@@ -104,9 +104,7 @@ func (u *upstreamResolverBase) Stop() {
 // ServeDNS handles a DNS request
 func (u *upstreamResolverBase) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
 	var err error
-	defer func() {
-		u.checkUpstreamFails(err)
-	}()
+	defer u.checkUpstreamFails(err)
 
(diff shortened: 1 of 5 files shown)


client/internal/conn_mgr.go
```diff
@@ -97,6 +98,15 @@ func (e *ConnMgr) UpdatedRemoteFeatureFlag(ctx context.Context, enabled bool) er
 	}
 }
 
+// UpdateRouteHAMap updates the route HA mappings in the lazy connection manager
+func (e *ConnMgr) UpdateRouteHAMap(haMap route.HAMap) {
+	if !e.isStartedWithLazyMgr() {
+		return
```
hakansa, line 105: a debug log would be helpful

client/internal/dns/upstream.go
```diff
@@ -104,9 +104,7 @@ func (u *upstreamResolverBase) Stop() {
 // ServeDNS handles a DNS request
 func (u *upstreamResolverBase) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
 	var err error
-	defer func() {
-		u.checkUpstreamFails(err)
-	}()
+	defer u.checkUpstreamFails(err)
```
hakansa: `defer u.checkUpstreamFails(err)` captures the current value of err so it will be always nil.

See the example:
https://go.dev/play/p/IWhXrHMZ8Nu

lixmal: Ah yes, completely forgot, will fix

PR #3910, opened 2025-06-02
[client] Refactor peer state change subscription mechanism

pappz:
Because the code generated new channel for every single event, was easy to miss notification. Use single channel.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/peer/status_test.go
+++ b/client/internal/peer/status_test.go
@@ -87,7 +87,7 @@ func TestGetPeerStateChangeNotifierLogic(t *testing.T) {
 	status := NewRecorder("https://mgm")
 	_ = status.AddPeer(key, "abc.netbird", ip)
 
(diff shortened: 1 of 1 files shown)


client/internal/peer/status.go
```diff
@@ -939,13 +978,17 @@ func (d *Status) onConnectionChanged() {
...
 		return
 	}
-
-	close(ch)
-	delete(d.changeNotify, peerID)
+	for _, sub := range subs {
+		// block the write because we do not want to miss notification
+		// must have to be sure we will run the GetPeerState() on separated thread
+		go func() {
+			sub.eventsChan <- struct{}{}
```
lixmal: Let's add a select with ctx done here like suggested here https://github.com/netbirdio/netbird/pull/3907#discussion_r2120511177

lixmal (approved):

PR #3917, opened 2025-06-03
[client] Add systemd netbird logs to debug bundle

lixmal:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/internal/debug/debug_linux.go
```diff
@@ -4,17 +4,106 @@ package debug
...
 	"strings"
+	"time"
 
 	"github.com/google/nftables"
 	"github.com/google/nftables/expr"
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	maxLogEntries = 10000
```
nazarewk: I'd give it more like 30-50k lines, some of the samples I have on my computer are ~2MB @ 13k lines, 5MB @ 30k lines,

I'm not sure I would limit the maxLogAge at all (or put something like 60 days), we don't do that for other logs.

client/internal/debug/debug_linux.go
```diff
@@ -4,17 +4,106 @@ package debug
...
+		return fmt.Errorf("add systemd logs to bundle: %w", err)
+	}
+
+	log.Infof("Added systemd journal logs for %s to debug bundle", serviceName)
+	return nil
+}
+
+// getServiceName gets the service name from environment or defaults to netbird
+func getServiceName() string {
+	if unitName := os.Getenv("SYSTEMD_UNIT"); unitName != "" {
```
nazarewk, line 58: Not sure if it can be pulled in, but there is a service name set at https://github.com/netbirdio/netbird/blob/f16f0c78318add10cef54f9b70d58ee2f2d97376/client/cmd/root.go#L136-L136 (later sourced from envvar) that we could default to

PS: this should probably go to `return` statement

client/internal/debug/debug_linux.go
nazarewk: I'd put it even higher at 100k log lines, just analyzing a bundle having 60k lines (incl. gzipped) that take up only 500kb as a bundle

lixmal: done

client/internal/debug/debug_linux.go
lixmal: done

client/cmd/service.go
```diff
@@ -27,12 +28,20 @@ func newProgram(ctx context.Context, cancel context.CancelFunc) *program {
...
-	return &service.Config{
+	config := &service.Config{
 		Name:        serviceName,
 		DisplayName: "Netbird",
 		Description: "Netbird mesh network client",
 		Option:      make(service.KeyValue),
+		EnvVars:     make(map[string]string),
 	}
+
+	// Set SYSTEMD_UNIT environment variable for Linux systems
```
lixmal:
```suggestion
```

PR #3922, opened 2025-06-04
[client] Add output similar to wg show to the debug package

pappz:
## Describe your changes
Add output similar to wg show to the debug package.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/iface/configurer/wgshow.go
+++ b/client/iface/configurer/wgshow.go
@@ -12,6 +12,7 @@ type Peer struct {
 	TxBytes       int64
 	RxBytes       int64
 	LastHandshake time.Time
+	PresharedKey  bool
 }
 
 type Stats struct {
--- a/client/iface/configurer/wgshow.go
+++ b/client/iface/configurer/wgshow.go
@@ -1,0 +1,23 @@
+package configurer
+
+import (
+	"net"
+	"time"
+)
+
+type Peer struct {
+	PublicKey     string
+	Endpoint      net.UDPAddr
+	AllowedIPs    []net.IPNet
+	TxBytes       int64
+	RxBytes       int64
+	LastHandshake time.Time
+}
+
+type Stats struct {
+	DeviceName string
+	PublicKey  string
+	ListenPort int
+	FWMark     int
+	Peers      []Peer
+}
--- a/client/internal/debug/wgshow.go
+++ b/client/internal/debug/wgshow.go
@@ -1,0 +1,61 @@
+package debug
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+	"time"
+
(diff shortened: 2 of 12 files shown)


client/internal/debug/wgshow.go
```diff
@@ -0,0 +1,58 @@
...
+type WGIface interface {
+	FullStats() (*configurer.Stats, error)
+}
+
+func (g *BundleGenerator) addWgShow() error {
+	result, err := g.statusRecorder.PeersStatus()
+	if err != nil {
+		return err
+	}
+
```
lixmal, line 21: We're missing the anonymization step here

client/iface/configurer/wgshow.go
```diff
@@ -0,0 +1,23 @@
...
+	LastHandshake time.Time
+}
+
+type Stats struct {
+	DeviceName string
+	PublicKey  string
+	ListenPort int
+	FWMark     int
+	Peers      []Peer
+}
```
lixmal: Can we add a bool indicating whether a PSK is set here? Also keepalive

client/internal/debug/wgshow.go
pappz: What should we anonymize here? IP?

client/iface/configurer/wgshow.go
pappz: I think yes, we can I will extend with the PSK I do not see the sense of keepalive period because it is hardcoded.

client/internal/debug/wgshow.go
lixmal: Yep

client/iface/configurer/wgshow.go
lixmal: In case it is missing for whatever reason, just a full view of the setup

client/internal/debug/wgshow.go
```diff
@@ -0,0 +1,66 @@
...
+	for _, peer := range s.Peers {
+		sb.WriteString(fmt.Sprintf("\npeer: %s\n", peer.PublicKey))
+		if peer.Endpoint.IP != nil {
+			if g.anonymize {
+				anonEndpoint := g.anonymizer.AnonymizeUDPAddr(peer.Endpoint)
+				sb.WriteString(fmt.Sprintf("  endpoint: %s\n", anonEndpoint.String()))
+			} else {
+				sb.WriteString(fmt.Sprintf("  endpoint: %s\n", peer.Endpoint.IP.String()))
```
lixmal:
```suggestion
				sb.WriteString(fmt.Sprintf("  endpoint: %s\n", peer.Endpoint.String()))
```

lixmal (dismissed):

lixmal (approved):

PR #3929, opened 2025-06-05
[client] Carry the peer's actual state with the notification.

pappz:
## Describe your changes

This PR addresses an unnecessary invocation of GetStates in a separate thread during notification handling. The notifier now provides the relevant state data directly with each notification, eliminating the need for redundant or parallel state queries. This simplifies the logic, reduces potential race conditions.

Changes:

- Removed separate thread execution of GetStates during notifications.
- Updated notification handler to rely on state data included in the notification payload.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


Barberky859 (approved):

client/internal/peer/status.go
```diff
@@ -991,15 +998,29 @@ func (d *Status) notifyPeerStateChangeListeners(peerID string) {
...
 		// block the write because we do not want to miss notification
-		// must have to be sure we will run the GetPeerState() on separated thread
-		go func() {
-			select {
-			case sub.eventsChan <- struct{}{}:
-			case <-sub.ctx.Done():
-			}
-		}()
+		select {
+		case sub.eventsChan <- changedPeers:
```
lixmal: Now that this is sync, should we use a buffered channel in case the route manager is too slow?

client/internal/routemanager/client/client.go
```diff
@@ -232,8 +249,8 @@ func (w *Watcher) watchPeerStatusChanges(ctx context.Context, peerKey string, pe
 			return
 		case <-closer:
 			return
-		case <-subscription.Events():
-			peerStateUpdate <- struct{}{}
+		case routersStates := <-subscription.Events():
```
lixmal: I would call it routerStates

client/internal/peer/status.go
pappz: It is buffered. The comment was outdated.
```
eventsChan: make(chan map[string]RouterState, 1),
```

We have an issue here. If we add new statuses to this channel while simultaneously receiving a new network map with changes to the peers in the channel, the router manager state will become inconsistent.

client/internal/peer/status.go
```diff
@@ -995,15 +1002,28 @@ func (d *Status) notifyPeerStateChangeListeners(peerID string) {
 	if !ok {
 		return
 	}
+
+	// collect the relevant data for router peers
+	changedPeers := make(map[string]RouterState, len(d.changeNotify))
```
hakansa: am i missing something or we're not considering the peers didnt changed when recalculating routes? `convertRouterPeerStatuses` only gets updated peers and passes to `recalculateRoutes`

lixmal (approved):

PR #3943, opened 2025-06-08
[client] Handle lazy routing peers that are part of HA groups 

lixmal:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/peer/conn.go
+++ b/client/internal/peer/conn.go
@@ -319,6 +319,9 @@ func (conn *Conn) WgConfig() WgConfig {
 
 // IsConnected returns true if the peer is connected
 func (conn *Conn) IsConnected() bool {
+	conn.mu.Lock()
+	defer conn.mu.Unlock()
+
 	return conn.evalStatus() == StatusConnected
 }
 


pappz:
What will be if we call the `UpdateRouteHAMap` middle of the `activateHAGroupPeers` iteration? Probably does not cause any panic but is it fine if we active peers while we purge that peer from the list by the `UpdateRouteHAMap`?

client/internal/conn_mgr.go
```diff
@@ -175,7 +175,7 @@ func (e *ConnMgr) AddPeerConn(ctx context.Context, peerKey string, conn *peer.Co
 		PeerConnID: conn.ConnID(),
 		Log:        conn.Log,
 	}
-	excluded, err := e.lazyConnMgr.AddPeer(lazyPeerCfg)
+	excluded, err := e.lazyConnMgr.AddPeer(ctx, lazyPeerCfg)
```
pappz: Here we should use the lazy context

lixmal: fixed

client/internal/peer/conn.go
```diff
@@ -317,12 +317,9 @@ func (conn *Conn) WgConfig() WgConfig {
 	return conn.config.WgConfig
 }
 
-// IsConnected unit tests only
-// refactor unit test to use status recorder use refactor status recorded to manage connection status in peer.Conn
+// IsConnected returns true if the peer is connected
 func (conn *Conn) IsConnected() bool {
```
pappz, line 321: Shouldn't we use the mutex? The ice and relay status can change at the same time.

lixmal: good point, let me check

lixmal: yep, fixed

pappz: Deadlock. Better to lock in the exported function

lixmal: done

Issue #3760, opened 2025-04-30
Add process name as posture check


Lamera:
**Is your feature request related to a problem? Please describe.**
Windows Defender gets deployed very frequently with a new version.
To check if Defender is active, I need to provide the full path to the executable.
For ex. ``"C:\ProgramData\Microsoft\Windows Defender\Platform\4.18.25030.2-2\MpDefenderCoreService.exe"``.
So I need to change the path with every new version from Defender.

**Describe the solution you'd like**
I would like to just provide the process name, not the full executable path.
Otherwise, it may be possible to check for Windows services running?

synedraIT:
We are currently evaluating netbird as a possible solution for us.

We also have some tools (rmm, EDR, ..) that use the version number in their install path. 

So this feature with process names or maybe wildcards for the path would be great (or lets say its a must have for us).

I hope to see this feature soon ;)

Best regards,
Markus

OmarAboulMakarem:
As a fintech organization, this feature is critical for us. We're implementing
  NetBird posture checks to enforce security policies, including mandatory Kaspersky
  Endpoint Security (KES) deployment across all Windows endpoints.

  However, we've encountered a significant operational challenge: Kaspersky's
  installation path includes the version number, which means every time the software
  is updated, the posture check breaks.

  Currently, we must hardcode:
  C:\Program Files (x86)\Kaspersky Lab\KES.12.12.0\avp.exe

  But when Kaspersky updates to v13.x, this path becomes invalid:
  C:\Program Files (x86)\Kaspersky Lab\KES.13.x.x\avp.exe

  With this PR's regex capabilities, we could use a pattern like:
  C:\Program Files (x86)\Kaspersky Lab\KES.[0-9]+.[0-9]+.[0-9]+\avp.exe

  This would allow our posture check to remain valid across version updates without
  manual intervention. For fintech organizations managing thousands of endpoints,
  this is essential for maintaining continuous compliance verification.

  This feature would significantly reduce operational overhead and improve security
  policy enforcement stability. Looking forward to this being merged.

OmarAboulMakarem:
any news or alternative solution about this, its a big blocker

PR #3945, opened 2025-06-09
Added regex-capabilities to process posture checks

volkermauel:
## Describe your changes
This PR aims at giving more flexibility for process-type posture checks by allowing the users to use regex in the process path.

Potentially also helps with https://github.com/netbirdio/netbird/issues/3760

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/posture/process.go
+++ b/management/server/posture/process.go
@@ -3,6 +3,7 @@ package posture
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"slices"
 
 	nbpeer "github.com/netbirdio/netbird/management/server/peer"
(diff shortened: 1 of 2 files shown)


mlsmaycon:
@volkermauel thanks for your contribution. 

By itself, it is not enough to get the functionality in place. We would need to update the client to support it as well. See:

https://github.com/netbirdio/netbird/blob/0f050e5fe1b2f34b484a6f082cc483f0712a3135/client/system/process.go#L55

Another option is to have a separate field for regex paths in the network map update, but I didn't wrap my head around it yet.

volkermauel:
@mlsmaycon i could see two ways for this:
1. keep this on the server completely and the clients provides only a list of running processes (probably including details, such as pid, uid of the user running the process, ...), the the server can do the filtering
2. do this on both sides of the connection.

let me know if you want this to be implemented in the client too, and i'll give it a try.

OmarAboulMakarem:
As a fintech organization, this feature is critical for us. We're implementing
  NetBird posture checks to enforce security policies, including mandatory Kaspersky
  Endpoint Security (KES) deployment across all Windows endpoints.

  However, we've encountered a significant operational challenge: Kaspersky's
  installation path includes the version number, which means every time the software
  is updated, the posture check breaks.

  Currently, we must hardcode:
  C:\Program Files (x86)\Kaspersky Lab\KES.12.12.0\avp.exe

  But when Kaspersky updates to v13.x, this path becomes invalid:
  C:\Program Files (x86)\Kaspersky Lab\KES.13.x.x\avp.exe

  With this PR's regex capabilities, we could use a pattern like:
  C:\Program Files (x86)\Kaspersky Lab\KES.[0-9]+.[0-9]+.[0-9]+\avp.exe

  This would allow our posture check to remain valid across version updates without
  manual intervention. For fintech organizations managing thousands of endpoints,
  this is essential for maintaining continuous compliance verification.

  This feature would significantly reduce operational overhead and improve security
  policy enforcement stability. Looking forward to this being merged.

OmarAboulMakarem:
any news or alternative solution about this, its a big blocker

PR #3948, opened 2025-06-09
[client] Prevent from lazy peers going to idle if ha group members are active

lixmal:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor


client/internal/lazyconn/manager/manager.go
```diff
@@ -513,6 +558,17 @@ func (m *Manager) onPeerInactivityTimedOut(peerConnID peerid.ConnID) {
 
 	mp.peerCfg.Log.Infof("connection timed out")
 
+	if m.shouldDeferIdleForHA(mp.peerCfg.PublicKey) {
```
pappz, line 555: This breaks the logic of the internal states.
Active peer: run the inactivity monitor
Inactive peer: run the activity monitor
Excluded peers: stop the inactivity monitor and keep the persistent connection

It makes no sense to run an unused timer logic in the background if we know we do not want to put back to idle state.

client/internal/lazyconn/manager/manager.go
```diff
@@ -513,6 +558,17 @@ func (m *Manager) onPeerInactivityTimedOut(peerConnID peerid.ConnID) {
...
+		iw, ok := m.inactivityMonitors[peerConnID]
+		if ok {
+			mp.peerCfg.Log.Debugf("resetting inactivity timer due to HA group requirements")
+			iw.ResetMonitor(m.ctx, m.onInactive)
```
pappz: Better if you pass the context in line 148.

client/internal/lazyconn/manager/manager.go
```diff
@@ -136,13 +137,13 @@ func (m *Manager) UpdateRouteHAMap(haMap route.HAMap) {
 func (m *Manager) Start(ctx context.Context) {
 	defer m.close()
 
-	ctx, m.cancel = context.WithCancel(ctx)
+	m.ctx, m.cancel = context.WithCancel(ctx)
```
pappz: I ran a nil pointer exception with this m.ctx. From the conn_mgr.go we run the Start(...) from a go rutine. So we can not be sure the m.ctx already has been written before we run any other exported function.

client/internal/lazyconn/manager/manager.go
lixmal: >It makes no sense to run an unused timer logic in the background if we know we do not want to put back to idle state.

At some point we want to put the peer back to idle; when all other peers are not connected either (last peer standing).
I think adding to excluded peers and especially removing it from there again is more complex. What do you think?

client/internal/lazyconn/manager/manager.go
lixmal: I have now passed the original ctx without storing it. I missed before that we still have access to it.

client/internal/lazyconn/manager/manager.go
lixmal: Should be resolved now

pappz:
Let's continue the review in lazy-routing/new-peers. Approved.

pappz (approved):

PR #3967, opened 2025-06-12
[client] Always register NetBird with plain Linux DNS and use original servers as upstream

lixmal:
## Describe your changes

- Removes the restriction that only all-domains resolvers can be registered on Linux with `resolvconf` and `file` managers
- Registers NetBird as the sole nameserver
- Registers original nameservers as upstream with low priority: NetBird-assigned "all domains" resolvers can take precedence
- Replace DNS-related server IP strings with `netip.Addr`
- Use `resolvconf` as a fallback instead of `openresolv`
- Return `RcodeRefused` if no handler was found in the chain

### Checklist
- [x] Is it a bug fix
- [x] It is a refactor


client/internal/dns/host_darwin.go
```diff
@@ -165,13 +165,13 @@ func (s *systemConfigurator) removeKeyFromSystemConfig(key string) error {
 }
 
 func (s *systemConfigurator) addLocalDNS() error {
-	if s.systemDNSSettings.ServerIP == "" || len(s.systemDNSSettings.Domains) == 0 {
```
hakansa, line 168: `ServerIP.IsValid()` or `!ServerIP.IsValid()` is correct here? previous state was `ServerIP == ""`

lixmal: Oof no. Good catch thank you! Fixed

hakansa (approved):

buzzzo:
Please modity this by permit sysadmin to override it. It should cause issues on some scenarios.

PR #3980, opened 2025-06-16
[client] Feat/multiple profile

hakansa:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/internal/profilemanager/profilemanager.go
```diff
@@ -0,0 +1,131 @@
...
+	profileName = sanitazeProfileName(profileName)
+
+	if err := pm.setActiveProfileState(profileName); err != nil {
+		return fmt.Errorf("failed to switch profile: %w", err)
+	}
+	return nil
+}
+
+// sanitazeProfileName sanitizes the username by removing any invalid characters and spaces.
+func sanitazeProfileName(profileName string) string {
```
lixmal: This should have a unit tests and I would prefer an allow list over a deny list

client/internal/profilemanager/state.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type ProfileState struct {
+	Email string `json:"email"`
+}
+
+func (pm *ProfileManager) GetProfileState(profileName string) (*ProfileState, error) {
+	configDir, err := getConfigDir()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get config directory: %w", err)
```
lixmal: All these `failed to` are redundant (we know it's an error) in `fmt.Errorf` and just clutter logs
```suggestion
		return nil, fmt.Errorf("get config directory: %w", err)
```

client/internal/debug/debug.go
```diff
@@ -24,8 +24,8 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/netbirdio/netbird/client/anonymize"
-	"github.com/netbirdio/netbird/client/internal"
 	"github.com/netbirdio/netbird/client/internal/peer"
+	"github.com/netbirdio/netbird/client/internal/profilemanager"
 	"github.com/netbirdio/netbird/client/internal/statemanager"
```
lixmal: Not really related to this line.

Currently, the state manager is still acting globally, which is fine for the most part (DNS, routes, etc.).
However, it also stores the route selection, which should be per-profile and not global

client/cmd/profile.go
```diff
@@ -0,0 +1,249 @@
...
+	return nil
+}
+func listProfilesFunc(cmd *cobra.Command, _ []string) error {
+	if err := setupCmd(cmd); err != nil {
+		return err
+	}
+
+	conn, err := DialClientGRPCServer(cmd.Context(), daemonAddr)
+	if err != nil {
+		log.Errorf("failed to connect to service CLI interface %v", err)
```
lixmal: Why do we log this + return an error instead of returning an error with `fmt.Errorf`? Also, logging with `log` will have a user-unfriendly log level and timestamp prefixed when done in the `cmd` package.

Same applies to the whole file

client/internal/debug/debug.go
hakansa: do we reset state manager on "Down" state? If yes then it should not be a problem because we call down when switching between profiles. If not, I can reset the state manager when switching profile, wdyt?

lixmal: We clean most of it, but not the route selection. The first part can stay global, no problem. The route selection shouldn't be cleared, but it should exist per profile. Otherwise selections leak between profiles

Fantu:
Big thanks for adding this very useful feature!

sh00t3r:
Thank you very much!!

sandstrom:
Awesome! 💯

PR #3981, opened 2025-06-16
Stop having management depend on dashboard

Silex:
## Describe your changes

I can wrap my head around why management service would depend on dashboard, so here is a PR removing the dependency.

If the dependency is actually needed, I might transform this PR into one where I add `HEALTHCHECK` to the dashboard.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


mlsmaycon:
is mainly because we mount the certificates generated by the dashboard into the management

Silex:
@mlsmaycon ok thanks, I removed the commit. The rest should still apply, ask if you have questions.

Silex:
@mlsmaycon wait actually, why is it needed to mount the certificates generated by the dashboard into the management? is it if they both use the same domain on different ports?

I don't quite get it because by default it looks like only the dashboard serves HTTPS, and management is used internally on port 33080.

Or maybe it's simply so the letsencrypt folder contains all certificates, and the dependencies forces some kind of order for the file creation.

infrastructure_files/docker-compose.yml.tmpl
```diff
@@ -1,8 +1,16 @@
+x-default: &default
+  restart: 'unless-stopped'
+  logging:
+    driver: 'json-file'
+    options:
+      max-size: '50m'
```
r0b2g1t: why do you set the log max-size to 50m and the max-file to 10 instead of the 500m and 2?

Silex: It's subjective, but 500m and 2 feels much more fragile than 50m and 10. I can remove that commit if you want, or make it 50m and 20 so it has the same "2*500 = 20*50 = 1000" retention.

My rationale was that giants logfiles are more difficult to work with.

Silex: Ok for the record I removed the commit.

PR #3989, opened 2025-06-17
[client] Implement dns routes for Android

lixmal:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/android/client.go
```diff
@@ -203,10 +203,6 @@ func (c *Client) Networks() *NetworkArray {
 			continue
 		}
 
-		if routes[0].IsDynamic() {
```
pappz, line 206: If we remove this, what will happen with the HA routes on Android? I think I will get notifications all the time.

client/firewall/uspfilter/nat.go
```diff
@@ -0,0 +1,408 @@
...
+
+	dstIP := netip.AddrFrom4([4]byte{packetData[16], packetData[17], packetData[18], packetData[19]})
+
+	translatedIP, exists := m.getDNATTranslation(dstIP)
+	if !exists {
+		return false
+	}
+
+	if err := m.rewritePacketDestination(packetData, d, translatedIP); err != nil {
+		m.logger.Error("Failed to rewrite packet destination: %v", err)
```
pappz, line 214: It is not formatted Error -> Errorf

client/internal/routemanager/fakeip/fakeip.go
```diff
@@ -0,0 +1,93 @@
...
+		allocated:  make(map[netip.Addr]netip.Addr),
+		fakeToReal: make(map[netip.Addr]netip.Addr),
+		baseIP:     baseIP,
+		maxIP:      maxIP,
+	}
+}
+
+// AllocateFakeIP allocates a fake IP for the given real IP
+// Returns the fake IP, or existing fake IP if already allocated
+func (m *Manager) AllocateFakeIP(realIP netip.Addr) (netip.Addr, error) {
```
pappz, line 35: It's not fully clear yet when and how we associate these addresses. We allocated a fake IP, but it seems like you never released it. How does this work in real life?

lixmal: Yes, you are right, we never release them. That's something that can be improved, it would also make fake ip management much more complex

client/firewall/uspfilter/nat.go
lixmal: You confused me for a moment! this logger is always formatted, there's no Errorf

pappz:
Screenshot_20250626-144834 (1)
Looks like we do not propagate any name for the network list

client/firewall/uspfilter/nat.go
pappz: Sorry, I did not realise it is a different logger

pappz (approved):

PR #4003, opened 2025-06-18
[management] Add backward compatibility for older clients without firewall rules port range support

bcmmbaga:
## Describe your changes

This PR adds backward compatibility for clients with versions prior to `v0.48.0` that do not support port range firewall rules. For these older clients, port ranges are expanded into individual ports during network map generation.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor


lixmal:
I believe we can split this in two. Port ranges in peer ACLs on the client are relatively recent, but they've been there from the beginning for route ACLs.

Issue #3883, opened 2025-05-27
Filter-by-status for relayed & P2P state


florian-obradovic:
I'd like to filter the detailed status for all peers which are connected via relay or P2P
Example: `netbird status -d --filter-by-status P2P`

Silex:
In the meantime you can use this:

``` shell
netbird status --json | jq '.peers.details[] | select(.connectionType == "P2P")'
```

florian-obradovic:
@Silex awesome - thank you very much!
`netbird status --json | jq -r '.peers.details[] | select(.connectionType == "P2P") | [.fqdn, .netbirdIp, .status, .connectionType] | @tsv' | column -t`

or 

`netbird status --json | jq -r '.peers.details[] | select(.connectionType == "Relayed") | [.fqdn, .netbirdIp, .status, .connectionType] | @tsv' | column -t
`

![Image](https://github.com/user-attachments/assets/92b11e5e-5b65-4fe1-b9bd-e8cbcf1a5307)

PR #4010, opened 2025-06-19
Add --filter-by-connection-type flag to status command

aliamerj:
## Describe your changes
This PR introduces a new flag `--filter-by-connection-type` to the status command.
It allows users to filter peers by connection type (`P2P` or `Relayed`) in both JSON and detailed views.

Input validation is added in `parseFilters()` to ensure proper usage, and `--detail` is auto-enabled if no output format is specified (consistent with other filters).

#3883

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


mlsmaycon:
> @aliamerj I think we can limit this change to the status command only, similar to what we have for the other flags.
> 
> There is not much gain in filtering in the daemon/client

It seems like I am becoming blind... please disregard my comment

aliamerj:
@mlsmaycon 
OK :)

aliamerj:
@mlsmaycon 

Is there anything should I update here ??

client/status/status.go
```diff
@@ -222,6 +223,9 @@ func mapPeers(
 			if pbPeerState.Relayed {
 				connType = "Relayed"
 			}
+			if connectionTypeFilter != "" && !strings.EqualFold(connType, connectionTypeFilter) {
```
mlsmaycon: we should move this check into the `skipDetailByFilters` to align with current filter processing and output.

aliamerj:
I just remembered why I initially didn’t include `connectionTypeFilter` in `skipDetailByFilters`.
To apply this filter, I first need to determine the connection type (connType), which depends on the Relayed flag inside the peer’s state. Since this logic didn’t exist in `skipDetailByFilters` yet, I needed to derive `connType` before calling the filter

we only want to skip a peer if its actual connection type doesn't match the filter. So instead of duplicating logic, I introduced a small `getConnectionType()` helper to keep things clean and consistent in both the filtering and output steps.

mlsmaycon (approved):
Thanks @aliamerj for the contribution

PR #4038, opened 2025-06-23
[management] Remove store locks

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -280,9 +280,6 @@ func (am *DefaultAccountManager) GetIdpManager() idp.Manager {
 // User that performs the update has to belong to the account.
 // Returns an updated Settings
 func (am *DefaultAccountManager) UpdateAccountSettings(ctx context.Context, accountID, userID string, newSettings *types.Settings) (*types.Settings, error) {
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
-	defer unlock()
-
 	allowed, err := am.permissionsManager.ValidateUserPermissions(ctx, accountID, userID, modules.Settings, operations.Update)
 	if err != nil {
 		return nil, fmt.Errorf("failed to validate user permissions: %w", err)
@@ -460,8 +457,6 @@ func (am *DefaultAccountManager) peerLoginExpirationJob(ctx context.Context, acc
 		ctx := context.WithValue(ctx, nbcontext.AccountIDKey, accountID)
 		//nolint
 		ctx = context.WithValue(ctx, hook.ExecutionContextKey, fmt.Sprintf("%s-PEER-EXPIRATION", hook.SystemSource))
-		unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
-		defer unlock()
 
 		expiredPeers, err := am.getExpiredPeers(ctx, accountID)
 		if err != nil {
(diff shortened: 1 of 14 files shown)


management/server/account.go
```diff
@@ -640,8 +632,6 @@ func (am *DefaultAccountManager) isCacheCold(ctx context.Context, store cacheSto
 
 // DeleteAccount deletes an account and all its users from local store and from the remote IDP if the requester is an admin and account owner
 func (am *DefaultAccountManager) DeleteAccount(ctx context.Context, accountID, userID string) error {
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
-	defer unlock()
 	account, err := am.Store.GetAccount(ctx, accountID)
 	if err != nil {
```
crn4, line 636: I think that here we can have possible race. somewhere between user are being removed and actual removal of an account, still possible that user could be created

crn4 (commented):
I added some comments where I think better to add a transaction. let's discuss

management/server/account.go
```diff
@@ -1007,9 +997,6 @@ func (am *DefaultAccountManager) updateAccountDomainAttributesIfNotUpToDate(ctx
 		return nil
 	}
 
-	unlockAccount := am.Store.AcquireWriteLockByUID(ctx, accountID)
-	defer unlockAccount()
-
```
crn4, line 1012: I think that here better to add a transaction

management/server/networks/manager.go
```diff
@@ -104,9 +101,6 @@ func (m *managerImpl) UpdateNetwork(ctx context.Context, userID string, network
 		return nil, status.NewPermissionDeniedError()
 	}
 
-	unlock := m.store.AcquireWriteLockByUID(ctx, network.AccountID)
-	defer unlock()
-
```
crn4, line 109: I think that we need to add a transaction here, as some changes could be made between select and actual update

management/server/setupkey.go
```diff
@@ -55,8 +55,6 @@ type SetupKeyUpdateOperation struct {
 // and adds it to the specified account. A list of autoGroups IDs can be empty.
 func (am *DefaultAccountManager) CreateSetupKey(ctx context.Context, accountID string, keyName string, keyType types.SetupKeyType,
 	expiresIn time.Duration, autoGroups []string, usageLimit int, userID string, ephemeral bool, allowExtraDNSLabels bool) (*types.SetupKey, error) {
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
-	defer unlock()
```
crn4, line 58: seems that here we have a transaction which runs another transaction (SaveSetupKey). this is not about the  removal of the lock, but maybe worth it to change

management/server/user.go
```diff
@@ -227,9 +221,6 @@ func (am *DefaultAccountManager) DeleteUser(ctx context.Context, accountID, init
 		return status.Errorf(status.InvalidArgument, "self deletion is not allowed")
 	}
 
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
-	defer unlock()
-
 	initiatorUser, err := am.Store.GetUserByUserID(ctx, store.LockingStrengthShare, initiatorUserID)
```
crn4, line 224: I propose to add a transaction here

management/server/user.go
```diff
@@ -328,9 +316,6 @@ func (am *DefaultAccountManager) InviteUser(ctx context.Context, accountID strin
 
 // CreatePAT creates a new PAT for the given user
 func (am *DefaultAccountManager) CreatePAT(ctx context.Context, accountID string, initiatorUserID string, targetUserID string, tokenName string, expiresIn int) (*types.PersonalAccessTokenGenerated, error) {
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
-	defer unlock()
-
```
crn4, line 333: I think that better to add here too

management/server/user.go
```diff
@@ -379,9 +364,6 @@ func (am *DefaultAccountManager) CreatePAT(ctx context.Context, accountID string
 
 // DeletePAT deletes a specific PAT from a user
 func (am *DefaultAccountManager) DeletePAT(ctx context.Context, accountID string, initiatorUserID string, targetUserID string, tokenID string) error {
-	unlock := am.Store.AcquireWriteLockByUID(ctx, accountID)
-	defer unlock()
-
```
crn4, line 384: and here. we can have a race if these methods will be called concurrently

crn4 (changes requested):

PR #4042, opened 2025-06-24
[management] add uniqueness constraint for peer ip and label and optimize generation

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/migration/migration.go
```diff
@@ -373,3 +373,42 @@ func DropIndex[T any](ctx context.Context, db *gorm.DB, indexName string) error
 	log.WithContext(ctx).Infof("dropped index %s from table %T", indexName, model)
 	return nil
 }
+
+func CreateIndexIfNotExists[T any](ctx context.Context, db *gorm.DB, indexName string, columns ...string) error {
```
pnmcosta, line 377: why the custom migration instead of https://gorm.io/docs/indexes.html#uniqueIndex ?

management/server/account.go
```diff
@@ -104,6 +106,20 @@ type DefaultAccountManager struct {
 	updateAccountPeersBufferInterval atomic.Int64
 }
 
+func isUniqueConstraintError(err error) bool {
```
pnmcosta, line 107: can we have a test that covers this method on all db types?

management/server/peer/peer.go
```diff
@@ -20,14 +20,14 @@ type Peer struct {
 	// WireGuard public key
 	Key string `gorm:"index"`
 	// IP address of the Peer
-	IP net.IP `gorm:"serializer:json"`
+	IP net.IP `gorm:"serializer:json"` // uniqueness index per accountID (check migrations)
```
pnmcosta, line 23: net.IP is stringable, any reason by this is being json serialized instead of plain text? even so, can't we use gorm's uniqueIndex here instead of a custom migration?

management/server/peer/peer.go
```diff
@@ -20,14 +20,14 @@ type Peer struct {
...
 	// Meta is a Peer system meta data
 	Meta PeerSystemMeta `gorm:"embedded;embeddedPrefix:meta_"`
 	// Name is peer's name (machine name)
 	Name string
 	// DNSLabel is the parsed peer name for domain resolution. It is used to form an FQDN by appending the account's
 	// domain to the peer label. e.g. peer-dns-label.netbird.cloud
-	DNSLabel string
+	DNSLabel string // uniqueness index per accountID (check migrations)
```
pnmcosta, line 30: can't we use gorm's uniqueIndex here instead of a custom migration?

management/server/migration/migration.go
pascal-fischer: Well I started with gorm unique index and it worked well for postgres and sqlite but for mysql you get greeted with
```
Error 1170 (42000): BLOB/TEXT column 'ip' used in key specification without a key length
```
if you try to run management. As far as I could tell there is no way of defining the key length via gorm.
Since postgres and sqlite don't have the keylength I had to choose between engine specifc migration or "global" migration so I decided to go this way. I would have loved to use the gorm tags primarily because it would be simpler and we would have a reference on the objects but I didn't find a good way to support mysql with this.

management/server/account.go
pascal-fischer: Yes I will add one

management/server/peer/peer.go
pascal-fischer: This one I don't know, actually. I think there have been issues in the past with string parsing.
I would much rather like to go with something like the inet type on postgres as this whole IP <-> string parsing is a performance nightmare but then how to support the other engines as they do not have anything similar

pnmcosta (approved):

management/server/store/sql_store.go
```diff
@@ -967,15 +970,15 @@ func (s *SqlStore) GetTakenIPs(ctx context.Context, lockStrength LockingStrength
...
+func (s *SqlStore) GetPeerLabelsInAccount(ctx context.Context, lockStrength LockingStrength, accountID string, dnsLabel string) ([]string, error) {
 	tx := s.db
 	if lockStrength != LockingStrengthNone {
 		tx = tx.Clauses(clause.Locking{Strength: string(lockStrength)})
 	}
 
 	var labels []string
 	result := tx.Model(&nbpeer.Peer{}).
-		Where("account_id = ?", accountID).
+		Where("account_id = ? AND dns_label LIKE ?", accountID, dnsLabel+"%").
```
pnmcosta, line 981: @pascal-fischer why is this filtering now needed? AFAIK dns_label is a FQDN, e.g. `peer1.domain.test` are we ever suffixing it?

pnmcosta (dismissed):

PR #4050, opened 2025-06-25
[client] refactor lazy detection

pappz:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


client/iface/configurer/usp.go
```diff
@@ -87,7 +90,16 @@ func (c *WGUSPConfigurer) UpdatePeer(peerKey string, allowedIps []netip.Prefix,
 		Peers: []wgtypes.PeerConfig{peer},
 	}
 
-	return c.device.IpcSet(toWgUserspaceString(config))
+	if ipcErr := c.device.IpcSet(toWgUserspaceString(config)); ipcErr != nil {
+		return ipcErr
+	}
+
+	if endpoint != nil {
+		addr := netip.MustParseAddr(endpoint.IP.String())
```
lixmal: Let's not use `Must` unless it's test code. Isn't `endpoint.IP` already some proper type?

client/iface/bind/activity.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+type PeerRecord struct {
+	Address      netip.AddrPort
+	LastActivity atomic.Int64 // UnixNano timestamp
+}
+
+type ActivityRecorder struct {
+	mu         sync.RWMutex
+	peers      map[string]*PeerRecord         // publicKey to PeerRecord map
```
lixmal, line 25: Ideally, we should use something clearer than string in maps for new code, the type from `wgtypes`, or a new type based on string/[x]byte.
But not a dealbreaker

client/iface/bind/activity.go
```diff
@@ -0,0 +1,93 @@
...
+		addrToPeer: make(map[netip.AddrPort]*PeerRecord),
+	}
+}
+
+// GetLastActivities returns a snapshot of peer last activity
+func (r *ActivityRecorder) GetLastActivities() map[string]time.Time {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	activities := make(map[string]time.Time, len(r.peers))
```
lixmal, line 41: We should benchmark this part, it has potential to block the hot path for a while if there are many peers.  Maybe we can move the allocation out of the lock. map snapshot, iterators?

client/iface/bind/activity.go
```diff
@@ -0,0 +1,93 @@
...
+		log.Warnf("could not find record for address %s", address)
+		return
+	}
+
+	now := monotime.Now()
+	if now-record.LastActivity.Load() < saveFrequency {
+		return
+	}
+
+	record.LastActivity.Store(now)
```
lixmal: This should be a CompareAndSwap to avoid replacing this with an older value (race with another goroutine), no need to check the returned bool

client/iface/configurer/usp.go
pappz: The IP is not enough, we need the ip+port. Is it possible the endpoint.IP.String() returns with an invalid IP string?

client/iface/bind/activity.go
pappz: Good point. One time, I started working on a refactor where we used a custom byte[32] type, but I gave it up after a while. We should figure out a common type for it at some point.
In the Relay code, I already removed the strings, but I'm not completely satisfied with where I moved the peer ID type

client/iface/configurer/usp.go
lixmal: >The IP is not enough, we need the ip+port.

Right, but here you're only parsing the IP, although it's a `net.IP` :/

>Is it possible the endpoint.IP.String() returns with an invalid IP string?

Probably not, but why take the risk

client/iface/bind/activity.go
pappz: I think the chance of causing any performance issue here is low.

- We make this call only once every minute
- It uses a read lock only
- We apply a write lock only when updating a peer's IP address (e.g., changing from ICE to Relay).

So if this function is slow it means that the address change will be slow, not the data transfer.

client/iface/bind/activity.go
pappz: I do not think it is a valid case.
1. We use mono time
2. We write this value only from the same thread, in a loop.

client/iface/configurer/usp.go
pappz: MustParseAddr parse the IP only. But ok, I will update the error handling

lixmal (dismissed):

Issue #4054, opened 2025-06-26
Users can not be added / re-authed , JWT Token expired, max_age=0,  self-hosted EntraID


natechit:
Hello to all,

I'm self hosting an instance with EntraID as my SSO provider. This setup has always worked fine for the last couple of months but the last few days although my users are getting logged in correctly, the client errors out with :

`Error: waiting sso login failed with: rpc error: code = InvalidArgument desc = invalid jwt token, err: token could not be parsed: Token is expired`

![Image](https://github.com/user-attachments/assets/a4078cfc-baa5-4b5d-a381-66482a6c21b4)

By researching the issue I found the following from : https://stackoverflow.com/questions/77307924/azure-entra-id-tokens-endpoint-issues-an-expired-token

> The problem was in the max_age query parameter of the authorization URL provided by our client (the one that a client redirects a user to in order to get an authorization code). The max_age had a value of 0 (e.g. max_age=0) which for some reason caused Entra to issue an authorization code that would provide a token that was seemingly issued 5 minutes in the past and immediately expired in the present. We fixed it by removing the query parameter altogether. This resulted into getting a token with the default 60-90 minutes lifetime. More about the query parameter can be read in the [OIDC specification](https://openid.net/specs/openid-connect-core-1_0.html).

So I tried it, and when I start the auth procedure from a terminal, grab the url string it creates and instead of max_age=0 in the string, I change it >5 say max_age=60 then the client authenticates and connects as it should.

To exhaust all of the troubleshooting, I nuked my whole EntraID app registration/enterprise app for Netbird and recreated it, still the same issue.
Also, tried a fresh install of the management plane, still the same issue.

Could not find anywhere to change the max_age value, seems to be hardcoded (on the client?)

My Entra sign-in logs show the users doing successful logins, so it should not be a matter of MFA, Conditional Policies or other controls.
I also created custom token (as described here: https://learn.microsoft.com/en-us/graph/api/tokenlifetimepolicy-post-tokenlifetimepolicies?view=graph-rest-1.0&tabs=powershell#request), assigned it to the Netbird object but still the issue persists.

Logs from the management when trying with the altered max_age=60 url string (when a new user logs in, if a user is already created I'm not seeing the successful connection logs):
[...]

PR #4055, opened 2025-06-26
fix: self-hosting: the wrong default for NETBIRD_AUTH_PKCE_LOGIN_FLAG

nazarewk:
fixes https://github.com/netbirdio/netbird/issues/4054


infrastructure_files/base.setup.env
```diff
@@ -60,7 +60,7 @@ NETBIRD_TOKEN_SOURCE=${NETBIRD_TOKEN_SOURCE:-accessToken}
 NETBIRD_AUTH_PKCE_REDIRECT_URL_PORTS=${NETBIRD_AUTH_PKCE_REDIRECT_URL_PORTS:-"53000"}
 NETBIRD_AUTH_PKCE_USE_ID_TOKEN=${NETBIRD_AUTH_PKCE_USE_ID_TOKEN:-false}
 NETBIRD_AUTH_PKCE_DISABLE_PROMPT_LOGIN=${NETBIRD_AUTH_PKCE_DISABLE_PROMPT_LOGIN:-false}
-NETBIRD_AUTH_PKCE_LOGIN_FLAG=${NETBIRD_AUTH_PKCE_LOGIN_FLAG:-1}
+NETBIRD_AUTH_PKCE_LOGIN_FLAG=${NETBIRD_AUTH_PKCE_LOGIN_FLAG:-"0"}
```
mlsmaycon:
```suggestion
NETBIRD_AUTH_PKCE_LOGIN_FLAG=${NETBIRD_AUTH_PKCE_LOGIN_FLAG:-0}
```

nazarewk: the `"` is actually more "correct", I just use it like that everywhere to avoid weird mistakes (double-quoting prevents treating the characters after `:` as part of the variable substitution operator)

mlsmaycon: besides NETBIRD_AUTH_PKCE_REDIRECT_URL_PORTS we are using without quotes. Let's keep this way everywhere for numbers

mlsmaycon (approved):

PR #4096, opened 2025-07-03
[management] migrate group peers into seperate table

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/management/server/store/sql_store.go
+++ b/management/server/store/sql_store.go
@@ -1521,7 +1521,7 @@ func (s *SqlStore) GetPeerGroups(ctx context.Context, lockStrength LockingStreng
 	return groups, nil
 }
 
-// GetPeerGroupIDs retrives all group IDs assigned to a specific peer in a given account.
+// GetPeerGroupIDs retrieves all group IDs assigned to a specific peer in a given account.
 func (s *SqlStore) GetPeerGroupIDs(ctx context.Context, lockStrength LockingStrength, accountId string, peerId string) ([]string, error) {
 	tx := s.db
 	if lockStrength != LockingStrengthNone {


management/server/types/group.go
```diff
@@ -26,14 +26,38 @@ type Group struct {
 	Issued string
 
 	// Peers list of the group
-	Peers []string `gorm:"serializer:json"`
+	Peers      []string    `gorm:"-"`
+	GroupPeers []GroupPeer `gorm:"foreignKey:GroupID;references:id;constraint:OnDelete:CASCADE;"`
```
pnmcosta, line 30: As discussed previously, foreign key and reference should be accountId instead, to prevent large `IN group id` queries when fetching group peers association.

And we should instead leverage a `GetAccountGroupPeers(context, accountId, groupId string)` method instead, so that we do not rely on the Group type to work with groups peers.

The `GroupPeers` can remain, but only as a GORM hint for FK reference and migration, can perhaps remain on a group fetch (GetAccountGroup) but depends on the use cases, if any.

management/server/types/group.go
```diff
@@ -26,14 +26,38 @@ type Group struct {
...
 
 	// Resources contains a list of resources in that group
 	Resources []Resource `gorm:"serializer:json"`
 
 	IntegrationReference integration_reference.IntegrationReference `gorm:"embedded;embeddedPrefix:integration_ref_"`
 }
 
+type GroupPeer struct {
+	GroupID string `gorm:"primaryKey"`
```
pnmcosta: add accountid as part of composite primarykey

management/server/types/account.go
```diff
@@ -73,7 +73,7 @@ type Account struct {
 	Users                  map[string]*User                  `gorm:"-"`
 	UsersG                 []User                            `json:"-" gorm:"foreignKey:AccountID;references:id"`
 	Groups                 map[string]*Group                 `gorm:"-"`
```
pnmcosta, line 75: I think we should remove this map, and refactor everywhere it's being used to `GetAccountGroups(context, accountId)` instead, otherwise having the table split will provide no performance advantage if we're still having to manage groups via the Account type.

The `GroupsG` can remain, but only as a GORM hint for FK reference and migration and be omitted from all account(s) query(ies).

management/server/store/sql_store.go
```diff
@@ -455,19 +459,40 @@ func (s *SqlStore) SaveGroups(ctx context.Context, lockStrength LockingStrength,
...
-			clause.Locking{Strength: string(lockStrength)},
-			clause.OnConflict{
-				Where:     clause.Where{Exprs: []clause.Expression{clause.Eq{Column: "groups.account_id", Value: accountID}}},
-				UpdateAll: true,
-			},
-		).
-		Create(&groups)
-	if result.Error != nil {
-		return status.Errorf(status.Internal, "failed to save groups to store: %v", result.Error)
+	for _, g := range groups {
```
pnmcosta: The `SaveGroup` and `SaveGroups` methods should be decomposed to leverage the `Add/RemoveGroupPeer` methods instead, to properly take advantage of the table split

management/server/types/group.go
pascal-fischer: This is a very valid point and I would love to do this. But this means the PR will get so big. Groups are used everywhere, to change this whole behaviour in all the places will become quite a lot.

management/server/store/sql_store.go
pascal-fischer: I agree on this one. This gonna be part of followup PRs as those methods are used in many places that will require slight adjustments. In this PR I wanna focus on the migration itself and the `Add/RemoveGroupPeer` in the addPeer method. But this should definitely come

management/server/types/account.go
pascal-fischer: That would be the ultimate goal but for now this needs to stay. There are too many things that depend on this. We gonna move that step by step.

management/server/types/group.go
pascal-fischer: This will follow once we can drop the Account object or at least the groups references. For now I added the accountID just as an index and replaced the GetAccount call to ignore the peer groups (therefor avoid the `IN` query). Instead I do  getAccountGroupPeers within the get account and map them afterwards.

management/server/account.go
```diff
@@ -1379,28 +1379,20 @@ func (am *DefaultAccountManager) SyncUserJWTGroups(ctx context.Context, userAuth
...
-			}
-
-			if err = transaction.SaveGroups(ctx, store.LockingStrengthUpdate, userAuth.AccountId, updatedGroups); err != nil {
-				return fmt.Errorf("error saving groups: %w", err)
+			for _, peer := range peers {
+				for _, g := range addNewGroups {
+					if err := transaction.AddPeerToGroup(ctx, userAuth.AccountId, peer.ID, g); err != nil {
+						return fmt.Errorf("error adding peer %s to group %s: %w", peer.ID, g, err)
+					}
+					if err := transaction.RemovePeerFromGroup(ctx, peer.ID, g); err != nil {
```
pnmcosta, line 1397: @pascal-fischer shouldn't this remove be done to the groups from `removeOldGroups` slice instead? a second iteration? as it stands, unless I'm misunderstanding it, it's just adding and then removing the peer from the same group 😅

pascal-fischer: Oh yeah, you are absolutely right.

management/server/store/sql_store.go
```diff
@@ -1311,55 +1368,73 @@ func (s *SqlStore) IncrementSetupKeyUsage(ctx context.Context, setupKeyID string
...
+// AddPeerToGroup adds a peer to a group
+func (s *SqlStore) AddPeerToGroup(ctx context.Context, accountID, peerID, groupID string) error {
+	peer := &types.GroupPeer{
+		AccountID: accountID,
+		GroupID:   groupID,
+		PeerID:    peerID,
+	}
 
-		return status.Errorf(status.Internal, "issue finding group: %s", result.Error)
+	err := s.db.WithContext(ctx).Create(peer).Error
```
pnmcosta: @pascal-fischer like on `AddPeerToAllGroup` I think it might be worth having a on conflict cause here to do nothing in case it already exists. Especially cause of group propagation, I think there might be a case where the peer can be re-added to the same group, I think.

management/server/types/group.go
```diff
@@ -26,14 +26,41 @@ type Group struct {
 	Issued string
 
 	// Peers list of the group
-	Peers []string `gorm:"serializer:json"`
+	Peers      []string    `gorm:"-"`
+	GroupPeers []GroupPeer `gorm:"foreignKey:GroupID;references:id;constraint:OnDelete:CASCADE;"`
```
pnmcosta, line 30: can these be documented as read only, and that `[Create/Update]Group` will ignore any changes to these, instead `[Add/Remove]Peer[To/From]Group` should be used instead?

management/server/store/sql_store.go
```diff
@@ -1775,33 +1857,53 @@ func (s *SqlStore) GetGroupsByIDs(ctx context.Context, lockStrength LockingStren
 	}
 
 	var groups []*types.Group
-	result := tx.Find(&groups, accountAndIDsQueryCondition, accountID, groupIDs)
+	result := tx.Preload(clause.Associations).Find(&groups, accountAndIDsQueryCondition, accountID, groupIDs)
```
pnmcosta, line 1887: @pascal-fischer I think we need to be careful with `Preload`, I remember having issues before, where instead of loading only the associations for the PK of the type, it was loading everything. Please debug these queries and if needed specify a join entry.

pascal-fischer: Confirmed, gorm will only preload the required groups
```
[1.142ms] [rows:1] SELECT * FROM `group_peers` WHERE `group_peers`.`group_id` = 'group-id'
```

pnmcosta (dismissed):

Issue #2345, opened 2024-07-29
netbird client inside docker: can not access to status or route list: context deadline exceeded


bravosierrasierra:
**Describe the problem**

Can not access to netbird daemon from outside docker container.
Daemon successfully started and registered in mesh with 
docker compose up -d

But i am can not access to daemon stats and routes:

$ docker compose exec netbird_client netbird --daemon-addr=unix:///var/run/netbird.sock status
Error: failed to connect to daemon error: context deadline exceeded
If the daemon is not running please run: 
netbird service install 
netbird service start

$ docker compose exec netbird_client ls -laF /var/run/
total 8
drwxr-xr-x    2 root     root          4096 Jun 18 15:37 ./
drwxr-xr-x    1 root     root          4096 Jul 29 14:48 ../
ubuntu@netbird-gw-dev-dc-01 .../docker-compose/netbird $ 

**To Reproduce**

use "docker compose up -d" and then "docker compose exec netbird_client netbird status"
with this docker-compose.yaml:

```
services:
  netbird_client:
    image: "netbirdio/netbird:latest"
    restart: always
    logging:
      driver: local
    labels:
      logging: "promtail"
      logging_jobname: "containerlogs"
    container_name: netbird_client
    cap_add:
      - NET_ADMIN
      - SYS_ADMIN
      - SYS_RESOURCE
    environment:
      - NB_SETUP_KEY=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
      - NB_HOSTNAME=netbird-gw-dev-dc.domain.com
      - NB_SERVICE=netbird
      - NB_DAEMON_ADDR=unix:///var/run/netbird.sock
      - NB_MANAGEMENT_URL=https://mgmt.domain.com:33073
      - NB_LOG_LEVEL=debug
    volumes:
      - netbird-client:/etc/netbird
    network_mode: host
    privileged: true

volumes:
  netbird-client:
    driver: local
```

**Expected behavior**

netbird status should be fetched and printed

**Are you using NetBird Cloud?**

self-host NetBird's control plane

**NetBird version**

latest, 0.28.4, 0.27.7

itoffshore:
I noticed the same issue under a privileged podman container / cloud hosted netbird

Maybe a good solution here would be to add a small function to detect if `netbird` is running in a container (`cat /proc/1/cgroup`) - & if so skip the daemon check when querying `status`

houckham:
[...]

PR #4126, opened 2025-07-09
[client] client: container: support CLI with entrypoint addition

nazarewk:
## Describe your changes

This will allow running netbird commands (including debugging) against the daemon and provide a flow similar to non-container usages.

It will by default both log to file and `stderr` so it can be handled more uniformly in container-native environments.

fixes https://github.com/netbirdio/netbird/issues/2345

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/cmd/up.go
+++ b/client/cmd/up.go
@@ -79,7 +79,7 @@ func upFunc(cmd *cobra.Command, args []string) error {
 
 	cmd.SetOut(cmd.OutOrStdout())
 
-	err := util.InitLog(logLevel, "console")
+	err := util.InitLog(logLevel, util.LogConsole)
 	if err != nil {
 		return fmt.Errorf("failed initializing log %v", err)
 	}
@@ -484,7 +484,7 @@ func parseCustomDNSAddress(modified bool) ([]byte, error) {
 		if !isValidAddrPort(customDNSAddress) {
 			return nil, fmt.Errorf("%s is invalid, it should be formatted as IP:Port string or as an empty string like \"\"", customDNSAddress)
 		}
-		if customDNSAddress == "" && logFile != "console" {
+		if customDNSAddress == "" && util.FindFirstLogPath(logFiles) != "" {
 			parsed = []byte("empty")
 		} else {
 			parsed = []byte(customDNSAddress)
--- a/client/cmd/up.go
+++ b/client/cmd/up.go
@@ -484,7 +484,7 @@ func parseCustomDNSAddress(modified bool) ([]byte, error) {
 		if !isValidAddrPort(customDNSAddress) {
 			return nil, fmt.Errorf("%s is invalid, it should be formatted as IP:Port string or as an empty string like \"\"", customDNSAddress)
 		}
-		if customDNSAddress == "" && util.FindFirstLogPath(logFiles...) != "" {
+		if customDNSAddress == "" && util.FindFirstLogPath(logFiles) != "" {
 			parsed = []byte("empty")
 		} else {
 			parsed = []byte(customDNSAddress)
--- a/client/cmd/up.go
+++ b/client/cmd/up.go
@@ -484,7 +484,7 @@ func parseCustomDNSAddress(modified bool) ([]byte, error) {
 		if !isValidAddrPort(customDNSAddress) {
 			return nil, fmt.Errorf("%s is invalid, it should be formatted as IP:Port string or as an empty string like \"\"", customDNSAddress)
 		}
-		if customDNSAddress == "" && util.FindFirstLogPath(logFiles) != "" {
+		if customDNSAddress == "" && util.FindFirstLogPath(logFiles...) != "" {
 			parsed = []byte("empty")
 		} else {
 			parsed = []byte(customDNSAddress)
--- a/client/netbird-entrypoint.sh
+++ b/client/netbird-entrypoint.sh
@@ -3,7 +3,6 @@ set -eEuo pipefail
 
 : ${NB_ENTRYPOINT_SERVICE_TIMEOUT:="5"}
 : ${NB_ENTRYPOINT_LOGIN_TIMEOUT:="1"}
-: ${NB_ENTRYPOINT_TAIL_LOG_FILE:="true"}
 NETBIRD_BIN="${NETBIRD_BIN:-"netbird"}"
 export NB_LOG_FILE="${NB_LOG_FILE:-"/var/log/netbird/client.log"}"
 service_pids=()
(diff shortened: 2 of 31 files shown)


nazarewk:
as a piece of background: `service run` doesn't support all of the `up` flags.

nazarewk:
TODO: there is some issue with goreleaser's Docker context https://github.com/netbirdio/netbird/actions/runs/16205993143/job/45756297182?pr=4126#step:16:183 

```
 ⨯ release failed after 4m19s               error=docker build failed: seems like you tried to copy a file that is not available in the build context.
```

edit: solved this

client/cmd/up.go
```diff
@@ -484,7 +484,7 @@ func parseCustomDNSAddress(modified bool) ([]byte, error) {
 		if !isValidAddrPort(customDNSAddress) {
 			return nil, fmt.Errorf("%s is invalid, it should be formatted as IP:Port string or as an empty string like \"\"", customDNSAddress)
 		}
-		if customDNSAddress == "" && logFile != "console" {
+		if customDNSAddress == "" && util.FindFirstLogPath(logFiles) != "" {
```
nazarewk, line 487: @mlsmaycon I'm not sure what the motivation was for checking logFile against `console` here, but I hope I preserved the intention

client/netbird-entrypoint.sh
```diff
@@ -0,0 +1,113 @@
...
+  else
+    info "there are no service processes to terminate"
+  fi
+}
+
+on_signal_propagate() {
+  local signal="${1}"
+  if test "${#service_pids[@]}" -gt 0; then
+    info "Propagating ${signal} to NetBird daemon..."
+    kill -"${signal}" "${#service_pids[@]}"
```
pappz: I do not understand this expression. What are we killing?
```
${#service_pids[@]}
```

client/netbird-entrypoint.sh
```diff
@@ -0,0 +1,113 @@
...
+  fi
+}
+
+wait_for_message() {
+  local timeout="${1}" message="${2}"
+  if test "${timeout}" -eq 0; then
+    info "not waiting for info message '${message}' due to zero timeout."
+  elif test -n "${log_file_path}"; then
+    info "waiting for info message '${message}' for ${timeout} seconds..."
+    timeout "${timeout}" grep -q "${message}" <(tail -F "${log_file_path}" 2>/dev/null)
```
pappz: How does this line work?
With the () operator, start a subshell (so a new process). What will happen with this shell after the timeout? Is it not orphaned?

util/log.go
```diff
@@ -55,6 +77,41 @@ func InitLog(logLevel string, logPath string) error {
 	return nil
 }
 
+// IterateLogs parses and iterates over logging entries
+func IterateLogs(logs []string) iter.Seq[string] {
```
pappz: The cobra + pflag library already manages the `--log-file` flag as a slice. It handles environment variables, separators, and multiple flags. There is no need to parse the results again.

client/netbird-entrypoint.sh
nazarewk: `${#service_pids[@]}` returns number of entries for `$service_pids` array

we're just forwarding the signal to the NetBird client daemon, the function isn't used anymore (could delete it), I originally thought I'd need pass `SIGUSR1` through to the daemon, but it's not true in the `service run` mode

client/netbird-entrypoint.sh
nazarewk: grep's `stdin` get's wired up from tail's `stdout` (in this case I'm pretty sure it could just be piped),
once `timeout` is reached, the `stdin` gets closed, then tail's `stdout` closes automatically and exits cleanly

I am not 100% sure about that "stdout closes automatically" part, but it **does** exit cleanly (I do have that `set -eEuo pipefail` stanza that makes the whole script fail fast whenever anything returns non-zero exit code)

util/log.go
nazarewk: I'll link it Monday if I don't forget, but I think you could **both** pass it out as multiple arguments and as a single `,` separated argument, can look into it and test later

nazarewk: oh right, I thought it's for another function.

(not that it matters here) Is it not actually less performant in Go to use `...` operator instead of passing the slice directly?

client/netbird-entrypoint.sh
pappz: Then do not hesitate to delete. Any way it was incorrect.
You used the **number of entries** in the kill command instead of the pid

client/netbird-entrypoint.sh
pappz: Nope, it left behind an orphaned process forever. This is why do not prefer the scripts. Proof:
```
pzoli@machine:~$ timeout 5 grep -q "waitsomething" <(tail -F "/tmp/src.txt" 2>/dev/null)
pzoli@machine:~$ ps ax | grep tail
37622 s004  S      0:00.01 tail -F /tmp/src.txt
```

util/log.go
pappz: I suppose there are performance differences between using the variadic expansion operator and just passing a slice.

Eliminate this function if the upper layer does exactly the same thing!

client/netbird-entrypoint.sh
nazarewk: you're right, should probably be more carefull about it, this works a lot better:

```text
kdn@brys ~> bash -c 'set -eEuo pipefail; touch /tmp/src.txt; timeout 2 tail -F /tmp/src.txt | grep -q wait || true; pstree $$'
--= 121124 kdn pstree 121124 
kdn@brys ~> 
```

client/netbird-entrypoint.sh
nazarewk: dropped it

client/netbird-entrypoint.sh
nazarewk: turns out we needed a middle ground after all (just moving the `timeout`), grep was not picking up the last line (it was looking for) at all when `tail` was interrupted while using piping:

```shell
grep -q "${message}" <(timeout "${timeout}" tail -F "${log_file_path}" 2>/dev/null)
```

util/log.go
nazarewk: I dropped the iteration and switched the `FindFirstLogPath()` to variadic to be consistent across file.

pappz: Did you push something? I do not see changes.

nazarewk: committed, but forgot to push the latest version

util/log.go
```diff
@@ -55,6 +73,29 @@ func InitLog(logLevel string, logPath string) error {
 	return nil
 }
 
+// FindFirstLogPath returns the first logs entry that could be a log path, that is neither empty, nor a special value
+func FindFirstLogPath(logs ...string) string {
```
pappz: Why do you prefer ...string instead of []string?

nazarewk: thought it would be more consistent with `InitLog()`, but otherwise I don't prefer it any specific way.

client/netbird-entrypoint.sh
```diff
@@ -0,0 +1,105 @@
...
+    case "${log_file}" in
+    console | syslog) ;;
+    *)
+      log_file_path="${log_file}"
+      return
+      ;;
+    esac
+  done < <(sed 's#,#\n#g' <<<"${log_files_string}")
+
+  warn "log files parsing parsing for '${log_files_string}' is not supported by debug bundles"
```
pappz: Double **parsing** word

nazarewk: fixed, also thought I'd print escaped variables (see the commit message), those turned out to be quite nicely displayed in all cases

PR #4177, opened 2025-07-18
[management, client] Add API to change the network range

lixmal:
## Describe your changes

- Adds an API to change the account's subnet and individual peer IPs
- Disables IP update on client to avoid stale config (the own IP is used in routes, firewalls, etc)

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


management/server/account.go
```diff
@@ -2018,3 +2037,156 @@ func propagateUserGroupMemberships(ctx context.Context, transaction store.Store,
...
+	}
+
+	var takenIPs []net.IP
+	peersToUpdate := make([]*nbpeer.Peer, 0, len(peers))
+
+	for _, peer := range peers {
+		newIP, err := types.AllocatePeerIP(*newIPNet, takenIPs)
+		if err != nil {
+			return status.Errorf(status.Internal, "allocate IP for peer %s: %v", peer.ID, err)
+		}
```
crn4: seems that we don't need here another peers slice (`peersToUpdate`) as it will be filled by the same peers

management/server/account.go
```diff
@@ -2018,3 +2037,156 @@ func propagateUserGroupMemberships(ctx context.Context, transaction store.Store,
...
+	}
+
+	updateNetworkMap, err := am.updatePeerIPInTransaction(ctx, accountID, userID, peerID, newIP)
+	if err != nil {
+		return fmt.Errorf("update peer IP transaction: %w", err)
+	}
+
+	if updateNetworkMap {
+		am.UpdateAccountPeers(ctx, accountID)
+	}
```
crn4: better to use `BufferUpdateAccountPeers`

management/server/account.go
```diff
@@ -2018,3 +2037,156 @@ func propagateUserGroupMemberships(ctx context.Context, transaction store.Store,
...
+		peersToUpdate = append(peersToUpdate, peer)
+	}
+
+	if err = transaction.SaveAccount(ctx, account); err != nil {
+		return err
+	}
+
+	for _, peer := range peersToUpdate {
+		if err = transaction.SavePeer(ctx, store.LockingStrengthUpdate, accountID, peer); err != nil {
+			return status.Errorf(status.Internal, "save updated peer %s: %v", peer.ID, err)
```
crn4, line 2085: we shouldn't have any errors here. but if we do - maybe worth it to change to continue to confirm that we looped through all peers?

lixmal: Should we not rollback on error?

management/server/account.go
lixmal: fixed

management/server/account.go
lixmal: fixed

management/server/account.go
crn4: ah, my bad, transaction will be rolled back, sure

crn4 (approved):
lgtm

PR #4213, opened 2025-07-24
[client] Add flag to configure MTU

lixmal:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


pappz:
The Relay code has multiple hardcoded buffer sizes (client, server too). It might be a good idea to define a maximum MTU size.

reimarstier:
Great to see there is progress with this feature! :-)

wiwiw-wiwiw:
Thanks, I also needed to adjust the MTU because I encountered a problem where GitHub repositories weren't downloading from a Docker container due to a difference in the MTU size of the Netbird network and the Docker network (as far as I understand). So, everything works with Netbird disabled, but not with it enabled.
But now I'll go test it.

Issue #1584, opened 2024-02-15
Netbird Management HA


JaSei:
**Is your feature request related to a problem? Please describe.**
The problem is that the management component lacks high availability (HA) support. Currently, the management component is central to other components like signal and dashboard, and stores data in a JSON file or, experimentally, in SQLite. However, these storage options cannot be shared across multiple instances. This limitation inhibits the ability to achieve high availability for the whole Netbird.

**Describe the solution you'd like**
Therefore, I propose introducing a new database connector for PostgreSQL as an alternative to SQLite. By adding support for PostgreSQL, the service can become stateless and run seamlessly on container orchestration platforms like Kubernetes or Docker Swarm across multiple instances. This change would enable high availability (HA) by allowing the management component to distribute its load and ensure resilience through redundancy.

**Describe alternatives you've considered**
* I tried storing the JSON file and SQLite database on AWS EFS (mounted as NFS) to share storage across multiple instances. However, this approach was unsuccessful as it did not support concurrent access effectively, leading to operational failures in multi-instance setups (only one instance was able to handle requests successfully).
* Another option is to explore other distributed database solutions or storage mechanisms that support concurrent access and are compatible with the existing architecture. PostgreSQL is the preferred option due to its robustness, scalability, and wide adoption in the industry.

**Additional context**
Achieving high availability for the management component is crucial for ensuring the reliability and scalability of services that depend on NetBirdIO. By enabling the management component to operate across multiple instances without storage bottlenecks, users can leverage container orchestration platforms to achieve better resilience and load distribution. This enhancement would greatly improve NetBirdIO's operational capabilities, particularly in production environments where uptime and scalability are crucial.

surik:
Hi @JaSei, thank you for the request and details explanation. I want to let you know that we have plans for PostgreSQL support, see our public roadmap: https://github.com/netbirdio/netbird/projects/2

JaSei:
That's amazing. Thanks for the roadmap. In this case, this ticket is probably useless.

ez1976:
Hi.
until this is implemented, any other way except replicating the instance with its sql DB.
creating another copy with a dedicated relay servers and any change made to one instance be changed also on the replicated one.
that way we can use a load balancer and any peer can connect to any instance and have the same routes and policies (of course each instance will use its own relay).
i am afraid of relying only on one instance for the whole solution if we plan to replace traditional VPN with NetBird.

awapf:
[...]

PR #4228, opened 2025-07-28
[client] Remove endpoint address on peer disconnect, retain status for activity recording

pappz:
When a peer disconnects, remove the endpoint address to avoid sending traffic to a non-existent location, but retain the status for the activity recorder.

Previously, before implementing lazy connections, the complete endpoint was removed. Now, the endpoint itself must be retained

### Checklist
- [x] Is it a bug fix
- [ ] Is a typo/documentation fix
- [ ] Is a feature enhancement
- [ ] It is a refactor
- [ ] Created tests that fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

> By submitting this pull request, you confirm that you have read and agree to the terms of the [Contributor License Agreement](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT.md).

PR #4245, opened 2025-07-30
[client] Add support for retain existing AllowedIPs when no alternative paths are available

gamerslouis:
We are using Netbird to manage our WireGuard network (without relay nodes). Our topology is relatively stable and does not change frequently.

We have analyzed potential network instability issues that may occur when either the ICE connection drops or the management API becomes temporarily unavailable. As noted in [this comment](https://github.com/netbirdio/netbird/issues/1584#issuecomment-2179294070), in such cases, network routes (e.g., 10.0.0.0/24) are removed from AllowedIPs, effectively cutting off connectivity.

In our scenario, we prefer:
	•	When ICE fails or the management service is temporarily down, the existing network connectivity should remain unaffected. The system should not proactively remove AllowedIPs, especially when no alternative paths are available.
	•	Even if the management service remains operational, the temporary unavailability of routes during ICE reconnection is still unacceptable, as it causes unnecessary and avoidable disruptions.

We propose adding an option to enable a failsafe routing mode, where:
	•	Route manager only updates AllowedIPs when a valid, reachable path is available.
	•	If no valid path is detected, the current AllowedIPs are kept unchanged.

This behavior would help prevent unnecessary network disconnections caused by transient ICE or management issues.

Example environment variable:
```
NB_ROUTE_STICKY_ON_FAILURE=true
```

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/internal/routemanager/client/client.go
+++ b/client/internal/routemanager/client/client.go
@@ -336,6 +336,11 @@ func (w *Watcher) recalculateRoutes(rsn reason, routerPeerStatuses map[route.ID]
 			return nil
 		}
 
+		if isRouteStickyOnFailure() {
+			log.Warnf("No available routes for network [%v], keep current route %s", w.handler, w.currentChosen.Peer)
+			return nil
+		}
+
 		if err := w.removeAllowedIPs(w.currentChosen, rsn); err != nil {
 			return fmt.Errorf("remove obsolete: %w", err)
 		}
--- a/client/internal/routemanager/client/env.go
+++ b/client/internal/routemanager/client/env.go
@@ -1,0 +1,24 @@
+package client
+
+import (
+	"os"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	// envRouteStickyOnFailure is used to configure if routes should be kept on failure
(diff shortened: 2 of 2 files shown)


gamerslouis:
#4228 might broke our changes. 

Currently, ICE disconnection does not remove the endpoint, so we only retain the routes.
If #4228 is merged, we should also disable endpoint removal.

gamerslouis:
Hi @pappz @lixmal,

Since #4228 has been merged, I will update this PR to also skip endpoint removal when NB_ROUTE_STICKY_ON_FAILURE is enabled.

At the same time, I’d like to confirm whether this feature aligns with NetBird’s overall direction, and whether you see any concerns with this approach.

In our environment, maintaining stable connectivity during temporary management API outages is essential. When the management service is unavailable, clients are unable to re-establish ICE connections, and such ICE disconnections can naturally occur for several hours under our deployment conditions (as also described in the related [issue](https://github.com/netbirdio/netbird/issues/1584#issuecomment-2179294070)). Because of this, preserving the existing AllowedIPs and endpoints becomes critical to avoid unnecessary disruptions.

Thanks, and I appreciate any guidance or suggestions you can provide.

Issue #4019, opened 2025-06-20
Auto update for desktop clients


PeakeIT:
**Is your feature request related to a problem? Please describe.**
I want to be able to deploy this to clients however my hesitation is that I wont have a dependable way to update end client devices.

**Describe the solution you'd like**
Some method to either trigger an update remotely or set the device to automatically update when a new release comes out.

**Describe alternatives you've considered**
I've considered using Winget or UniGet but they don't seem to reliably update the netbird client. 

**Additional context**
Nothing additional.

sahps:
I have the same issue, not just with Netbird but lots of apps that require admin-rights to update.

Here is my attempt to resolve this for Netbird, feel free to use.

https://gist.github.com/sahps/37fcfb1edc5241a879a0a7df4275cace

HammyHavoc:
Would publishing via the Microsoft Store be an option for Windows users? Seeing more and more FOSS available officially via it.

I also see plenty of peer machines running old versions of NetBird frequently as users generally don't update the client.

xihefeng:
> Would publishing via the Microsoft Store be an option for Windows users? Seeing more and more FOSS available officially via it.
> 
> I also see plenty of peer machines running old versions of NetBird frequently as users generally don't update the client.

In many Windows systems—such as Windows Server—or when users prefer not to rely on the Microsoft Store, having automated updates would save a great deal of unnecessary manual effort. This is a highly important feature that we are eagerly looking forward to.

HammyHavoc:
> > Would publishing via the Microsoft Store be an option for Windows users? Seeing more and more FOSS available officially via it.
> > 
> > I also see plenty of peer machines running old versions of NetBird frequently as users generally don't update the client.
> 
> In many Windows systems—such as Windows Server—or when users prefer not to rely on the Microsoft Store, having automated updates would save a great deal of unnecessary manual effort. This is a highly important feature that we are eagerly looking forward to.

You are presenting it like a false dichotomy when every other FOSS project just treats it like any other download mirror.

xihefeng:
> > > Would publishing via the Microsoft Store be an option for Windows users? Seeing more and more FOSS available officially via it.
> > > I also see plenty of peer machines running old versions of NetBird frequently as users generally don't update the client.
> > 
> > 
[...]

Issue #832, opened 2023-05-03
Auto-Update functionality for netbird client


giezi:
**Is your feature request related to a problem? Please describe.**
Client Updates are currently manual.

**Describe the solution you'd like**
It would be great if the netbird client had an auto update function. Users who don't want to should be able to disable it and update the client manually with a button in the client settings.

**Describe alternatives you've considered**
-

support-tt:
yes this feature would be super nice because if you have like 100 peers the update stuff is very nasty especially on clients with users.

lfarkas:
if your os packaging system can do it for you (rpm, deb) from it's repo. android, ios also do it (will do it). what else?

giezi:
> if your os packaging system can do it for you (rpm, deb) from it's repo. android, ios also do it (will do it). what else?

MacOS - I don't know if netbird would like to enter the Mac App Store... An auto-update of just the client would also be useful for Linux systems which are used for subnet routing.

lfarkas:
auto update overwrite some file on the filesystem. in a normal os the filesystem integrity is very critical. which means replacing a file which checksum is calculated at build time is a security nightmare. of i know for windows the security is never been so critical, but from a security software...
may be a beta version can do it, may be some as playground it can be done, but in a real production environment it's never a option...
to show a popup or log a notice it's ok. to automatically do the update it's not.

giezi:
> auto update overwrite some file on the filesystem. in a normal os the filesystem integrity is very critical. which means replacing a file which checksum is calculated at build time is a security nightmare. of i know for windows the security is never been so critical, but from a security software... may be a beta version can do it, may be some as playground it can be done, but in a real production environment it's never a option... to show a popup or log a notice it's ok. to automatically do the update it's not.

Of course, the auto-update functionality should be an optional feature.

Rob787:
+1 From our side as well. It is now a nightmare to manage. If at least the Windows client would trigger/ask users to update to new version that would help enormously already.

ruairinewman:
> auto update overwrite some file on the filesystem. in a normal os the filesystem integrity is very critical. which means replacing a file which checksum is calculated at build time is a security nightmare. of i know for windows the security is never been so critical, but from a security software... may be a beta version can do it, may be some as playground it can be done, but in a real production environment it's never a option... to show a popup or log a notice it's ok. to automatically do the update it's not.

[...]

PR #4256, opened 2025-07-30
Feature: Auto-update client

mohamed-essam:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Summary by CodeRabbit

* **New Features**
  * Configurable auto-update: clients can auto-update to latest, a specific version, or be disabled; initial auto-update applied at startup.
  * Update progress UI: displays live progress during client upgrades.
  * Platform update support: macOS, Windows, Linux (and placeholders for other targets).

* **Server & Sync**
  * Management surfaces auto-update settings in sync payloads and account APIs; emits events when account auto-update version changes.

* **Tests**
  * Added coverage for update manager and version decision logic

* **Documentation**
  * auto_update_version exposed in account settings API

### DIFF
--- a/management/server/types/settings.go
+++ b/management/server/types/settings.go
@@ -54,7 +54,7 @@ type Settings struct {
 	LazyConnectionEnabled bool `gorm:"default:false"`
 
 	// AutoUpdateVersion client auto-update version
-	AutoUpdateVersion string `gorm:"default:'latest'"`
+	AutoUpdateVersion string
 }
 
 // Copy copies the Settings struct


client/internal/updatemanager/update_darwin.go
```diff
@@ -0,0 +1,109 @@
...
+		return err
+	}
+
+	currentPID := os.Getpid()
+
+	// Restart netbird service after the fact
+	// This is a workaround since attempting to restart using launchctl will kill the service and die before starting
+	// the service again as it's a child process
+	// using SigTerm should ensure a clean shutdown
+	cmd = exec.Command("kill", "-15", fmt.Sprintf("%d", currentPID))
```
lixmal: Let's use Go APIs

client/internal/updatemanager/update_darwin.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return err
+}
+
+func (u *UpdateManager) updateHomeBrew() error {
+	// Homebrew must be run as a non-root user
+	// To find out which user installed NetBird using HomeBrew we can check the owner of our brew tap directory
+	fileInfo, err := os.Stat("/opt/homebrew/Library/Taps/netbirdio/homebrew-tap/")
+	if err != nil {
+		return err
```
lixmal: All these `err`s should have some minimal context to distinguish them in logs

client/internal/updatemanager/update_windows.go
```diff
@@ -0,0 +1,51 @@
...
+	"os/exec"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+
+	"golang.org/x/sys/windows/registry"
+)
+
+const (
+	msiDownloadURL = "https://github.com/netbirdio/netbird/releases/download/v%s/netbird_installer_%s_windows_amd64.msi"
```
lixmal: Should take arch into account

client/internal/updatemanager/update_windows.go
```diff
@@ -0,0 +1,51 @@
...
+	exeDownloadURL = "https://github.com/netbirdio/netbird/releases/download/v%s/netbird_installer_%s_windows_amd64.exe"
+)
+
+func (u *UpdateManager) triggerUpdate(targetVersion string) error {
+	k, err := registry.OpenKey(registry.LOCAL_MACHINE, `SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\Netbird`, registry.QUERY_VALUE)
+	if err != nil && strings.Contains(err.Error(), "system cannot find the file specified") {
```
lixmal: String matching will fail on localized error messages.

client/internal/updatemanager/update_darwin.go
```diff
@@ -0,0 +1,109 @@
...
+	cmd = exec.Command("sudo", "-u", userName, "sh", "-c", "echo $HOME")
+	out, err = cmd.CombinedOutput()
+	if err != nil {
+		return err
+	}
+
+	homeDir := strings.TrimSpace(string(out))
+	// Homebrew does not support installing specific versions
+	// Thus it will always update to latest and ignore targetVersion
+	cmd = exec.Command("sudo", "-u", userName, "/opt/homebrew/bin/brew", "upgrade", "netbirdio/tap/netbird")
```
lixmal: Do we handle netbirb-ui?

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func NewUpdateManager(statusRecorder *peer.Status) *UpdateManager {
+	update := version.NewUpdate("nb/client")
+	manager := &UpdateManager{
+		update:         update,
+		lastTrigger:    time.Now().Add(-10 * time.Minute),
+		statusRecorder: statusRecorder,
+	}
+	manager.version = "disabled"
```
lixmal: Let's use a const here

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,118 @@
...
+	update.SetDaemonVersion(version.NetbirdVersion())
+	update.SetOnUpdateListener(manager.CheckForUpdates)
+	return manager
+}
+
+func (u *UpdateManager) SetVersion(v string) {
+	if u.version != v {
+		log.Errorf("############## Version set to %s", v)
```
lixmal: That's not the typical error format we use

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,118 @@
...
+		u.version = v
+		go u.CheckForUpdates()
+	}
+}
+
+func (u *UpdateManager) CheckForUpdates() {
```
lixmal: This should take ctx to avoid leaking goroutines

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,118 @@
...
+	if u.version == "disabled" {
+		log.Trace("Skipped checking for updates, auto-update is disabled")
+		return
+	}
+	currentVersionString := version.NetbirdVersion()
+	updateVersionString := u.version
+	if updateVersionString == "latest" || updateVersionString == "" {
```
lixmal: Const here too

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,118 @@
...
+		return "", err
+	}
+	fileNameParts := strings.Split(fileURL, "/")
+	out, err := os.Create(filepath.Join(tempDir, fileNameParts[len(fileNameParts)-1]))
+	if err != nil {
+		return "", err
+	}
+	defer out.Close()
+
+	resp, err := http.Get(fileURL)
```
lixmal: This should take context and/or timeout

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
```
lixmal: Let's log errors

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,118 @@
...
+	tempDir, err := os.MkdirTemp("", "netbird-installer-*")
+	if err != nil {
+		return "", err
+	}
+	fileNameParts := strings.Split(fileURL, "/")
+	out, err := os.Create(filepath.Join(tempDir, fileNameParts[len(fileNameParts)-1]))
+	if err != nil {
+		return "", err
+	}
+	defer out.Close()
```
lixmal: Let's log errors

client/internal/updatemanager/update_darwin.go
```diff
@@ -0,0 +1,109 @@
...
+		return err
+	}
+
+	fileSysInfo, ok := fileInfo.Sys().(*syscall.Stat_t)
+	if !ok {
+		return fmt.Errorf("Error checking file owner, sysInfo type is %T not *syscall.Stat_t", fileInfo.Sys())
+	}
+
+	// Get user name from UID
+	cmd := exec.Command("id", "-nu", fmt.Sprintf("%d", fileSysInfo.Uid))
```
lixmal: Let's use Go APIs

client/internal/updatemanager/update_darwin.go
```diff
@@ -0,0 +1,109 @@
...
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return err
+	}
+	userName := strings.TrimSpace(string(out))
+
+	// Get user HOME, required for brew to run correctly
+	// https://github.com/Homebrew/brew/issues/15833
+	cmd = exec.Command("sudo", "-u", userName, "sh", "-c", "echo $HOME")
```
lixmal: Let's use Go APIs

robertgro:
Related

https://github.com/netbirdio/netbird/issues/1793

https://github.com/netbirdio/netbird/issues/4019

client/internal/updatemanager/update_windows.go
mohamed-essam: There are no installers for arm64 in the release pages though

client/internal/updatemanager/update_windows.go
mohamed-essam: Makes sense, what would you suggest instead? Simply matching on `err != nil`?
I'm worried any random error could cause an EXE installation to be mistaken for MSI

mohamed-essam:
@lixmal  All comments should be resolved, will need to conduct another round of testing however, so will keep the PR as Draft until testing is complete

client/internal/updatemanager/update_windows.go
robertgro: Excuse the intervention. To bypass localization issues here...

Either go for another reg key like e.g. 
```
SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\Netbird
```

which returns a reliable result - in my debugging test case - when trying to ```registry.OpenKey``` with ```registry.READ```:
- if installed via .exe, no error thrown (always)
- if installed via .msi, an/the error "_The system cannot find the file specified_." is always thrown

but this approach could be troublesome (as Microsoft states) concerning the availability of the key on x86 machines: 

_[only]_

> If they're present


(see https://learn.microsoft.com/en-us/troubleshoot/windows-client/application-management/wow6432node-registry-key-present-32-bit-machine#cause)

The original provided path
```
SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\Netbird
````
on the other hand always throws a "_The system cannot find the file specified._" when ```registry.OpenKey``` and ```registry.READ``` on my system, regardless if I installed it via .exe or .msi.

---

Alternatively replace ```registry.READ``` completely at this point here due to the Golang language authors presumably not giving this a priority as their language obviously seems to have a hard time getting a reliable read operation on a windows registry key done. Down the rabbit hole, here we go: https://github.com/golang/go/issues/35730 and https://ohlinger.co/posts/golang-and-windows-registry/ (credits/kudos to the author who fixed this by switching to Python). I'm currently lacking ideas on how to query this better in Golang.

Comment (robertgro) [reply client/internal/updatemanager/update_windows.go]: So a possible solution may be a nested one:
```
	const uninstallKeyPath = `SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\Netbird`
	const uninstallKeyPath2 = `SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Netbird`

	k, err := registry.OpenKey(registry.LOCAL_MACHINE, uninstallKeyPath, registry.READ)
	if err != nil {
		kv, err := registry.OpenKey(registry.LOCAL_MACHINE, uninstallKeyPath2, registry.READ)
		if err != nil {
			return err
		}
		defer kv.Close()
                return kv
	}
	defer k.Close()
        return k
```

Why do we need to double check here? Obviously when a 32-bit process tries to access certain registry keys, Windows redirects it to the WOW6432Node. Sometimes there can even be redirecting issues, so this extra verification is required.

In case Netbird has been installed via .msi:
- this always throws an error

In case Netbird has been installed via .exe:
- no error thrown

File: client/internal/updatemanager/update_windows.go (hunk 3 above)
Comment (lixmal) [reply client/internal/updatemanager/update_windows.go]: There will be https://github.com/netbirdio/netbird/pull/4206

File: client/internal/updatemanager/update_windows.go (hunk 4 above)
Comment (mohamed-essam) [reply client/internal/updatemanager/update_windows.go]: Thank you @robertgro for the suggestion :pray: 
I implemented your method and pushed it, will also double check it on my local machine

File: client/internal/updatemanager/update_darwin.go [hunk 15]
```diff
@@ -0,0 +1,111 @@
...
+	// Thus it will always update to latest and ignore targetVersion
+	upgradeArgs := []string{"-u", userName, "/opt/homebrew/bin/brew", "upgrade", "netbirdio/tap/netbird"}
+	// Check if netbird-ui is installed
+	cmd := exec.Command("brew", "info", "--json", "netbirdio/tap/netbird-ui")
+	err = cmd.Run()
+	if err == nil {
+		// netbird-ui is installed
+		upgradeArgs = append(upgradeArgs, "netbirdio/tap/netbird-ui")
+	}
+	cmd = exec.Command("sudo", upgradeArgs...)
```
lixmal: Can this get stuck if sudo is configured to ask for a password?

client/internal/updatemanager/update_darwin.go
```diff
@@ -0,0 +1,111 @@
...
+func (u *UpdateManager) triggerUpdate(targetVersion string) error {
+	cmd := exec.Command("pkgutil", "--pkg-info", "io.netbird.client")
+	outBytes, err := cmd.Output()
+	if err != nil && cmd.ProcessState.ExitCode() == 1 {
+		// Not installed using pkg file, thus installed using Homebrew
+
+		return u.updateHomeBrew()
+	}
+	// Installed using pkg file
+	path, err := downloadFileToTemporaryDir(u.ctx, urlWithVersionArch(pkgDownloadURL, targetVersion))
```
lixmal: I suggest we also add an optional hash to mgmt that we can compare here after download. At least for non-latest, it will be tricky with latest.
Thoughts?

lixmal (commented):
Have you put any thoughts into how we could do a rollback if the new version breaks connectivity?

client/internal/updatemanager/update_darwin.go
mohamed-essam: It _can_ but I don't believe I've ever seen a system where sudo is configured to ask for password from root, and as far as I understand, the NetBird daemon always runs as root, no?

client/internal/updatemanager/update_darwin.go
mohamed-essam: It would make sense, but I believe this can be added in a later release, but I believe it'd make more sense to verify the hash from Github since it's already there (and we already know the exact version we want to upgrade to) no?

client/internal/updatemanager/update_darwin.go
lixmal: It can in theory run without root in netstack mode. But I doubt anyone has deployed this with a non-root launch agent. Maybe we can add a reasonable timeout (minutes?) after which we terminate the cmd

mohamed-essam:
@lixmal 
> Have you put any thoughts into how we could do a rollback if the new version breaks connectivity?

Based on my discussions with @mlsmaycon , the upgrade mechanism is forward-moving only

client/internal/updatemanager/update_darwin.go
lixmal: The attack vector here is a manipulated release in github (even if signed). Pulling hashes from github in this case would only help file integrity, not security.
But yes, can be a follow-up

mohamed-essam:
@lixmal If it's alright with you, I will mark the PR as no documentation needed for the time being as the dashboard changes are not yet done, as soon as dashboard can modify the auto upgrade version I will open a PR in docs

client/internal/updatemanager/update_darwin.go
lixmal: Maybe we can run `go u.CheckForUpdates()` with a deadline `context.Context` in general, and it will pass/check the context where  possible (e.g. `cmd`, `http.Client`, ...).

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,154 @@
...
+	update.SetDaemonVersion(version.NetbirdVersion())
+	update.SetOnUpdateListener(manager.CheckForUpdates)
+	return manager
+}
+
+func (u *UpdateManager) SetVersion(v string) {
+	if u.version != v {
+		log.Tracef("Auto-update version set to %s", v)
+		u.version = v
+		go u.CheckForUpdates()
```
lixmal: This will run once per network map update, right? It should  debounce internally to avoid running multiple of these.

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,154 @@
...
+		cancel:         cancel,
+		version:        disableAutoUpdate,
+	}
+	update.SetDaemonVersion(version.NetbirdVersion())
+	update.SetOnUpdateListener(manager.CheckForUpdates)
+	return manager
+}
+
+func (u *UpdateManager) SetVersion(v string) {
+	if u.version != v {
```
lixmal: Let's either add a mutex here or a comment on top of the method that this is synchronized externally (`syncMsgMux`). I'd prefer a mutex because we also access it in line 66 outside the syncMsgMux lock

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,154 @@
...
+	lastTrigger    time.Time
+	statusRecorder *peer.Status
+}
+
+func NewUpdateManager(statusRecorder *peer.Status) *UpdateManager {
+	update := version.NewUpdate("nb/client")
+	ctx, cancel := context.WithCancel(context.Background())
+	manager := &UpdateManager{
+		update:         update,
+		lastTrigger:    time.Now().Add(-10 * time.Minute),
```
lixmal: This field is not thread-safe

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (u *UpdateManager) CheckForUpdates() {
+	if u.version == disableAutoUpdate {
+		log.Trace("Skipped checking for updates, auto-update is disabled")
+		return
+	}
+	currentVersionString := version.NetbirdVersion()
+	updateVersionString := u.version
+	if updateVersionString == latestVersion || updateVersionString == "" {
+		if u.update.LatestAvailable == nil {
```
lixmal: `LatestAvailable` is accessed without the lock of `update`

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,154 @@
...
+				"Your client version is older than auto-update version set in Management, updating client now.",
+				nil,
+			)
+			err = u.triggerUpdate(updateVersionString)
+			if err != nil {
+				log.Errorf("Error triggering auto-update: %v", err)
+			}
+		}
+	} else {
+		log.Trace("Current version is equal to or higher than auto-update version")
```
lixmal: Let's make this debug and add which versions are involved

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,154 @@
...
+		return
+	}
+	currentVersionString := version.NetbirdVersion()
+	updateVersionString := u.version
+	if updateVersionString == latestVersion || updateVersionString == "" {
+		if u.update.LatestAvailable == nil {
+			log.Tracef("Latest version not fetched yet")
+			return
+		}
+		updateVersionString = u.update.LatestAvailable.String()
```
lixmal: There's a race here. If the client starts up, gets a network map (and runs this method), the version fetcher is most likely not ready yet, so this will end up as `0.0.0` -> update won't start.

mohamed-essam: That's alright, because as soon as the version fetcher is ready, it will call the onUpdateListener, in which case the update will run
The method is structured so that change in either update fetcher or autoUpdateVersion will trigger the method

client/internal/updatemanager/manager.go
mohamed-essam: Locked entire `CheckForUpdates` method with a mutex

client/internal/updatemanager/manager.go
mohamed-essam: Locked entire `CheckForUpdates` method with a mutex

client/internal/updatemanager/manager.go
mohamed-essam: Locked entire `CheckForUpdates` method with a mutex

client/internal/updatemanager/manager.go
lixmal: Alright

client/internal/updatemanager/manager.go
mohamed-essam: Locked entire `CheckForUpdates` method with a mutex, if there's an update we won't need to worry about it anyway since the entire Process will shutdown

client/internal/updatemanager/manager.go
lixmal:
```
	if u.version != v {
```
is accessed here outside that lock

client/internal/updatemanager/manager.go
lixmal: Right, but that's a different lock than the one `update` uses when it accesses LatestAvailable. I don't think it should be exported in the first place

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (u *UpdateManager) SetVersion(v string) {
+	if u.version != v {
+		log.Tracef("Auto-update version set to %s", v)
+		u.version = v
+		go u.CheckForUpdates()
+	}
+}
+
+func (u *UpdateManager) Stop() {
```
pappz, line 138: It is not safe. Just because you cancel the context does not mean it will stop gracefully. The engine can be restarted at any time, and it may run a parallel update. The engine must wait for the UpdateManager to stop gracefully.

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,158 @@
...
+	u.update.StopWatch()
```
pappz: The `u.update.StopWatch() `contains a bug. It can miss the "channel close" event because the channel is unbuffered. It leak a go thread.

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,158 @@
+package updatemanager
+
+import (
+	"context"
+	"fmt"
+	"github.com/netbirdio/netbird/client/internal/peer"
```
pappz: Please follow our order rules of import (std lib, thirparty, NetBird)

client/internal/updatemanager/manager.go
mohamed-essam: I don't understand your point, as I can see, since the channel is unbuffered, it will wait for the other goroutine to recieve the event in https://github.com/netbirdio/netbird/pull/4256/files#diff-9bf0202e6bc7e6197efaa70257c72730aac7b94dd106bf73ec4dce218fc58e29R62 and won't return until the other goroutine is dead, and I don't see the channel closed anywhere

pappz: We have two goroutines. If you run the 2. routine at that moment when the u.checkUpdate is in progress, then the 2. thread writes nothing to the channel, just use the default path. You are right, do not have channels closed. I mean that "close" event on channels.
1.
```
select {
		case <-u.fetchDone:
			return
		case <-u.fetchTicker.C:
			if changed := u.fetchVersion(); changed {
				u.checkUpdate()
			}
		}
```
2.
```
select {
	case u.fetchDone <- struct{}{}:
	default:
	}
```

mohamed-essam:
@pappz @lixmal I've changed the way goroutines and contexts are handled for the PR:

1. version.Update now closes the channel synchronously, so `updatemanager.Stop()` should now wait for version.Update's goroutine to exit.
2. Un-exported version.Update.latestVersion and made it so that it would send a copy of the latest version to the onUpdate handler instead.
3. Added a mutex lock around the entire update logic.
4. Added a child context with Deadline for all update logic.
5. Added a waitGroup for all update logic to ensure `updatemanager.Stop()` completely waits for updatemanager to exit gracefully.
6. Ensured context done would fast-fail update logic if context was done before it obtained mutex lock.

version/update.go
```diff
@@ -57,11 +57,7 @@ func NewUpdate(httpAgent string) *Update {
 // StopWatch stop the version info fetch loop
 func (u *Update) StopWatch() {
 	u.fetchTicker.Stop()
-
-	select {
-	case u.fetchDone <- struct{}{}:
-	default:
-	}
+	u.fetchDone <- struct{}{}
```
pappz: The StopWatch is exported, which means it can be run multiple times. This can cause a deadlock because the channel write may block (based on a quick investigation, this could originate from the engine code). I recommend using a context to handle stopping.

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func (u *UpdateManager) Stop() {
+	u.update.StopWatch()
+	u.cancel()
+	u.waitGroup.Wait()
+}
+
+func (u *UpdateManager) Updated(latestVersion string) {
+	u.waitGroup.Add(1)
```
pappz: If you increase the `waitGroup` inside the goroutine where you decrease it, it can lead to an inconsistent state.
Imagine this situation: you run the `SetVersion(v)` function, which unlocks the mutex and returns, but the scheduler hasn’t yet picked up the go `u.Updated("N/A")` goroutine. If `Stop()` runs immediately afterward, your `u.Updated` will execute after Stop().

To avoid this, simply call `u.waitGroup.Add(1) `before starting the goroutine.

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,183 @@
...
+	defer u.waitGroup.Done()
+	u.mutex.Lock()
+	defer u.mutex.Unlock()
+	select {
+	case <-u.ctx.Done():
+		return
+	default:
+	}
+	if latestVersion != "N/A" {
```
pappz: "N/A" looks strange. Use a const, or better, if you pass a nil

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,183 @@
...
+	u.cancel()
+	u.waitGroup.Wait()
+}
+
+func (u *UpdateManager) Updated(latestVersion string) {
+	u.waitGroup.Add(1)
+	defer u.waitGroup.Done()
+	u.mutex.Lock()
+	defer u.mutex.Unlock()
+	select {
```
pappz: What is the goal of this select logic? It basically doing nothing because you can cancel the context outside of the locked scope

client/internal/updatemanager/manager.go
mohamed-essam: One problem here is that u.Updated is also called from inside update.Version, which I can't pass the waitGroup to

version/update.go
mohamed-essam: Changed slightly, since each UpdateManager instance retains its own copy of version.Update, this should prevent multiple calls
```go
	u.mutex.Lock()
	defer u.mutex.Unlock()
	if u.update != nil {
		u.update.StopWatch()
		u.update = nil
	}
```

client/internal/updatemanager/manager.go
hakansa: aggre with @pappz, current behaviour against the best practices. Instead of passing goroutine to `*UpdateManager.Updated`, you should increment the wg before you call it. e.g you should do sth like this on line 64:

```go
go func (wg *sync.WaitGroup){
    wg.Add(1)
    defer wg.Done()
    u.Updated(unknownVersion)
}(wg)


```

mohamed-essam: I understand what to do regarding line 64's call of Updated, but I'm not sure what to do regarding version.Update's call

hakansa: I'm not super familiar with the design but, why do you need a waitGroup there and a mutex it not enough? There will be only one goroutine handle updates right?

mohamed-essam: There is also the goroutine from version.Update (which handles fetching the latest version)
I'm thinking of revamping the design here a bit by changing version.Update to send latest version to a channel, and have one goroutine handling actual update logic
What do you think?

hakansa: Yeah I was also thinking a similar approach. Just I'm not sure about the channel thing. `version.Update` already gets a callback with `SetOnUpdateListener`right? Does setting a callback and updating the latest works?

mohamed-essam: This is the current setup, the issue is that version.Update runs in its own goroutine and spawns a new goroutine for each onUpdate, which causes the number of goroutines running Updated to be (theoretically) unlimited, practically it should only trigger once every release (+ once on connect for getting the latest release then)
And for the SetVersion it's triggered per Sync (but only calls Updated when the version is different) 
In practice this update logic isn't highly parallel, so everything we're talking about here is an extreme edge case

lixmal:
@mohamed-essam what happens if the upgrade fails? Since we don't have a rollback planned, could this end up in an unusable state?

For example, the windows installer uninstalls netbird first, what happens if the install part fails?

mohamed-essam:
@lixmal There's nothing we can do from inside the service, especially that (for example) the windows installer stops the service, then uninstalls, and installs.

I feel like this should be more in the responsibility of the installer itself, (i.e the installer should rollback its own changes on failure), I'm not sure if other installation methods are able to handle this that way, but that will require a lot of work on multiple installation methods

Although I'm sure some installation methods do in fact handle failure more gracefully (e.g apt doesn't restart the service until after installation is complete)

version/update.go
pappz: What is about with the server.go,` func (s *BaseServer) Stop() error`? The behavior of StopWatch has been changed. It it fully compatible with the upper layers?

mohamed-essam: I quickly checked all references to StopWatch and I believe it should be fine that way

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,193 @@
...
+		log.Tracef("Auto-update version set to %s", v)
+		u.version = v
+		u.mutex.Unlock()
+		u.updateChannel <- unknownVersion
+	} else {
+		u.mutex.Unlock()
+	}
+}
+
+func (u *UpdateManager) Stop() {
```
pappz, line 138: It is an exported function, so be sure it operates correctly and never causes a deadlock. If the engine calls Stop() multiple times (and it does, as I mentioned earlier), the second call will block forever.

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,193 @@
...
+	u.mutex.Lock()
+	defer u.mutex.Unlock()
+	if u.update != nil {
+		u.update.StopWatch()
+		u.update = nil
+	}
+	<-u.doneChannel
+}
+
+func (u *UpdateManager) UpdateLoop() {
```
pappz: Looks like you never call it from externally. It should not be exported.

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,193 @@
...
+			}
+			u.mutex.Unlock()
+			ctx, cancel := context.WithDeadline(u.ctx, time.Now().Add(time.Minute))
+			u.CheckForUpdates(ctx)
+			cancel()
+		}
+	}
+}
+
+func (u *UpdateManager) CheckForUpdates(ctx context.Context) {
```
pappz: Looks like you never call it from externally. It should not be exported.

version/update.go
```diff
@@ -169,6 +181,9 @@ func (u *Update) checkUpdate() bool {
 
 	u.listenerLock.Lock()
 	defer u.listenerLock.Unlock()
+	if u.onUpdateChannel != nil {
+		u.onUpdateChannel <- u.latestAvailable.String()
```
pappz: If you don’t drain the channel outside of this package, it will cause a deadlock in this struct. It’s risky to operate on a channel provided by an external source. Fortunately, since you only need to maintain one listener, you can create the channel internally and read from it through a function—or handle it however you prefer.
```
func (u *Update) Updates() <-chan string {
    return u.updates
}
```

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,193 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	manager := &UpdateManager{
+		update:         update,
+		lastTrigger:    time.Now().Add(-10 * time.Minute),
+		statusRecorder: statusRecorder,
+		ctx:            ctx,
+		cancel:         cancel,
+		version:        disableAutoUpdate,
+		latestVersion:  unknownVersion,
+		updateChannel:  make(chan string, 4),
```
pappz: Why 4, why not 1? If I understand the logic, you do not care about all updates, you care only about the latest.

mohamed-essam: Yes, but this is just to make sure that it doesn't lock the engine when it's setting a new autoUpdate version if the autoUpdate thread is busy, I thought about making it 2, but then I thought what's a few more bytes

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,193 @@
...
+	latestVersion  string
+	update         *version.Update
+	lastTrigger    time.Time
+	statusRecorder *peer.Status
+	mutex          sync.Mutex
+	updateChannel  chan string
+	doneChannel    chan struct{}
+}
+
+func NewUpdateManager(ctx context.Context, statusRecorder *peer.Status) *UpdateManager {
```
pappz: You pass a context, but it is not clarified how this works. How can we free up this resource?
- Do we have to call Stop() explicitly?
- Or can we free it just by canceling the context from the upper layer?

client/internal/engine.go
```diff
@@ -659,6 +668,9 @@ func (e *Engine) handleSync(update *mgmProto.SyncResponse) error {
 	e.syncMsgMux.Lock()
 	defer e.syncMsgMux.Unlock()
 
+	if update.GetAutoUpdateVersion() != "skip" {
```
pappz: Where come from this "skip" string? Can we use a const?

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,193 @@
...
+		u.mutex.Unlock()
+		u.updateChannel <- unknownVersion
+	} else {
+		u.mutex.Unlock()
+	}
+}
+
+func (u *UpdateManager) Stop() {
+	u.cancel()
+	u.mutex.Lock()
```
pappz: What do you protect here? If I see well, you do not write the update just in the constructor.

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+}
+
+func (u *UpdateManager) CheckForUpdates(ctx context.Context) {
+	if u.version == disableAutoUpdate {
+		log.Trace("Skipped checking for updates, auto-update is disabled")
+		return
+	}
+	currentVersionString := version.NetbirdVersion()
+	updateVersionString := u.version
```
pappz: u.version is not protected by a lock

management/server/token_mgr.go
```diff
@@ -210,6 +210,7 @@ func (m *TimeBasedAuthSecretsManager) pushNewTURNAndRelayTokens(ctx context.Cont
 		NetbirdConfig: &proto.NetbirdConfig{
 			Turns: turns,
 		},
+		AutoUpdateVersion: "skip",
```
pappz: It makes no sense to mix the AutoUpdateVersion feature with the `TimeBasedAuthSecretsManager`
@pascal-fischer How can we handle these cases in protobuf? I do not think the "optional" keyword is usable there. Should we resend the expected version in every single network map update?

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,247 @@
...
+
+func (u *UpdateManager) handleUpdate(ctx context.Context) {
+	var updateVersion *v.Version
+
+	u.expectedVersionMutex.Lock()
+	expectedVersion := u.expectedVersion
+	u.expectedVersionMutex.Unlock()
+
+	// Resolve "latest" to actual version
+	if u.updateToLatestVersion {
```
pappz: It is outside of the protected scope. Create a temp variable for this, too.

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,247 @@
...
+		if !u.isVersionAvailable() {
```
pappz: Please use a new variable for this, because if you run it twice, the data might become deprecated during this period.

management/server/token_mgr.go
```diff
@@ -210,6 +210,7 @@ func (m *TimeBasedAuthSecretsManager) pushNewTURNAndRelayTokens(ctx context.Cont
 		NetbirdConfig: &proto.NetbirdConfig{
 			Turns: turns,
 		},
+		AutoUpdateVersion: skipAutoUpdateVersion,
```
pascal-fischer: Why do we have to send this even on messages that should not care about this? The absence of the string should indicate that we do not handle it. There is no need to send a skip string. AutoUpdateVersion should be optional then

mohamed-essam: It was agreed with @mlsmaycon that empty == latest, this is why

pascal-fischer: @mlsmaycon Why???

management/server/account.go
```diff
@@ -471,6 +475,15 @@ func (am *DefaultAccountManager) handleGroupsPropagationSettings(ctx context.Con
...
 }
 
+func (am *DefaultAccountManager) handleAutoUpdateVersionSettings(ctx context.Context, oldSettings, newSettings *types.Settings, userID, accountID string) {
+	if oldSettings.AutoUpdateVersion != newSettings.AutoUpdateVersion {
+		am.StoreEvent(ctx, userID, accountID, accountID, activity.AccountAutoUpdateVersionUpdated, map[string]any{
+			"old_value": oldSettings.AutoUpdateVersion,
+			"new_value": newSettings.AutoUpdateVersion,
+		})
+	}
+}
```
bcmmbaga, line 515: I think we  could simplifying the meta to just store current version instead of both `old_value` and `new_value`. The event sequence already captures the change history, adding it again would be redundant.
 
 ```suggestion
func (am *DefaultAccountManager) handleAutoUpdateVersionSettings(ctx context.Context, oldSettings, newSettings *types.Settings, userID, accountID string) {
	if oldSettings.AutoUpdateVersion != newSettings.AutoUpdateVersion {
		am.StoreEvent(ctx, userID, accountID, accountID, activity.AccountAutoUpdateVersionUpdated, map[string]any{
			"version": newSettings.AutoUpdateVersion,
		})
	}
}
```

**Review by bcmmbaga — CHANGES_REQUESTED** [review]

**Comment by ghost** [comment]
Hey,

Just wondering when is this expected to be released in production ?

Best,

**Comment by yshah-aromatech** [comment]
Hi Team, 

Do we have an expected date for this release? 

Thanks

**Comment by wgetnz** [comment]
Do we have an expected date for this release?

**Comment by BashBandito** [comment]
Very much looking forward to this feature. 

Thanks for the good work.

File: client/internal/updatemanager/manager.go [hunk 44]
```diff
@@ -0,0 +1,261 @@
...
+		mgmUpdateChan:  make(chan struct{}, 1),
+		updateChannel:  make(chan struct{}, 1),
+		currentVersion: version.NetbirdVersion(),
+		updateFunc:     triggerUpdate,
+		update:         version.NewUpdate("nb/client"),
+	}
+	return manager
+}
+
+func (u *UpdateManager) WithCustomVersionUpdate(versionUpdate UpdateInterface) *UpdateManager {
```
pappz: Used by test only. Do not mix prod and test codes

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (u *UpdateManager) Stop() {
+	if u.cancel == nil {
+		return
+	}
+
+	u.cancel()
+	if u.update != nil {
+		u.update.StopWatch()
+		u.update = nil
```
pappz, line 147: You set u.update = nil in Stop() while updateLoop or handleUpdate could still try to access it.

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		log.Errorf("Error checking for update, error parsing version `%s`: %v", u.currentVersion, err)
+		return false
+	}
+	if currentVersion.GreaterThanOrEqual(updateVersion) {
+		log.Debugf("Current version (%s) is equal to or higher than auto-update version (%s)", u.currentVersion, updateVersion)
+		return false
+	}
+
+	if time.Since(u.lastTrigger) < 5*time.Minute {
```
pappz, line 323: I am not sure but we can miss a potential upgrade if we just ignore. What is the goal with this condition?

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,261 @@
...
+	default:
+		log.Debugf("No expected version information set")
+		return
+	}
+
+	if !u.shouldUpdate(updateVersion) {
+		return
+	}
+
+	ctx, cancel := context.WithDeadline(ctx, time.Now().Add(time.Minute))
```
pappz: WithTimeout simpler

pappz (changes requested):

client/internal/updatemanager/manager.go
mohamed-essam: This is so it doesn't spam retrying to upgrade with any trigger since each failure will show an error to the user, so this is just a way to backoff failures

client/internal/updatemanager/manager.go
mohamed-essam: This is only referenced inside `handleUpdate` once inside a mutex, so I used this to prevent a race and also added a check so that if u.update is nil it should return

nazarewk:
fixes https://github.com/netbirdio/netbird/issues/4019 https://github.com/netbirdio/netbird/issues/1793 https://github.com/netbirdio/netbird/issues/832

PR #4268, opened 2025-08-01
[management, client] Add logout feature

lixmal:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


crn4 (approved):
LGTM

client/server/server.go
```diff
@@ -811,26 +792,201 @@ func (s *Server) Down(ctx context.Context, _ *proto.DownRequest) (*proto.DownRes
...
+
+func (s *Server) Logout(ctx context.Context, msg *proto.LogoutRequest) (*proto.LogoutResponse, error) {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+
+	if msg.ProfileName != nil && *msg.ProfileName != "" {
+		return s.handleProfileLogout(ctx, msg)
+	}
+
+	return s.handleActiveProfileLogout(ctx)
```
hakansa, line 838: should we allow logout for other users' profiles? 

we allow "down" for other users' profiles but logout is more destructive since we delete the peer from the management.

lixmal: >should we allow logout for other users' profiles?

I agree we shouldn't. However, currently, anyone who can connect to the daemon can claim to be any user, so this enforcement on the daemon would be toothless.

hakansa (approved):

PR #4272, opened 2025-08-04
[management,client] Feat/exit node selection

hakansa:
## Describe your changes
[management,client] Feat/exit node selection
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor


shared/management/http/api/openapi.yml
```diff
@@ -1323,6 +1323,10 @@ components:
           items:
             type: string
             example: "chacbco6lnnbn6cg5s91"
+        is_not_forced:
```
mlsmaycon: Let's use `skip_auto_apply` to avoid confusion with the client behavior where selected or deselected routes dismiss this.

shared/management/proto/management.proto
```diff
@@ -400,6 +400,7 @@ message Route {
   string NetID = 7;
   repeated string Domains = 8;
   bool keepRoute = 9;
+  bool isNotForced = 10;
```
mlsmaycon: same as openapi

client/internal/routemanager/vars/vars.go
```diff
@@ -13,4 +13,6 @@ var (
 
 	Defaultv4 = netip.PrefixFrom(netip.IPv4Unspecified(), 0)
 	Defaultv6 = netip.PrefixFrom(netip.IPv6Unspecified(), 0)
+
+	ExitNodeCIDR = "0.0.0.0/0"
```
mlsmaycon, line 17: should this be a const?

management/server/store/store.go
```diff
@@ -368,6 +368,9 @@ func getMigrationsPostAuto(ctx context.Context) []migrationFunc {
 				}
 			})
 		},
+		func(db *gorm.DB) error {
```
mlsmaycon: don't think this is necessary, default is false value

mlsmaycon (dismissed):

client/internal/routemanager/vars/vars.go
hakansa: it's getting hard to keep track hardcoded values

client/internal/routemanager/manager.go
```diff
@@ -583,3 +587,106 @@ func resolveURLsToIPs(urls []string) []net.IP {
...
+	} else {
+		info.userDeselected = append(info.userDeselected, netID)
+	}
+}
+
+func (m *DefaultManager) checkManagementSelection(routes []*route.Route, netID route.NetID, info *exitNodeInfo) {
+	for _, route := range routes {
+		if !route.SkipAutoApply {
+			info.selectedByManagement = append(info.selectedByManagement, netID)
+			break
```
pappz, line 646: If a routes slice has mixed SkipAutoApply values, the first non-skipped one wins and the rest are ignored. Is it the expected?

hakansa: yep, this part only works exit nodes and if user didn't override the selection

pappz (approved):

PR #4275, opened 2025-08-04
[management] expire invalid peer

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -63,7 +63,7 @@ require (
 	github.com/miekg/dns v1.1.59
 	github.com/mitchellh/hashstructure/v2 v2.0.2
 	github.com/nadoo/ipset v0.5.0
-	github.com/netbirdio/management-integrations/integrations v0.0.0-20250804141010-70162f3e9444
+	github.com/netbirdio/management-integrations/integrations v0.0.0-20250805075620-df61803a4461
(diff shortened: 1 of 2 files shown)


management/server/account.go
```diff
@@ -1700,8 +1700,24 @@ func (am *DefaultAccountManager) GetDNSDomain(settings *types.Settings) string {
...
+		peers = append(peers, peer)
+	}
+	if len(peers) > 0 {
+		err := am.expireAndUpdatePeers(ctx, accountID, peers)
+		if err != nil {
+			log.WithContext(ctx).Errorf("failed to expire and update invalidated peers for account %s: %v", accountID, err)
+		}
+	}
+	log.WithContext(ctx).Debugf("validated peers have been invalidated for account %s", accountID)
 	am.BufferUpdateAccountPeers(ctx, accountID)
```
bcmmbaga: When expiring peers, we also call `am.UpdateAccountPeers(ctx, accountID)`. Can we ensure that `am.BufferUpdateAccountPeers(ctx, accountID)` is only called when there are no peers to expire?

pascal-fischer: You are right. The Expiration should always have at least one peer in the list so we can depend on the peer expiration to do the buffer update. Updated.

PR #4276, opened 2025-08-04
[client]: Add config flag to service to override default profile path

hakansa:
## Describe your changes
[client]: Add config flag to service to override default profile path
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

### DIFF
--- a/client/cmd/up.go
+++ b/client/cmd/up.go
@@ -89,11 +89,6 @@ func upFunc(cmd *cobra.Command, args []string) error {
 
 	cmd.SetOut(cmd.OutOrStdout())
 
-	// Check if deprecated config flag is set and show warning
-	if cmd.Flag("config").Changed && configPath != "" {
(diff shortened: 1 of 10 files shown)


nazarewk:
We should probably default to `/etc/netbird/config.json` instead of `/var/lib/netbird/default.json`, so we don't break stuff we're not aware of (community packages, router OS configurations being lost etc.).

client/cmd/up.go
```diff
@@ -89,6 +89,11 @@ func upFunc(cmd *cobra.Command, args []string) error {
 
 	cmd.SetOut(cmd.OutOrStdout())
 
+	// Check if deprecated config flag is set and show warning
+	if cmd.Flag("config").Changed && configPath != "" {
+		cmd.PrintErrf("Warning: Config flag is deprecated on up command, it should be set as a service argument with $NB_CONFIG environment or with \"-config\" flag; netbird service reconfigure --service-env=\"NB_CONFIG=<file_path>\" or netbird service run --config=<file_path>\n")
```
lixmal: How do we handle foreground mode?

hakansa: oh nice catch! Just pushed the change for that.

PR #4281, opened 2025-08-04
[client] Fix/ice handshake

pappz:
## Describe your changes

This PR improves the handling of Trickle ICE offer/answer exchange in cases where both peers send offers simultaneously before an ICE connection is established. While simultaneous offers are allowed, gRPC signaling buffers these messages, and under high message volume, outdated offers may still be delivered to the ICE agent. In the worst-case scenario, one side may create an ICE agent after receiving a deprecated answer to its offer, while the other side is still waiting for a timeout. After a timeout time roles are reversed.

In this PR, speed up the GRPC message processing, force the recreation of the ICE agent when getting a new, remote offer (do not wait for local STUN timeout).

### Key Improvements
- Improved gRPC message processing speed to reduce latency and minimize stale signaling data.
- ICE agent is now forcefully recreated upon receiving a new remote offer, rather than waiting for a local STUN timeout.
- Introduced a dedicated worker thread for decrypting signaling messages.
- Implemented a session ID mechanism that updates on ICE agent disconnection. This helps discard outdated signaling messages from the gRPC buffer, preventing the reprocessing of stale or irrelevant ICE offers.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

### DIFF
--- a/client/internal/peer/worker_ice.go
+++ b/client/internal/peer/worker_ice.go
@@ -234,6 +234,7 @@ func (w *WorkerICE) connect(ctx context.Context, agent *ice.Agent, remoteOfferAn
 	w.log.Debugf("gather candidates")
 	if err := agent.GatherCandidates(); err != nil {
 		w.log.Warnf("failed to gather candidates: %s", err)
+		w.closeAgent(agent, w.agentDialerCancel)
 		return
 	}
 
@@ -241,12 +242,14 @@ func (w *WorkerICE) connect(ctx context.Context, agent *ice.Agent, remoteOfferAn
 	remoteConn, err := w.turnAgentDial(ctx, remoteOfferAnswer)
 	if err != nil {
 		w.log.Debugf("failed to dial the remote peer: %s", err)
+		w.closeAgent(agent, w.agentDialerCancel)
 		return
 	}
 	w.log.Debugf("agent dial succeeded")
 
 	pair, err := agent.GetSelectedCandidatePair()
 	if err != nil {
+		w.closeAgent(agent, w.agentDialerCancel)
 		return
 	}
 


client/internal/peer/worker_ice.go
```diff
@@ -96,63 +132,20 @@ func (w *WorkerICE) OnNewOffer(remoteOfferAnswer *OfferAnswer) {
...
-		LocalIceCandidateType:      pair.Local.Type().String(),
-		RemoteIceCandidateType:     pair.Remote.Type().String(),
-		LocalIceCandidateEndpoint:  fmt.Sprintf("%s:%d", pair.Local.Address(), pair.Local.Port()),
-		RemoteIceCandidateEndpoint: fmt.Sprintf("%s:%d", pair.Remote.Address(), pair.Remote.Port()),
-		Relayed:                    isRelayed(pair),
-		RelayedOnLocal:             isRelayCandidate(pair.Local),
-	}
-	w.log.Debugf("on ICE conn is ready to use")
-	go w.conn.onICEConnectionIsReady(selectedPriority(pair), ci)
+	go w.connect(dialerCtx, agent, remoteOfferAnswer)
```
lixmal, line 148: If w.connect fails internally and returns early, should we reset agentConnection?

client/internal/peer/worker_ice.go
```diff
@@ -190,76 +187,124 @@ func (w *WorkerICE) Close() {
...
+	return w.sessionID
+}
+
+// will block until connection succeeded
+// but it won't release if ICE Agent went into Disconnected or Failed state,
+// so we have to cancel it with the provided context once agent detected a broken connection
+func (w *WorkerICE) connect(ctx context.Context, agent *ice.Agent, remoteOfferAnswer *OfferAnswer) {
+	w.log.Debugf("gather candidates")
+	if err := agent.GatherCandidates(); err != nil {
+		w.log.Debugf("failed to gather candidates: %s", err)
```
lixmal: Should the errors in this method really only be Debug and not Warn/Error?

shared/signal/client/grpc.go
```diff
@@ -398,24 +413,37 @@ func (c *GrpcClient) receive(stream proto.SignalExchange_ConnectStreamClient,
...
 	}
 }
 
+func (c *GrpcClient) startEncryptionWorker(handler func(msg *proto.Message) error) {
+	if c.decryptionWorker != nil {
+		return
+	}
+
+	c.decryptionWorker = NewWorker(c.decryptMessage, handler)
+	workerCtx, workerCancel := context.WithCancel(context.Background())
```
lixmal, line 436: Think passing a parent ctx here might be better? Would cancel everything immediately instead of waiting for an explicit cancellation.

client/internal/peer/session_id.go
```diff
@@ -0,0 +1,47 @@
+package peer
+
+import (
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"io"
+)
+
+const sessionIDSize = 5
```
lixmal, line 10: This might be a little low for collisions?

shared/signal/client/worker.go
```diff
@@ -0,0 +1,55 @@
...
+	handler        func(msg *proto.Message) error
+
+	encryptedMsgPool chan *proto.EncryptedMessage
+}
+
+func NewWorker(decryptFn func(msg *proto.EncryptedMessage) (*proto.Message, error), handlerFn func(msg *proto.Message) error) *Worker {
+	return &Worker{
+		decryptMessage:   decryptFn,
+		handler:          handlerFn,
+		encryptedMsgPool: make(chan *proto.EncryptedMessage, 1),
```
lixmal, line 22: Higher buffer for bursts?

shared/signal/client/grpc.go
pappz: Please check this commit:
https://github.com/netbirdio/netbird/pull/4281/commits/0fcb272f0574fcc9fdccc15211298564bdea61c1

shared/signal/client/worker.go
pappz: Yeah, I was thinking about it, but with this approach we risk providing deprecated packages after a reconnection. Context cancellation is not trivial (see: [link](https://github.com/netbirdio/netbird/pull/4281/commits/0fcb272f0574fcc9fdccc15211298564bdea61c1)). Maybe the signal has been reconnected, but the worker is still operating. We currently don’t have logic to drain this channel.

Anyway, we have three stages:
1. GRPC package reading (producer side)
2. Decryption
3. Processing the package (consumer)

In case of a burst, the first stage is slower than stages 2 and 3 combined.

My first idea was to add more worker threads, but that would cause us to lose package ordering. In the future, I’d like to investigate why the GRPC connection is slow and explore possible optimizations here.

client/internal/peer/worker_ice.go
pappz: Good point. I will fix it

client/internal/peer/session_id.go
pappz: It generates 5 random bytes. Each byte can have a value from 0 to 255. That means there are 256^5
 possible combinations just to distinguish between two IDs, only two, no more than two instances.
 The collision chance is one to 1.1 trillion

PR #4283, opened 2025-08-04
[client] Shorten peer public key in connection logs

pappz:
## Describe your changes

Shorter pub keys in logs:
```
2025-08-04T22:26:06Z INFO 285 [peer: 5ylZjz6] client/internal/peer/conn.go:478: created new wgProxy for relay connection: 127.0.0.1:1
2025-08-04T22:26:06Z INFO 64 [peer: 5ylZjz6] client/internal/peer/wg_watcher.go:88: WireGuard watcher started
2025-08-04T22:26:07Z INFO 285 [peer: 5ylZjz6] client/internal/peer/conn.go:509: start to communicate with peer via relay
2025-08-04T22:26:07Z INFO 453 [peer: 5ylZjz6] client/internal/peer/conn.go:349: set ICE to active connection
2025-08-04T22:26:07Z INFO 64 [peer: 5ylZjz6] client/internal/peer/wg_watcher.go:112: WireGuard watcher stopped
2025-08-04T22:26:09Z INFO 294 [peer: 5ylZjz6] client/internal/peer/guard/guard.go:74: start reconnect loop...
```
### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

### DIFF
--- a/client/internal/peer/key.go
+++ b/client/internal/peer/key.go
@@ -1,5 +1,8 @@
 package peer
 
 func shortenKey(pubKey string) string {
+	if len(pubKey) < 7 {
+		return pubKey
+	}
 	return pubKey[:7]
 }


client/internal/peer/key.go
```diff
@@ -0,0 +1,8 @@
+package peer
+
+func shortenKey(pubKey string) string {
```
nazarewk, line 3: I would like to make it wrapped (`'` / `"`) or prefixed (eg: `pk:{pubKey}` or `pk({pubKey})`), so it's possible to identify when processing logs.

As an example I currently have a general Base64 regex of length 44 that occasionally even matches file paths with specific lengths due to large scope of Base64 character set:
```
peersRe = regexp.MustCompile(`[a-zA-Z0-9+/=]{44}`)
```

This is not much of an issue for 44 characters that consist a WireGuard Public Key, but will very likely return too much false positives for just a 7 character key.

pappz: The `peerid: ` prefix is not good enough?

nazarewk: for `peer: Key` it is ok-ish, but I won't be able to locate it in automation if somebody forgot about it and used it somewhere else later on

nazarewk: let's at least make it clear in the function's documentation that this is meant to be wrapped if it were to be used anywhere else

nazarewk: How about we shorten it the way I do my markdown notes? A few leading characters, some form of ellipsis and a few trailing characters:
```
2025-03-24T09:07:31+01:00 INFO [peer: IoCN.../HA=] client/internal/peer/handshaker.go:91: received connection confirmation, running version 0.39.1 and with remote WireGuard listen port 51820
2025-07-29T16:00:36-04:00 INFO [peer: a64b...jGM=] client/internal/peer/conn.go:509: start to communicate with peer via relay
```

would be both readable, easy to identify explicitly and search for

PS: I just put triple dots there, but we could use UTF ellipsis https://www.compart.com/en/unicode/U+2026

PR #4314, opened 2025-08-10
[client] Use native windows sock opts to avoid routing loops

lixmal:
## Describe your changes

- Move `util/grpc` and `util/net` to `client` so `internal` packages can be accessed
 - Add methods to return the next best interface after the NetBird interface.
- Use `IP_UNICAST_IF` sock opt to force the outgoing interface for the NetBird `net.Dialer` and `net.ListenerConfig` to avoid routing loops. The interface is picked by the new route lookup method.
- Some refactoring to avoid import cycles
- Old behavior is available through `NB_USE_LEGACY_ROUTING=true` env var

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

Interal routing change, no user facing code change


client/internal/routemanager/systemops/systemops_windows.go
```diff
@@ -821,6 +877,75 @@ func getInterfaceMetric(interfaceIndex uint32, family int16) int {
...
+}
+
+// GetBestInterface finds the best interface for reaching a destination,
+// excluding the VPN interface to avoid routing loops.
+//
+// Route selection priority:
+// 1. Longest prefix match (most specific route)
+// 2. Lowest route metric
+// 3. Lowest interface metric
+func GetBestInterface(dest netip.Addr, vpnIntf string) (*net.Interface, error) {
```
hakansa, line 905: Do we call this on every network change (e.g Wifi-> Ethernet) ? If we don't refresh the best iface we might keep pinning to a non-best interface as i understand.

client/net/net_windows.go
```diff
@@ -0,0 +1,275 @@
...
+		return nil, fmt.Errorf("find route for %s: %w", dest, err)
+	}
+
+	if dest.Is6() {
+		return &interfaceSelection{iface6: iface}, nil
+	}
+	return &interfaceSelection{iface4: iface}, nil
+}
+
+func setIPv4UnicastIF(fd uintptr, iface *net.Interface) error {
```
hakansa, line 176: do we need to fallback `NB_USE_LEGACY_ROUTING` if sth goes wrong here?

lixmal: In theory yes, but it's too late for the setup at this point.

client/internal/routemanager/systemops/systemops_windows.go
lixmal: We call this every time a socket is created. For network changes we have the network monitor to restart the engine.

hakansa (dismissed):

hakansa (dismissed):

mlsmaycon (approved):

PR #4367, opened 2025-08-18
 [management/client] create job channel between management and client

aliamerj:
## Describe your changes

This PR sets up a communication channel between the management server and the client to handle job requests and responses.

### What was done:

- Created a Job channel for each peer in JobManager.
- Send messages from management to the client.
- Send replies from the client back to management.
- Prepared client-side Job() stream to receive messages from the server.
- Integrated peer lifecycle management with ephemeral and secrets managers.

## Stack

- `feature/remote-debug` - :warning: No PR associated with branch 
  - \#4367 :point\_left:

### Checklist

- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation

Select exactly one:

Paste the PR link from <https://github.com/netbirdio/docs> here:

<https://github.com/netbirdio/docs/pull/>\_\_

### DIFF
--- a/management/server/grpcserver.go
+++ b/management/server/grpcserver.go
@@ -335,7 +335,6 @@ func (s *GRPCServer) handleUpdates(ctx context.Context, accountID string, peerKe
 			}
 			log.WithContext(ctx).Debugf("received an update for peer %s", peerKey.String())
 			if err := s.sendUpdate(ctx, accountID, peerKey, peer, update, srv); err != nil {
-				s.cancelPeerRoutines(ctx, accountID, peer)
 				return err
 			}
 
--- a/management/server/permissions/operations/operation.go
+++ b/management/server/permissions/operations/operation.go
@@ -7,5 +7,4 @@ const (
 	Read   Operation = "read"
 	Update Operation = "update"
 	Delete Operation = "delete"
-	Job    Operation = "job"
 )


shared/management/proto/management.proto
```diff
@@ -63,29 +63,40 @@ message EncryptedMessage {
   int32 version = 3;
 }
 
-message JobRequest {
-  JobType type = 1;
-  bytes ID = 2;
-  bytes Payload =3;
-}
+message JobCreateRequest {
```
pascal-fischer: Sorry, this was not my intention to rename. This can stay JobRequest. I was looking at the API objects when writing this

aliamerj: No problem 🙃

aliamerj:
@pascal-fischer 
can i merge this pr if there is no other feedback ?

pascal-fischer:
Ah I just had a look at the proto. Let me review the whole thing now

client/internal/engine.go
```diff
@@ -886,6 +887,51 @@ func (e *Engine) updateConfig(conf *mgmProto.PeerConfig) error {
...
+			e.config.RosenpassPermissive,
+			&e.config.ServerSSHAllowed,
+			e.config.DisableClientRoutes,
+			e.config.DisableServerRoutes,
+			e.config.DisableDNS,
+			e.config.DisableFirewall,
+			e.config.BlockLANAccess,
+			e.config.BlockInbound,
+			e.config.LazyConnectionEnabled,
+		)
```
pascal-fischer: why do you need to prepare system info here? This is part of the Sync registration, thats handling the peer lifecycle

management/server/grpcserver.go
```diff
@@ -74,10 +78,14 @@ func NewServer(
...
+		if err := appMetrics.GRPCMetrics().RegisterConnectedStreams(func() int64 {
 			return int64(len(peersUpdateManager.peerChannels))
-		})
-		if err != nil {
+		}); err != nil {
+			return nil, err
+		}
+		if err = appMetrics.GRPCMetrics().RegisterConnectedStreams(func() int64 {
+			return int64(len(jobManager.jobChannels))
+		}); err != nil {
```
pascal-fischer: `RegisterConnectedStreams` is a gauge. If you just overwrite both of those calls will interfere with each other now. If you wanna monitor job channels you need to create a new metric

management/server/grpcserver.go
```diff
@@ -132,6 +141,48 @@ func getRealIP(ctx context.Context) net.IP {
...
+	defer func() {
+		if unlock != nil {
+			unlock()
+		}
+	}()
+
+	accountID, peer, _, _, err := s.syncPeer(ctx, nil, peerKey)
+	if err != nil {
+		return err
+	}
```
pascal-fischer: This whole logic is handling peer lifecycle. The `s.acquirePeerLockByUID` is used to handle overwrites where reconnects are faster than the cleanup routine. This is all handled by the main Sync stream

management/server/grpcserver.go
```diff
@@ -132,6 +141,48 @@ func getRealIP(ctx context.Context) net.IP {
...
+	// nolint:staticcheck
+	ctx = context.WithValue(ctx, nbContext.AccountIDKey, accountID)
+
+	// Start background response handler
+	s.startResponseReceiver(ctx, accountID, srv)
+
+	// Prepare per-peer state
+	updates := s.jobManager.CreateJobChannel(peer.ID)
+	s.ephemeralManager.OnPeerConnected(ctx, peer)
+	s.secretsManager.SetupRefresh(ctx, accountID, peer.ID)
```
pascal-fischer: same here. This is all peer lifecycle, no need to handle this twice, it will just interfere. The jobs stream gonna be a side channel, Sync gonna stay the main stream handling peer lifecycle

management/server/grpcserver.go
```diff
@@ -158,55 +208,132 @@ func (s *GRPCServer) Sync(req *proto.EncryptedMessage, srv proto.ManagementServi
...
 	realIP := getRealIP(ctx)
-	log.WithContext(ctx).Debugf("Sync request from peer [%s] [%s]", req.WgPubKey, realIP.String())
 
-	if syncReq.GetMeta() == nil {
+	if syncReq != nil && syncReq.GetMeta() == nil {
 		log.WithContext(ctx).Tracef("peer system meta has to be provided on sync. Peer %s, remote addr %s", peerKey.String(), realIP)
 	}
 
-	peer, netMap, postureChecks, err := s.accountManager.SyncAndMarkPeer(ctx, accountID, peerKey.String(), extractPeerMeta(ctx, syncReq.GetMeta()), realIP)
+	peer, netMap, postureChecks, err := s.accountManager.SyncAndMarkPeer(ctx, accountID, peerKey.String(), nbpeer.PeerSystemMeta{}, realIP)
```
pascal-fischer: This is also not needed on jobs. This is all Sync code which does not need to be touched at all. Also sending an empty meta object would overwrite the actual metadata that we received from Sync with empty values

management/server/grpcserver.go
```diff
@@ -158,55 +208,132 @@ func (s *GRPCServer) Sync(req *proto.EncryptedMessage, srv proto.ManagementServi
...
+	hello, err := srv.Recv()
 	if err != nil {
-		log.WithContext(ctx).Debugf("error while sending initial sync for %s: %v", peerKey.String(), err)
-		return err
+		return wgtypes.Key{}, status.Errorf(codes.InvalidArgument, "missing hello: %v", err)
 	}
 
-	updates := s.peersUpdateManager.CreateChannel(ctx, peer.ID)
+	jobReq := &proto.JobRequest{}
+	peerKey, err := s.parseRequest(ctx, hello, jobReq)
```
pascal-fischer, line 257: You can already send an encryptedMessage during registration, pretty much what sync is doing. This way you do not send a separate handshake

management/server/grpcserver.go
```diff
@@ -158,55 +208,132 @@ func (s *GRPCServer) Sync(req *proto.EncryptedMessage, srv proto.ManagementServi
...
+				log.WithContext(ctx).Warnf("invalid job response: %v", err)
+				return
+			}
 
-	unlock()
-	unlock = nil
+			if err := s.jobManager.HandleResponse(ctx, accountID, jobResp); err != nil {
+				log.WithContext(ctx).Errorf("handle job response failed: %v", err)
+				return
+			}
```
pascal-fischer: You only have one response receiver. Just because we failed to handle one message we should not close the listener. There is no way to recover from here

management/server/grpcserver.go
```diff
@@ -246,34 +373,69 @@ func (s *GRPCServer) handleUpdates(ctx context.Context, accountID string, peerKe
...
+// Only closes update channels (used when sync stream ends / errors)
+func (s *GRPCServer) closeUpdateChannel(ctx context.Context, accountID string, peer *nbpeer.Peer) {
+	s.peersUpdateManager.CloseChannel(ctx, peer.ID)
+	s.disconnectPeer(ctx, accountID, peer)
+}
+
+// Only closes job channels (used when job stream ends / errors)
+func (s *GRPCServer) closeJobChannels(ctx context.Context, accountID string, peer *nbpeer.Peer) {
+	s.jobManager.CloseChannel(ctx, accountID, peer.ID)
+	s.disconnectPeer(ctx, accountID, peer)
```
pascal-fischer: Same here. Disconnect peer logic is just interesting on sync, not jobs. Sync can stay as it was (just add an addition on Sync cleanup to also disconnect jobs)

management/server/jobChannel.go
```diff
@@ -0,0 +1,148 @@
...
+	if !ok {
+		return fmt.Errorf("job %s not found", resp.ID)
+	}
+
+	event.Response = resp
+	//TODO: update the store for job response
+	// jm.store.CompleteJob(ctx,accountID, string(resp.GetID()), string(resp.GetResult()),string(resp.GetReason()))
+	close(event.Done)
+
+	return nil
```
pascal-fischer: if a job is done we should clean it from the pending table no?

management/server/jobChannel.go
```diff
@@ -0,0 +1,148 @@
...
+	select {
+	case <-event.Done:
+		return nil
+	case <-time.After(jm.responseWait):
+		jm.cleanup(ctx, accountID, string(req.ID))
+		return fmt.Errorf("job %s timed out", req.ID)
+	case <-ctx.Done():
+		jm.cleanup(ctx, accountID, string(req.ID))
+		return ctx.Err()
+	}
```
pascal-fischer: cleanup will write reason 'timeout'. Might make sense to pass the exact error here (and maybe do sanatize of the context error in cleanup)

pascal-fischer (changes requested):

management/server/grpcserver.go
aliamerj: i did not want the pr be large so i was thing to make separate pr for just the Metrics but i will remove this for now and once we fix all issues with this pr i can make pr for Metrics and to be handle just like sync , what do you think ?

pascal-fischer: sure you can do separate PRs for this

management/server/grpcserver.go
aliamerj: just to clarify how I handled the handshake:
- In Sync, the first EncryptedMessage comes in as a request argument.
- In Job (since it’s bidirectional), we don’t get that as a parameter, so I just call `srv.Recv()` once to grab the first message.
After that I run it through parseRequest, same as Sync, just with a JobRequest.
So the flow is basically the same, the only difference is how we read that first message.

management/server/grpcserver.go
aliamerj: my bad , this should be continue

management/server/grpcserver.go
pascal-fischer: Ah true, I did not think of that. Then lets keep it like this

management/server/grpcserver.go
```diff
@@ -150,33 +145,28 @@ func (s *GRPCServer) Job(srv proto.ManagementService_JobServer) error {
...
+		ctx = context.WithValue(ctx, nbContext.AccountIDKey, "UNKNOWN")
+		log.WithContext(ctx).Tracef("peer %s is not registered", peerKey.String())
+		if errStatus, ok := internalStatus.FromError(err); ok && errStatus.Type() == internalStatus.NotFound {
+			return status.Errorf(codes.PermissionDenied, "peer is not registered")
+		}
 		return err
 	}
 	// nolint:staticcheck
 	ctx = context.WithValue(ctx, nbContext.AccountIDKey, accountID)
+	peer, err := s.accountManager.GetStore().GetPeerByPeerPubKey(ctx, store.LockingStrengthKeyShare, peerKey.String())
```
pascal-fischer: LockingStrengthNone

aliamerj:
@pascal-fischer 
can i merge it or still not ready ?

pascal-fischer:
Almost. I just accepted sonar for this case (usually, there should be no code quality issues, but I loosened the rules a bit here) If you could just fix the linter complaint should be ready

pascal-fischer:
The Docs test is failing because you did not set a mark here. In the PR description just set "Documentation is not needed for this change (explain why)" then this failed test is gone as well. The Benchmarks are a bit flaky depending on the runner they are executed on (we can rerun those, usually is fixed on the next run). And when all the tests are green you can merge

aliamerj:
> Almost. I just accepted sonar for this case (usually, there should be no code quality issues, but I loosened the rules a bit here) If you could just fix the linter complaint should be ready

the linter complaint comes from the codespell in the generated code from .proto

pascal-fischer:
```Error: ./shared/management/proto/management.proto:77: successed ==> succeeded, success, successful```
The source proto file already has the same typo. Once that one is fixed and proto is generated it will be fixed in all 3 places.

File: management/server/grpcserver.go [hunk 13]
```diff
@@ -171,17 +211,14 @@ func (s *GRPCServer) Sync(req *proto.EncryptedMessage, srv proto.ManagementServi
 
 	// nolint:staticcheck
 	ctx = context.WithValue(ctx, nbContext.AccountIDKey, accountID)
-
 	realIP := getRealIP(ctx)
-	log.WithContext(ctx).Debugf("Sync request from peer [%s] [%s]", req.WgPubKey, realIP.String())
-
```
pascal-fischer: Why are you removing the log messages?

management/server/grpcserver.go
```diff
@@ -171,17 +211,14 @@ func (s *GRPCServer) Sync(req *proto.EncryptedMessage, srv proto.ManagementServi
...
 	if syncReq.GetMeta() == nil {
 		log.WithContext(ctx).Tracef("peer system meta has to be provided on sync. Peer %s, remote addr %s", peerKey.String(), realIP)
 	}
 
 	peer, netMap, postureChecks, err := s.accountManager.SyncAndMarkPeer(ctx, accountID, peerKey.String(), extractPeerMeta(ctx, syncReq.GetMeta()), realIP)
 	if err != nil {
-		log.WithContext(ctx).Debugf("error while syncing peer %s: %v", peerKey.String(), err)
```
pascal-fischer: This one also does not need to chnage (the log level is fine)

pascal-fischer:
Ok no maybe we are not done. I just reviewed the thing as a whole again. One sec I will share

shared/management/client/grpc.go
```diff
@@ -238,12 +378,12 @@ func (c *GrpcClient) connectToStream(ctx context.Context, serverPubKey wgtypes.K
 	return sync, nil
 }
 
-func (c *GrpcClient) receiveEvents(stream proto.ManagementService_SyncClient, serverPubKey wgtypes.Key, msgHandler func(msg *proto.SyncResponse) error) error {
+func (c *GrpcClient) receiveUpdatesEvents(stream proto.ManagementService_SyncClient, serverPubKey wgtypes.Key, msgHandler func(msg *proto.SyncResponse) error) error {
 	for {
 		update, err := stream.Recv()
 		if err == io.EOF {
 			log.Debugf("Management stream has been closed by server: %s", err)
-			return err
```
pascal-fischer: why do you remove the error?

management/server/grpcserver.go
```diff
@@ -246,34 +352,63 @@ func (s *GRPCServer) handleUpdates(ctx context.Context, accountID string, peerKe
...
 	if err != nil {
-		s.cancelPeerRoutines(ctx, accountID, peer)
+		s.closeUpdateChannel(ctx, accountID, peer)
 		return status.Errorf(codes.Internal, "failed sending update message")
 	}
 	log.WithContext(ctx).Debugf("sent an update to peer %s", peerKey.String())
 	return nil
 }
 
-func (s *GRPCServer) cancelPeerRoutines(ctx context.Context, accountID string, peer *nbpeer.Peer) {
```
pascal-fischer, line 264: The cancelPeerRoutines method should stay as it was. There is no need to chnage anything here. The jobs stream should only deal with the jobs channels nothing more. The sync side of thing should be untouched. The only thing that should happen on s.cancelPeerRoutines is that we additionally close the jobs channel. On jobs stream disconnect we only deal with jobs. There is a hierarchy, Sync gonna control the connections, jobs is just a side channel.

management/server/grpcserver.go
```diff
@@ -226,8 +334,8 @@ func (s *GRPCServer) handleUpdates(ctx context.Context, accountID string, peerKe
 				return nil
 			}
 			log.WithContext(ctx).Debugf("received an update for peer %s", peerKey.String())
-
 			if err := s.sendUpdate(ctx, accountID, peerKey, peer, update, srv); err != nil {
+				s.cancelPeerRoutines(ctx, accountID, peer)
```
pascal-fischer: this here is wrong

management/server/permissions/operations/operation.go
```diff
@@ -7,4 +7,5 @@ const (
 	Read   Operation = "read"
 	Update Operation = "update"
 	Delete Operation = "delete"
+	Job    Operation = "job"
```
pascal-fischer: this is also not right. Those are CRUD operations. You need this as a module and the Create for starting a job and get to check the status. This stays pretty much in sync with the http operation GET, POST, PUT, DELETE. We do plan on moving the whole permission validation into the http router middleware then it might get even clearer.

This repo contains a new design for our management service. You can read this repo, this is where we want to go towards with out code. Here is the example how how permissions gonna be used in the future https://github.com/netbirdio/management-refactor/blob/main/internals/modules/peers/manager/api.go

PR #4370, opened 2025-08-19
[client] Add WireGuard interface lifecycle monitoring

hakansa:
Implement a background watcher that restarts the engine when the WireGuard interface is created or deleted. This addition enhances the engines resilience to external changes in the interface state.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

- [x] Documentation is **not needed** for this change (explain why):
Internal change

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -479,9 +479,11 @@ func (e *Engine) Start() error {
 	go func() {
 		defer e.wgIfaceMonitorWg.Done()
 
-		if err := e.wgIfaceMonitor.Start(e.ctx, e.wgInterface.Name()); err != nil {
+		if shouldRestart, err := e.wgIfaceMonitor.Start(e.ctx, e.wgInterface.Name()); shouldRestart {
 			log.Infof("WireGuard interface monitor: %s, restarting engine", err)
 			e.restartEngine()
+		} else if err != nil {
+			log.Warnf("WireGuard interface monitor: %s", err)
 		}
 	}()
 
--- a/client/internal/wg_iface_monitor.go
+++ b/client/internal/wg_iface_monitor.go
@@ -26,27 +26,25 @@ func NewWGIfaceMonitor() *WGIfaceMonitor {
 
 // Start begins monitoring the WireGuard interface.
 // It relies on the provided context cancellation to stop.
-// Start only returns an error if the interface is deleted or recreated,
-// context cancellation does not return an error.
-func (m *WGIfaceMonitor) Start(ctx context.Context, ifaceName string) error {
+func (m *WGIfaceMonitor) Start(ctx context.Context, ifaceName string) (shouldRestart bool, err error) {
 	defer close(m.done)
 
 	// Skip on mobile platforms as they handle interface lifecycle differently
 	if runtime.GOOS == "android" || runtime.GOOS == "ios" {
 		log.Debugf("Interface monitor: skipped on %s platform", runtime.GOOS)
-		return nil
+		return false, errors.New("not supported on mobile platforms")
 	}
 
 	if ifaceName == "" {
 		log.Debugf("Interface monitor: empty interface name, skipping monitor")
-		return nil
+		return false, errors.New("empty interface name")
 	}
 
 	// Get initial interface index to track the specific interface instance
 	expectedIndex, err := getInterfaceIndex(ifaceName)
 	if err != nil {
 		log.Debugf("Interface monitor: interface %s not found, skipping monitor", ifaceName)
-		return nil
+		return false, fmt.Errorf("interface %s not found: %w", ifaceName, err)
 	}
 
 	log.Infof("Interface monitor: watching %s (index: %d)", ifaceName, expectedIndex)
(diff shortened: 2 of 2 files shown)


client/internal/engine.go
```diff
@@ -1701,6 +1704,63 @@ func (e *Engine) startNetworkMonitor() {
...
+// startWGIfaceMonitor starts a background watcher that restarts the engine if
+// the WireGuard interface is created or deleted externally while the engine is running.
+// It relies on the engine context cancellation to stop.
+func (e *Engine) startWGIfaceMonitor() {
+	// wgInterface should be initialized at this point
+	if e.wgInterface == nil {
+		return
+	}
+
+	name := e.wgInterface.Name()
```
lixmal: If the interface is recreated, this will continue to succeed, even though we won't be able to use the interface.
I suggest we look up the index initially and keep polling for that.

client/internal/engine.go
```diff
@@ -1701,6 +1704,63 @@ func (e *Engine) startNetworkMonitor() {
...
+	if name == "" {
+		return
+	}
+
+	// Polling approach for cross-platform simplicity
+	go func(ctx context.Context, ifaceName string) {
+		log.Infof("Interface monitor: watching %s", ifaceName)
+
+		// Determine initial state
+		prevExists := interfaceExists(ifaceName)
```
lixmal: Don't think we need this, if the interface is not there there's no point even starting this monitor. We also don't need to watch for created events

client/internal/engine.go
```diff
@@ -1701,6 +1704,63 @@ func (e *Engine) startNetworkMonitor() {
 	}()
 }
 
+// startWGIfaceMonitor starts a background watcher that restarts the engine if
+// the WireGuard interface is created or deleted externally while the engine is running.
+// It relies on the engine context cancellation to stop.
+func (e *Engine) startWGIfaceMonitor() {
```
lixmal: Shouldn't run on mobile and will fail without using the mobile dep for interface retrieval.

client/internal/engine.go
```diff
@@ -1701,6 +1704,63 @@ func (e *Engine) startNetworkMonitor() {
...
+	}(e.ctx, name)
+}
+
+// interfaceExists checks whether a network interface with the given name exists.
+func interfaceExists(name string) bool {
+	if name == "" {
+		return false
+	}
+	ifi, err := net.InterfaceByName(name)
+	if err != nil || ifi == nil {
```
lixmal: Can this be more granular, i.e. only "not found" errors?

client/internal/engine.go
hakansa: fixed

client/internal/engine.go
hakansa: done

client/internal/engine.go
hakansa: done

client/internal/engine.go
```diff
@@ -1701,6 +1704,87 @@ func (e *Engine) startNetworkMonitor() {
 	}()
 }
 
+// startWGIfaceMonitor starts a background watcher that restarts the engine if
+// the WireGuard interface is deleted externally while the engine is running.
+// It relies on the engine context cancellation to stop.
+func (e *Engine) startWGIfaceMonitor() {
```
pappz: Please move this new feature into a new struct. The engine structure is already a huge monster.

client/internal/engine.go
```diff
@@ -1701,6 +1704,87 @@ func (e *Engine) startNetworkMonitor() {
...
+
+	// Get initial interface index to track the specific interface instance
+	initialIndex, err := getInterfaceIndex(name)
+	if err != nil {
+		log.Debugf("Interface monitor: interface %s not found, skipping monitor", name)
+		return
+	}
+
+	// Polling approach for cross-platform simplicity
+	go func(ctx context.Context, ifaceName string, expectedIndex int) {
```
pappz: When closing, the engine should wait to exit from this goroutine. If we don’t do this, the engine could accidentally restart after closing because the thread is still running in the background.

client/internal/engine.go
hakansa: done 😄

client/internal/engine.go
hakansa: done

client/internal/wg_iface_monitor.go
```diff
@@ -0,0 +1,124 @@
...
+		done:          make(chan struct{}),
+		restartEngine: restartEngine,
+	}
+}
+
+// Start begins monitoring the WireGuard interface.
+// It relies on the provided context cancellation to stop.
+func (m *WGIfaceMonitor) Start(ifaceName string) {
+	// Skip on mobile platforms as they handle interface lifecycle differently
+	if runtime.GOOS == "android" || runtime.GOOS == "ios" {
```
nazarewk, line 33: I'd probably put this check in the code calling `NewWGIfaceMonitor()` instead of here, it's not the Monitor's concern whether it should run or not, especially when you're already passing an unspecified callback function.

client/internal/wg_iface_monitor.go
```diff
@@ -0,0 +1,124 @@
...
+					m.restartEngine()
+					return
+				}
+			}
+		}
+	}(m.ctx, ifaceName, initialIndex)
+}
+
+// Stop stops the monitor and waits for the goroutine to exit.
+func (m *WGIfaceMonitor) Stop() {
```
pappz: If you run `Stop` and `m.retartEngine()` at the same time, then it can cause a deadlock for 5 sec because both of them use the same mutex in the engine.

PR #4428, opened 2025-09-01
[management/client] Integrate Job API with Job Stream and Client Engine

aliamerj:
## Describe your changes

This PR connects the Job API with the peer job stream and updates both server and client code to work together.

#### What was done:

- Integrated Job API with the JobManager stream channel.
- Updated server logic to send and track jobs through the channel.
- Implemented client-side handling in the engine for job requests and responses.
- Adjusted existing code to align with the new unified job flow.

## Stack

- `feature/remote-debug` - :warning: No PR associated with branch 
  - \#4428 :point\_left:

### Checklist

- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation

Select exactly one:

Paste the PR link from <https://github.com/netbirdio/docs> here:

<https://github.com/netbirdio/docs/pull/>\_\_

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -14,7 +14,6 @@ import (
 	"sort"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/hashicorp/go-multierror"
@@ -199,9 +198,10 @@ type Engine struct {
 	stateManager *statemanager.Manager
 	srWatcher    *guard.SRWatcher
 
-	// Sync response persistence
-	persistSyncResponse atomic.Bool
-	latestSyncResponse  atomic.Pointer[mgmProto.SyncResponse]
+	// Sync response persistence (protected by syncRespMux)
+	syncRespMux         sync.RWMutex
+	persistSyncResponse bool
+	latestSyncResponse  *mgmProto.SyncResponse
 	connSemaphore       *semaphoregroup.SemaphoreGroup
 	flowManager         nftypes.FlowManager
 
@@ -662,8 +662,10 @@ func (e *Engine) removePeer(peerKey string) error {
 	return nil
 }
 
-// 1. xyzzz
 func (e *Engine) handleSync(update *mgmProto.SyncResponse) error {
+	e.syncMsgMux.Lock()
+	defer e.syncMsgMux.Unlock()
+
 	if update.GetNetbirdConfig() != nil {
 		wCfg := update.GetNetbirdConfig()
 		err := e.updateTURNs(wCfg.GetTurns())
@@ -702,11 +704,19 @@ func (e *Engine) handleSync(update *mgmProto.SyncResponse) error {
 	if nm == nil {
 		return nil
 	}
-	persistSyncResponse := e.persistSyncResponse.Load()
+
+	// Persist sync response under the dedicated lock (syncRespMux), not under syncMsgMux.
+	// Read the storage-enabled flag under the syncRespMux too.
+	e.syncRespMux.RLock()
+	enabled := e.persistSyncResponse
+	e.syncRespMux.RUnlock()
 
 	// Store sync response if persistence is enabled
-	if persistSyncResponse {
-		e.latestSyncResponse.Store(update)
+	if enabled {
+		e.syncRespMux.Lock()
+		e.latestSyncResponse = update
+		e.syncRespMux.Unlock()
+
 		log.Debugf("sync response persisted with serial %d", nm.GetSerial())
 	}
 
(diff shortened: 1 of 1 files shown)


shared/management/client/grpc.go
```diff
@@ -298,7 +298,7 @@ func (c *GrpcClient) handleSyncStream(ctx context.Context, serverPubKey wgtypes.
 
 	// blocking until error
 	err = c.receiveUpdatesEvents(stream, serverPubKey, msgHandler)
-	if err != nil {
+	if err != nil && err != io.EOF{
```
aliamerj: I noticed that when sending a job, I was repeatedly seeing logs from the sync with EFO. To fix this, I updated this not ideal but for now,
the problem is when sending stream request for peer both sync and job triggered and one get valid request and the other get nothing , so we get this error and disconnect the peer , I think we need we to drop the `io.EOF` check in sync as i did for the job

client/internal/engine.go
```diff
@@ -887,19 +892,45 @@ func (e *Engine) updateConfig(conf *mgmProto.PeerConfig) error {
...
+						Reason: []byte(err.Error()),
+					}
+				}
+				return &mgmProto.JobResponse{
+					ID:     msg.ID,
+					Status: mgmProto.JobStatus_succeeded,
+					WorkloadResults: &mgmProto.JobResponse_Bundle{
+						Bundle: &mgmProto.BundleResult{
+							UploadKey: uploadKey,
+						},
```
pascal-fischer: Here you are creating unnecessary code duplication. The only generic part is  WorkloadResults. The rest is identical and can be build after the switch

management/server/store/sql_store.go
```diff
@@ -136,18 +137,39 @@ func (s *SqlStore) CreatePeerJob(ctx context.Context, job *types.Job) error {
 	return nil
 }
 
+func (s *SqlStore) CompletePeerJob(ctx context.Context, jobResponse *proto.JobResponse) error {
+	var job types.Job
+	if err := job.ApplyResponse(jobResponse); err != nil {
+		return status.Errorf(status.Internal, err.Error())
+	}
```
pascal-fischer: I would move this conversion to the GRPC layer to not drag proto sturct to deep into the code

client/internal/engine.go
```diff
@@ -887,19 +896,45 @@ func (e *Engine) updateConfig(conf *mgmProto.PeerConfig) error {
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to daemon error: %v\n"+
+			"If the daemon is not running please run: "+
+			"\nnetbird service install \nnetbird service start\n", err)
+	}
+
+	return conn, nil
+}
+
```
pascal-fischer: I don't think this is the way to go. The engine is part of the daemon so we are basically connecting to ourself.

aliamerj: yes, i understand , that is why i needed help to understand what is the best why to call bundle ,
is passing grpc client and call it from it or passing address string and call the bundle or without the grpc and call the debugging directly ?
and also considering the other os

client/internal/engine.go
aliamerj: The generic part are 
 1. Status could be succeeded or failed
 2. Reason could be nil or error message
 3. WorkloadResults could be nil or the result
 so the identical part is just job id

client/android/client.go
```diff
@@ -112,7 +112,7 @@ func (c *Client) Run(urlOpener URLOpener, dns *DNSList, dnsReadyListener DnsRead
 
 	// todo do not throw error in case of cancelled context
 	ctx = internal.CtxInitState(ctx)
-	c.connectClient = internal.NewConnectClient(ctx, cfg, c.recorder)
+	c.connectClient = internal.NewConnectClient(ctx, cfg, c.recorder, "")
```
aliamerj, line 115: 1. Do we need to pass the `logFile` here

client/cmd/up.go
```diff
@@ -196,7 +196,8 @@ func runInForegroundMode(ctx context.Context, cmd *cobra.Command, activeProf *pr
 	r := peer.NewRecorder(config.ManagementURL.String())
 	r.GetFullStatus()
 
-	connectClient := internal.NewConnectClient(ctx, config, r)
+	//todo: do we need to pass logFile here ?
+	connectClient := internal.NewConnectClient(ctx, config, r, "")
```
aliamerj, line 200: 2. Do we need to pass the `logFile` here

client/embed/embed.go
```diff
@@ -131,7 +131,9 @@ func (c *Client) Start(startCtx context.Context) error {
 	}
 
 	recorder := peer.NewRecorder(c.config.ManagementURL.String())
-	client := internal.NewConnectClient(ctx, c.config, recorder)
+
+	//todo: do we need to pass logFile here ?
+	client := internal.NewConnectClient(ctx, c.config, recorder, "")
```
aliamerj, line 136: 3. Do we need to pass the `logFile` here

client/ios/NetBirdSDK/client.go
```diff
@@ -127,7 +127,8 @@ func (c *Client) Run(fd int32, interfaceName string) error {
 	c.onHostDnsFn = func([]string) {}
 	cfg.WgIface = interfaceName
 
-	c.connectClient = internal.NewConnectClient(ctx, cfg, c.recorder)
+	//todo: do we need to pass logFile here ?
+	c.connectClient = internal.NewConnectClient(ctx, cfg, c.recorder, "")
```
aliamerj, line 131: 4. Do we need to pass the `logFile` here

client/server/server.go
```diff
@@ -235,7 +233,7 @@ func (s *Server) connectWithRetryRuns(ctx context.Context, config *profilemanage
 
 	runOperation := func() error {
 		log.Tracef("running client connection")
-		s.connectClient = internal.NewConnectClient(ctx, config, statusRecorder)
+		s.connectClient = internal.NewConnectClient(ctx, config, statusRecorder, s.logFile)
```
aliamerj, line 236: here the only place we we pass the `logFile`

pappz (approved):

PR #4430, opened 2025-09-01
[client] Fix/v4 stun routing

pappz:
## Describe your changes

Deduplicate STUN package sending.
Originally, because every peer shared the same UDP address, the library could not distinguish which STUN message was associated with which candidate. As a result, the Pion library responded from all candidates for every STUN message.

- During candidate exchange, send the Candidate ID in the extension part.
- In STUN messages, include both the source and destination Candidate IDs (this is required for peer-reflexive candidates).
- Move the extra server reflexive candidate handling from the sender to the receiver side.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

### DIFF
--- a/client/android/client.go
+++ b/client/android/client.go
@@ -4,6 +4,7 @@ package android
 
 import (
 	"context"
+	"os"
 	"slices"
 	"sync"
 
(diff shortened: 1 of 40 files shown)


lixmal (approved):

client/iface/udpmux/mux.go
```diff
@@ -260,12 +264,14 @@ func (m *UDPMuxDefault) GetConn(ufrag string, addr net.Addr) (net.PacketConn, er
...
 	}
 
-	c := m.createMuxedConn(ufrag)
+	c := m.createMuxedConn(ufrag, candidateID)
 	go func() {
 		<-c.CloseChannel()
 		m.RemoveConnByUfrag(ufrag)
 	}()
 
+	m.candidateConnMap[candidateID] = c
```
hakansa, line 273: would having a dedicated mutex to `m.candidateConnMap`make any difference in a good way?

pappz: I think we should protect this variable in the same way we protect **connsIPv4** and **connsIPv6**. If I add a new lock, I risk causing a deadlock. This code is simpler.

PR #4517, opened 2025-09-18
[management] Filter DNS records to include only peers to connect

bcmmbaga:
### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:


management/server/types/account.go
```diff
@@ -1651,3 +1655,28 @@ func peerSupportsPortRanges(peerVer string) bool {
 	meetMinVer, err := posture.MeetsMinVersion(firewallRuleMinPortRangesVer, peerVer)
 	return err == nil && meetMinVer
 }
+
+// filterZoneRecordsForPeers filters DNS records to only include peers to connect.
+func filterZoneRecordsForPeers(peer *nbpeer.Peer, customZone nbdns.CustomZone, peersToConnect []*nbpeer.Peer) []nbdns.SimpleRecord {
+	if peer == nil || customZone.Domain == "" {
+		return nil
+	}
+
```
crn4: seems that we don't need these checks since for nil peer we are not building the map, and empty domain was checked in `if` clause

PR #4545, opened 2025-09-25
[client, signal] Refactor/reduce signaling

pappz:
## Describe your changes

When the client sends a signaling offer or answer, the Signal server may return an error if the message cannot be delivered. The client then checks the error type to decide how to proceed:

- If the error suggests the remote peer is unavailable, the client waits for the peer instead of retrying immediately.
- If the method is not implemented, then switch back to the original retry mechanism (backward compatibility case)
- If the error indicates another error, the client retries by sending a new offer.

Whenever a peer reconnects to the signaling server and no active peer connection exists, it must send a new offer. This ensures that any signaling messages lost while the peer was offline are recovered and the connection can be re-established.

The NB_DISABLE_SEND_WITH_DELIVERY_CHECK: true|false option can be used to enforce that clients always use the server’s original methods.

For the review, keep in mind that these Signal server changes must support both old and new clients, and old clients must still connect properly with each other.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

### DIFF
--- a/signal/server/signal.go
+++ b/signal/server/signal.go
@@ -147,8 +147,10 @@ func (s *Server) SendWithDeliveryCheck(ctx context.Context, msg *proto.Encrypted
 	if _, found := s.registry.Get(msg.RemoteKey); found && !s.directSendDisabled {
 		if err := s.forwardMessageToPeer(ctx, msg); err != nil {
 			if errors.Is(err, ErrPeerNotConnected) {
+				log.Tracef("remote peer [%s] not connected", msg.RemoteKey)
 				return nil, status.Errorf(codes.NotFound, "remote peer not connected")
 			}
+			log.Errorf("error sending message with delivery check to peer [%s]: %v", msg.RemoteKey, err)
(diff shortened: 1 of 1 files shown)


client/internal/peer/conn.go
```diff
@@ -772,6 +806,17 @@ func (conn *Conn) rosenpassDetermKey() (*wgtypes.Key, error) {
 	return &key, nil
 }
 
+func (conn *Conn) switchGuard() {
+	conn.guardCtxCancel()
+	conn.wgGuard.Wait()
+	conn.wg.Add(1)
+	go func() {
+		defer conn.wg.Done()
+		conn.guard = guard.NewRetryGuard(conn.Log, conn.isConnectedOnAllWay, conn.config.Timeout, conn.srWatcher)
```
hakansa, line 826: this could be called twice and `conn.guard` reassigned racy, isn't it?

hakansa (approved):

PR #4563, opened 2025-10-01
Modify client-side behavior

mohamed-essam:
## Describe your changes

* Add an option for only updating on startup (To be implemented on management-side, defaults to true)
* Add timeout for updating on startup
* Add progress window for auto-updates
* Persist state for auto-update to show after-update notification

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:


client/internal/updatemanager/manager.go
```diff
@@ -3,6 +3,7 @@ package updatemanager
 import (
 	"context"
 	"fmt"
+	"github.com/netbirdio/netbird/client/internal/statemanager"
```
pappz: Please keep the order of the imported packages as usual

pappz (changes requested):

client/internal/updatemanager/manager.go
pappz: And everywhere else.

client/internal/connect.go
```diff
@@ -272,6 +273,18 @@ func (c *ConnectClient) run(mobileDependency MobileDependency, runningChan chan
 
 		c.engineMutex.Lock()
 		c.engine = NewEngine(engineCtx, cancel, signalClient, mgmClient, relayManager, engineConfig, mobileDependency, c.statusRecorder, checks)
+		if loginResp.PeerConfig != nil && loginResp.PeerConfig.AutoUpdate != nil {
+			if c.engine.updateManager == nil && loginResp.PeerConfig.AutoUpdate.Version != "disabled" {
```
pappz: Where does this string come from? It should be const

pappz: Code duplication with the engine. Move this logic into a function

pappz (changes requested):

pappz:
Please fix the build issues

client/internal/updatemanager/manager.go
```diff
@@ -66,12 +79,35 @@ func (u *UpdateManager) WithCustomVersionUpdate(versionUpdate UpdateInterface) *
 	return u
 }
 
+func (u *UpdateManager) StartWithTimeout(ctx context.Context, timeout time.Duration) {
```
pappz, line 77: What will happen to the fetcher after a timeout? It seems like it will never be freed

client/internal/updatemanager/manager.go
```diff
@@ -132,12 +188,24 @@ func (u *UpdateManager) Stop() {
 	u.wg.Wait()
 }
 
+func (u *UpdateManager) onContextCancel() {
```
pappz, line 188: You clean up updateManager with Stop and also with context cancellation. This means we have a race condition in the onContextCancel and Stop() functions.

client/internal/updatemanager/manager.go
mohamed-essam: Since timeout will cancel the context, onContextCancel will handle fetcher cleanup

client/internal/updatemanager/manager.go
mohamed-essam: Added the locking mechanism from the original PR modifications

PR #4564, opened 2025-10-01
[client] Recreate agent when receive new session id

pappz:
## Describe your changes

When an ICE agent connection was in progress, new offers were being ignored. This was incorrect logic because the remote agent could be restarted at any time.
In this change, whenever a new session ID is received, the ongoing handshake is closed and a new one is started.

- Populate initial candidates in the ICE monitor.
- Ensure the session ID is updated when the agent is recreated. Because of this, the Offer listener is now synchronized with the ICE listener.
- Set the NB_ICE_MONITOR_PERIOD parameter to be configurable for end-to-end testing purposes

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

### DIFF
--- a/client/internal/peer/conn_test.go
+++ b/client/internal/peer/conn_test.go
@@ -118,10 +118,10 @@ func TestConn_OnRemoteAnswer(t *testing.T) {
 		return
 	}
 
-	onNewOffeChan := make(chan struct{})
+	onNewOfferChan := make(chan struct{})
 
 	conn.handshaker.AddRelayListener(func(remoteOfferAnswer *OfferAnswer) {
-		onNewOffeChan <- struct{}{}
+		onNewOfferChan <- struct{}{}
 	})
 
 	conn.OnRemoteAnswer(OfferAnswer{
@@ -136,7 +136,7 @@ func TestConn_OnRemoteAnswer(t *testing.T) {
 	defer cancel()
 
 	select {
-	case <-onNewOffeChan:
+	case <-onNewOfferChan:
 		// success
 	case <-ctx.Done():
 		t.Error("expected to receive a new offer notification, but timed out")
--- a/client/internal/peer/conn_test.go
+++ b/client/internal/peer/conn_test.go
@@ -81,7 +81,7 @@ func TestConn_OnRemoteOffer(t *testing.T) {
 
 	onNewOffeChan := make(chan struct{})
 
-	conn.handshaker.AddOnNewOfferListener(func(remoteOfferAnswer *OfferAnswer) {
+	conn.handshaker.AddRelayListener(func(remoteOfferAnswer *OfferAnswer) {
 		onNewOffeChan <- struct{}{}
 	})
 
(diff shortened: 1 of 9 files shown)


client/internal/peer/guard/env.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+const (
+	envICEMonitorPeriod = "NB_ICE_MONITOR_PERIOD"
+)
+
+func GetICEMonitorPeriod() time.Duration {
+	if envVal := os.Getenv(envICEMonitorPeriod); envVal != "" {
+		if seconds, err := strconv.Atoi(envVal); err == nil && seconds > 0 {
+			return time.Duration(seconds) * time.Second
```
hakansa, line 16: why we dont use `seconds * time.Second` here?

client/internal/peer/conn_test.go
```diff
@@ -81,7 +81,7 @@ func TestConn_OnRemoteOffer(t *testing.T) {
 
 	onNewOffeChan := make(chan struct{})
```
hakansa:
```suggestion
	onNewOfferChan := make(chan struct{})
```

client/internal/peer/worker_ice.go
```diff
@@ -92,30 +92,29 @@ func NewWorkerICE(ctx context.Context, log *log.Entry, config ConnConfig, conn *
 func (w *WorkerICE) OnNewOffer(remoteOfferAnswer *OfferAnswer) {
 	w.log.Debugf("OnNewOffer for ICE, serial: %s", remoteOfferAnswer.SessionIDString())
 	w.muxAgent.Lock()
+	defer w.muxAgent.Unlock()
 
-	if w.agentConnecting {
```
hakansa, line 96: could you explain why we create a new session if status is agent connection now?

client/internal/peer/guard/env.go
pappz: Type mismatch: `seconds` is an `int`, `time.Second `is a `time.Duration`, and we expect a `time.Duration`.

client/internal/peer/worker_ice.go
pappz: The purpose of the session ID is to establish an agent connection with the correct and most recent offer. Since gRPC may cache a large number of deprecated signal messages, it is not guaranteed that a response will be provided for every offer. The session ID therefore functions as a synchronization mechanism between the our agent and the remote agent.

We need to recreate the agent in an edge case. A synchronization issue can occur when, after a successful offer–answer exchange, both peers start the agent creation process, but the remote peer has been restarted for some reason. Because the remote peer has been reset, its credentials have changed, so we cannot reuse the current session. The remote peer sends a completely new session ID, which indicates that something has happened on the remote side. Therefore, it makes no sense to continue the current agent creation process.

We try to enforce creating agents with the same session ID pairs on both sides.

hakansa (approved):

PR #4574, opened 2025-10-02
[client] Implement DNS query caching in DNSForwarder

hakansa:
## Describe your changes
[client] Implement DNS query caching in DNSForwarder
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

.

### DIFF
--- a/client/internal/dnsfwd/cache.go
+++ b/client/internal/dnsfwd/cache.go
@@ -33,9 +33,9 @@ func (c *cache) get(domain string, reqType uint16) ([]netip.Addr, bool) {
 
 	switch reqType {
 	case dns.TypeA:
-		return cloneAddrs(entry.ip4Addrs), entry.ip4Addrs != nil
+		return cloneAddrs(entry.ip4Addrs), len(entry.ip4Addrs) > 0
 	case dns.TypeAAAA:
-		return cloneAddrs(entry.ip6Addrs), entry.ip6Addrs != nil
+		return cloneAddrs(entry.ip6Addrs), len(entry.ip6Addrs) > 0
 	default:
 		return nil, false
 	}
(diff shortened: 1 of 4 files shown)


client/internal/dnsfwd/cache.go
```diff
@@ -0,0 +1,79 @@
...
+	norm := normalizeDomain(domain)
+	entry, exists := c.records[norm]
+	if !exists {
+		entry = &cacheEntry{}
+		c.records[norm] = entry
+	}
+
+	switch reqType {
+	case dns.TypeA:
+		entry.ip4Addrs = cloneAddrs(addrs)
```
lixmal: We can use slices.Clone here

lixmal (commented):
- Currently the cache is unbounded and there is no way to manually clear it, can we do something about it?
- NXDOMAIN is not cached, but it should be
- `UpdateDomains` might remove some domains in which case we should remove those from the cache as well

hakansa:
@lixmal added cache for NXDOMAIN, also added a method to remove stale cache entries on UpdateDomains. 

Regarding to the clearing cache, maybe we can expire not used cache items after some time but i think it can be implemented later /cc @mlsmaycon

lixmal (approved):

solariz:
Thanks for implementing that, can help quite a lot.
Not a go coder but if I may ask some things I did not figured out looking at the code:
- is it only cached for the time netbird is running? 
- On long running instances `records` could potentially get quite big
- so for later improvements cache should respect ttl
- for now it is only implemented as fallback if the initial lookup failed, would be great to have it in future -maybe first optional- to answer from cache first if the record is fresh. But for sure this needs a garbage collection and reliant ttl cleanup


Again thanks for implementing this fallback.

Issue #2317, opened 2024-07-24
Add an arm64 build for Windows


PeakeIT:
**Is your feature request related to a problem? Please describe.**
Reduced performance and battery life on Snapdragon X Elite processors (arm64).

**Describe the solution you'd like**
Provide a Windows installer with a build compiled for arm64.

**Describe alternatives you've considered**
The amd64 build works at the expense of reduced performance and battery life.

**Additional context**
N/A.

m0nji:
would love to know if there is a ETA for the arm64 release. the  x64 release does weird reconnection to its peers

glitchsys:
+1 this request/feature

xan-it:
+1

webash:
Yes please!

sh00t3r:
+1

kiwibro8:
+1

ordovice:
I'd like to +1 this now that Microsoft Surface laptops are moving only to ARM territory.

ThorTheStorm:
+1

si458:
would love an arm64 version here too! if you can point me to build docs, i can always trying building one myself haha

grantcurell:
+1 - the x86 client doesn't work correctly. On Mac silicon everything works great, but you have to use the x86 client on Windows and it definitely doesn't. I've gotten it to fire up once or twice but the rest of the time it just gets stuck on connecting.

mlsmaycon:
we are building arm64 versions for windows: https://github.com/netbirdio/netbird/releases/download/v0.56.1/netbird_installer_0.56.1_windows_arm64.exe

We will add the links this week.

xan-it:
Great!

grantcurell:
> we are building arm64 versions for windows: https://github.com/netbirdio/netbird/releases/download/v0.56.1/netbird_installer_0.56.1_windows_arm64.exe
> 
> We will add the links this week.

Works like a champ!

webash:
#4206 appears to have fully merged this in and releases are being generated as of 0.54.0 (https://github.com/netbirdio/netbird/releases/tag/v0.54.0)

Perhaps the amd64 client should detect if its running on a arm64 machine and suggest changing to that version on an update?

But I guess this request can be closed!

mlsmaycon:
This won't be possible yet. But we will add the link to download on our dashboard really soon. 

In the meantime, please use these links for the latest downloads:

EXE:
https://pkgs.netbird.io/windows/arm64
MSI:
[...]

PR #4586, opened 2025-10-04
Update url_windows.go to offer arm64 executable download

webash:
## Describe your changes

When the Windows arm64 client detects an update, it offers the amd64 download URL instead of the arm64 one; so I borrowed the selection pattern from the [macOS client's url_darwin.go ](https://github.com/netbirdio/netbird/blob/e7b5537dcc280384470668f461bbb1f7d2f41218/version/url_darwin.go)

Relates to #2317 

### Checklist
- [X] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

### DIFF
--- a/version/url_windows.go
+++ b/version/url_windows.go
@@ -1,9 +1,13 @@
 package version
 
-import "golang.org/x/sys/windows/registry"
+import (
+	"golang.org/x/sys/windows/registry"
+	"runtime"
+)
 
 const (
 	urlWinExe = "https://pkgs.netbird.io/windows/x64"
+	urlWinExeArm = "https://pkgs.netbird.io/windows/arm64"
 )
 
 var regKeyAppPath = "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\Netbird"
@@ -14,6 +18,16 @@ func DownloadUrl() string {
 	if err == nil {
 		return urlWinExe
 	} else {
+		return downloadURL
+	}
+
+PKGINSTALL:
+	switch runtime.GOARCH {
+	case "amd64":
+		return urlWinExe
+	case "arm64":
+		return urlWinExeArm
+	default:
 		return downloadURL
 	}
 }


version/url_windows.go
```diff
@@ -12,8 +16,18 @@ var regKeyAppPath = "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\Ne
...
+
+PKGINSTALL:
+	switch runtime.GOARCH {
+	case "amd64":
+		return urlWinExe
+	case "arm64":
+		return urlWinExeArm
+	default:
+		return downloadURL
+	}
```
mlsmaycon:
```suggestion
  if err != nil {
    return downloadURL
  }

  url := urlWinExe
  if untime.GOARCH == arm64 {
    url = urlWinExeArm
  }

  return url
```

mlsmaycon (changes requested):
Thanks. I've left a small suggestion that alignes better to Go's style.

webash:
Thanks @mlsmaycon - I'm not at all a 'go' coder, so was just doing my best to copy a pattern used by netbird in one of the other distros; but I can definitely see how this is much better.

Should all be ready to go now.

version/url_windows.go
```diff
@@ -1,19 +1,28 @@
...
 	_, err := registry.OpenKey(registry.LOCAL_MACHINE, regKeyAppPath, registry.QUERY_VALUE)
-	if err == nil {
-		return urlWinExe
-	} else {
+	if err != nil {
 		return downloadURL
 	}
+	
+	url := urlWinExe
+	if untime.GOARCH == arm64 {
```
mlsmaycon:
```suggestion
	if runtime.GOARCH == "arm64" {
```

mlsmaycon: my bad on this one @webash , it seems like I removed a few letters and quotes.

PR #4607, opened 2025-10-08
[android] allow selection/deselection of network resources on android peers

doromaraujo:
## Describe your changes

Allows user to toggle network resources on android client via exported functions.

Since we can't update the routes table on android's tun interface after creating a VPN when a network route change is detected, this also exports a function to renew the tun.Device's file descriptor used in android devices with a new implementation of WireGuard's tun.Device interface.

RenewableTUN is used in place of the created unmonitored tun on client/iface/device/device_android's Create function when passing a tun.Device to WGTunDevice's filteredDevice field. The unmonitored tun is added as a device to RenewableTUN, which manages a queue of tun.Devices internally. 

Every call to a tun.Device method implemented by RenewableTUN uses the latest queued tun.Device's implementation.

When renewing a file descriptor in device_android, a new unmonitored tun is created with it, then it is added to RenewableTUN, deferring closure of the oldest tun.Device managed by it.

This also fixes a bug that didn't allow android clients to properly save its current profile's state with the same approach
used in iOS.

https://linear.app/netbird/issue/NET-500/route-selection-on-android

### Checklist
- [X] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:


client/android/client.go
```diff
@@ -4,6 +4,10 @@ package android
 
 import (
 	"context"
+	"fmt"
+	"github.com/netbirdio/netbird/route"
```
pappz: Please keep the orders of the imports:
- std pkgs
- external pkgs
- our pkgs

client/iface/device/renewable_tun.go
```diff
@@ -0,0 +1,173 @@
...
+	log.Debug("reading from device.")
+
+	device := r.peekLast()
+	if device == nil {
+		return 0, errors.New("no available devices")
+	}
+
+	n, err = device.Read(bufs, sizes, offset)
+
+	if err != nil {
```
pappz: Why do you ignore the Read error?

client/iface/device/renewable_tun.go
```diff
@@ -0,0 +1,173 @@
...
+	"errors"
+	"os"
+	"sync"
+
+	log "github.com/sirupsen/logrus"
+	"golang.zx2c4.com/wireguard/tun"
+)
+
+type RenewableTUN struct {
+	devices []tun.Device
```
pappz: Do we really need to a list? Maybe I do not understand well, but looks like we maintain only one active device at the same time.

client/iface/device/renewable_tun.go
pappz: Why do you ignore the Read error?

client/iface/device/renewable_tun.go
```diff
@@ -0,0 +1,173 @@
...
+	log.Debug("writing to device.")
+
+	device := r.peekLast()
+	if device == nil {
+		return 0, nil
+	}
+
+	n, err := device.Write(bufs, offset)
+
+	if err != nil {
```
pappz: Why do you ignore the Write error?

client/iface/device/renewable_tun.go
```diff
@@ -0,0 +1,173 @@
...
+	return err
+}
+
+func (r *RenewableTUN) BatchSize() int {
+	log.Debug("returning batch size.")
+
+	return 1
+}
+
+func (r *RenewableTUN) addDevice(device tun.Device) {
```
pappz: Please export this function too

client/iface/device/renewable_tun.go
```diff
@@ -0,0 +1,173 @@
...
+	if len(r.devices) == 0 {
+		return nil
+	}
+
+	first := r.devices[0]
+	r.devices = r.devices[1:]
+	return first
+}
+
+func NewRenewableTUN() *RenewableTUN {
```
pappz: Please move this after the type definition

pappz (changes requested):

client/iface/device/renewable_tun.go
doromaraujo: The list is used to defer closing the first dequeued tun.Device after a new one is added. The most recent one is always chosen to be picked to be read from/written to. Tried just "hot-swapping" them, but that doesn't work.

client/iface/device/renewable_tun.go
doromaraujo: Because if it returns an error while reading after the file descriptor is closed, the engine calls RenewableTUN's Close implementation of tun.Device and it screws everything up, connection never gets reestablished even if you add a new device. 

I made a commit wrapping the tun.Device into a new type that keeps a flag to check if the device is already closed.

client/iface/device/renewable_tun.go
doromaraujo: Same reason as above.

client/iface/device/renewable_tun.go
doromaraujo: Done.

client/iface/device/renewable_tun.go
doromaraujo: Done.

client/iface/device/renewable_tun.go
```diff
@@ -0,0 +1,220 @@
...
+		return 0, errors.New("no available devices")
+	}
+
+	n, err = device.Read(bufs, sizes, offset)
+
+	if err != nil {
+		log.Debugf("error reading from device: %v", err)
+
+		if device.IsClosed() {
+			err = nil
```
pappz: In this case, we will lose packages. Right? We should solve this somehow.
I see multiple options.
- Syncrhonize the AddDevice and Read, Write functions. So the r.peekLast will provide an up-to-date device.
- In case of error, retry to read, write the package into, from a "newer" device if it is available. So rerun the r.peekLast() in a for loop while there are new devices.

pappz (changes requested):

client/iface/device/renewable_tun.go
doromaraujo: I refactored the code to make use of these for loops.
Please take a look at it again.

client/iface/device/device_netstack_android.go
```diff
@@ -5,3 +5,7 @@ package device
 func (t *TunNetstackDevice) Create(routes []string, dns string, searchDomains []string) (WGConfigurer, error) {
 	return t.create()
 }
+
+func (t *TunNetstackDevice) RenewTun(fd int) error {
+	return t.RenewTun(fd)
+}
```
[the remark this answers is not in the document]
pappz: @doromaraujo Could you dobule check this?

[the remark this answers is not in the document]
doromaraujo: On commit 61af625735ca5211be90da6d495328963649d472 I changed this to return an error as we discussed yesterday.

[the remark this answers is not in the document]
pappz: @coderabbitai is it fine now?

client/iface/device/renewable_tun.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+
+		n, err = dev.Read(bufs, sizes, offset)
+		if err == nil {
+			return n, nil
+		}
+
+		// swap in progress; retry on the newest instead of returning the error
+		if dev.IsClosed() {
+			time.Sleep(1 * time.Millisecond)
```
pappz, line 148: Why do we need to this? If you remove the sleep then in worst case just hang for a while on r.waitForDevice(). Am I right?

client/iface/device/renewable_tun.go
```diff
@@ -0,0 +1,302 @@
...
+			select {
+			case ev, ok := <-c.Device.Events():
+				if !ok {
+					return
+				}
+				select {
+				case out <- ev:
+				case <-c.closeEventCh:
+					return
+				default:
```
pappz: Here we loose events silently. I think we could remofe the default because it does not cause any deadlock

client/iface/device/renewable_tun.go
```diff
@@ -0,0 +1,302 @@
...
+	if len(r.devices) > 0 {
+		toClose = r.devices[len(r.devices)-1]
+	}
+
+	cad := newClosableDevice(device)
+	cad.redirectEvents(r.events)
+	r.devices = []*closeAwareDevice{cad}
+	r.cond.Broadcast()
+	r.mu.Unlock()
+
```
pappz, line 282: At this moment two or more devices can write to the events channel. It means that the events from the reader perspective can be mixed. For example you call the cad.redirectEvents() on the new device and it send a UP event inmediatelly. The deprecated device just close the cannel a littlebit later and it send a DOWN event. I do not know what could it cause.

client/iface/device/renewable_tun.go
```diff
@@ -0,0 +1,302 @@
...
+	close(r.events)
+	return lastErr
+}
+
+func (r *RenewableTUN) BatchSize() int {
+	return 1
+}
+
+func (r *RenewableTUN) AddDevice(device tun.Device) {
+	if r.closed.Load() {
```
pappz, line 264: TOCTOU (Time-of-Check-Time-of-Use) issue.
If you run parallel a `Close` call after line 262 then add a device to the list but the struct never clean it up.

client/iface/device/renewable_tun.go
doromaraujo: Resolved, default is removed.

client/iface/device/renewable_tun.go
doromaraujo: waitForDevice only waits when the list of devices is empty, so if a device exists but is in the closed state it will just loop until a new one is added. This sleep is used just as a micro backoff.

client/iface/device/renewable_tun.go
doromaraujo: Changed. Lock is acquired right at the start now.

client/iface/device/renewable_tun.go
doromaraujo: tun.EventDown is now being ignored when redirecting events.

client/iface/device/renewable_tun.go
```diff
@@ -49,6 +49,11 @@ func (c *closeAwareDevice) redirectEvents(out chan tun.Event) {
 				if !ok {
 					return
 				}
+
```
pappz, line 52: if you fix the gracefull device renew then make no sense for this line I think.

client/iface/device/renewable_tun.go
pappz: With this you hardcoded a fix delay time for the Engine restart procedure.

client/iface/device/renewable_tun.go
pappz: This still doesn’t look completely safe. If you run AddDevice multiple times in parallel, it could happen that one thread closes the deprecated device while another thread runs cad.redirectEvents. When you leave the protected scope, the order of execution becomes unpredictable.

client/iface/device/renewable_tun.go
doromaraujo: The events channel of the underlying tun.Device might close outside of our scope, as in the routineNetlinkListener inside tun_linux.go, which is the tun.Device implementation being used for android. Best to keep it.

client/iface/device/renewable_tun.go
doromaraujo: I changed it so the events will still be redirected inside the lock. Device closure is outside the lock still, though.

client/iface/device/renewable_tun.go
doromaraujo: It's preferable from it looping non-stop in the scenario a device is closed, in my opinion.

pappz: The waitForDevice prevent the loop. But fine

PR #4709, opened 2025-10-27
[client] Block on all subsystems on shutdown

lixmal:
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -926,17 +926,18 @@ func (e *Engine) updateSSH(sshConf *mgmProto.SSHConfig) error {
 				listenAddr = fmt.Sprintf("127.0.0.1:%d", nbssh.DefaultSSHPort)
 			}
 			// nil sshServer means it has not yet been started
-			var err error
-			e.sshServer, err = e.sshServerFunc(e.config.SSHKey, listenAddr)
-
+			server, err := e.sshServerFunc(e.config.SSHKey, listenAddr)
 			if err != nil {
 				e.sshMux.Unlock()
 				return fmt.Errorf("create ssh server: %w", err)
 			}
+
+			e.sshServer = server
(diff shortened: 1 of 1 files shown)


client/internal/peer/guard/sr_watcher.go
```diff
@@ -52,22 +52,28 @@ func (w *SRWatcher) Start() {
 	w.cancelIceMonitor = cancel
 
 	iceMonitor := NewICEMonitor(w.iFaceDiscover, w.iceConfig, GetICEMonitorPeriod())
-	go iceMonitor.Start(ctx, w.onICEChanged)
+	w.shutdownWg.Add(1)
```
pappz: I don’t think it makes sense to wait for this. This struct is already used by the guard code, which waits for the exit and removes the listeners beforehand. It has no effect if it keeps running in the background for a while. This wait group only slows down the restart, and restart speed will be important on mobile due to route changes.

lixmal: Reverted that part

client/internal/engine.go
```diff
@@ -925,32 +930,37 @@ func (e *Engine) updateSSH(sshConf *mgmProto.SSHConfig) error {
...
 			if err != nil {
+				e.sshMux.Unlock()
 				return fmt.Errorf("create ssh server: %w", err)
 			}
-			e.shutdownWg.Add(1)
+			e.sshMux.Unlock()
 			go func() {
-				defer e.shutdownWg.Done()
 				// blocking
 				err = e.sshServer.Start()
```
pappz: touch the e.sshServer in unprotected scope

lixmal: fixed

PR #4717, opened 2025-10-29
[client] Add quick actions window

doromaraujo:
## Describe your changes

Adds a window that will pop up when running NetBird on desktop after it is already running.
It sends a SIGUSR1 to a previously set up signal handler (platform-dependent) that will handle opening the window itself.

The window only contains the NetBird logo with a single button indicating the action to connect or disconnect.
When connected, the NetBird logo is colored, when disconnected, the logo is greyed out.

On connection change, the window closes on its own.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Quick Actions UI window for expedited connection management
  * --quick-actions command-line mode
  * Disconnected-state icon and visual indicator

* **Improvements**
  * Better daemon detection with clearer messaging and show-window signaling
  * Cross-platform signal/event handling to trigger the Quick Actions window
  * Faster icon refresh on theme/settings changes
  * Upload success flow now integrates clipboard behavior with the app context

* **Chores**
  * Updated project dependencies

### DIFF
--- a/client/ui/signal_windows.go
+++ b/client/ui/signal_windows.go
@@ -5,6 +5,7 @@ package main
 import (
 	"context"
 	"errors"
+	"fmt"
 	"os"
 	"os/exec"
 	"time"


client/ui/quickactions.go
```diff
@@ -0,0 +1,292 @@
...
+		uiChan:     uiChan,
+		pauseChan:  make(chan struct{}),
+		resumeChan: make(chan struct{}),
+	}
+
+	// base UI status
+	uiChan <- newQuickActionsUiState()
+
+	// this retrieves the current connection status
+	go func() {
```
lixmal: This goroutine should take a ctx and check for its cancellation in the for loop to avoid leaking the it.
Also  move the function body to a named function to improve readability

client/ui/quickactions.go
```diff
@@ -0,0 +1,292 @@
...
+				toggleConnectionButton.Refresh()
+
+				// second position in the content's objects array is the NetBird logo.
+				content.Objects[1] = logo
+				content.Refresh()
+			})
+
+			toggleConnectionButton.OnTapped = func() {
+				go func() {
+					uiState.toggleAction()
```
lixmal: Why do we set this in a goroutine? That looks like it can race with something else.

client/ui/quickactions.go
```diff
@@ -0,0 +1,292 @@
...
+					uiState.toggleAction = func() {
+						viewModel.executeCommand(connect)
+					}
+				}
+
+				uiState.isToggleButtonEnabled = true
+				uiState.connectionStatus = connectionStatus
+				//uiState.statusLabelText = fmt.Sprintf("Connection status: %s", connectionStatus)
+				uiChan <- uiState
+				time.Sleep(500 * time.Millisecond)
```
lixmal: Please use a ticker here and add is as select case.

client/ui/quickactions.go
```diff
@@ -0,0 +1,292 @@
...
+
+	err := command.execute()
+
+	if err != nil {
+		log.Errorf("Status: Error - %v", err)
+	} else {
+		uiState = newQuickActionsUiState()
+		uiState.isConnectionChanged = true
+		q.uiChan <- uiState
+		//q.resumeChan <- struct{}{}
```
lixmal: This doesn't seem to be fully implemented

client/ui/quickactions.go
```diff
@@ -0,0 +1,292 @@
...
+			return s.menuDownClick()
+		},
+	}
+
+	if err != nil {
+		log.Errorf("get service client: %v", err)
+		return
+	}
+
+	uiChan := make(chan quickActionsUiState, 1)
```
lixmal, line 285: This channel should be closed, e.g. in a defer. Consider using `sync.Once` to avoid panics if it can be closed in more than one code path

client/ui/quickactions.go
```diff
@@ -0,0 +1,292 @@
...
+				if !getConnectionStatus {
+					continue
+				}
+
+				uiState := newQuickActionsUiState()
+				connectionStatus, err := provider.connectionStatus()
+
+				if err != nil {
+					log.Errorf("Status: Error - %v", err)
+					uiChan <- uiState
```
lixmal: After this we should `continue` or `return`, right?

client/ui/signal_windows.go
```diff
@@ -0,0 +1,171 @@
...
+			return
+		}
+	}
+
+	if eventHandle == windows.InvalidHandle {
+		log.Errorf("Obtained an invalid handle for quick actions trigger event '%s'", quickActionsTriggerEventName)
+		return
+	}
+
+	log.Infof("Debug handler waiting for signal on event: %s", quickActionsTriggerEventName)
```
lixmal: Log still says debug handler

client/ui/signal_windows.go
```diff
@@ -0,0 +1,171 @@
...
+
+	eventHandle, err := windows.OpenEvent(desiredAccesses, false, eventNamePtr)
+	if err != nil {
+		return err
+	}
+
+	err = windows.SetEvent(eventHandle)
+	if err != nil {
+		log.Printf("Error setting event: %v", err)
+		return err
```
lixmal: Return meaningful error context instead of logging an error + returning `err`



```suggestion
		return fmt.Errorf("set event: %w", err)
```

client/ui/quickactions.go
```diff
@@ -0,0 +1,292 @@
...
+
+func (c disconnectCommand) execute() error {
+	return c.disconnectClient()
+}
+
+type quickActionsViewModel struct {
+	provider   clientConnectionStatusProvider
+	connect    clientCommand
+	disconnect clientCommand
+	uiChan     chan quickActionsUiState
```
lixmal: This field is accessed from multiple goroutines but it's not protected

lixmal (changes requested):

client/ui/quickactions.go
```diff
@@ -0,0 +1,292 @@
...
+type clientConnectionStatusProvider interface {
+	connectionStatus() (string, error)
+}
+
+type daemonClientConnectionStatusProvider struct {
+	client proto.DaemonServiceClient
+}
+
+func (d daemonClientConnectionStatusProvider) connectionStatus() (string, error) {
+	status, err := d.client.Status(context.Background(), &proto.StatusRequest{})
```
lixmal: This method should accept a `ctx` paramenter instead of using Background

mlsmaycon:
@coderabbitai review

client/ui/quickactions.go
```diff
@@ -0,0 +1,292 @@
...
+
+			if uiState.connectionStatus == "Connected" {
+				buttonText = connectedLabelText
+				buttonIcon = connectedCircle.Resource
+				logo = connectedImage
+			} else if uiState.connectionStatus == "Idle" {
+				buttonText = disconnectedLabelText
+				buttonIcon = disconnectedCircle.Resource
+				logo = disconnectedImage
+			}
```
lixmal: What about the other states? How do we handle those?

doromaraujo: Other states won't be displayed in this UI. It's either "Connect " when disconnected or "Disconnect" when connected, and when clicked the button will be disabled until it either switches connection status or an error occurs.

client/ui/quickactions.go
doromaraujo: It's the command to be used (connect, disconnect) when user taps the button. It's updated via the UI channel; it's also the only place where it is set.

client/ui/quickactions.go
doromaraujo: Not using it anymore, please the updated changes.

client/ui/quickactions.go
doromaraujo: Done.

client/ui/quickactions.go
doromaraujo: Done.

client/ui/quickactions.go
doromaraujo: Yup. Done.

client/ui/signal_windows.go
doromaraujo: Fixed.

client/ui/signal_windows.go
doromaraujo: Done.

client/ui/quickactions.go
doromaraujo: Which field? What protection is needed?

client/ui/quickactions.go
doromaraujo: Done.

client/ui/quickactions.go
doromaraujo: The goroutine that uses it is now making use of ctx.Done to return from it. More than one place in the view model send UI states to it.

lixmal (approved):

PR #4732, opened 2025-10-31
[client] auto-update

pappz:
## Describe your changes

- Only execute the update when there is human interaction with the "Connect" button in the UI.
- Run the installation process from a detached binary

### Checklist
- [x] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Background auto-update with per-account/version controls and in-app update progress UI
  * Cross-platform updater/installer with installer result reporting to the daemon and UI
  * New signer CLI for creating, bundling, signing, and verifying artifact keys and revocation lists

* **Documentation**
  * Added docs describing updater/installer and signing/verification workflows; embedded verification keys

* **Other**
  * Debug bundles now include updater logs; downloader/installer utilities and extensive tests added

### DIFF
--- a/management/server/http/handlers/accounts/accounts_handler.go
+++ b/management/server/http/handlers/accounts/accounts_handler.go
@@ -11,6 +11,7 @@ import (
 	"github.com/gorilla/mux"
 
 	goversion "github.com/hashicorp/go-version"
+
 	"github.com/netbirdio/netbird/management/server/account"
 	nbcontext "github.com/netbirdio/netbird/management/server/context"
 	"github.com/netbirdio/netbird/management/server/settings"
@@ -248,6 +249,7 @@ func (h *handler) updateAccount(w http.ResponseWriter, r *http.Request) {
 	settings, err := h.updateAccountRequestSettings(req)
 	if err != nil {
 		util.WriteError(r.Context(), err, w)
+		return
 	}
 	if req.Settings.NetworkRange != nil && *req.Settings.NetworkRange != "" {
 		prefix, err := netip.ParsePrefix(*req.Settings.NetworkRange)


pappz:
@coderabbitai review this PR

client/internal/updatemanager/manager.go
```diff
@@ -61,139 +60,173 @@ type UpdateManager struct {
...
-	case u.mgmUpdateChan <- struct{}{}:
+	case m.mgmUpdateChan <- struct{}{}:
 	default:
 	}
 }
 
-func (u *UpdateManager) Stop() {
-	if u.cancel == nil {
+func (m *Manager) Stop() {
+	if m.cancel == nil {
```
lixmal, line 188: This reads `m.cancel` outside the lock.

client/internal/updatemanager/manager.go
```diff
@@ -61,139 +60,173 @@ type UpdateManager struct {
...
 		return
 	}
 
-	u.cancel()
-	u.updateMutex.Lock()
-	if u.update != nil {
-		u.update.StopWatch()
-		u.update = nil
+	m.cancel()
```
lixmal, line 192: Here too

client/internal/updatemanager/installer/installer_run_darwin.go
```diff
@@ -0,0 +1,240 @@
...
+		if strings.HasPrefix(trimmed, "volume: ") {
+			volume = strings.Split(trimmed, ": ")[1]
+		}
+	}
+
+	cmd := exec.CommandContext(ctx, "installer", "-pkg", path, "-target", volume)
+	if err := cmd.Start(); err != nil {
+		return fmt.Errorf("error running pkg file: %w", err)
+	}
+	res, err := cmd.CombinedOutput()
```
lixmal: This is weird, why do we have both `Start` and `CombinedOutput`? It should only be one of them, not both

client/internal/updatemanager/manager.go
```diff
@@ -211,176 +244,122 @@ func (u *UpdateManager) handleUpdate(ctx context.Context) {
...
+	m.statusRecorder.PublishEvent(
+		cProto.SystemEvent_CRITICAL,
 		cProto.SystemEvent_SYSTEM,
 		"",
 		"",
-		map[string]string{"progress_window": "show"},
+		map[string]string{"progress_window": "show", "version": updateVersion.String()},
 	)
 
 	updateState := UpdateState{
```
lixmal, line 286: This should probably go after successful `triggerUpdateFn` (line 200), right?

client/internal/updatemanager/installer/result.go
```diff
@@ -0,0 +1,223 @@
...
+		return result, nil
+	}
+
+	dir := filepath.Dir(rh.resultFile)
+
+	if err := rh.waitForDirectory(ctx, dir); err != nil {
+		return Result{}, err
+	}
+
+	return rh.watchForResultFile(ctx, dir)
```
lixmal, line 89: There could be a race between these two (line 78, 89)

pappz: I do not get it

client/internal/updatemanager/manager.go
pappz: The mutex protects the update and expectedVersion fields (as you can see in the comment). This part of the code is not thread-safe by design. The upper layer is responsible for synchronization.

client/internal/updatemanager/manager.go
pappz: The mutex protects the update and expectedVersion fields (as you can see in the comment). This part of the code is not thread-safe by design. The upper layer is responsible for synchronization.

client/internal/updatemanager/installer/installer_run_darwin.go
pappz: Good question. The Start command will kill this service while the installer keeps running in the background. That’s why I chose to use Start(), it detaches and allows it to run in the background. However, if something goes wrong, I want to be able to see the logs. Given these conditions, what’s the best approach?

client/internal/updatemanager/manager.go
pappz: No. With this action open the progress window in the UI.

lixmal: But why write the state before success?

client/internal/updatemanager/installer/result.go
lixmal: tryReadResult -> another thread changes something -> watching the result file

client/internal/updatemanager/manager.go
pappz: On line 274, we only notify the UI about the update. We write the version to disk so we can determine whether the update succeeded after the restart. Correct me if this logic is wrong, it's legacy code.

client/internal/updatemanager/installer/installer_run_darwin.go
lixmal: You'll need to use `cmd.Stderr` and `cmd.Stdout` after `cmd.Start()`. `CombinedOutput` will run another `Start()`

client/internal/updatemanager/installer/result.go
pappz: What do you think about the update?

lixmal: Looks good

management/server/http/handlers/accounts/accounts_handler.go
```diff
@@ -206,25 +187,67 @@ func (h *handler) updateAccount(w http.ResponseWriter, r *http.Request) {
...
+	err = json.NewDecoder(r.Body).Decode(&req)
+	if err != nil {
+		util.WriteErrorResponse("couldn't parse JSON request", http.StatusBadRequest, w)
+		return
+	}
+
+	settings, err := h.updateAccountRequestSettings(req)
+	if err != nil {
+		util.WriteError(r.Context(), err, w)
 	}
```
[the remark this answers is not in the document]
pascal-fischer: This is a valid concern. If we do not fix it we might panic

client/internal/updatemanager/manager.go
```diff
@@ -0,0 +1,361 @@
...
+		m.statusRecorder.PublishEvent(
+			cProto.SystemEvent_INFO,
+			cProto.SystemEvent_SYSTEM,
+			"Auto-update completed",
+			fmt.Sprintf("Your NetBird Client was auto-updated to version %s", m.currentVersion),
+			nil,
+		)
+		return
+	}
+}
```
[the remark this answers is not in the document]
lixmal: Looks valid @pappz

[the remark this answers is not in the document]
pappz: I don’t think so. After updateState, err := m.loadUpdateState(ctx), we just check the version and send the notification, but we don’t have any error handling there. So it doesn’t make sense to prevent the delete operation.

PR #4812, opened 2025-11-19
add apt install resolvconf for fix resolvconf missing make wt0 interface

### LINKED ISSUE #4811: systemd service repeatedly destroys wt0 when **resolvconf** is missing → exit 99 → self-stop loop
## Describe the bug  
On Debian/Ubuntu systems **without `resolvconf` installed**, NetBird’s systemd service enters a **4-second self-restart loop**:

1. Engine configures DNS for `wt0` → `resolvconf -a wt0 …` **returns exit status 99**
2. NetBird treats this as **fatal**, logs `stopped NetBird Engine`, and **actively calls `free_netdev()`**
3. `wt0` is destroyed → reconnect → loop repeats  
   *(Docker images are unaffected because they usually ship `resolvconf`)*

---

## To Reproduce  
1. `sudo apt install resolvconf`  
2. `sudo systemctl restart netbird`  
3. `ip link show wt0` → interface **appears and disappears every 4–5 s**  
4. `journalctl -u netbird -f` → endless **“stopped NetBird Engine”**

---

## Expected behavior  
DNS setup failure should be **logged only**, not **fatal**.  
`wt0` should **remain up** and retry DNS later, **not self-stop**.

---

## Screenshots / Logs  
```text
Nov 19 10:46:39 host netbird[374638]: resolvconf: exit status 99
Nov 19 10:46:39 host netbird[374638]: stopped Netbird Engine
Nov 19 10:46:43 host netbird[374638]: WireGuard interface monitor: interface wt0 deleted: failed to lookup interface: route ip+net: no such network interface, restarting engine

yimiaoxiehou:
as title to fix #4811 

## Summary by CodeRabbit

* **Chores**
  * Updated installation process to include an additional system dependency, ensuring all required components are installed during setup.

### DIFF
--- a/release_files/install.sh
+++ b/release_files/install.sh
@@ -223,7 +223,7 @@ install_netbird() {
     case "$PACKAGE_MANAGER" in
     apt)
         add_apt_repo
-        ${SUDO} apt-get install netbird -y
+        ${SUDO} apt-get install resolvconf netbird -y
 
         if ! $SKIP_UI_APP; then
             ${SUDO} apt-get install netbird-ui -y


mlsmaycon:
Thanks for your PR @yimiaoxiehou . We need to investigate the issue before moving forward with it. Also, we should make this installation optional since the default behavior on Debian and Ubuntu is usually fine.

lixmal:
This would collide with e.g. systemd-resolved, so we cannot be opinionated about it, sorry!

PR #4853, opened 2025-11-24
[client] Add excluded port range handling for PKCE flow

mlsmaycon:
## Describe your changes
Introduce support for handling system-excluded port ranges in PKCE flow. Update port validation to skip excluded and in-use ports, including integration of Windows-specific logic to retrieve excluded ranges from the registry. Add tests for various usage scenarios across platforms.
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Bug Fixes**
  * Detect and avoid system-reserved port ranges when selecting redirect URLs on Windows.
  * Port availability checks now respect OS-specific excluded port ranges for more consistent redirect selection.

* **Tests**
  * Added tests covering port exclusion parsing and redirect URL selection behavior across scenarios.

### DIFF
--- a/client/internal/auth/pkce_flow_windows.go
+++ b/client/internal/auth/pkce_flow_windows.go
@@ -3,61 +3,83 @@
 package auth
 
 import (
+	"bufio"
 	"fmt"
+	"os/exec"
 	"strconv"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
-	"golang.org/x/sys/windows/registry"
 )
 
-// getSystemExcludedPortRanges retrieves the excluded port ranges from Windows.
+// getSystemExcludedPortRanges retrieves the excluded port ranges from Windows using netsh.
 func getSystemExcludedPortRanges() []excludedPortRange {
-	ranges, err := getExcludedPortRangesFromRegistry()
-	if err == nil && len(ranges) > 0 {
-		return ranges
+	ranges, err := getExcludedPortRangesFromNetsh()
+	if err != nil {
+		log.Debugf("failed to get Windows excluded port ranges: %v", err)
+		return nil
 	}
 
 	return ranges
 }
 
-// getExcludedPortRangesFromRegistry retrieves excluded port ranges from Windows registry.
-func getExcludedPortRangesFromRegistry() ([]excludedPortRange, error) {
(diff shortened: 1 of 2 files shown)


client/internal/auth/pkce_flow_windows.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil
+	}
+
+	return ranges
+}
+
+// getExcludedPortRangesFromNetsh retrieves excluded port ranges using netsh command.
+func getExcludedPortRangesFromNetsh() ([]excludedPortRange, error) {
+	cmd := exec.Command("netsh", "interface", "ipv4", "show", "excludedportrange", "protocol=tcp")
+	output, err := cmd.Output()
```
lixmal, line 29: This should be CombinedOutput to get stderr as well

client/internal/auth/pkce_flow.go
```diff
@@ -282,15 +284,22 @@ func createCodeChallenge(codeVerifier string) string {
...
 }
 
-// isRedirectURLPortUsed checks if the port used in the redirect URL is in use.
-func isRedirectURLPortUsed(redirectURL string) bool {
+// isRedirectURLPortUsed checks if the port used in the redirect URL is in use or excluded on Windows.
+func isRedirectURLPortUsed(redirectURL string, excludedRanges []excludedPortRange) bool {
 	parsedURL, err := url.Parse(redirectURL)
 	if err != nil {
 		log.Errorf("failed to parse redirect URL: %v", err)
 		return true
```
lixmal, line 292: Should we actually return false here if we don't manage to parse the url?

client/internal/auth/pkce_flow_windows.go
mlsmaycon: For parsing, Output is the best. Its docs also have this in case of error:

```Go
// Output runs the command and returns its standard output. Any returned error will usually be of type ExitError.
//  If c.Stderr was nil and the returned error is of type ExitError, Output populates the Stderr field of the returned error.
```

client/internal/auth/pkce_flow.go
mlsmaycon: no, if we can't parse the URL we should skip it

lixmal (approved):

PR #4866, opened 2025-11-27
feat(client): auto-reconfigure WireGuard interface on peer IP change

fahrishih:
### Added

- Automatic WireGuard interface reconfiguration when peer IP changes
- The NetBird client now detects changes to the self peer’s IP address delivered by the Management Service and applies them live without requiring netbird down / netbird up. This is a more scalable approach for someone who has many peers and impossible to restart 1 by 1.

### Improved

- Enhanced client engine logic to watch for self-IP updates from network map diffs.
- Added logic to update interface address, routes, and internal state when the IP changes.
- Reduced downtime during CIDR migrations and tenant-level IP reassignment.
- Improved consistency between Management-assigned IPs and the active local WireGuard configuration.

### Fixed

- Fixed issue where the client received new IP configuration from Management but did not reconfigure the local interface until a manual reconnect.
- Fixed mismatch between API IP and active wtX/utunX interface, resulting in routing inconsistencies.

### Notes
- Implementation tested on macOS and Linux client builds.
- WG interface name handling remains platform-specific:
- macOS uses utun*
- Linux uses wg0 / kernel interface path

**This enhancement significantly improves NetBird’s dynamic IP behavior and avoids downtime during network CIDR changes.**

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Hot-update of the local WireGuard IP during config changes.
  * Automatic SSH rebind and firewall DNAT refresh when the node IP changes.
  * Local IP state is refreshed to reflect address changes.

* **Bug Fixes**
  * Config updates now continue if an IP update fails; failures are logged rather than aborting.

* **Tests**
  * Added tests for IP-update scenarios, SSH restart lifecycle, and firewall/DNAT interactions.

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -940,6 +940,9 @@ func (e *Engine) updateSelfPeerIP(oldAddr, newAddr string) error {
 
 	log.Infof("peer IP address has changed from %s to %s, updating local WireGuard interface", oldAddr, newAddr)
 
+	// Parse old IP for firewall rule updates (before UpdateAddr changes the interface)
+	oldIP := e.wgInterface.Address().IP
+
 	// Update the WireGuard interface with the new address
 	// The UpdateAddr method handles removing the old address and adding the new one
 	if err := e.wgInterface.UpdateAddr(newAddr); err != nil {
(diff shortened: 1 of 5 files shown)


lixmal:
Hi @fahrishih 

Unfortunately that is not enough. The IP is also used in various subsystems (DNS, Firewall, ...) that would also require updating

fahrishih:
hi @lixmal , which firewall are you referring to? if by dns you mean the netbird dns, i tested and it was able to resolve to the new machine's IP

lixmal:
I mean the firewall implementations (userspace, nftables, iptables) and the DNS server. Pretty much everything that calls `Address()` on the wg interface is affected, especially if it was only called once on setup

fahrishih:
sure @lixmal , but anyways those you mentioned would still not work if the network range is changed from the console, until the user runs netbird down and netbird up again.

at least on this MR, what I aim is to avoid running that command on all the peers deployed and keep the connectivity going by this autoupdate.

fahrishih:
hi @lixmal , i added a new commit to automatically update the nft

When the management server changes a peer's IP address, the SSH server now properly restarts to bind to the new IP.

### Changes:
- Update DNAT rule to match the new IP
- Add Restart method to SSH server for rebinding to new address
- Update updateFirewallForAddressChange to restart SSH server when IP changes
- Add test coverage for SSH restart on IP change

Issue #4869, opened 2025-11-28
SSH broken on Debian 13 due to new login package


afonsofrancof:
# Environment
- **NetBird Version**: 0.60.3
- **OS**: Debian 13 (Trixie)
- **Architecture**: amd64
- **Installation Method**: apt repository

# Description

NetBird SSH connections fail immediately on Debian 13 (Trixie) systems. After successful authentication and PTY allocation, the session closes immediately without spawning an interactive shell. Non-interactive commands work correctly.

# Root Cause

After looking through logs and debugging this issue for **hours** I finally found the cause. 

NetBird's SSH server uses `login` with some flags for interactive sessions which works fine with shadow-login (Debian 12) but fails with util-linux login (Debian 13).

Debian 13 switched from [shadow's](https://packages.debian.org/bookworm/login) to [util-linux](https://packages.debian.org/trixie/login) implementation of the `login` binary (check the right side of the page on those two links. It also shows it on the man page of the packages). The util-linux version behaves differently when invoked by NetBird and exits instead of spawning an interactive shell.

# Behavior

- Connection authenticates successfully
- PTY allocation is accepted
- Shell request is accepted
- Session immediately closes
- Non-interactive commands work: `ssh user@peer 'echo test'` succeeds

## Receiving peer logs (without prefix because of horizontal space):
```
... client/ssh/server/server.go:607: SSH connection from NetBird peer 100.64.0.5:62942 allowed

... client/ssh/server/session_handlers.go:35: SSH session started

... client/ssh/server/command_execution_unix.go:147: starting interactive shell: /usr/bin/login
```

## Connecting peer output

```
❯ ssh afonso@togepi.netbird.selfhosted
Shared connection to togepi.netbird.selfhosted closed.
```

### Workaround

The only way I could get NetBird's ssh to work again was to do the following:

1 - Edit `/etc/apt/sources.list` and change the repositories back to `bookworm`
2 - `sudo apt update`
3 - `sudo apt install login=1:4.13+dfsg1-1+deb12u1`
4 - Use NetBird's SSH (here I didn't use `netbird.selfhosted` suffix but it's the same)

<img width="930" height="246" alt="Image" src="https://github.com/user-attachments/assets/65c7654e-199c-4f60-b8df-dd799f33a614" />

This is obviously not desired and I only used it for testing.

# Change

Something needs to change here

[...]

PR #4873, opened 2025-11-28
SSH: prefer using `runuser` instead of `login`

afonsofrancof:
## Describe your changes

Some distros use util-linux's `login` binary, which doesn't work with NetBird's SSH.
Debian, for example, started using this binary recently in Debian 13, which broke NetBird's SSH.
`runuser` is the recommended utility for this use case and is now tried first on all Linux systems.
If `runuser` doesn't exist, it falls back to the standard login method.
I also removed the special handling for arch-linux, since it just uses `runuser` for that distro now.

#4869

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

### DIFF
--- a/client/ssh/server/userswitching_unix.go
+++ b/client/ssh/server/userswitching_unix.go
@@ -59,8 +59,8 @@ func isFullyNumeric(username string) bool {
 	return true
 }
 
-// createPtyLoginCommand creates a Pty command using login for privileged processes
-func (s *Server) createPtyLoginCommand(localUser *user.User, ptyReq ssh.Pty, session ssh.Session) (*exec.Cmd, error) {
+// createUserSwitchPtyCommand creates a Pty command using login for privileged processes
+func (s *Server) createUserSwitchPtyCommand(localUser *user.User, ptyReq ssh.Pty, session ssh.Session) (*exec.Cmd, error) {
 	loginPath, args, err := s.getLoginCmd(localUser.Username, session.RemoteAddr())
 	if err != nil {
 		return nil, fmt.Errorf("get login command: %w", err)
@@ -75,11 +75,6 @@ func (s *Server) createPtyLoginCommand(localUser *user.User, ptyReq ssh.Pty, ses
 
 // getLoginCmd returns the login command and args for privileged Pty user switching
 func (s *Server) getLoginCmd(username string, remoteAddr net.Addr) (string, []string, error) {
-	loginPath, err := exec.LookPath("login")
-	if err != nil {
-		return "", nil, fmt.Errorf("login command not available: %w", err)
-	}
-
 	addrPort, err := netip.ParseAddrPort(remoteAddr.String())
 	if err != nil {
 		return "", nil, fmt.Errorf("parse remote address: %w", err)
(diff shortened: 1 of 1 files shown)


afonsofrancof:
Will fix these implementation details

afonsofrancof:
Changed it :)

alexmoras:
This PR appears to have fixed the issue.

I started my testing on an Ubuntu 24.04 LTS VM which was using `shadow-login`. SSH with the current production build worked fine. I upgraded the VM to 25.10 and confirmed that it was using `util-linux`. Production client no longer worked.

I have built the client from this PR and the issues appear to have been resolved, it now SSH's straight away with zero problems.

I have also tested this PR on a 24.04 server with `shadow-login` and there do not appear to be any regressions caused.

afonsofrancof:
@alexmoras I changed the PR to always use runuser if available on linux systems.
`runuser` is the recommended utility for this use case, not `login`.
If `runuser` isn't available, `login` is used.

I know this works on recent versions, but could you try it on your Ubuntu 24.04 again, just to make sure there are no regressions? Sorry for all the trouble. (I squashed some commits, so you might have to reclone or delete the commits and then pull from my branch)

afonsofrancof:
I can confirm that it doesn't have any regression on Debian 12 (which uses shadow-login)

afonsofrancof:
I changed the branch name and it closed the PR... perfect 😅

alexmoras:
Likewise, no regression introduced on 24.04 LTS or Debian 11 (using `shadow-login`) for me.

afonsofrancof:
Nice, it should be ready to merge then

lixmal:
Hi @afonsofrancof, thanks for the PR!

Unfortunately, it has one flaw: `runuser` as the default loses some audit trail and bypasses `pam`. E.g., `loginuid` is not set, so all commands will look like they were run by root/netbird (check `w`/`who` when logged in through netbird w/ and w/o the PR).

The proper way should be to try `login` first and fall back if it fails (ideally to `su`, which should be more widespread,`pam`, not exclusive to Linux).
 This is not a straightforward task, as there is ambiguity in exit codes (did `login`'s `execle` fail or did the spawned shell fail?).

So we need an intermediate process that can check righ after a (manual, not through the Go stdlib) `exec` and fall back to something else. Fortunately we already have the executor code in place, so this mostly needs some careful refactoring.

afonsofrancof:
Hey @lixmal !
Thanks for the heads up, I didn't know about the loginuid thing.
From what I researched `su` is indeed a good option

Is there anyone working on this ATM? Could I implement this?

Thanks :)

lixmal:
Nobody works on it right now; that's more like a mid-term fix. As a short-term fix, omitting the `-h` flag should work (basically remove the arch check)

afonsofrancof:
That won't work for distros that use util-linux's login binary, like debian 13 and Ubuntu 25.10.

Login does not allow that behavior, so it can't be used at all in those cases

lixmal:
I have tested it on Debian 13, and it works for me. Although it seems there is a race condition that fails something like 4 of 5 attempts, at least for me.

afonsofrancof:
@lixmal It never works for me, and another user has also reported it  #4869 
On my debian 12 machine it works every time ( it uses shadow-login ) but they changed it to util-linux on 13, so it doesn't work there at all for me.

alexmoras:
As above, exact same behaviour for me. Any system running `shadow-login` works fine but I can never get a system with `util-linux` to work.

alexmoras:
Testing with `su` doesn't seem to amend the `loginuid` either. From some cursory research, the only way to change this is by utilising the full PAM stack or by having the `CAP_AUDIT_CONTROL` or `CAP_AUDIT_WRITE` capabilities to write to `/process/self/loginuid`. The latter option is a significant security consideration and the prior will take considerable work.

Given that this is affecting so many users, would it be a beneficial temporary measure to place the `runuser` variant of the login implementation behind a feature flag? That way, users have to intentionally enable it and would only be aware of the feature flag after reading a disclaimer regarding security. It could be something like `--enable-ssh-modern-compatibility`. I don't understand the authentication system well enough to know whether this would expose a system to unacceptable levels of risk, or whether it is something that could be tolerated in the interim - advice from someone more knowledgeable would be brilliant.

afonsofrancof:
@alexmoras I tried it with `su` as well, and it was better, because it actually showed a session in the `w` or `who` command, while `runuser` didn't.

I am trying to understand and work on the solution that lixmal suggested, but it will take a bit of time.
From what I understood he wants to first try `login`, and if the command itself fails, fallback to `su`, and if that fails, fallback to manual privilege dropping (setting uid, gid, etc) 

I agree that this should probably be shipped, as it fixes a lot of people's workflows, and we could keep working on a more permanent fix meanwhile.

EDIT: It should probably be shipped with `su` and not with `runuser`

afonsofrancof:
I think the way to do this is by using `unix.Exec`, which replaces the current process with the new command (be it `login`, `su`, etc)
That way, if the command returns, we know it failed (because the only way to continue executing the code after the `unix.Exec` is if it fails).

I think this is how tailscale does it, and if it works for them, I think the same approach should work here.

afonsofrancof:
A better fix was implemented in #4900.
Closing this as there is no need for it anymore.

PR #4900, opened 2025-12-03
[client] Use setsid to avoid the parent process from being killed via HUP by login

lixmal:
https://github.com/netbirdio/netbird/issues/4869
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Detects util-linux login on Linux and adapts how login commands are constructed and wrapped for sessions.
  * Platform-specific no-op detection added for non-Linux environments.

* **Bug Fixes**
  * Validates and handles empty remote commands to prevent erroneous executions.
  * Logs full executed command lines and ensures session exit status is recorded for clearer troubleshooting.

### DIFF
--- a/client/ssh/server/userswitching_unix.go
+++ b/client/ssh/server/userswitching_unix.go
@@ -87,7 +87,8 @@ func (s *Server) getLoginCmd(username string, remoteAddr net.Addr) (string, []st
 
 	switch runtime.GOOS {
 	case "linux":
-		return s.getLinuxLoginCmd(loginPath, username, addrPort.Addr().String())
+		p, a := s.getLinuxLoginCmd(loginPath, username, addrPort.Addr().String())
+		return p, a, nil
 	case "darwin", "freebsd", "openbsd", "netbsd", "dragonfly":
 		return loginPath, []string{"-fp", "-h", addrPort.Addr().String(), username}, nil
 	default:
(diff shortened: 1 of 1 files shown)


client/ssh/server/userswitching_unix.go
```diff
@@ -87,19 +87,45 @@ func (s *Server) getLoginCmd(username string, remoteAddr net.Addr) (string, []st
...
 		return loginPath, []string{"-fp", "-h", addrPort.Addr().String(), username}, nil
 	default:
 		return "", nil, fmt.Errorf("unsupported Unix platform for login command: %s", runtime.GOOS)
 	}
 }
 
-// fileExists checks if a file exists (helper for login command logic)
+// getLinuxLoginCmd returns the login command for Linux systems.
+// Handles differences between util-linux and shadow-utils login implementations.
+func (s *Server) getLinuxLoginCmd(loginPath, username, remoteIP string) (string, []string, error) {
```
pappz: err is meaningless, it is always nil

lixmal: fixed

alexmoras:
Working on Ubuntu 25.10 on multiple devices. `/proc/self/loginuid` seems to be showing a random ID rather than `0` I get when logging in as `root` normally, but it doesn't seem to be affecting anything. Its resolved the underlying issue for me.

afonsofrancof:
This is working perfectly for me.
`/proc/self/loginuid` is set to 1000 when I login with my user, which is the correct one.

alexmoras:
Interesting! For me, I get:
```shell
root@vps:~# cat /proc/self/loginuid
4294967295
```
`4294967295` is the decimal version of `0xFFFFFFFF` which is the "unset" value.

It doesn't matter whether I log in as root or my normal user, I still get the unset value. This is running on Ubuntu 25.10 with the binaries from this PR.

I'm not sure if this expected behaviour?

Issue #4755, opened 2025-11-06
Clients fail to connect on self hosted server - possibly signal certificate issue following #4644?


fishboy25uk:
Attempted to deploy a new server on a VPS this morning (version 0.59.12, Azure as IdP and Let's Encrypt certs which I've obtained on host using certbot myself). Although I can authenticate ok and clients get registered, clients then fail to connect to the network:

Client log:
`2025-11-06T12:26:50Z INFO ./caller_not_available:0: 2025/11/06 12:26:50 WARNING: [core] [Channel #116 SubChannel #117]grpc: addrConn.createTransport failed to connect to {Addr: "<SERVER IP>:33073", ServerName: "vpn.<SERVER DOMAIN>:33073", BalancerAttributes: {"<%!p(pickfirstleaf.managedByPickfirstKeyType={})>": "<%!p(bool=true)>" }}. Err: connection error: desc = "transport: Error while dialing: nbnet.NewDialer().DialContext: dial tcp <SERVER IP>:33073: connectex: No connection could be made because the target machine actively refused it."`

I've modified the docker-compose.yml to change volumes from netbird-letsencrypt to /etc/netbird for dashboard, signal, and management containers. Otherwise it's the standard docker-compose as generated, following the Netbird self hosted guide for Azure IdP

Certificates appear to work correctly for dashboard at least, as I can successfully log in and I can see the "offline" client, but certs are not working with signal container. Initially signal logs showing it couldn't find the certs due to certs on docker host being symbolically linked, but even if I address this, client connection still fails. This is the output from the signal container:

```
2025-11-06T09:42:39Z INFO signal/cmd/run.go:236: setting up TLS with custom certificates.
2025-11-06T09:42:39Z INFO signal/cmd/run.go:150: running HTTPS server with WebSocket proxy: [::]:443
2025-11-06T09:42:39Z INFO signal/cmd/run.go:161: running gRPC backward compatibility server: [::]:10000
2025-11-06T09:42:39Z INFO signal/cmd/run.go:164: signal server version 0.59.12
2025-11-06T09:42:39Z INFO signal/cmd/run.go:165: started Signal Service
2025-11-06T09:42:39Z INFO signal/cmd/run.go:111: running metrics server: :9090/metrics
```

[...]

PR #4763, opened 2025-11-07
fix: correct Signal service port mapping for custom TLS certificates

krth1k:
## Describe your changes

This PR fixes a critical port mapping issue in the Signal service configuration that prevents clients from connecting when using custom TLS certificates.

The Signal service with custom TLS certificates runs HTTPS/WebSocket on **internal port 443**, not port 80. The current docker-compose template incorrectly maps `$NETBIRD_SIGNAL_PORT:80`, causing connection failures.

**Updated port mapping:**
```yaml
signal:
  ports:
    - $NETBIRD_SIGNAL_PORT:443  # Changed from :80 to :443
```

**Impact:**
- ✅ Signal service connects successfully  
- ✅ Clients can establish full NetBird connectivity
- ✅ Uses modern HTTPS/WebSocket approach (port 443)
- ℹ️ Port 10000 is kept for backward compatibility only

## Issue ticket number and link

Fixes #4762

### Checklist
- [x] Is it a bug fix
- [x] Is a typo/documentation fix
- [ ] Is a feature enhancement
- [ ] It is a refactor
- [ ] Created tests that fail without the change (if possible)

> By submitting this pull request, you confirm that you have read and agree to the terms of the [Contributor License Agreement](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT.md).

## Documentation
Select exactly one:

- [x] I added/updated documentation for this change [ https://github.com/netbirdio/netbird/pull/4763 ]
- [ ] Documentation is **not needed** for this change 
---

### Problem Analysis

When the Signal service runs with custom TLS certificates (via `--cert-file` and `--cert-key` flags), it starts the HTTPS/WebSocket server on **internal port 443**, not port 80.

**Evidence from Signal service logs:**
```
signal-1  | INFO signal/cmd/run.go:180: running HTTPS server with WebSocket proxy: [::]:443
signal-1  | INFO signal/cmd/run.go:161: running gRPC backward compatibility server: [::]:10000
```

This mismatch causes:
- ❌ Signal service shows as "Disconnected" in `netbird status`
- ❌ Connection refused errors when clients attempt to connect
- ❌ Client timeout errors: "daemon up failed: rpc error: code = DeadlineExceeded"
- ✅ Management service connects successfully (only Signal fails)

### Testing Results

**Before the fix:**
```bash
$ nc -zv netbird.example.com 10000
nc: connect to netbird.example.com port 10000 (tcp) failed: Connection refused

$ netbird status
[...]

PR #4941, opened 2025-12-11
[infrastructure] Change NETBIRD_SIGNAL_PORT mapping from 80 to 10000

alfkonee:
I found that the signal port is 10000 in the container logs. That's why this change is being made.

## Describe your changes
During my deployment I found that peers were not able to connect upon checking the logs, I saw that the service was running on the same port as the default signal port. As such, I made this change to fix the port 80 issue. 
#4755 and probably a few others
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

the default port specified is the same as this fix

## Summary by CodeRabbit

* **Chores**
  * Updated Signal service port mapping from 80 to 10000 for improved network port management in deployment environments.

### DIFF
--- a/infrastructure_files/docker-compose.yml.tmpl
+++ b/infrastructure_files/docker-compose.yml.tmpl
@@ -46,7 +46,7 @@ services:
       - $SIGNAL_VOLUMENAME:/var/lib/netbird
       - $LETSENCRYPT_VOLUMENAME:/etc/letsencrypt:ro
     ports:
-      - $NETBIRD_SIGNAL_PORT:80
+      - $NETBIRD_SIGNAL_PORT:10000
   #      # port and command for Let's Encrypt validation
   #      - 443:443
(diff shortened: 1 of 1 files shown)


diegocn:
Hi @alfkonee 

Are you running with lets encrypt (https) enabled?
If so, that might be the issue with port 80 hardcoded. 
If the certs are set then the signal goes for https/443 instead of 80, causing this mismatch of ports.

But port 10000 will always work due to a fallback for old client versions that rely on it for grpc.

I wonder if you can test the same scenario you are trying to achieve, but with port 443 hardcoded in the place of 10000.
If that works it will confirm this point of havint the setup based on TLS/HTTPS that causes signal to listen on port 443.

Then we can figure out a way to handle the internal container port dinamically.

diegocn:
The port 443 hardcoded is suggested on this other PR -> https://github.com/netbirdio/netbird/pull/4763

PR #4947, opened 2025-12-13
Add automatic WebSocket fallback for HTTP/1.1-only environments for Management and Signal Server Connection

shyam0904a:
## Describe your changes

This PR adds automatic WebSocket fallback support to the NetBird client for environments where native gRPC (HTTP/2) connections fail due to ALPN stripping or HTTP/1.1-only proxies (e.g., SASE solutions, corporate proxies).

**Changes:**

- client/grpc/dialer.go: Modified CreateConnection to first attempt native gRPC, then automatically fall back to WebSocket if specific errors are detected (ALPN issues, DeadlineExceeded, Unavailable, Internal)
- client/grpc/dialer_generic.go: Added dialWebSocket and dialNative helper functions; WebSocket connections now use context.Background() to prevent premature connection closure
- client/grpc/dialer_js.go: Updated to match new WithCustomDialer signature (JS/WASM always uses WebSocket)
- client/grpc/fallback.go: New file managing fallback state with IsWebSocketFallbackEnabled(), EnableWebSocketFallback(), and ShouldFallbackToWebSocket(err) functions
- flow/client/client.go: Updated to use the new dialer signature

**Behavior:**

1. Client attempts native gRPC connection (5-10s timeout)
2. If it fails with ALPN/HTTP2 errors, WebSocket fallback is enabled globally
3. Subsequent connections automatically use WebSocket proxy (/ws-proxy/{management,signal})

N/A - Addresses connectivity issues for users behind restrictive proxies that strip ALPN or enforce HTTP/1.1

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

The feature is automatic and transparent to users - no configuration flags or user action required. The client automatically detects HTTP/2 issues and falls back to WebSocket without user intervention.

## Summary by CodeRabbit

* **New Features**
  * Automatic WebSocket fallback for gRPC connections when native dialing fails.
  * Runtime controls to enable/query the fallback state; client creation respects the fallback flag.
  * Platform-aware transport selection with distinct native and WebSocket timeouts.

* **Bug Fixes / Reliability**
  * TLS handling adjusted so native and WebSocket paths use appropriate TLS behavior.
  * WebSocket fallback disabled for browser builds to avoid unsupported behavior.


mlsmaycon:
Hello @shyam0904a, thanks for the contribution. To document this before adding the support, can you share more about the motivation behind this change? Have you encountered such issues? Can you detail more, and if possible, in a GitHub issue?

shyam0904a:
> Hello @shyam0904a, thanks for the contribution. To document this before adding the support, can you share more about the motivation behind this change? Have you encountered such issues? Can you detail more, and if possible, in a GitHub issue?

Hi @mlsmaycon,
Yep — I hit this in a real environment.
I’m self-hosting NetBird and have multiple peers joined. When I tried to add my work laptop as a peer, it kept failing during the gRPC TLS handshake to the Management API (and later Signal). After digging through client logs and packet behavior, the root cause was our corporate SASE/secure web gateway: it forces HTTP/1.1 and strips ALPN, so gRPC (HTTP/2) can’t be negotiated. 

Because the environment effectively cannot do HTTP/2 end-to-end, the only reliable option is to tunnel gRPC over WebSocket (HTTP/1.1). Netbird backend already supports Websocket but cli and desktop clients didn't have the options; this PR makes it automatic: attempt native gRPC first, and if the error indicates HTTP/2 negotiation/proxy restrictions, retry using WebSocket.
This keeps the default behavior unchanged for normal networks, but unblocks users behind SASE / restrictive proxies without requiring extra configuration (not having to ask IT admins to whitelist :) ).

shyam0904a:
Any update on this :)

MrCzz:
Another situation where this would help with a failure mode that differs slightly from "HTTP/1.1-only", so it may be worth checking whether the current trigger set catches it.

**Setup:** self-hosted NetBird 0.77.1 (combined image) behind nginx, behind a Cloudflare Tunnel public hostname.
**Symptom:** no peer ever comes up. Agent logs:
```
    WARN shared/signal/client/grpc.go:155: disconnected from the Signal Exchange due to an
         error: didn't receive a registration header from the Signal server whille
         connecting to the streams
    Error: daemon up failed: call service up method: rpc error: code = DeadlineExceeded
```

**The nuance:** HTTP/2 is available end to end, and *unary* gRPC works fine `management.ManagementService/GetServerKey` returns 200 over the public hostname. Only server-streaming breaks: the proxy withholds the response headers until the request body completes, and `ConnectStream` never completes its body.

Measured with a request body deliberately held open for 75 s:

    # through the Cloudflare Tunnel
    starttransfer=74.98s  http=520     <- first byte only once the body is closed

    # identical request straight to the origin
    HTTP/2 200
    x-wiretrustee-peer-registered: 1   <- immediately, stream stays open

Cloudflare documents the limitation: "Cloudflare Tunnel supports gRPC traffic via private
subnet routing. Public hostname deployments are not currently supported."
https://developers.cloudflare.com/network/grpc-connections/

**Why the fallback might not fire here:** it triggers on ALPN failures,
`context.DeadlineExceeded` and gRPC `Unavailable`/`DeadlineExceeded`. In this case ALPN
succeeds and the RPC never returns a status — the client just blocks in `stream.Header()`
until the proxy tears the stream down. Would it make sense to also trigger on "no initial
metadata received within N seconds on ConnectStream / the management Sync stream"? That is
the observable signal in any buffering-proxy setup, and it covers the HTTP/1.1-only case as
a side effect.

(Note: Unlike gRPC, WebSockets do work through Cloudflare Tunnels. The 0.77.1 server already supports WebSocket upgrades on /ws-proxy/*, but the Go agent does not seem to utilize them yet).

PR #5038, opened 2026-01-06
[client] add option to prevent remove peers and routes when management services are not available

gamerslouis:
## Describe your changes

NetBird clients rely on the **management service** and **signal service** to establish and maintain connections.
When the management service is temporary unavailable, a NetBird client is unable to re-establish existing connections after an ICE disconnect.

Introduce a new environment variable option: `NB_KEEP_CONNECTION_ON_MANAGEMENT_DOWN`.

When this flag is enabled and the management service is unavailable:

* The client will **not remove peers** after an ICE disconnect.
* If a network route cannot select a new peer, the client will **retain the existing peer** instead of removing it.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection stability during management server maintenance by preserving active endpoints and routes under specific conditions, preventing unnecessary disconnections when the management service is temporarily unavailable.

* **New Features**
  * Added configuration option to keep connections alive during management downtime scenarios.

### DIFF
--- a/client/internal/peer/conn.go
+++ b/client/internal/peer/conn.go
@@ -433,8 +433,12 @@ func (conn *Conn) onICEStateDisconnected() {
 	} else {
 		conn.Log.Infof("ICE disconnected, do not switch to Relay. Reset priority to: %s", conntype.None.String())
 		conn.currentConnPriority = conntype.None
-		if err := conn.config.WgConfig.WgInterface.RemoveEndpointAddress(conn.config.WgConfig.RemoteKey); err != nil {
-			conn.Log.Errorf("failed to remove wg endpoint: %v", err)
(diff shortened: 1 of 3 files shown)


gamerslouis:
The issue has now been resolved, and the original approach has some inherent problems. Ignoring ICE disconnection events and forcibly keeping the existing WireGuard configuration may lead to issues during reconnection.
Adjusting NB_ICE_FAILED_TIMEOUT_SEC and NB_ICE_DISCONNECTED_TIMEOUT_SEC proves to be a more appropriate solution for our use case.

PR #5044, opened 2026-01-06
[client/android] - Fix Rosenpass connectivity for Android peers

doromaraujo:
## Describe your changes

Rosenpass implementation for Android's sandboxed environment was failing to acquire a new WireGuard client due to a lack of permission using the kernel configurer (socket: permission denied). There already exists a WireGuard userspace configurer implementation, though.

This PR makes use of the existing WGUSConfigurer for android/ios, while keeping the kernel configurer for all other platforms.

### Checklist
- [X] Is it a bug fix
- [X] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Set or update a peer preshared key with an update-only option.
  * Query if a peer's preshared key has been initialized; manager now receives interface once active.

* **Bug Fixes**
  * Startup propagates the active WireGuard interface to the remote manager, improving integration reliability.

* **Refactor**
  * WireGuard access abstracted behind a preshared-key setter; handler wiring simplified.

* **Tests**
  * Added coverage for Rosenpass-managed preshared-key behavior.

### DIFF
--- a/shared/management/client/grpc.go
+++ b/shared/management/client/grpc.go
@@ -185,10 +185,10 @@ func (c *GrpcClient) handleJobStream(
 	for {
 		jobReq, err := c.receiveJobRequest(ctx, stream, serverPubKey)
 		if err != nil {
-			c.notifyDisconnected(err)
 			if s, ok := gstatus.FromError(err); ok {
 				switch s.Code() {
 				case codes.PermissionDenied:
+					c.notifyDisconnected(err)
 					return backoff.Permanent(err) // unrecoverable error, propagate to the upper layer
 				case codes.Canceled:
 					log.Debugf("management connection context has been canceled, this usually indicates shutdown")
@@ -198,11 +198,13 @@ func (c *GrpcClient) handleJobStream(
 						"Please update the management service to use this feature.")
 					return nil
 				default:
+					c.notifyDisconnected(err)
 					log.Warnf("disconnected from the Management service but will retry silently. Reason: %v", err)
 					return err
 				}
 			} else {
 				// non-gRPC error
+				c.notifyDisconnected(err)
 				log.Warnf("disconnected from the Management service but will retry silently. Reason: %v", err)
 				return err
 			}


client/iface/iface.go
```diff
@@ -297,6 +297,13 @@ func (w *WGIface) FullStats() (*configurer.Stats, error) {
 	return w.configurer.FullStats()
 }
 
+// GetConfigurer returns the WireGuard configurer for direct peer configuration.
+func (w *WGIface) GetConfigurer() device.WGConfigurer {
```
pappz: Do not expose the WGConfigurer. This variable should keep internally. Create a new function on the WGIface itself, like:
`func (w *WGIface) SetPresharedKey(peerid, presharedkey) error 
`
In the rosenpass pkg do not need to import the full wgIface if you do not want, enough if you create a new interface just for this function.

pappz (changes requested):

client/iface/configurer/usp.go
```diff
@@ -33,6 +33,7 @@ const (
 	listenPort                  = "listen_port"
 	publicKey                   = "public_key"
 	presharedKey                = "preshared_key"
+	persistentKeepaliveInterval = "persistent_keepalive_interval"
```
pappz: I think this interval is hardcoded. Does it make sense to keep this value in memory?

doromaraujo: Currently it's set to defaultWgKeepAlive = 25 * time.Second in peer/endpoint.go. If this isn't subjected to change, I can use the same value as well when updating the preshared key in iface (I'm migrating the logic you suggested there, too).

doromaraujo: Changed it to use the same value in the set preshared key method.

client/iface/iface.go
doromaraujo: Done.

client/iface/iface.go
```diff
@@ -297,6 +297,91 @@ func (w *WGIface) FullStats() (*configurer.Stats, error) {
 	return w.configurer.FullStats()
 }
 
+// SetPresharedKey sets or updates the preshared key for a peer.
+// If the peer has no PSK or has the original NetBird PSK, it restarts the connection
+// to ensure the new PSK takes effect. Otherwise, it updates the PSK in place.
+func (w *WGIface) SetPresharedKey(peerKey string, psk wgtypes.Key, originalPSK [32]byte) error {
```
pappz: Reading the full peer list is extremely heavy for the WG. It can take several seconds when there are many peers in the list. If every peer performs key rotation, this operation will block the flow.

Can we use the UpdateOnly flag to speed up this operation?
```
	// Simple UpdateOnly path - no need to read peer config, just update PSK
	config := wgtypes.Config{
		Peers: []wgtypes.PeerConfig{
			{
				UpdateOnly:   true,
				PublicKey:    peerKeyParsed,
				PresharedKey: psk,
			},
		},
	}
```
Maybe a signature like this in worst case:
```
func (c *WGUSPConfigurer) UpdatePresharedKey(peerKey string, psk *wgtypes.Key, forceRestart bool) 
```

client/iface/device/interface.go
```diff
@@ -13,6 +13,7 @@ import (
 
 type WGConfigurer interface {
 	ConfigureInterface(privateKey string, port int) error
+	ConfigureDevice(config wgtypes.Config) error
```
pappz: I do not favor these too general functions. They could give too much power to make mistakes. If you see an option to rename it to SetPresharedKey(...) without duplicating code between the usp and kernel_unix implementations, please refactor it. Otherwise, it’s fine.

client/iface/iface.go
pappz, line 151: When we call this function, don't we overwrite the preSharedKey from Rosenpass?

pappz (changes requested):

client/iface/iface.go
doromaraujo: This actually happens, it would overwrite the preshared key set by Rosenpass via the handler when called again.
It is called every time ConfigureWGEndpoint is called.

Which would be better, have a specific method to set the initial presharedKey (and make UpdatePeer not update it afterwards), or remove the presharedKey parameter from ConfigureWGEndpoint (and set the initial one during peer creation alone, instead of in endpoint configuration)?

client/internal/peer/conn.go
```diff
@@ -759,10 +767,24 @@ func (conn *Conn) presharedKey(remoteRosenpassKey []byte) *wgtypes.Key {
...
+	// If Rosenpass has already set a PSK for this peer, return nil to prevent
+	// UpdatePeer from overwriting the Rosenpass-managed key.
+	if conn.rosenpassInitializedPresharedKeyValidator != nil && conn.rosenpassInitializedPresharedKeyValidator(conn.config.Key) {
+		return nil
+	}
+
+	// Use NetBird PSK as the seed for Rosenpass. This same PSK is passed to
+	// Rosenpass as PeerConfig.PresharedKey, ensuring the derived post-quantum
+	// key is cryptographically bound to the original secret.
+	if conn.config.WgConfig.PreSharedKey != nil {
```
mlsmaycon, line 779: is possible to use preshared key with rosenpass. So we need to check for:
```
conn.rosenpassInitializedPresharedKeyValidator == nil
```

mlsmaycon: the previous if would return already

client/iface/configurer/common.go
```diff
@@ -3,8 +3,26 @@ package configurer
 import (
 	"net"
 	"net/netip"
+
+	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
 )
 
+// buildPresharedKeyConfig creates a wgtypes.Config for setting a preshared key on a peer.
+// This is a shared helper used by both kernel and userspace configurers.
+func buildPresharedKeyConfig(peerKey string, psk wgtypes.Key, updateOnly bool) (wgtypes.Config, error) {
```
pappz: I think in the conn.go we already manage the peerKey in key format. Use that type instead of string.

doromaraujo: Changed.

client/iface/iface.go
doromaraujo: Update: a check is now being done in conn.go's presharedKey method. If the instance's rosenpassInitializedPresharedKeyValidator is set and returns true to currently configured key it means it's managed by Rosenpass, so it returns nil (as to not overwrite psk by the time UpdatePeer is called).

pappz (approved):

PR #5098, opened 2026-01-13
Detect stale ICE handshakes

### LINKED ISSUE #4769: Intermittent connectivity loss to routing peers requiring manual restart
**Describe the problem**

Users intermittently lose connectivity to NetBird routing peers (VPC gateway instances), making infrastructure resources unreachable. The issue affects both redundant routing peers simultaneously, despite
each being on separate EC2 instances with independent network paths.

Currently, the only reliable recovery method is manually executing `netbird down && netbird up` on the affected routing peer(s). We have implemented automated daily restarts via cron (with retry logic and
randomized delays), but connectivity issues still occur randomly between scheduled restarts, requiring manual intervention from operations team.

**To Reproduce**

1. Set up self-hosted NetBird with routing peers configured as VPC gateways
2. Configure redundant routing peers (2 per VPC) for high availability
3. Connect client machines to access resources through routing peers
4. Normal operation works initially
5. After random intervals (hours to days), clients lose connectivity to resources behind routing peers
6. Routing peer shows Status: Idle or Status: Connecting for affected client peers with "Last WireGuard handshake: -"
7. Client-side netbird status shows routing peer in "Connecting" state
8. Only resolution: SSH to routing peer and run `netbird down && netbird up`

**Expected behavior**

- Routing peers should maintain stable connectivity to client peers

Self-hosted NetBird control plane (on-premise deployment)

**NetBird version**

Routing peers: 0.59.12 (Amazon Linux 2023)
Clients: Mixed versions 0.59.10 - 0.59.12 (all affected randomly regardless of version)

No other VPN software installed on routing peers or affected clients.

**Debug output**

**Additional context**

Environment:
- Deployment: AWS EC2 routing peers (Amazon Linux 2023.9.20251014, kernel 6.12.46)
- Architecture: 2 routing peers per VPC (dev + production VPCs)
- Use case: VPC gateway routing to access RDS, EC2, ...
- Network: Behind AWS NLB with HTTP2Optional ALPN policy for gRPC support

Mitigation attempts:
- Implemented automated daily reconnection via cron with 10 retry attempts
- Deployed redundant routing peers for failover
- Neither prevents the issue - both peers can become unreachable simultaneously for some users.

Pattern observations:
- Issue affects users randomly regardless of client version (0.59.10-0.59.12)
- Both routing peers can fail simultaneously despite being on separate instances
- Manual netbird down && netbird up on routing peer immediately resolves connectivity

Question: Have you seen this pattern before in other deployments? It seems like a connection state management issue where NetBird doesn't detect or recover from stale peer connections automatically.

Have you tried these troubleshooting steps?
- Reviewed https://docs.netbird.io/how-to/troubleshooting-client
- Checked for newer NetBird versions (running latest 0.59.x)
- Searched for similar issues on GitHub
- Restarted the NetBird client (this is the only working fix)
- Disabled other VPN software (none installed)
- Checked firewall settings (NLB ALPN policy configured correctly)

Silex:
## Describe your changes

Add a WireGuard handshake watchdog for ICE connections so stale handshakes trigger a disconnect/reconnect, matching the relay path behavior.

**WARNING**: this code is untested

This would close #4769 (and maybe others)

### Checklist

- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation

Select exactly one:

Reason: internal connectivity watchdog behavior only; no user-facing change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection state monitoring and disconnection handling during ICE (Interactive Connectivity Establishment) transitions to enhance connection reliability.

### DIFF
--- a/client/internal/peer/conn.go
+++ b/client/internal/peer/conn.go
@@ -98,6 +98,7 @@ type Conn struct {
 
 	workerICE   *WorkerICE
 	workerRelay *WorkerRelay
+	wgWatcherICE *WGWatcher
 	wgWatcherWg sync.WaitGroup
 
 	// used to store the remote Rosenpass key for Relayed connection in case of connection update from ice
@@ -125,6 +126,7 @@ func NewConn(config ConnConfig, services ServiceDependencies) (*Conn, error) {
 	}
 
 	connLog := log.WithField("peer", config.Key)
+	dumpState := newStateDump(config.Key, connLog, services.StatusRecorder)
 
 	var conn = &Conn{
 		Log:             connLog,
@@ -137,9 +139,10 @@ func NewConn(config ConnConfig, services ServiceDependencies) (*Conn, error) {
 		semaphore:       services.Semaphore,
 		statusRelay:     worker.NewAtomicStatus(),
 		statusICE:       worker.NewAtomicStatus(),
-		dumpState:       newStateDump(config.Key, connLog, services.StatusRecorder),
+		dumpState:       dumpState,
 		endpointUpdater: NewEndpointUpdater(connLog, config.WgConfig, isController(config)),
 	}
+	conn.wgWatcherICE = NewWGWatcher(connLog, config.WgConfig.WgInterface, config.Key, dumpState)
 
 	return conn, nil
 }
@@ -232,6 +235,7 @@ func (conn *Conn) Close(signalToRemote bool) {
 	conn.ctxCancel()
 
 	conn.workerRelay.DisableWgWatcher()
+	conn.wgWatcherICE.DisableWgWatcher()
 	conn.workerRelay.CloseConn()
 	conn.workerICE.Close()
 
@@ -368,6 +372,7 @@ func (conn *Conn) onICEConnectionIsReady(priority conntype.ConnPriority, iceConn
 
 	conn.workerRelay.DisableWgWatcher()
 	// todo consider to run conn.wgWatcherWg.Wait() here
+	conn.wgWatcherICE.DisableWgWatcher()
 
 	if conn.wgProxyRelay != nil {
 		conn.wgProxyRelay.Pause()
@@ -384,6 +389,12 @@ func (conn *Conn) onICEConnectionIsReady(priority conntype.ConnPriority, iceConn
 		return
 	}
 	wgConfigWorkaround()
+
+	conn.wgWatcherWg.Add(1)
+	go func() {
+		defer conn.wgWatcherWg.Done()
+		conn.wgWatcherICE.EnableWgWatcher(conn.ctx, conn.onICEStateDisconnected)
+	}()
 
 	if conn.wgProxyRelay != nil {
 		conn.Log.Debugf("redirect packets from relayed conn to WireGuard")
@@ -405,6 +416,7 @@ func (conn *Conn) onICEStateDisconnected() {
 	}
 
 	conn.Log.Tracef("ICE connection state changed to disconnected")
+	conn.wgWatcherICE.DisableWgWatcher()
 
 	if conn.wgProxyICE != nil {
 		if err := conn.wgProxyICE.CloseConn(); err != nil {


pappz:
@Silex Why do you think this is necessary? ICE has keep-alives, so it can detect connectivity issues faster than the WireGuard key exchange.

Silex:
In my experience if ICE has keep-alives then it does not work:

```
foo.bar.baz:
  NetBird IP: 100.70.205.198
  Public key: oWxONSNjVcitGblH6DA5OGDTh6PEIM6m4UwU37CODUE=
  Status: Connected
  -- detail --
  Connection type: P2P
  ICE candidate (Local/Remote): host/prflx
  ICE candidate endpoints (Local/Remote): 192.168.168.106:51820/116.202.18.146:58980
  Relay server address:
  Last connection update: 3 hours, 36 minutes ago
  Last WireGuard handshake: -
  Transfer status (received/sent) 0 B/354.8 KiB
  Quantum resistance: false
  Networks: -
  Latency: 51.9013ms
```
  
See `Last WireGuard handshake: -` that stays like this for days. This is for a peer that is an AXIS camera running the netbird client, but I've seen it happen on teltonika routers as well. It just stops doing handshakes and the only solution is to "netbird down/up" on one of the peers.

And when it happens it's not with all peers, e.g A cannot ping C but B can ping C just fine. Restarting netbird on either A or C fixes the problem of A not being able to ping C.
  
Unfortunately right now I don't have this problem anymore but when it happens again in a few days I'll do `netbird debug bundle -S -U` as requested in the linked issue.
  
Problem is I need to wait for the problem to come back and it usually takes around 24h/48h to trigger.

Silex:
Here's a summary of the issue by ChatGPT, but we have yet to confirm that it correctly assed the problem:

---

## ICE vs WireGuard — Why NetBird Can Lose Handshakes

- **ICE and WireGuard are separate layers**
  - ICE handles NAT traversal and connectivity discovery.
  - WireGuard handles encrypted data transport.

- **ICE keep-alives do not validate WireGuard health**
  - They keep NAT mappings open and confirm STUN reachability.
  - They do *not* confirm that WireGuard packets are flowing or that a handshake is active.

- **WireGuard has no automatic recovery mechanism**
  - No per-peer reset
  - No forced re-handshake
  - No liveness timeout that triggers a reconnect

- **On Windows, UDP sockets can silently break**
  - Triggered by sleep, NIC power saving, or NAT rebinding
  - Packets are sent but replies never arrive
  - WireGuard remains stuck with `Last handshake: -`

- **ICE can report “connected” while WireGuard is effectively dead**
  - ICE sees a valid candidate
  - NetBird considers the peer connected
  - The data plane is black-holed

- **Only a full teardown recovers the connection**
  - `netbird down/up` or restarting the NetBird service
  - This recreates sockets and forces a new WireGuard handshake

**Bottom line:**  
ICE keep-alives do not protect WireGuard from stale or broken UDP paths, and WireGuard cannot recover on its own once this happens.

pappz:
@Silex I just want to figure out the root cause of your issue. I see that a complete reconnection resolves the WireGuard connection, but it only covers up a bug. Basically, we have two conditions: an established ICE connection, and the WireGuard endpoint configured with the correct ICE candidate address. So, as a first step, we need to check the endpoint settings in the debug bundle.

Silex:
Ah the problem is finally back! debug bundle incoming 🥳

Silex:
Upload file key:
c33053488251f90ea4683596101892e50e3e31e8030de3052d05b2b3fe6d2468/232a5f43-a1cf-4e11-a0d6-9b1be0d824b6

Here are more information:

```
camera.peer.debug:
  NetBird IP: 100.70.205.198
  Public key: oWxONSNjVcitGblH6DA5OGDTh6PEIM6m4UwU37CODUE=
  Status: Connected
  -- detail --
  Connection type: P2P
  ICE candidate (Local/Remote): srflx/prflx
  ICE candidate endpoints (Local/Remote): 83.173.249.42:51820/116.202.18.146:45889
  Relay server address:
  Last connection update: 3 hours, 42 minutes ago
  Last WireGuard handshake: 3 hours, 42 minutes ago
  Transfer status (received/sent) 180 B/309.9 KiB
  Quantum resistance: false
  Networks: -
  Latency: 62.3291ms

 router.peer.debug
  NetBird IP: 100.70.221.132
  Public key: +oBb+Eno1lY0y0iXdQruOGQsWNV4HiM02xZIgRpyvnQ=
  Status: Connected
  -- detail --
  Connection type: P2P
  ICE candidate (Local/Remote): srflx/prflx
  ICE candidate endpoints (Local/Remote): 83.173.249.42:51820/116.202.18.146:40372
  Relay server address:
  Last connection update: 3 hours, 42 minutes ago
  Last WireGuard handshake: 16 seconds ago
  Transfer status (received/sent) 8.5 GiB/282.2 MiB
  Quantum resistance: false
  Networks: -
  Latency: 46.5375ms
```

As you see the last wireshark handshake was 3 hours ago. Tomorrow it will probably display `-` and never recover.

The camera has internet through a Teltonika (openWRT) router which also runs netbird and zerotier (we are in the process of switching to netbird once stability is there). Both are accessed from a NVR, and the NVR can ping the teltonika "forever" but the camera is lost after 1-2 days.

Both run fairly old versions of netbird (`0.37.1`), the NVR runs `0.62.2`. The self hosted server runs `0.48.0`. I believe this should not matter much but I can try to upgrade to recent versions if you think something recent explicitely fixes these kind of problems.

The camera runs in userspace mode like this, because it cannot run as root at all:

``` shell
export NB_USE_NETSTACK_MODE=true
export NB_ENABLE_NETSTACK_LOCAL_FORWARDING=true
```

I'm pretty sure I also noticed this behavior with the teltonika routers, but given the routers have ping reboot in place it's harder to notice.

Silex:
@pappz: do you need something else?

I'll soon upgrade everything to latest netbird version if that helps.

pappz:
Thank you for the logs! It’s good to know about the old versions. This could be a hidden backward-compatibility issue, but I can’t find an explanation in the logs. Let’s go deeper.
Could you please retest with these environment variables? This will provide more detailed logs:
```
PIONS_LOG_DEBUG=all
NB_WG_DEBUG=true
```
I also started extending the WireGuard watcher to cover ICE connections for another use case. It will achieve similar results, but through a different approach. Instead of instantiating a new watcher, I’m planning to move the existing one out of the Relay worker. I’m hoping to finish this today or within the next couple of days.

pappz:
@Silex [here](https://github.com/netbirdio/netbird/pull/5133) is the PR. If you’d like, you can take an early look and test this version.

Silex:
@pappz thanks. First waiting for the problem to trigger again with `PIONS_LOG_DEBUG=all` so I can provide another trace...

Silex:
@pappz: ok problem finally happened again:

```
PS C:\Users\stvs> netbird debug bundle -S -U
Local file:
C:\Windows\SystemTemp\netbird.debug.2019101234.zip
Upload file key:
c33053488251f90ea4683596101892e50e3e31e8030de3052d05b2b3fe6d2468/03221c7f-c077-438c-9023-1df49c7ed2f7
```

[netbird.debug.2019101234.zip](https://github.com/user-attachments/files/24791117/netbird.debug.2019101234.zip)

By the way it's not very clear if the "Upload file key" is enough or if I should also upload the zip here like I did.

The camera peer that interests you is 100.70.205.198. The router it's attached to is 100.70.221.132.

pappz:
> @pappz: ok problem finally happened again:
> 
> ```
> PS C:\Users\stvs> netbird debug bundle -S -U
> Local file:
> C:\Windows\SystemTemp\netbird.debug.2019101234.zip
> Upload file key:
> c33053488251f90ea4683596101892e50e3e31e8030de3052d05b2b3fe6d2468/03221c7f-c077-438c-9023-1df49c7ed2f7
> ```
> 
> [netbird.debug.2019101234.zip](https://github.com/user-attachments/files/24791117/netbird.debug.2019101234.zip)
> 
> By the way it's not very clear if the "Upload file key" is enough or if I should also upload the zip here like I did.
> 
> The camera peer that interests you is 100.70.205.198. The router it's attached to is 100.70.221.132.

Unfortunately, the logs did not contain the Pion logs for some reason :/
I will close this PR because the same changes have already been merged elsewhere and achieve similar behavior. Let’s continue the discussion on the issues page.

It’s enough to share the ID of the debug bundle. Based on that, we can download it.

pappz:
Fixed in: https://github.com/netbirdio/netbird/pull/5133

Silex:
Will test 0.64.1 and report

PR #5125, opened 2026-01-16
[client] Add support to wildcard custom records

mlsmaycon:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Wildcard DNS fallback for eligible query types (excluding NS/SOA): attempts wildcard records when no exact match, rewrites wildcard names back to the original query, and rotates responses; preserves CNAME resolution.

* **Tests**
  * Vastly expanded coverage for wildcard behaviors, precedence, multi-record round‑robin, multi-type chains, multi-hop and cross-zone scenarios, and edge cases (NXDOMAIN/NODATA, fallthrough).

* **Chores**
  * CI lint config updated to ignore an additional codespell entry.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -19,7 +19,7 @@ jobs:
       - name: codespell
         uses: codespell-project/actions-codespell@v2
         with:
-          ignore_words_list: erro,clienta,hastable,iif,groupd,testin,groupe,cros
+          ignore_words_list: erro,clienta,hastable,iif,groupd,testin,groupe,cros,ans
           skip: go.mod,go.sum
   golangci:
     strategy:


client/internal/dns/local/local.go
```diff
@@ -216,18 +226,58 @@ func (d *Resolver) lookupRecords(logger *log.Entry, question dns.Question) looku
...
+
+func supportsWildcard(queryType uint16) bool {
+	return queryType != dns.TypeNS && queryType != dns.TypeSOA
+}
+
+func responseFromWildRecords(originalName, wildName string, wildRecords []dns.RR) lookupResult {
+	records := make([]dns.RR, len(wildRecords))
+
+	for i, record := range wildRecords {
+		recordString := strings.Replace(record.String(), wildName, originalName, 1)
```
lixmal: You should modify the field in Header() instead of re-parsing the string

client/internal/dns/local/local.go
```diff
@@ -216,18 +226,58 @@ func (d *Resolver) lookupRecords(logger *log.Entry, question dns.Question) looku
...
+		newRecord, err := dns.NewRR(recordString)
+		if err != nil {
+			log.WithField("recordString", recordString).Warnf("failed to parse wildcard record: %v", err)
+			return lookupResult{rcode: dns.RcodeNameError}
```
lixmal: This should not be NXDOMAIN (will be cached) if we are going to keep this line

client/internal/dns/local/local_test.go
```diff
@@ -1341,6 +1370,1208 @@ func TestLocalResolver_FallthroughCaseInsensitive(t *testing.T) {
...
+		assert.Equal(t, "10.0.0.1", a.A.String())
+
+		// AAAA query for host should return NODATA (specific A exists, no AAAA for host.example.com)
+		msgAAAA := new(dns.Msg).SetQuestion("host.example.com.", dns.TypeAAAA)
+		var respAAAA *dns.Msg
+		resolver.ServeDNS(&test.MockResponseWriter{WriteMsgFunc: func(m *dns.Msg) error { respAAAA = m; return nil }}, msgAAAA)
+
+		require.NotNil(t, respAAAA)
+		// host.example.com exists (has A), so AAAA query returns NODATA, not wildcard
+		assert.Equal(t, dns.RcodeSuccess, respAAAA.Rcode, "Should return NODATA for existing host without AAAA")
```
lixmal, line 2510: The comment is correct, but this is missing the assertion on the answer length (should be 0). 


However, if you add it, it will fail, uncovering a bug in the code:
When host.example.com has a specific `A` record and `*.example.com` has a wildcard `AAAA`, querying `host.example.com AAAA` incorrectly returns the wildcard instead of `NODATA`.                                       

The fix in `lookupRecords()` should check if the name exists before wildcard fallback. 

See [RFC 4592](https://datatracker.ietf.org/doc/html/rfc4592#section-2.2.1)

mlsmaycon: Makes sense. I think we can fix it in a different PR post 0.64.0

PR #5133, opened 2026-01-19
[client] Extend WG watcher for ICE connection too

pappz:
## Describe your changes

Extend the WireGuard watcher to monitor handshakes over ICE and relayed connections as well.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WireGuard timeout handling and graceful failover between connection paths.

* **Refactor**
  * Reworked watchdog lifecycle and state-dump integration to centralize enable/disable control and reduce duplicated goroutine/context management.
  * Relay worker no longer directly manages watcher lifecycle.

* **Tests**
  * Updated watcher tests to use synchronized re-enable flows and context cancellation scenarios.

### DIFF
--- a/client/internal/peer/wg_watcher.go
+++ b/client/internal/peer/wg_watcher.go
@@ -30,9 +30,8 @@ type WGWatcher struct {
 	peerKey       string
 	stateDump     *stateDump
 
-	enabledTime time.Time
-	enabled     bool
-	muEnabled   sync.RWMutex
+	enabled   bool
+	muEnabled sync.RWMutex
 }
 
 func NewWGWatcher(log *log.Entry, wgIfaceStater WGInterfaceStater, peerKey string, stateDump *stateDump) *WGWatcher {
@@ -54,7 +53,7 @@ func (w *WGWatcher) EnableWgWatcher(ctx context.Context, onDisconnectedFn func()
 	}
 
 	w.log.Debugf("enable WireGuard watcher")
-	w.enabledTime = time.Now()
+	enabledTime := time.Now()
 	w.enabled = true
 	w.muEnabled.Unlock()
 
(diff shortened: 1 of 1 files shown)


client/internal/peer/wg_watcher.go
```diff
@@ -164,3 +155,13 @@ func (w *WGWatcher) wgState() (time.Time, error) {
 	}
 	return wgState.LastHandshake, nil
 }
+
+// calcElapsed calculates elapsed time since watcher was enabled.
+// The watcher started after the wg configuration happens, because of this need to normalise the negative value
+func (w *WGWatcher) calcElapsed(handshake *time.Time) float64 {
+	elapsed := handshake.Sub(w.enabledTime).Seconds()
```
lixmal: `w.enabledTime` is accessed outside the lock (but that hasn't changed)

client/internal/peer/wg_watcher.go
lixmal: This log should be modified since it doesn't apply to relay only anymore

lixmal (changes requested):

client/internal/peer/wg_watcher.go
pappz: Because do not need to protect it. We touch it from a single thread.

lixmal: Unless EenableWgWatcher is called more than once. Let's remove the field and pass it to periodicHandshakeCheck to remove that risk

pappz: No, it is still safe. We can call EnableWgWatcher multiple times, and we already do. We also cannot talk about a race condition because this is a single-threaded design.

lixmal: EnableWgWatcher is spawned within go func {} blocks but the check "if already enabled" happens after the write to w.enabledTime and the reads in periodicHandshakeCheck happen without the lock. We can move the enabledTime further down or remove it as a field.

lixmal: Discard, `enabled` comes first. I was looking at old code

PR #5185, opened 2026-01-26
[management] Add heartbeat to Job endpoint to prevent proxy timeouts

### LINKED ISSUE #5184: Management gRPC /management.ManagementService/Job intermittently times out (~every 2 minutes) with 504 Gateway Timeout, causing client disconnects / frequent “Network map updated” events
**Describe the problem**

In a self-hosted NetBird control-plane setup, the client intermittently disconnects from Management roughly every ~2 minutes when calling the gRPC endpoint /management.ManagementService/Job. At the same time the client logs an event “Network map updated”. When relay was unavailable, this behavior caused frequent/continuous loss of connectivity inside the VPN. After fixing relay, the VPN data-plane is stable, but the periodic Management 504 / gRPC transport failure still occurs and looks abnormal.

Client error at the moment of failure:
```
Management: Disconnected, reason: rpc error: code = Unavailable desc = unexpected HTTP status code received from server: 504 (Gateway Timeout); transport: received unexpected content-type "text/plain; charset=UTF-8"
```

On the management side, logs show:
```
2026-01-26T15:54:10Z WARN [context: GRPC, requestID: d5rorphfs40s73bt9h20, accountID: REDACTED] management/internals/shared/grpc/server.go:354: recv job response error: rpc error: code = Canceled desc = context canceled
```

This suggests the gRPC stream/long-poll is being interrupted (possibly by reverse proxy / LB / HTTP2 settings / timeouts), because the client receives an HTTP 504 text/plain instead of a proper gRPC response.

**To Reproduce**
	1.	Use a self-hosted NetBird control plane (Management + Signal) behind HTTPS on :443.
	2.	Connect a macOS client to the network.
	3.	Run watch -n 2 'netbird status --detail' (or keep the client running and observe logs/status for several minutes).
	4.	Observe that approximately every ~2 minutes the Management connection briefly drops with:
	•	unexpected HTTP status code received from server: 504 (Gateway Timeout)
	•	received unexpected content-type "text/plain; charset=UTF-8"
	5.	At the same moment, the client emits “Network map updated”.

**Expected behavior**

Management gRPC calls (including /management.ManagementService/Job) should remain stable without periodic 504 Gateway Timeout responses. The client should not show recurring “Management: Disconnected” events or receive non-gRPC HTTP responses (text/plain) on a gRPC endpoint.

Self-hosted NetBird control plane.

**NetBird version**

Client:
	•	OS: darwin/arm64
	•	Daemon version: 0.64.1
	•	CLI version: 0.64.1

Server:
	•	OS: Ubuntu 24.04
	•	Daemon version: 0.64.1

No

**Additional context**

This is a self-hosted NetBird control plane, but **Cloudflare is placed in between clients and the Management/Signal endpoints** (clients connect to https://vpn.example.com:443 via Cloudflare, not directly to the origin).

Also, traefik is used on between the management and CloudFlare:
```
"POST /management.ManagementService/Job HTTP/2.0" 499 21 "-" "-" 54 "netbird-management@docker" "h2c://172.18.0.6:33073" 125017ms
```

maxpain:
## Summary

- Add 30-second heartbeat to the `/management.ManagementService/Job` streaming endpoint to keep connections alive through reverse proxies

## Problem

The Job streaming endpoint blocks indefinitely when there are no pending jobs, causing reverse proxies (Traefik, Cloudflare, Nginx) to timeout the connection after their idle timeout period (typically 60-120 seconds).

While the gRPC server has keepalive configured (HTTP/2 PING frames), reverse proxies measure idle time based on HTTP/2 DATA frames, not PING frames. This causes **504 Gateway Timeout** errors for self-hosted deployments behind proxies.

## Solution

Add a periodic heartbeat (every 30 seconds) that sends an empty `JobRequest` to keep the stream active. The client already handles empty messages gracefully — it logs "received unknown or empty job request, skipping" and continues.

## Changes

- `management/server/job/channel.go`: Add `EventChan()` method to expose channel for select
- `management/internals/shared/grpc/server.go`: Modify `sendJobsLoop` to use ticker + add `sendHeartbeat` method

## Test plan

- [x] Verified build passes
- [x] Tested with Traefik proxy (timeout issue resolved)
- [ ] Run existing tests

Fixes #5184

## Summary by CodeRabbit

* **New Features**
  * Periodic heartbeat messaging added to keep job streams active and reduce unexpected disconnects.
  * Streaming loop upgraded for more responsive, event-driven job delivery.
  * Exposed a read-only events channel for safer integration with select-based consumers.

* **Bug Fixes**
  * Improved error handling and connection lifecycle control for more reliable job delivery.

### DIFF
--- a/management/internals/shared/grpc/server.go
+++ b/management/internals/shared/grpc/server.go
@@ -52,6 +52,11 @@ const (
 	envConcurrentSyncs = "NB_MAX_CONCURRENT_SYNCS"
 
 	defaultSyncLim = 1000
+
+	// jobStreamHeartbeatInterval defines how often to send heartbeat messages
+	// on the Job gRPC stream to keep connections alive through reverse proxies
+	// (e.g., Traefik, Cloudflare) that may have idle timeouts.
+	jobStreamHeartbeatInterval = 30 * time.Second
 )
 
 // Server an instance of a Management gRPC API server
@@ -372,10 +377,9 @@ func (s *Server) sendJobsLoop(ctx context.Context, accountID string, peerKey wgt
 	// todo figure out better error handling strategy
 	defer s.jobManager.CloseChannel(ctx, accountID, peer.ID)
 
-	// Send heartbeat every 30 seconds to keep stream alive through proxies (e.g., Traefik, Cloudflare)
-	// that may have idle timeouts. The client ignores empty messages with no ID.
-	heartbeatInterval := 30 * time.Second
-	ticker := time.NewTicker(heartbeatInterval)
+	// Send heartbeat to keep stream alive through proxies that have idle timeouts.
+	// The client ignores empty messages with no ID.
+	ticker := time.NewTicker(jobStreamHeartbeatInterval)
 	defer ticker.Stop()
 
 	for {
@@ -396,8 +400,10 @@ func (s *Server) sendJobsLoop(ctx context.Context, accountID string, peerKey wgt
 			}
 			if err := s.sendJob(ctx, peerKey, event, srv); err != nil {
 				log.WithContext(ctx).Warnf("send job failed: %v", err)
-				return nil
+				return err
 			}
+			// Reset ticker after successful job send to avoid unnecessary heartbeats
+			ticker.Reset(jobStreamHeartbeatInterval)
 		}
 	}
 }
@@ -506,7 +512,17 @@ func (s *Server) sendHeartbeat(ctx context.Context, peerKey wgtypes.Key, srv pro
 		Body:     encryptedResp,
 	})
 	if err != nil {
-		return status.Errorf(codes.Internal, "failed sending heartbeat")
+		// Preserve original error for expected shutdown scenarios
+		if errors.Is(err, context.Canceled) || errors.Is(err, io.EOF) {
+			return err
+		}
+		if st, ok := status.FromError(err); ok {
+			switch st.Code() {
+			case codes.Canceled, codes.Unavailable:
+				return err
+			}
+		}
+		return status.Errorf(codes.Internal, "failed sending heartbeat: %v", err)
 	}
 
 	log.WithContext(ctx).Tracef("sent heartbeat to peer: %s", peerKey.String())


golem131:
Confirming this bug in production with HAProxy in front of Management

  Running NetBird 0.68.3 self-hosted (management + signal + relay + router peer all on 0.68.3). HAProxy 3.3 fronts the management backend over HTTP/2 with alpn h2,http/1.1
   and http-reuse always. HAProxy config has timeout server 3600s on the management backend — the commonly recommended value for gRPC.

  Symptom is perfectly deterministic:

  1. Router peer container starts at time T.
  2. A management.ManagementService/Sync long-lived server-streaming RPC is established through HAProxy.
  3. At exactly T + 3600s (within ~100 ms), HAProxy closes the stream with termination state sD-- (server-side inactivity timeout, data phase). HAProxy log excerpt:

  [dd/Mon/yyyy:hh:mm:ss.sss] https~ netbird_mgmt/... 0/0/0/113/3600114 200 5120 - - sD-- ...
    "POST /management.ManagementService/Sync HTTP/2.0"
  4. TCP connection from the router to HAProxy stays ESTABLISHED post-event (verified with ss). HAProxy just closed the HTTP/2 stream, not the TCP conn.
  5. Peer goes "offline N min ago" in dashboard and never recovers — stays offline indefinitely until netbird service restart on the router. No amount of waiting helps.

  The underlying issue appears to be on the client side: the management gRPC Sync client in shared/management/client/grpc.go doesn't recreate the grpc.ClientConn when the
  stream is closed by the proxy — it reuses the (now zombie) conn, and new Sync RPCs opened on it either hang or return immediately in a failure loop, so the outer
  backoff.Retry never actually recovers. This pattern was fixed for the telemetry flow client in #5750 (close+re-dial on codes.Internal/RST_STREAM, drop WaitForReady,
  reset backoff on success) but that fix was not ported to the management Sync client, so any reverse proxy with a finite grpc_read_timeout / timeout server / idle timeout
   triggers this.

pappz:
@golem131 The #5750 was created for a special case. When the gRPC stream (not the connection) was broken (i.e., due to a protocol error), the original code attempted to reuse the same connection. In this patch, any stream-related error forces the connection itself to be recreated.

The long-lived reverse proxy configuration is a different topic. The preferred approach is for the protocol to manage keep-alive signals, not the application layer. Of course, if the reverse proxy is not configured for long-lived, silent connections, then it becomes a problem.

We plan to migrate the stream reconnection logic to the Signal and Management servers in the future.

PR #5202, opened 2026-01-28
[client] Add NAT-PMP/UPnP support

lixmal:
## Describe your changes

This adds a NAT mapper that tries to discover NAT-PMP and UPnP gateways to acquire a mapping for a forwarded port.
This mapping is then signaled to other peers.

Example logs
```
2026-01-28T14:50:22Z INFO client/internal/portforward/manager.go:189: created port mapping: 51820 -> 50497 via NAT-PMP (external IP: 45.45.45.2)

2026-01-28T14:50:38Z DEBG [peer: Fc5Y0orhGvvoTD9BDpwUpi9h8O6icW29EzJumvt7izg=] client/internal/peer/worker_ice.go:407: injecting port-forwarded candidate: udp4 srflx 45.45.45.2:50497 related 45.45.45.2:51820 (resolved: 45.45.45.2:50497) candidate:XnfKu4TIX7PFxUgCB/AWryc5TRVvPosU (mapping: 51820 -> 50497 via NAT-PMP, priority: 1694499815)
```

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Port‑forwarding manager: NAT discovery, UDP mapping creation/renewal, env toggle, crash‑recovery state, and graceful lifecycle tied to engine/peer.
  * ICE now injects port‑forwarded server‑reflexive candidates to improve connectivity.
  * JS/WASM builds include a no‑op port‑forwarding stub for parity.

* **Tests**
  * Unit tests covering manager mapping lifecycle, availability, and state cleanup.

* **Chores**
  * Added NAT/UPnP dependencies and clarified state registration comments.


client/internal/portforward/manager_test.go
```diff
@@ -0,0 +1,156 @@
...
+func TestState_Cleanup(t *testing.T) {
+	state := &State{
+		Protocol:     "udp",
+		InternalPort: 51820,
+	}
+
+	// Cleanup should not error even if NAT discovery fails
+	err := state.Cleanup()
+	assert.NoError(t, err)
+}
```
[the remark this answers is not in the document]
lixmal: Fixed

client/internal/portforward/manager.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	m.wg.Wait()
+
+	m.mu.Lock()
+	if err := m.persistStateLocked(); err != nil {
+		log.Debugf("persist state on stop: %v", err)
+	}
+	m.mu.Unlock()
+}
```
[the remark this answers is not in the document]
lixmal: Fixed

client/internal/portforward/manager.go
```diff
@@ -0,0 +1,289 @@
...
+
+	if uint16(externalPort) != m.mapping.ExternalPort {
+		log.Warnf("external port changed on renewal: %d -> %d (candidate may be stale)",
+			m.mapping.ExternalPort, externalPort)
+		m.mapping.ExternalPort = uint16(externalPort)
+	}
+
+	log.Debugf("renewed port mapping: %d -> %d", m.mapping.InternalPort, m.mapping.ExternalPort)
+	return nil
+}
```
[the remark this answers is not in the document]
lixmal: This is a known issue indeed, but mapped port changes should be rare, and in that case, the peer will disconnect and reconnect with the new port for now.

go.mod
```diff
@@ -63,6 +63,7 @@ require (
 	github.com/hashicorp/go-version v1.6.0
 	github.com/jackc/pgx/v5 v5.5.5
 	github.com/libdns/route53 v1.5.0
+	github.com/libp2p/go-nat v0.2.0
```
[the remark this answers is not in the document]
lixmal: These libraries are indirect. We don't want to pull in the whole go-libp2p dependency

client/internal/peer/worker_ice.go
```diff
@@ -362,6 +368,77 @@ func (w *WorkerICE) onICECandidate(candidate ice.Candidate) {
...
+	}
+
+	// Per RFC 8445, the related address for srflx is the base (host candidate address).
+	// If the original srflx has unspecified related address, use its own address as base.
+	relAddr := srflxCandidate.RelatedAddress().Address
+	if relAddr == "" || relAddr == "0.0.0.0" || relAddr == "::" {
+		relAddr = srflxCandidate.Address()
+	}
+
+	priority := srflxCandidate.Priority() + 1000
```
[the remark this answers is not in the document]
lixmal: Fixed

client/internal/portforward/manager.go
```diff
@@ -0,0 +1,307 @@
...
+	m.ctx, m.cancel = context.WithCancel(ctx)
+	m.wgPort = wgPort
+
+	m.stateManager.RegisterState(&State{})
+
+	m.wg.Add(1)
+	go m.run()
+}
+
+func isDisabledByEnv() bool {
```
pappz: Please move the static functions to the end of the file.

client/internal/portforward/manager.go
```diff
@@ -0,0 +1,307 @@
...
+		NATType:      m.gateway.Type(),
+	}
+
+	log.Infof("created port mapping: %d -> %d via %s (external IP: %s)",
+		m.wgPort, externalPort, m.gateway.Type(), externalIP)
+
+	return m.persistStateLocked()
+}
+
+func (m *Manager) Stop() {
```
pappz: This Stop function is not thread-safe. After the first unlock, another thread can run Start again, which will overwrite all instance-related variables such as m.gateway and m.mapping. We can say that this manager is not reusable but in that case handle this rule in the Start() too

client/internal/portforward/manager.go
```diff
@@ -0,0 +1,307 @@
...
+		return
+	}
+
+	ctx, ctxCancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer ctxCancel()
+
+	if err := m.gateway.DeletePortMapping(ctx, m.mapping.Protocol, int(m.mapping.InternalPort)); err != nil {
+		log.Debugf("delete port mapping on stop: %v", err)
+		return
+	}
```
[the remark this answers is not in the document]
lixmal: The purpose of the state that it exists FOR cleanup. So if port mapping deletion fails we need to make sure it is tried on next cleanup

client/internal/portforward/manager.go
lixmal: done

client/internal/portforward/manager.go
lixmal: Fixed

client/internal/portforward/manager.go
```diff
@@ -152,35 +152,45 @@ func (m *Manager) cleanupResidual(state *State) error {
...
+	wgPort := m.wgPort
+	m.mu.Unlock()
 
-	if m.gateway == nil {
+	if gateway == nil {
 		return nil
 	}
 
 	ctx, cancel := context.WithTimeout(m.ctx, 30*time.Second)
 	defer cancel()
```
pappz, line 165: if at this moment (line 165) run the Stop function then the stop procedure will think the m.gateway is empty so it will return and does not clean up the proper gateway resource.
I do not see sense to always lock -> read m.gateway -> run "is nil logic" in non exported functions. I think would be less risky to create a struct for these variables [ctx, gateway, mapping] and pass it for the functions as parameter. Does not have chance to run into nil pointer.

In the previous version, my impression was that you wanted Stop() to return as soon as possible (i.e., not block an engine restart in case of network changes with slow uPNP cleanup) and to keep the potentially long-running cleanup task in the background. But now it’s different, and because of this I don’t understand why we don’t block Start() (remove the goroutine from the last line) and pass this responsibility to e.shutdownWg. Because in this case would not need to the internal tricks with mutex, the full sturcture is a single thread, except the cleanup routing but it is not important for the engine restart.

client/internal/peer/worker_ice.go
```diff
@@ -362,6 +368,93 @@ func (w *WorkerICE) onICECandidate(candidate ice.Candidate) {
...
+	w.muxAgent.Lock()
+	if w.portForwardAttempted {
+		w.muxAgent.Unlock()
+		return
+	}
+	w.portForwardAttempted = true
+	w.muxAgent.Unlock()
+
+	pfManager := w.conn.portForwardManager
+	if pfManager == nil || !pfManager.IsAvailable() {
```
pappz: IsAvailable() function is unccessary. The next nil check is a must.

client/internal/portforward/manager.go
```diff
@@ -0,0 +1,334 @@
...
+type Mapping struct {
+	Protocol     string
+	InternalPort uint16
+	ExternalPort uint16
+	ExternalIP   net.IP
+	NATType      string
+}
+
+type Manager struct {
+	ctx          context.Context
```
pappz: It would be better to pass the context to every function where we use it, especially when we’re using it in a goroutine.

client/internal/portforward/manager.go
```diff
@@ -0,0 +1,334 @@
...
+	cancel       context.CancelFunc
+	stateManager *statemanager.Manager
+
+	mu      sync.RWMutex
+	gateway nat.NAT
```
pappz: It would be better to pass the gateway to every function where we use it, especially when we’re using it in a goroutine. You will realise do not need to protect this variable any more and not not need to risk and check if it is nil or not.

client/internal/portforward/manager.go
```diff
@@ -0,0 +1,334 @@
...
+		return fmt.Errorf("clear state after cleanup: %w", err)
+	}
+
+	return nil
+}
+
+func (m *Manager) createMapping() error {
+	m.mu.Lock()
+	gateway := m.gateway
+	wgPort := m.wgPort
```
pappz: Why do you copy m.wgPort? It should not be in a protected scope because it will be written only once.

client/internal/portforward/manager.go
lixmal: Fixed in 4d6b810f8 - Start() guards with `m.cancel != nil`, GracefullyStop holds m.mu for the full duration.

client/internal/peer/worker_ice.go
lixmal: Fixed in ec50347d1 - removed IsAvailable(), the nil checks on pfManager and mapping are sufficient.

client/internal/portforward/manager.go
lixmal: Fixed in 4d6b810f8 - context is passed as parameter to all functions (setup, createMapping, renewLoop, renewMapping, cleanup). No more m.ctx field.

client/internal/portforward/manager.go
lixmal: Fixed in 4d6b810f8 - gateway is returned from setup() and passed as parameter to renewLoop(ctx, gateway), renewMapping(ctx, gateway), cleanup(ctx, gateway). No more m.gateway field, no mutex needed for it.

client/internal/portforward/manager.go
lixmal: Fixed in 4d6b810f8 - m.wgPort is set once in Start() before unlock and only read from the Start goroutine. No copy, no mutex protection needed.

client/internal/portforward/manager.go
lixmal: Removed GracefullyStop entirely. The engine's context cancellation triggers the whole chain: renewLoop exits, cleanup spawns in a background goroutine with a 10s timeout (bounded, won't leak), Start returns. No path blocks engine Stop on a slow NAT gateway.

Changes: removed GracefullyStop, stopCtx channel, startTearDown, done channel. deleteMapping now takes the mapping as a parameter (snapshot under lock before spawning) so the goroutine is self-contained.

lixmal: Addressed in 4d6b810f8 (PR #5342): gateway and context are now passed as parameters to all internal functions, no more m.gateway/m.ctx fields. Start runs synchronously with cleanup via stopCtx channel. GracefullyStop provides a bounded context for cleanup, non-graceful path runs cleanup in a background goroutine with 10s timeout.

pappz (approved):

Issue #5200, opened 2026-01-28
failed to verify certificate when downloading geoip from pkgs.netbird.io


hexclann:
**Describe the problem**

Unable to download, geoip database because of restricted environment.

 2026-01-28T05:44:25Z FATL management/internals/server/modules.go:42: could not initialize geolocation service: failed to get database filename: Head "https://pkgs.netbird.io/geolocation-dbs/GeoLite2-City/download?suffix=tar.gz": tls: failed to verify certificate: x509: certificate signed by unknown authority

Im trying to deploy netbird in a AWS EC2 VM. We have PaloAlto firewall enabled for the whole AWS environment. This requires us to install a certificate from them to intercept the traffic.

For all the connection, it uses that certificate. For example, the curl command works:

`curl -L -vv https://pkgs.netbird.io/geolocation-dbs/GeoLite2-City/download?suffix=tar.gz -o test.tar.gz`

```
05:52:28.297187 [0-0] * ALPN: server accepted http/1.1
05:52:28.297247 [0-0] * Server certificate:
05:52:28.297306 [0-0] *  subject: CN=pkgs.netbird.io
05:52:28.297365 [0-0] *  start date: Jan 23 05:51:48 2026 GMT
05:52:28.297425 [0-0] *  expire date: Apr 23 05:51:47 2026 GMT
05:52:28.297487 [0-0] *  subjectAltName: host "pkgs.netbird.io" matched cert's "pkgs.netbird.io"
05:52:28.297553 [0-0] *  issuer: C=US; O=org; OU=mycompany; CN=nonprod-fws.mycompany.org
05:52:28.297609 [0-0] *  SSL certificate verify ok.
```

**To Reproduce**

1. Download the get-started script for selfhosting netbird.
2. Run the script which will launch the containers
3. But after sometime, an error appears which is FATAL and container launch fails 
```
 2026-01-28T05:44:25Z FATL management/internals/server/modules.go:42: could not initialize geolocation service: failed to get database filename: Head "https://pkgs.netbird.io/geolocation-dbs/GeoLite2-City/download?suffix=tar.gz": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

This can be reproduced when having a Palo Alto Panorama certificate installed. I have not tried with any other certificate. This error cannot be override and makes netbird impossible to deploy.

**Expected behavior**

Netbird deployment should work. Option to bypass strict certificate check should be provided.

**Are you using NetBird Cloud?**

No

**NetBird version**

[...]

PR #5216, opened 2026-01-30
[server] Add an option to disable cert verification for external client requests

anudeepND:
## Describe your changes

This PR creates a central HTTP client with an option to disable SSL cert validation when making external requests (for example, downloading GeoIP database, check release versions) when `NB_DISABLE_CERT_VALIDATION` env variable is set to `true`. 

This is useful for self-hosted environments using internal CAs or self-signed certificates which is common when using firewalls that intercepts all the traffic.

Changes:

1. Created `util.NewTransport()` and `util.NewHTTPClient()` to standardize HTTP transport configuration
2. If `NB_DISABLE_CERT_VALIDATION=true`, the transport is configured with `InsecureSkipVerify: true` which will disable certificate validation
3. If it is set to false which is by default, secure transport is used just like previous versions
4. The new config is used in GeoIP download logic, OIDC requests, version check and IDP managers

Fixes https://github.com/netbirdio/netbird/issues/5200

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

https://github.com/netbirdio/docs/pull/582

## Summary by CodeRabbit

* **Infrastructure Improvements**
  * Standardized HTTP client usage across the platform for more consistent and resilient network calls.
  * Added configurable transport behavior (including optional certificate-validation bypass via environment variable) for specialized deployments.
  * Improved HTTP resource handling to ensure responses are properly closed and connections are managed.

### DIFF
--- a/management/server/identity_provider.go
+++ b/management/server/identity_provider.go
@@ -21,6 +21,7 @@ import (
 	"github.com/netbirdio/netbird/management/server/permissions/operations"
 	"github.com/netbirdio/netbird/management/server/types"
 	"github.com/netbirdio/netbird/shared/management/status"
+	"github.com/netbirdio/netbird/util"
 )
 
 // oidcProviderJSON represents the OpenID Connect discovery document
@@ -33,9 +34,8 @@ type oidcProviderJSON struct {
 func validateOIDCIssuer(ctx context.Context, issuer string) error {
 	wellKnown := strings.TrimSuffix(issuer, "/") + "/.well-known/openid-configuration"
 
-	httpClient := &http.Client{
-		Timeout: 10 * time.Second,
-	}
+	httpClient := util.NewHTTPClient()
+	httpClient.Timeout = 10 * time.Second
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, wellKnown, nil)
 	if err != nil {
--- a/shared/auth/jwt/validator.go
+++ b/shared/auth/jwt/validator.go
@@ -10,7 +10,6 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
-	"net/http"
 	"net/url"
 	"strconv"
 	"strings"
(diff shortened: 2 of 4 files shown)


braginini:
@anudeepND What I don't understand is how accessing websites even work in this setting? It looks like some sort of malware

anudeepND:
@braginini this kind of traffic inspection is pretty common I guess. I have seen this in multiple places with Fortient, Cisco and Paloalto which is mentioned in the issue. 

They do DPI (deep packet inspection) by performing authorised MITM. (The VM will be auto configured to trust the root CA generated in the firewall)

Basically, whatever the request I send to external clients is intercepted by the firewall, the firewall will connect to the external endpoint on my behalf with the actual CA cert, it then creates a new certificate for the external end point signed by its own internal CA.

The actual MITM attack is also done in a similar manner so that is the reason the flag is false by default. Users can set it to true for deploying netbird in this kind of strict environment.

Let me know your POV on this

braginini:
@anudeepND 

I understand the point but I think that it could be solved with using a proxy on a VM or docker. We can describe this case in our docs, would you mind testing this? 
https://docs.docker.com/engine/cli/proxy/

P.S. Don't get me wrong, I appreciate your contribution, but I am a bit hesitant to commit code that explicitly promotes not secure behaviour. That being said, I'm excited about your use case and would love to have a chat. Could you please ping me (Misha) on slack and we set up a call: https://docs.netbird.io/slack-url

anudeepND:
@braginini I agree with your concerns, will ping you in slack to have a conversation about this. I will also be able to show my use case as well

mlsmaycon:
@anudeepND can you try setting the proxy via HTTP_PROXY and HTTPS_PROXY environment variables in the management container?

anudeepND:
@mlsmaycon Docker proxy will not work in this case, the root CA is installed at the OS level. So even when Docker tries to perform SSL handshake through the tunnel, the firewall will intercept the traffic which will cause the same issue where cert verification fails in netbird 

Also, is there any other communication channel other than slack?

PR #5264, opened 2026-02-05
[client] Optimize Windows DNS performance with domain batching and batch mode

pappz:
Implement two-layer optimization to reduce Windows NRPT registry operations:

1. Domain Batching (host_windows.go):
  - Batch domains per NRPT
  - Reduces NRPT rules by ~97% (e.g., 184 domains: 184 rules → 4 rules)
  - Modified addDNSMatchPolicy() to create batched NRPT entries
  - Added comprehensive tests in host_windows_test.go

2. Batch Mode (server.go):
  - Added BeginBatch/EndBatch methods to defer DNS updates
  - Modified RegisterHandler/DeregisterHandler to skip applyHostConfig in batch mode
  - Protected all applyHostConfig() calls with batch mode checks
  - Updated route manager to wrap route operations with batch calls

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * DNS rules are applied in batches to reduce per-domain updates; a batch mode defers applying DNS/host/route changes until batching completes.

* **Bug Fixes**
  * Cleanup and reporting improved when domain lists change: rule counts reflect batch-based NRPT entries and errors are reported per rule/batch.

* **Tests**
  * Added tests to validate domain batching, expected rule counts, and cleanup across multiple domain set sizes.

### DIFF
--- a/client/internal/dns/server.go
+++ b/client/internal/dns/server.go
@@ -43,6 +43,7 @@ type Server interface {
 	DeregisterHandler(domains domain.List, priority int)
 	BeginBatch()
 	EndBatch()
+	CancelBatch()
 	Initialize() error
 	Stop()
 	DnsIP() netip.Addr
@@ -287,6 +288,16 @@ func (s *DefaultServer) EndBatch() {
 	log.Debugf("DNS batch mode disabled, applying accumulated changes")
 	s.batchMode = false
 	s.applyHostConfig()
+}
+
+// CancelBatch cancels batch mode without applying accumulated changes.
+// This is useful when operations fail partway through and you want to
+// discard partial state rather than applying it.
+func (s *DefaultServer) CancelBatch() {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	log.Debugf("DNS batch mode cancelled, discarding accumulated changes")
+	s.batchMode = false
 }
 
 func (s *DefaultServer) deregisterHandler(domains []string, priority int) {
--- a/client/internal/routemanager/manager.go
+++ b/client/internal/routemanager/manager.go
@@ -348,9 +348,19 @@ func (m *DefaultManager) updateSystemRoutes(newRoutes route.HAMap) error {
 	var merr *multierror.Error
 
 	// Begin batch mode to avoid calling applyHostConfig() after each DNS handler operation
+	batchStarted := false
 	if m.dnsServer != nil {
 		m.dnsServer.BeginBatch()
-		defer m.dnsServer.EndBatch()
+		batchStarted = true
+		defer func() {
+			if merr != nil {
+				// On error, cancel batch to discard partial DNS state
+				m.dnsServer.CancelBatch()
+			} else {
+				// On success, apply accumulated DNS changes
+				m.dnsServer.EndBatch()
+			}
+		}()
 	}
 
 	for id, handler := range toRemove {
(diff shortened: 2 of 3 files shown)


client/internal/dns/host_windows.go
```diff
@@ -42,6 +42,8 @@ const (
 	dnsPolicyConfigConfigOptionsKey     = "ConfigOptions"
 	dnsPolicyConfigConfigOptionsValue   = 0x8
 
+	nrptMaxDomainsPerRule = 50
```
lixmal: Let's add any kind of reference here if we have one

client/internal/dns/host_windows.go
```diff
@@ -264,8 +276,8 @@ func (r *registryConfigurator) addDNSMatchPolicy(domains []string, ip netip.Addr
 		}
 	}
 
-	log.Infof("added %d separate NRPT entries. Domain list: %s", len(domains), domains)
-	return len(domains), nil
+	log.Infof("added %d NRPT rules for %d domains. Domain list: %v", ruleIndex, len(domains), domains)
+	return ruleIndex, nil
```
lixmal, line 281: The returned count here is correct, but the caller in `applyDNSConfig` discards it on error. `nrptEntryCount` keeps the stale value from the last successful run and the second `updateState` never fires.

So if previous run had 1 rule and the new run wrote entries 0-2 before failing, `nrptEntryCount` stays 1. Next `removeDNSMatchPolicies` only cleans entry 0;  entries 1 and 2 (up to 50 domains each) are orphaned in the registry.

client/internal/dns/server.go
```diff
@@ -259,6 +264,28 @@ func (s *DefaultServer) DeregisterHandler(domains domain.List, priority int) {
...
+}
+
+// BeginBatch starts batch mode for DNS handler registration/deregistration.
+// In batch mode, applyHostConfig() is not called after each handler operation,
+// allowing multiple handlers to be registered/deregistered efficiently.
+// Must be followed by EndBatch() to apply the accumulated changes.
+func (s *DefaultServer) BeginBatch() {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	log.Infof("DNS batch mode enabled")
```
lixmal: Better debug?

client/internal/dns/server.go
```diff
@@ -83,6 +85,7 @@ type DefaultServer struct {
 	currentConfigHash  uint64
 	handlerChain       *HandlerChain
 	extraDomains       map[domain.Domain]int
+	batchMode          bool
```
lixmal, line 89: `batchMode` is global on the server so it also suppresses `applyHostConfig` in `upstreamCallbacks` and `applyConfiguration`. If a nameserver goes down during a route update the host config change gets deferred until `EndBatch`. Maybe only for`RegisterHandler`/`DeregisterHandler`?

client/internal/routemanager/manager.go
```diff
@@ -346,6 +346,13 @@ func (m *DefaultManager) updateSystemRoutes(newRoutes route.HAMap) error {
 	}
 
 	var merr *multierror.Error
+
+	// Begin batch mode to avoid calling applyHostConfig() after each DNS handler operation
+	if m.dnsServer != nil {
+		m.dnsServer.BeginBatch()
+		defer m.dnsServer.EndBatch()
```
lixmal: If the loop below errors out partway, `EndBatch` still fires via defer and calls `applyHostConfig` with partial state. Do we want that?

PR #5277, opened 2026-02-09
[client] Refactor WG endpoint setup with role-based proxy activation

pappz:
Update websocket to v1.8.14

Move wgProxy.Work() and wgConfigWorkaround() into EndpointUpdater to centralize the timing of proxy activation relative to WireGuard peer updates. The initiator calls Work() before updating the peer (so the proxy is ready for the immediate handshake), while the responder calls Work() after (to avoid premature activation).

Add SwitchWGEndpoint for relay fallback on ICE disconnect, which skips the initiator/responder distinction since WireGuard roaming wil overwrite the endpoint regardless.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Direct endpoint switching and explicit endpoint removal for faster manual control.
  * Option to apply endpoint updates immediately or defer them for smoother transitions.

* **Bug Fixes**
  * Improved relay vs direct upgrade handling with better sequencing and conditional proxy activation.
  * Removed timing workaround in favor of controller-aware update flow.

* **Chores**
  * Updated websocket dependency to latest patch release.

### DIFF
--- a/client/internal/peer/conn.go
+++ b/client/internal/peer/conn.go
@@ -410,7 +410,7 @@ func (conn *Conn) onICEConnectionIsReady(priority conntype.ConnPriority, iceConn
 	conn.doOnConnected(iceConnInfo.RosenpassPubKey, iceConnInfo.RosenpassAddr)
 }
 
-func (conn *Conn) onICEStateDisconnected() {
+func (conn *Conn) onICEStateDisconnected(sessionChanged bool) {
 	conn.mu.Lock()
 	defer conn.mu.Unlock()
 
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -40,7 +40,7 @@ require (
 	github.com/c-robinson/iplib v1.0.3
 	github.com/caddyserver/certmagic v0.21.3
 	github.com/cilium/ebpf v0.15.0
-	github.com/coder/websocket v1.8.13
+	github.com/coder/websocket v1.8.14
```
lixmal, line 43: Nit: add this to the PR description so we'll find it if it becomes an issue

client/internal/peer/conn.go
```diff
@@ -430,14 +432,10 @@ func (conn *Conn) onICEStateDisconnected() {
 	if conn.isReadyToUpgrade() {
 		conn.Log.Infof("ICE disconnected, set Relay to active connection")
 		conn.dumpState.SwitchToRelay()
-		conn.wgProxyRelay.Work()
```
lixmal, line 437: We don't need these anymore at all? Don't we need to activate the relay if it's paused?

lixmal (approved):

PR #5283, opened 2026-02-09
[client] Reset WireGuard endpoint on ICE session change during relay fallback

pappz:
When an ICE connection disconnects and falls back to relay, reset the WireGuard endpoint and handshake watcher if the remote peer's ICE session has changed. This ensures the controller re-establishes a fresh WireGuard handshake rather than waiting on a stale endpoint from the previous session.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced peer connection resilience during network disconnections and transitions between connection protocols.
  * Improved connection state management with better session change detection and endpoint cleanup.
  * Added reset mechanisms to ensure proper state recovery from connection disruptions.


client/internal/peer/worker_ice.go
```diff
@@ -306,13 +308,17 @@ func (w *WorkerICE) connect(ctx context.Context, agent *icemaker.ThreadSafeAgent
...
 	cancel()
 	if err := agent.Close(); err != nil {
 		w.log.Warnf("failed to close ICE agent: %s", err)
 	}
 
 	w.muxAgent.Lock()
+	defer w.muxAgent.Unlock()
+
+	sessionChanged := w.remoteSessionChanged
+	w.remoteSessionChanged = false
```
lixmal, line 321: Both old and new agents consume this flag. Intentional?

pappz: I do not get it

pappz: 1. Agent A is alive and ICE connection is established with a WireGuard endpoint configured.
  2. New offer arrives (OnNewOffer):
    - Remote peer has a new session ID (different from current), meaning it wants to reconnect.
    - remoteSessionChanged = true is set, this is the signal that the WG endpoint needs to be reset.
    - Agent A is closed directly and replaced by a new Agent B.
    - w.agent now points to Agent B.
  3. Deprecated Agent A's callback fires (onConnectionStateChange → closeAgent):
    - Agent A transitions to disconnected/failed/closed.
    - closeAgent is called with Agent A.
    - but we must to reset the WireGuard, consume the flag

The problem is we have a race between the Agent A close callback and the Agent B on connected callback. If the Agent B connected faster then disconnect the old one then we have an inconsistent state. This is why we do all the time this double check but it is not enough.

PR #5411, opened 2026-02-21
[client,management] Feature/client service expose

mlsmaycon:
## Describe your changes
- Add `netbird expose` CLI command allowing peers to expose local HTTP services via the NetBird reverse proxy with TTL-based session management
- Implement 3 unary management RPCs (`CreateExpose`, `RenewExpose`, `StopExpose`) replacing the initial bidirectional streaming design
- Add permission validation, per-peer rate limiting, and expiration handling for peer-exposed services
- Introduce `PeerExposeEnabled` and `PeerExposeGroups` account settings to control which peers can expose services

## Architecture

```
CLI (expose cmd) ──server-stream──► Daemon ──3 unary RPCs (encrypted)──► Management
  Ctrl+C cancels ctx                 CreateExpose → send Ready event       creates ephemeral service
                                     30s ticker → RenewExpose              updates lastRenewed in-memory
                                     defer → StopExpose (bg ctx)           deletes service, cleans up
                                                                           TTL reaper (30s interval, 90s TTL)
```

## Key changes

**CLI (`client/cmd/expose.go`)**
- New `netbird expose <port>` command with flags: `--with-pin`, `--with-password`, `--with-user-groups`, `--with-custom-domain`, `--with-name-prefix`, `--protocol`
- Streams events from daemon, displays service URL on ready, handles Ctrl+C gracefully

**Daemon (`client/server/server.go`)**
- Server-streaming `ExposeService` handler: calls `CreateExpose`, starts 30s renewal ticker, defers `StopExpose` for cleanup on all exit paths
- Uses background context for management client so `StopExpose` works after CLI disconnect

**Management gRPC (`expose_service.go`)**
- Three unary encrypted handlers: `CreateExpose`, `RenewExpose`, `StopExpose`
- In-memory `activeExposes` sync.Map with mutex-protected `lastRenewed` per session
- TTL reaper goroutine (30s interval, 90s TTL) expires stale sessions
- Per-peer rate limit (max 10 concurrent exposes)
- Delegates to reverseproxy manager for all business logic

**Reverse proxy manager (`manager.go`)**
- `ValidateExposePermission`: checks `PeerExposeEnabled` setting and peer group membership
- `CreateServiceFromPeer`: creates ephemeral service, resolves peer IP for targets, stores activity event
- `DeleteServiceFromPeer` / `ExpireServiceFromPeer`: validates source peer ownership, deletes service, stores activity event
- `FromExposeRequest`: constructs `Service` from proto request (mirrors `FromAPIRequest` pattern)
- Source types renamed from `api`/`peer` to `permanent`/`ephemeral`

**Account settings**
- `PeerExposeEnabled` (bool) and `PeerExposeGroups` ([]string) added to `ExtraSettings`
- OpenAPI spec updated with new fields
- Account handler passes settings through to API responses

**Activity codes**
- `PeerServiceExposed` (111), `PeerServiceUnexposed` (112), `PeerServiceExposeExpired` (113)

## Test plan

- [ ] `go test ./management/internals/modules/reverseproxy/...` — 17 tests including `TestGenerateExposeName`, `TestFromExposeRequest`, `TestValidateExposePermission`, `TestCreateServiceFromPeer`, `TestGetGroupIDsFromNames`
- [ ] `go test ./management/internals/shared/grpc/` — expose tests including `TestReapExpiredExposes`, `TestCountPeerExposes`, `TestCleanupExpose`, concurrent safety tests
- [ ] `go build ./...` and `go vet ./...` pass clean
- [ ] Manual: `netbird expose 8080` creates service, shows URL, Ctrl+C cleans up
- [ ] Manual: service expires after 90s without renewal
- [ ] Manual: disable peer expose in settings, verify CLI gets permission denied
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

https://github.com/netbirdio/docs/pull/633

## Summary by CodeRabbit

* **New Features**
  * CLI: new expose command to publish a local port with flags for PIN, password, user groups, custom domain, name prefix and protocol (HTTP default).
  * Management/API: create/renew/stop expose sessions (streamed status), automatic naming/domain, TTL renewals, background expiration, new management RPCs and client methods.
  * UI/API: account settings now include peer_expose_enabled and peer_expose_groups; new activity codes for peer expose events.

* **Tests**
  * Extensive tests covering lifecycle, naming, permissions, reaper/expiration and concurrency.

### DIFF
--- a/management/internals/modules/reverseproxy/manager/manager.go
+++ b/management/internals/modules/reverseproxy/manager/manager.go
@@ -623,21 +623,21 @@ func (m *managerImpl) ValidateExposePermission(ctx context.Context, accountID, p
 	settings, err := m.store.GetAccountSettings(ctx, store.LockingStrengthNone, accountID)
 	if err != nil {
 		log.WithContext(ctx).Errorf("failed to get account settings: %v", err)
-		return fmt.Errorf("get account settings: %w", err)
+		return status.Errorf(status.Internal, "get account settings: %v", err)
 	}
 
 	if !settings.PeerExposeEnabled {
-		return fmt.Errorf("peer expose is not enabled for this account")
+		return status.Errorf(status.PermissionDenied, "peer expose is not enabled for this account")
 	}
 
 	if len(settings.PeerExposeGroups) == 0 {
-		return fmt.Errorf("no group is set for peer expose")
+		return status.Errorf(status.PermissionDenied, "no group is set for peer expose")
 	}
 
 	peerGroupIDs, err := m.store.GetPeerGroupIDs(ctx, store.LockingStrengthNone, accountID, peerID)
 	if err != nil {
 		log.WithContext(ctx).Errorf("failed to get peer group IDs: %v", err)
-		return fmt.Errorf("get peer groups: %w", err)
+		return status.Errorf(status.Internal, "get peer groups: %v", err)
 	}
 
 	for _, pg := range peerGroupIDs {
@@ -646,7 +646,7 @@ func (m *managerImpl) ValidateExposePermission(ctx context.Context, accountID, p
 		}
 	}
 
-	return fmt.Errorf("peer is not in an allowed expose group")
+	return status.Errorf(status.PermissionDenied, "peer is not in an allowed expose group")
 }
 
 // CreateServiceFromPeer creates a service initiated by a peer expose request.


shared/management/http/api/openapi.yml
```diff
@@ -378,12 +378,24 @@ components:
...
+            type: string
+            example: ch8i4ug6lnn4g9hqv7m0
       required:
         - peer_approval_enabled
         - user_approval_required
         - network_traffic_logs_enabled
         - network_traffic_logs_groups
         - network_traffic_packet_counter_enabled
+        - peer_expose_enabled
+        - peer_expose_groups
```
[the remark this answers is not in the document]
mlsmaycon: This is a
  broader API design issue that exists for all AccountExtraSettings fields, not just ours. No fix needed

client/internal/engine.go
```diff
@@ -1824,11 +1828,16 @@ func (e *Engine) GetRouteManager() routemanager.Manager {
...
 
-// GetFirewallManager returns the firewall manager
+// GetFirewallManager returns the firewall manager.
 func (e *Engine) GetFirewallManager() firewallManager.Manager {
 	return e.firewall
 }
 
+// GetExposeManager returns the expose session manager.
+func (e *Engine) GetExposeManager() *expose.Manager {
+	return e.exposeManager
```
pappz, line 1840: returns e.exposeManager without holding any lock, while engine.go:425 writes it inside Start(). If the client is restarting while ExposeService is running, there's a data race.

client/cmd/expose.go
```diff
@@ -0,0 +1,182 @@
...
+func validateExposeFlags(cmd *cobra.Command, portStr string) (uint64, error) {
+	port, err := strconv.ParseUint(portStr, 10, 32)
+	if err != nil {
+		return 0, fmt.Errorf("invalid port number: %s", portStr)
+	}
+	if port == 0 || port > 65535 {
+		return 0, fmt.Errorf("invalid port number: must be between 1 and 65535")
+	}
+
+	if exposeProtocol != "http" {
```
pappz: If we only support HTTP, why has it been extracted to the command line? It sounds like an “Am I a human?” test where I have to copy and paste the correct answer.

client/server/server.go
```diff
@@ -1312,6 +1313,81 @@ func (s *Server) WaitJWTToken(
...
+	result, err := mgr.Expose(exposeCtx, mgmReq)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		stopCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if err := mgr.Stop(stopCtx, result.Domain); err != nil {
+			log.Debugf("failed to stop expose: %v", err)
```
pappz: Use Warn instead of Debug

client/internal/expose/manager.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Result holds the response from creating an expose session.
+type Result struct {
+	ServiceName string
+	ServiceURL  string
+	Domain      string
+}
+
+// Manager handles expose session lifecycle via the management client.
+type Manager struct {
```
pappz, line 37: It is not a manager. It just a wrapper of mgm client. Move the renew ticker into this struct and become a manager. Keep the business logic into the feautre package.

client/internal/expose/manager.go
```diff
@@ -0,0 +1,54 @@
...
+	mgmClient mgm.Client
+}
+
+// NewManager creates a new expose Manager using the given management client.
+func NewManager(mgmClient mgm.Client) *Manager {
```
pappz: Would be easier to define a interface then depend on a monster mgm.Client.
```
type Client interface {
	CreateExpose(ctx context.Context, req *mgmProto.ExposeServiceRequest) (*mgmProto.ExposeServiceResponse, error)
	RenewExpose(ctx context.Context, domain string) error
	StopExpose(ctx context.Context, domain string) error
}
```

client/internal/expose/manager.go
```diff
@@ -0,0 +1,54 @@
...
+	return &Manager{mgmClient: mgmClient}
+}
+
+// Expose creates a new expose session via the management server.
+func (m *Manager) Expose(ctx context.Context, req *mgmProto.ExposeServiceRequest) (*Result, error) {
```
pappz: You defined a custom Result type. Do the same for the request, create a new request struct. The Protobuf conversion is a concern of the upper layer, not the feature manager.

client/server/server.go
```diff
@@ -1312,6 +1313,81 @@ func (s *Server) WaitJWTToken(
...
+	ticker := time.NewTicker(30 * time.Second)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			if err := mgr.Renew(ctx, result.Domain); err != nil {
+				_ = srv.Send(&proto.ExposeServiceEvent{
```
pappz: Move this line to a function where the name represent what is the goal.

client/cmd/expose.go
```diff
@@ -0,0 +1,182 @@
...
+		if err := conn.Close(); err != nil {
+			log.Debugf("failed to close daemon connection: %v", err)
+		}
+	}()
+
+	client := proto.NewDaemonServiceClient(conn)
+
+	stream, err := client.ExposeService(ctx, &proto.ExposeServiceRequest{
+		Port:       uint32(port),
+		Protocol:   proto.ExposeProtocol_EXPOSE_HTTP,
```
pappz: expose.go:116 hardcodes proto.ExposeProtocol_EXPOSE_HTTP in the request. The exposeProtocol variable is validated to be "http" but then discarded — the enum is never derived from it. This creates a subtle disconnect:
   future protocol support would require changes in two places, and a protocol mismatch bug could slip in.

Avoid overengineering. When we start to support multiple protocols then we will take care about protocol mapping.

management/internals/modules/reverseproxy/manager/manager.go
```diff
@@ -537,3 +546,183 @@ func (m *managerImpl) GetServiceIDByTargetID(ctx context.Context, accountID stri
...
+
+	extraSettings, err := m.settingsManager.GetExtraSettings(ctx, accountID)
+	if err != nil {
+		log.WithContext(ctx).Errorf("failed to get extra settings: %v", err)
+		return fmt.Errorf("get account settings: %w", err)
+	}
+
+	if extraSettings == nil || !extraSettings.PeerExposeEnabled {
+		return fmt.Errorf("peer expose is not enabled for this account")
+	}
```
pascal-fischer: This setting should not be part of extra settings

PR #5448, opened 2026-02-25
[client] refactor auto update workflow

pappz:
## Describe your changes

                                                                                                                                     
  - **Auto-update logic moved out of the UI** into a dedicated `updatemanager.Manager` service that runs in the connection layer. The 
  UI no longer polls or checks for updates independently.                                                                             
  - The update manager supports three modes driven by the management server's auto-update policy:                                     
    1. **No policy set by mgm**: checks GitHub for the latest version and notifies the user (previous behavior, now centralized)      
    2. **mgm enforces update**: the "About" menu triggers installation directly instead of just downloading the file — user still initiates the action
    3. **mgm forces update**: installation proceeds automatically without user interaction
  - `updateManager` lifecycle is now owned by daemon, giving the daemon server direct control via a new `TriggerUpdate` RPC
  - Introduces `EngineServices` struct to group external service dependencies passed to `NewEngine`, reducing its argument count from 11 to 4

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * UI can trigger installation of a pending enforced update via the daemon.
  * Update notifications show version and whether the update is enforced.
  * GUI starts update handling earlier to surface available updates promptly.

* **Refactor**
  * Reworked update manager and engine wiring for clearer update states, install flow, and improved cross-platform auto-update behavior.

### DIFF
--- a/client/proto/daemon.proto
+++ b/client/proto/daemon.proto
@@ -230,8 +230,7 @@ message WaitSSOLoginResponse {
 message UpRequest {
   optional string profileName = 1;
   optional string username = 2;
-  // deprecated, reserve 3
-  // optional bool autoUpdate = 3;
+  reserved 3;
 }
 
 message UpResponse {}
--- a/client/proto/daemon.pb.go
+++ b/client/proto/daemon.pb.go
@@ -6096,12 +6096,12 @@ const file_daemon_proto_rawDesc = "" +
 	"\buserCode\x18\x01 \x01(\tR\buserCode\x12\x1a\n" +
 	"\bhostname\x18\x02 \x01(\tR\bhostname\",\n" +
 	"\x14WaitSSOLoginResponse\x12\x14\n" +
-	"\x05email\x18\x01 \x01(\tR\x05email\"p\n" +
+	"\x05email\x18\x01 \x01(\tR\x05email\"v\n" +
 	"\tUpRequest\x12%\n" +
 	"\vprofileName\x18\x01 \x01(\tH\x00R\vprofileName\x88\x01\x01\x12\x1f\n" +
 	"\busername\x18\x02 \x01(\tH\x01R\busername\x88\x01\x01B\x0e\n" +
 	"\f_profileNameB\v\n" +
-	"\t_username\"\f\n" +
+	"\t_usernameJ\x04\b\x03\x10\x04\"\f\n" +
 	"\n" +
 	"UpResponse\"\xa1\x01\n" +
 	"\rStatusRequest\x12,\n" +


client/internal/engine.go
```diff
@@ -793,39 +792,31 @@ func (e *Engine) PopulateNetbirdConfig(netbirdConfig *mgmProto.NetbirdConfig, mg
...
-		updateManager, err := updatemanager.NewManager(e.statusRecorder, e.stateManager)
-		if err != nil {
-			return
-		}
-		e.updateManager = updateManager
-		e.updateManager.Start(e.ctx)
+	if autoUpdateSettings.Version == disableAutoUpdate {
+		log.Infof("auto-update is disabled")
+		e.updateManager.SetDownloadOnly()
+		return
```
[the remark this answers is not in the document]
pappz: If autoUpdateSettings is nil, it means that the server does not want to modify the current settings.

client/server/server.go
```diff
@@ -135,6 +139,12 @@ func (s *Server) Start() error {
 		log.Warnf(errRestoreResidualState, err)
 	}
 
+	if s.updateManager == nil {
+		stateMgr := statemanager.New(s.profileManager.GetStatePath())
+		s.updateManager = updatemanager.NewManager(s.statusRecorder, stateMgr)
+		s.updateManager.CheckUpdateSuccess(s.rootCtx)
+	}
```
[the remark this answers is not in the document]
pappz: This logic should operate with GUI only. In headless mode no.

client/internal/engine.go
```diff
@@ -309,7 +315,7 @@ func (e *Engine) Stop() error {
 	}
 
 	if e.updateManager != nil {
-		e.updateManager.Stop()
+		e.updateManager.SetDownloadOnly()
 	}
```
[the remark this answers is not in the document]
pappz: This comment is outdated

client/internal/connect.go
```diff
@@ -59,17 +59,19 @@ func NewConnectClient(
...
-		doInitialAutoUpdate: doInitalAutoUpdate,
-		engineMutex:         sync.Mutex{},
+		ctx:            ctx,
+		config:         config,
+		statusRecorder: statusRecorder,
+		engineMutex:    sync.Mutex{},
 	}
 }
 
+func (c *ConnectClient) SetUpdateManager(um *updatemanager.Manager) {
```
mlsmaycon: race risk with c.UpdateManager()?

client/internal/updatemanager/manager.go
```diff
@@ -159,9 +179,13 @@ func (m *Manager) SetVersion(expectedVersion string) {
 		log.Errorf("empty expected version provided")
 		m.expectedVersion = nil
 		m.updateToLatestVersion = false
+		m.downloadOnly = true
 		return
 	}
 
+	m.downloadOnly = false
```
mlsmaycon: set m.downloadOnly = false and m.forceUpdate = forceUpdate before validating the version string. If v.NewVersion() fails at line 193, the method
  returns early with downloadOnly=false and forceUpdate set

client/internal/updater/manager.go
```diff
@@ -184,6 +208,73 @@ func (m *Manager) SetVersion(expectedVersion string) {
...
+
+	m.statusRecorder.PublishEvent(
+		cProto.SystemEvent_INFO,
+		cProto.SystemEvent_SYSTEM,
+		"New version available",
+		"",
+		map[string]string{"new_version_available": latestVersion.String()},
+	)
+}
+
```
mlsmaycon, line 293: Stop() doesn't reset m.cancel = nil, so Start() can never be called again after Stop() because Start() checks if m.cancel != nil { return }. The manager
  becomes permanently dead.

client/ui/event_handler.go
```diff
@@ -212,9 +212,40 @@ func (h *eventHandler) handleGitHubClick() {
...
+	enforced := h.client.isEnforcedUpdate
+	h.client.updateIndicationLock.Unlock()
+
+	if !enforced {
+		if err := openURL(version.DownloadUrl()); err != nil {
+			log.Errorf("failed to open download URL: %v", err)
+		}
+		return
 	}
+
```
mlsmaycon, line 224: we can disable the update button before calling the routine

mlsmaycon (changes requested):

client/internal/updater/manager.go
```diff
@@ -142,13 +144,31 @@ func (m *Manager) Start(ctx context.Context) {
...
+	m.forceUpdate = false
+	m.expectedVersion = nil
+	m.updateToLatestVersion = false
+	m.updateMutex.Unlock()
+
+	select {
+	case m.mgmUpdateChan <- struct{}{}:
+	default:
+	}
+}
```
lixmal, line 167: This should also clear `pendingVersion`

client/internal/engine.go
```diff
@@ -793,39 +792,31 @@ func (e *Engine) PopulateNetbirdConfig(netbirdConfig *mgmProto.NetbirdConfig, mg
 	return nil
 }
 
-func (e *Engine) handleAutoUpdateVersion(autoUpdateSettings *mgmProto.AutoUpdateSettings, initialCheck bool) {
-	if autoUpdateSettings == nil {
-		return
+// TriggerUpdate initiates installation of the pending enforced version via the update manager.
+// Called from the server when the user clicks the install button in the UI.
+func (e *Engine) TriggerUpdate(ctx context.Context) error {
```
lixmal: Not called anywhere

client/proto/daemon.proto
```diff
@@ -226,7 +230,6 @@ message WaitSSOLoginResponse {
 message UpRequest {
   optional string profileName = 1;
   optional string username = 2;
-  optional bool autoUpdate = 3;
```
lixmal, line 229: should probable reserve / deprecate this just in case of the rare version mismatch

client/ui/event_handler.go
```diff
@@ -212,9 +212,40 @@ func (h *eventHandler) handleGitHubClick() {
...
+	// prevent against of busy server
```
lixmal: Fix this comment grammar

client/internal/updater/manager.go
pappz: I am not sure. Just because the server change the expected version we should not break the pending installation.

client/internal/updater/manager.go
pappz: This is the goal. This manager is not reusable (would be hard to implement to be reusable and it is not a requirement at the moment). I will document it.

client/proto/daemon.proto
```diff
@@ -226,7 +230,8 @@ message WaitSSOLoginResponse {
...
+  // deprecated, reserve 3
```
lixmal: we can use the `reserved` keyword here

PR #5478, opened 2026-03-01
[management] Validate account ID in URL matches authenticated account

John-Dixon-IV:
## Summary
- The `updateAccount` and `deleteAccount` handlers extracted `accountId` from the URL parameter without validating it against the authenticated user's account from the auth context. Every other handler in the codebase (users, groups, routes, peers, dns) uses `userAuth.AccountId` from the auth context — these two were the only ones trusting the URL parameter.
- Added explicit validation that the URL `accountId` matches `userAuth.AccountId`, returning 403 if they differ.
- The backend `ValidateUserPermissions` check mitigates single-account exploitation, but handlers should enforce this as defense-in-depth.

## Test plan
- [x] `PutAccount with mismatched accountId returns forbidden` — URL has different accountId than auth context → 403
- [x] `TestDeleteAccount_CrossAccountForbidden` — DELETE with mismatched accountId → 403
- [x] All 9 existing account handler tests continue to pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened account access control to ensure users can only update or delete their own accounts. Attempts to modify accounts belonging to other users now return permission denied responses.

* **Tests**
  * Added comprehensive test coverage for cross-account access restriction scenarios during account updates and deletions.

### DIFF
--- a/management/server/http/handlers/accounts/accounts_handler.go
+++ b/management/server/http/handlers/accounts/accounts_handler.go
@@ -237,12 +237,17 @@ func (h *handler) updateAccount(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	_, userID := userAuth.AccountId, userAuth.UserId
+	accountID, userID := userAuth.AccountId, userAuth.UserId
 
 	vars := mux.Vars(r)
-	accountID := vars["accountId"]
-	if len(accountID) == 0 {
+	reqAccountID := vars["accountId"]
+	if len(reqAccountID) == 0 {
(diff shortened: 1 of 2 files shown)


mlsmaycon:
Hello @John-Dixon-IV, thanks for your PR. The userAuth.AccountId derives from the URL implicitly. So the check implemented here is done in the managers. We are currently moving the permissions validation in the PR https://github.com/netbirdio/netbird/pull/5482 and it will be better to focus the fix if needed after that is merged.

PR #5533, opened 2026-03-07
[proxy] refactor metrics and add usage logs

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Access logs now include bytes_upload and bytes_download (API and schemas updated, fields required).
  * Certificate issuance duration is now recorded as a metric.

* **Refactor**
  * Metrics switched from Prometheus client to OpenTelemetry-backed meters; health endpoint now exposes OpenMetrics via OTLP exporter.

* **Tests**
  * Metric tests updated to use OpenTelemetry Prometheus exporter and MeterProvider.

### DIFF
--- a/proxy/internal/metrics/metrics.go
+++ b/proxy/internal/metrics/metrics.go
@@ -13,13 +13,14 @@ import (
 )
 
 type Metrics struct {
-	ctx               context.Context
-	requestsTotal     metric.Int64Counter
-	activeRequests    metric.Int64UpDownCounter
-	configuredDomains metric.Int64UpDownCounter
-	totalPaths        metric.Int64UpDownCounter
-	requestDuration   metric.Int64Histogram
-	backendDuration   metric.Int64Histogram
+	ctx                      context.Context
+	requestsTotal            metric.Int64Counter
+	activeRequests           metric.Int64UpDownCounter
+	configuredDomains        metric.Int64UpDownCounter
+	totalPaths               metric.Int64UpDownCounter
+	requestDuration          metric.Int64Histogram
+	backendDuration          metric.Int64Histogram
+	certificateIssueDuration metric.Int64Histogram
 
 	mappingsMux  sync.Mutex
 	mappingPaths map[string]int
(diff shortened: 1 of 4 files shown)


proxy/internal/accesslog/logger.go
```diff
@@ -103,3 +146,82 @@ func (l *Logger) log(ctx context.Context, entry logEntry) {
...
+		}
+		l.domainUsage[domain] = usage
+	}
+
+	usage.lastActivity = now
+
+	usage.requestCount++
+	if usage.requestCount >= requestThreshold {
+		elapsed := time.Since(usage.requestStartTime)
+		l.logger.WithFields(log.Fields{
```
mlsmaycon, line 175: should log based on a module calculation

proxy/internal/accesslog/logger.go
```diff
@@ -103,3 +146,82 @@ func (l *Logger) log(ctx context.Context, entry logEntry) {
...
+
+		usage.requestCount = 0
+		usage.requestStartTime = now
+	}
+
+	usage.bytesTransferred += bytesTransferred
+	if usage.bytesTransferred >= bytesThreshold {
+		elapsed := time.Since(usage.bytesStartTime)
+		bytesInGB := float64(usage.bytesTransferred) / (1024 * 1024 * 1024)
+		l.logger.WithFields(log.Fields{
```
mlsmaycon, line 189: should log based on a module calculation

proxy/server.go
```diff
@@ -19,14 +19,17 @@ import (
...
 	"net/url"
 	"path/filepath"
+	"reflect"
 	"sync"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
 	"github.com/pires/go-proxyproto"
-	"github.com/prometheus/client_golang/prometheus"
+	prometheus2 "github.com/prometheus/client_golang/prometheus"
```
mlsmaycon, line 28: clientPrometheus is a better name

proxy/server.go
```diff
@@ -19,14 +19,17 @@ import (
...
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	log "github.com/sirupsen/logrus"
+	"go.opentelemetry.io/otel/exporters/prometheus"
```
mlsmaycon, line 31: exporterPrometheus is a better name

proxy/internal/metrics/metrics.go
```diff
@@ -108,44 +146,52 @@ func (f roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
...
-	m.pathsPerDomain.With(prometheus.Labels{
-		"domain": mapping.Host,
-	}).Set(float64(len(mapping.Paths)))
+	m.mappingsMux.Lock()
+	defer m.mappingsMux.Unlock()
+
+	newPathCount := len(mapping.Paths)
+	oldPathCount, exists := m.mappingPaths[mapping.Host]
+
+	if !exists {
```
mlsmaycon, line 166: Can you share more on why you chose to handle this here and not move the call from update?

pascal-fischer: because otherwise i could just count domains not paths. We could argue if we need the paths counter

proxy/internal/accesslog/logger.go
mlsmaycon: we discussed and the idea is to log on every 10k requests and reset after every 10k requests

PR #5576, opened 2026-03-11
[client] Fix DNS probe thread safety and avoid blocking engine sync

pappz:
Refactor ProbeAvailability to prevent blocking the engine's sync mutex during slow DNS probes.
- the probe now derives its context from the server's own context (s.ctx) instead of accepting one from the caller
- uses a mutex to ensure only one probe runs at a time, new calls cancel the previous probe before starting.
- fixes a data race in Stop() when accessing probeCancel without the probe mutex.
- fix multierr race condition handling

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented overlapping DNS probes and ensured probes are cleanly canceled during shutdown.

* **Performance**
  * DNS availability checks now run asynchronously so updates don’t block main flows.

* **Chores**
  * Improved probe lifecycle coordination and cancellation handling for safer, more reliable concurrent behavior.

### DIFF
--- a/client/internal/dns/upstream.go
+++ b/client/internal/dns/upstream.go
@@ -372,7 +372,9 @@ func (u *upstreamResolverBase) waitUntilResponse() {
 	log.Infof("upstreams %s are responsive again. Adding them back to system", u.upstreamServersString())
 	u.successCount.Add(1)
 	u.reactivate()
+	u.mutex.Lock()
 	u.disabled = false
+	u.mutex.Unlock()
 }
 
 // isTimeout returns true if the given error is a network timeout error.


client/internal/dns/upstream.go
```diff
@@ -398,10 +409,18 @@ func (u *upstreamResolverBase) upstreamServersString() string {
 	return strings.Join(servers, ", ")
 }
 
-func (u *upstreamResolverBase) testNameserver(server netip.AddrPort, timeout time.Duration) error {
-	ctx, cancel := context.WithTimeout(u.ctx, timeout)
+func (u *upstreamResolverBase) testNameserver(ctx context.Context, ctx2 context.Context, server netip.AddrPort, timeout time.Duration) error {
```
lixmal: Can we give these `ctx`s a better name? It's not clear why there are two

client/internal/dns/upstream.go
```diff
@@ -398,10 +409,18 @@ func (u *upstreamResolverBase) upstreamServersString() string {
...
+	ctx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
+	stop := context.AfterFunc(u.ctx, cancel)
+	defer stop()
+
+	if ctx2 != nil {
+		stop2 := context.AfterFunc(ctx2, cancel)
```
lixmal: This seems redundant? Canceling the parent cancels the children too

client/internal/dns/upstream.go
```diff
@@ -114,7 +115,11 @@ func (u *upstreamResolverBase) MatchSubdomains() bool {
 
 func (u *upstreamResolverBase) Stop() {
 	log.Debugf("stopping serving DNS for upstreams %s", u.upstreamServers)
+	u.mutex.Lock()
 	u.cancel()
+	u.wg.Wait()
```
lixmal, line 121: Do we want to wait while locked? This should be outside the lock, right?

client/internal/dns/server.go
```diff
@@ -492,15 +503,52 @@ func (s *DefaultServer) ProbeAvailability() {
...
+
+	// cancel any running probe
+	if s.probeCancel != nil {
+		s.probeCancel()
+		s.probeCancel = nil
+	}
+
+	// wait for the previous probe goroutines to finish while holding
+	// the mutex so no other caller can start a new probe concurrently
+	s.probeWg.Wait()
```
lixmal, line 522: Do we want to wait while locked?

client/internal/dns/upstream.go
pappz: Please check RabbitAI comment: Around line 116-120: Stop() calls u.wg.Wait() without synchronization while
disable() calls u.wg.Add(1) concurrently; this can panic.

client/internal/dns/server.go
pappz: Yes, we want this. With this logic, we can protect this exported function from running multiple probes in parallel. This wait should not block for a long time because the "old" probe should exit immediately as it respects the context cancellation.

client/internal/dns/upstream.go
lixmal: Right, because Add and Wait are in different goroutines, which is unusual

client/internal/dns/upstream.go
```diff
@@ -279,31 +279,39 @@ func (u *upstreamResolverBase) ProbeAvailability() {
...
 
 	wg.Wait()
 
+	select {
+	case <-ctx.Done():
+		return
+	case <-u.ctx.Done():
+		return
+	default:
+	}
```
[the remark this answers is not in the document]
pappz: @RabbitAI review agian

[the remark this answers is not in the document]
pappz: @coderabbitai review again

PR #5592, opened 2026-03-12
fix(dns): kill DnsCache svchost to flush stale NRPT policy on Windows

### LINKED ISSUE #5579: Custom Zone records are not removed when the Netbird client disconnects
**Custom Zones break after disconnection**

Custom zones work wonderfully when connected, but when the netbird client (at least in 0.66.3 on Windows 11) disconnects, the zone no longer reverts to the pre-connected state, and name resolution for the zone is broken.  I believe that whichever mechanism is being used on the the local adapter to intercept the requests, and redirect them to the netbird server DNS is not being properly cleaned up when disconnecting.

I use custom zone records to specify the netbird IP addresses (100.110,x,x) of known resources on my network that are running the netbird client, to cause netbird clients to communicate with those servers via the netbird IP range when connected. (I believe this is the intended purpose of the Custom Zones).

**To Reproduce**

1. Select a host/zone/DNS domain that already exists on your network/in the world without netbird (myhost.contoso.local).
2. from a client, before connecting to netbird, ping the host by name.. name resolution happens immediately via normal DNS, returning a public/private IP address (192.168.1.x)
3. In the netbird console, Create a Zone that matches a DNS domain that already exists outside of netbird (contoso.local)
4. In the netbird console, Create a host record in that zone (myhost.contoso.local) and specify a netbird ranged IP address (100.110.x.x)
5. Connect a netbird client that is scoped to the distribution group the zone applies to.
6. From the client, ping the host record you created in step 2. Name resolution works perfectly, but this time returns the netbird IP Address of the host (100.110.x.x), as expected.
7. Disconnect the netbird client.
8. Try to ping ANY host in the zone that was created.  Name resolution will fail.
    - A packet capture will show that all DNS requests for the zone/domain are being sent to 100.110.255.254 via the regular network adapter.  Because the netbird adapter is not up, it is not reachable, and ALL DNS for the zone now fails.
10. The only way to resolve network communication I've found is to reboot the client machine.

** Expected behaviour **

The DNS redirection/interception that is put in place when the netbird connection is brought up should be removed when the connection is terminated/disconnected.

**NetBird version**
WIndows 11 - Netbird 0.66.3

No

sbeaudry-qms:
After removing NRPT registry keys on disconnect, DnsCache retains stale rules in memory on Windows builds where the service is NOT_PAUSABLE and ignores PARAMCHANGE signals. This causes DNS queries for NetBird domains to continue resolving to the dead forwarder IP until reboot.

Fix by terminating the svchost process hosting DnsCache after registry cleanup, forcing a cold reload of NRPT policy from the now-clean registry. Windows Service Host respawns DnsCache immediately.

Fixes #5579

## Describe your changes

After `removeDNSMatchPolicies()` deletes NRPT registry keys on disconnect, the `DnsCache` service on many Windows 10/11 builds retains the stale rules in memory. The service reports `NOT_PAUSABLE` and silently ignores `SERVICE_CONTROL_PARAMCHANGE`, so the existing `DnsFlushResolverCache` call has no effect on loaded policy. This can be confirmed with `Get-DnsClientNrptPolicy -Effective` — stale rules pointing at the dead NetBird forwarder IP remain visible after disconnect despite the registry being clean.

This fix adds `killDNSCacheProcess()` which terminates the svchost process hosting DnsCache after registry cleanup. Windows Service Host respawns DnsCache immediately, and the fresh instance reads from the now-clean registry. Network connectivity is not interrupted.

- #5579 https://github.com/netbirdio/netbird/issues/5579

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:
- [x] Documentation is **not needed** for this change (internal Windows service lifecycle fix with no user-facing behaviour change)

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNS cache handling on Windows so DNS configuration and policy changes (NRPT) take effect immediately by forcing a fresh resolver instance.
  * If the forced refresh cannot be completed, the system logs a non-fatal warning instead of failing the operation.

### DIFF
--- a/client/internal/dns/host_windows.go
+++ b/client/internal/dns/host_windows.go
@@ -408,7 +408,9 @@ func (r *registryConfigurator) restoreHostDNS() error {
 		return fmt.Errorf("remove interface registry key: %w", err)
 	}
 
-	go r.flushDNSCache()
+	// Flush synchronously before killing DnsCache to avoid a race condition
+	// where the async flush targets a process that is already being terminated.
+	r.flushDNSCache()
 
 	if err := killDNSCacheProcess(); err != nil {
 		log.Warnf("failed to kill DnsCache process for NRPT flush, DNS may require reboot to clear: %v", err)
@@ -497,6 +499,11 @@ func refreshGroupPolicy() error {
 // forcing a cold reload of NRPT policy from registry. This is necessary because
 // DnsCache ignores PARAMCHANGE signals on many Windows builds (NOT_PAUSABLE),
 // and retains stale NRPT rules in memory even after registry keys are deleted.
+//
+// Note: This terminates the entire svchost.exe process, which may include other
+// services in the same process group. Windows Service Control Manager will
+// automatically respawn affected services. There may be a brief interruption
+// or state loss in co-hosted services during respawn.
 func killDNSCacheProcess() error {
 	scm, err := windows.OpenSCManager(nil, nil, windows.SC_MANAGER_CONNECT)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


wizdude:
hi all. someone to review this would be greatly appreciated. i've done a few rollouts with netbird for different people and i've found specifically that a bunch of HP laptops one client has is affected directly by this bug.

@sbeaudry-qms i'm not sure if this also fits your same scenario here. for whatever reason i cannot yet work out, the issue only affects specific HP laptops. if i force kill the dnscache process the machine recovers. doesn't happen on any of the lenovo machines i've pushed this out to. after going through a whole pile of diagnostic work, the only difference i can see are the WiFi drivers between the two machines.

sbeaudry-qms:
@wizdude - The issue affected me on both an MSI laptop and Dell desktop, so definitely not manufacturer specific (the Dell doesn't even have wireless).  My research had led me to believe it was a windows update that had prevented flush from completing properly.

rthill:
I am also impacted by this issue on Dell laptops. Same as @sbeaudry-qms, I believe it is since a specific windows update as when I started to use Netbird I did not notice this issue. It would be great to have this fix released.
Meanwhile, I found the command in the mentioned issue that helps:
```
$dnsPid = (Get-WmiObject Win32_Service -Filter "Name='Dnscache'").ProcessId
Write-Host "Will kill PID: $dnsPid"
Stop-Process -Id $dnsPid -Force
```

Issue #4471, opened 2025-09-09
Client Reauthentication After token expiration is not smooth and seamless.


anshumanFPII:
**Describe the problem**
When the NetBird client token expires, the re-authentication popup on Windows and macOS is unreliable and often does not appear at all. This confuses end-users, requiring IT support assistance to manually re-authenticate them. The process should be seamless and predictable. On Linux, there is also no straightforward prompt to re-authenticate via CLI, which complicates the user experience.

**To Reproduce**

1. Install and run NetBird client (Windows/macOS/Linux).
2. Let the client token expire (e.g., after inactivity or token lifetime).
3. Observe that the expected re-authentication popup often does not show (Windows/macOS).
4. On Linux, there is no clear CLI-driven re-authentication prompt.
5. User remains disconnected until manual IT intervention.

**Expected behavior**

* On Windows/macOS: A reliable popup should *always* appear when re-authentication is required, automatically opening the default browser for login.
* On Linux: A clear CLI prompt should guide the user through re-authentication.
* Overall: The re-authentication process should be seamless and user-friendly without IT support dependency.

**Are you using NetBird Cloud?**
Self-hosted NetBird control plane.

**NetBird version**
`0.56.1`

**Is any other VPN software installed?**
*No*.

**Screenshots**
N/A (popups often do not appear).

**Additional context**
This is a usability issue that leads to confusion among non-technical users. Making re-authentication seamless will significantly improve adoption and reduce IT support overhead.

**Have you tried these troubleshooting steps?**

* [x] Reviewed [[client troubleshooting](https://docs.netbird.io/how-to/troubleshooting-client)](https://docs.netbird.io/how-to/troubleshooting-client)
* [x] Checked for newer NetBird versions
* [x] Searched for similar issues on GitHub (including closed ones)
* [x] Restarted the NetBird client
* [x] Disabled other VPN software
* [x] Checked firewall settings

---

### Feature request 1: Reliable re-authentication flow

Introduce a reliable, cross-platform re-authentication mechanism:

* On Windows/macOS → Always trigger a popup that launches the default browser for login when a token expires.
* On Linux → Provide a clear CLI prompt or command to re-authenticate seamlessly.
* Ensure the process is consistent, predictable, and user-friendly to minimize IT support intervention.

### Feature request 2: Auto-update support for self-hosted setups

Add a mechanism for automatic updates:

[...]

Issue #2109, opened 2024-06-08
When the session expires (because of peer login expiration settings) the client looks like it's connected but it's not


andresemartinez:
**Describe the problem**

I'm using Netbird Selfhosted with Google as IDP. I have peer login expiration set to 24hs. When the time comes the session expires and I lose connection and DNS resolution but the client icon is still orange and if I right click it offers to "disconnect". I could reproduce this on Mac and Windows laptops.

**To Reproduce**

1. Have an instance of Netbird Control Plane with Google as IDP
2. Configure Peer login expiration to 1h
3. Connect with a Windows or Mac client using Google SSO 
4. Put the laptop in sleep mode (not sure if necessary)
5. Wake the laptop up after 1h
6. Try to access any resource on another peer

**Expected behavior**

The client icon and menu should show that you are disconected and need to authenticate again. A desktop notification like the one the OpenVPN client has would be awesome.

**Are you using NetBird Cloud?**

I'm using Selfhosted Control Plane with Google as IDP

**NetBird version**

Control Plane: 0.27.9
Clients for Mac and Windows: 0.27.9 and  0.27.10

nazarewk:
@andresemartinez can you confirm whether it is still the case on the latest NetBird versions?

andresemartinez:
Hey! I haven't used NetBird in a while due to a different issue. I'm going to update everything now and will report back once I've tested it.

janosmiko:
@nazarewk I can confirm - I'm on macOS - (and one of my colleagues - on Windows - reported this) when the session expired we did not get logged out or prompted to login again. After a manual disconnect and reconnect everything works again.
I raised the 
We are on a self-hosted instance using Azure OIDC.

Volemon:
I can confirm that this issue still persists on Linux systems on the latest versions.
On some clients, the NetBird connection stops working while the client still shows as "connected". In the NetBird management center, we don't see a "Login of the peer name is expired" message either. The only way to restore connectivity is to disconnect and reconnect (with re-authentication).

TechHutTV:
Session-expiration UX was fixed in PR #3955. The client now shows a login popup and updates the tray status when the peer login expires. Closing as fixed; please reopen with logs and your IdP details if you still see a stale "connected" state on the latest release.

PR #5608, opened 2026-03-17
fix(ui): open browser in user context on macOS/Linux when daemon runs as root

Nabat0:
## Describe your changes

When the NetBird daemon runs as root and triggers re-authentication after peer login expiry, the bare `open`/`xdg-open` command launches the browser in the root process context, ignoring the logged-in user's default browser preference. On macOS this consistently opens Safari in the background even when the user has Chrome/Firefox set as their default browser, causing the session expired re-auth flow to go unnoticed.

Fix by detecting when running as root (`os.Getuid() == 0`) and invoking `open`/`xdg-open` as the current logged-in user via `sudo -u`, ensuring macOS/Linux resolves the correct default browser in the user's session context. Falls back to bare `open`/`xdg-open` if `user.Current()` fails or if not running as root.

The existing `open-golang` dependency was considered but has the same root cause on darwin — it calls `open` directly without user context.

Tested on macOS (Intel + ARM64) by invalidating a peer session and confirming Chrome opens instead of Safari.

Related: #4471
Related: #2109

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation

- [x] Documentation is **not needed** for this change — internal bug fix for browser launch behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed URL-opening when the app runs with elevated privileges on macOS and Linux/FreeBSD so links open in the desktop user's browser instead of failing or opening as root.

* **Behavior**
  * Respects user-configured browser (BROWSER env) and falls back to the platform default, preserving normal behavior when not running as root.

### DIFF
--- a/client/ui/client_ui.go
+++ b/client/ui/client_ui.go
@@ -1795,19 +1795,58 @@ func (s *serviceClient) showLoginURL() context.CancelFunc {
 	return cancel
 }
 
-func openURL(url string) error {
+// desktopUser returns the username of the logged-in desktop user, even when the
+// process is running as root. On darwin it queries /dev/console; on Linux/FreeBSD
+// it checks the SUDO_USER environment variable, falling back to LOGNAME/USER.
+// Returns an empty string if the desktop user cannot be determined.
+func desktopUser() string {
+	switch runtime.GOOS {
+	case "darwin":
+		out, err := exec.Command("stat", "-f", "%Su", "/dev/console").Output()
+		if err == nil {
+			u := strings.TrimSpace(string(out))
+			if u != "" && u != "root" {
+				return u
+			}
+		}
+	case "linux", "freebsd":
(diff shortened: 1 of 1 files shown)


lixmal:
I don't think this does what you want. `netbird-ui` usually doesn't run as root, only the daemon does. The daemon never calls `open` by itself. 

>Tested on macOS (Intel + ARM64) by invalidating a peer session and confirming Chrome opens instead of Safari.

How did you test that?  I don't see how that works even if you run `netbird-ui` as root (as per coderabbitai's comment)

Nabat0:
After further investigation, I believe the issue I was observing was related to running an older client version (0.64.1) rather than the root cause I described. The fix as written doesn't address a real problem since the UI process
  runs as the logged-in user, not root. Closing this PR. I'll do more investigation before resubmitting.

PR #5611, opened 2026-03-17
[management] Add GET /reverse-proxies/clusters endpoint

lixmal:
## Describe your changes

Add a `/reverse-proxies/clusters` API endpoint that returns active proxy clusters with their connected proxy count. This is needed by the Terraform provider (v0.0.9) `netbird_clusters` data source.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

The endpoint schema is already defined in the OpenAPI spec.

## Summary by CodeRabbit

* **New Features**
  * Added GET /reverse-proxies/clusters to list active reverse-proxy clusters. Response includes each cluster's address and the number of connected proxies. Returns standard authorization and server error responses; access requires Services read permission.
  * API returns aggregated per-cluster counts (recently active proxies) rather than individual proxy addresses.

### DIFF
--- a/management/server/store/sql_store.go
+++ b/management/server/store/sql_store.go
@@ -5444,8 +5444,7 @@ func (s *SqlStore) GetActiveProxyClusterAddresses(ctx context.Context) ([]string
 func (s *SqlStore) GetActiveProxyClusters(ctx context.Context) ([]proxy.Cluster, error) {
 	var clusters []proxy.Cluster
 
-	result := s.db.WithContext(ctx).
-		Model(&proxy.Proxy{}).
+	result := s.db.Model(&proxy.Proxy{}).
 		Select("cluster_address as address, COUNT(*) as connected_proxies").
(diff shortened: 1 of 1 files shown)


management/server/store/sql_store.go
```diff
@@ -5440,6 +5440,25 @@ func (s *SqlStore) GetActiveProxyClusterAddresses(ctx context.Context) ([]string
 	return addresses, nil
 }
 
+// GetActiveProxyClusters returns all active proxy clusters with their connected proxy count.
+func (s *SqlStore) GetActiveProxyClusters(ctx context.Context) ([]proxy.Cluster, error) {
+	var clusters []proxy.Cluster
+
+	result := s.db.WithContext(ctx).
```
pascal-fischer: please no .WithContext() this was causing quite a few issues in the past

lixmal: done

PR #5623, opened 2026-03-18
[client] Support embed.Client on Android with netstack mode

tham-le:
## Describe your changes

`embed.Client.Start()` crashes on Android with nil pointer dereferences because `ConnectClient.Run()` passes an empty `MobileDependency{}`. The engine then dereferences nil `IFaceDiscover`, `NetworkChangeListener`, `DnsReadyListener`, and `TunAdapter`.

This PR makes `embed.Client` usable on Android with netstack/userspace mode (`NB_USE_NETSTACK_MODE=true`), without requiring `VpnService`, `TunAdapter`, or root access.

### Approach

Provide **complete no-op stubs** for all mobile dependencies so the engine's existing Android code paths work unchanged — **zero modifications to engine.go**.

On Android builds, an `init()` in `connect_android_default.go` sets `androidRunOverride` which routes `Run()` through `runOnAndroidEmbed()` with complete stubs:
- `noopIFaceDiscover` — returns empty interface list (relay-only, no P2P ICE)
- `noopNetworkChangeListener` — ignores network change events
- `noopDnsReadyListener` — ignores DNS readiness notifications

The engine sees a fully populated `MobileDependency` and behaves exactly like the real Android app.

### Changes (3 files, no engine.go modifications)

- **`connect.go`**: Add `androidRunOverride` hook in `Run()` — checked before the default `c.run(MobileDependency{})` path
- **`connect_android_default.go`** (new): Default `init()` wires complete no-op stubs via `androidRunOverride`. Includes `noopIFaceDiscover`, `noopNetworkChangeListener`, `noopDnsReadyListener`.
- **`connect_android_embed.go`** (new): Unexported `runOnAndroidEmbed()` that accepts all mobile deps + `runningChan` + `logPath`, builds `MobileDependency`, calls `c.run()`

### Testing

Tested on Android arm64 (Pixel 7 Pro, Android 15):
- `embed.New()` + `Start()` succeeds
- NetBird engine starts with netstack TUN interface
- Relay connection established (WebSocket)
- Peers discovered, WireGuard handshake completes
- Localhost TCP+UDP proxy forwards through mesh

### Non-Breaking

- Default (non-Android) code path unchanged — `androidRunOverride` is nil, `Run()` behaves exactly as before
- Engine code untouched — stubs provide complete `MobileDependency`
- No new dependencies

N/A — new feature, no existing issue.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal engine change with no user-facing API modification. The embed package API is unchanged — existing callers work as before. Android support is automatic via build tags.

N/A

## Summary by CodeRabbit

* **New Features**
  * Improved Android support with a configurable embedded startup path for the client.
  * Added default/no-op Android implementations for network, DNS, and interface handling so embedded mode runs in netstack environments.

* **Tests**
  * Updated service params path test to be OS/path-separator aware by using path-joining.

### DIFF
--- a/client/cmd/service_params_test.go
+++ b/client/cmd/service_params_test.go
@@ -25,10 +25,10 @@ func TestServiceParamsPath(t *testing.T) {
 	t.Cleanup(func() { configs.StateDir = original })
 
 	configs.StateDir = "/var/lib/netbird"
-	assert.Equal(t, "/var/lib/netbird/service.json", serviceParamsPath())
+	assert.Equal(t, filepath.Join("/var/lib/netbird", "service.json"), serviceParamsPath())
 
 	configs.StateDir = "/custom/state"
-	assert.Equal(t, "/custom/state/service.json", serviceParamsPath())
+	assert.Equal(t, filepath.Join("/custom/state", "service.json"), serviceParamsPath())
 }
 
 func TestSaveAndLoadServiceParams(t *testing.T) {


client/internal/engine.go
```diff
@@ -1766,17 +1770,28 @@ func (e *Engine) newDnsServer(dnsConfig *nbdns.Config) (dns.Server, error) {
...
-			e.wgInterface,
-			e.mobileDep.HostDNSAddresses,
-			*dnsConfig,
-			e.mobileDep.NetworkChangeListener,
-			e.statusRecorder,
-			e.config.DisableDNS,
-		)
-		go e.mobileDep.DnsReadyListener.OnReady()
-		return dnsServer, nil
+		if e.mobileDep.NetworkChangeListener != nil {
```
pappz: noopNetworkChangeListener{} is non-nil, so engine.go:1770 (if e.mobileDep.NetworkChangeListener  != nil) will always take the full Android app path (NewDefaultServerPermanentUpstream), and the NewDefaultServer fallback below is dead code. Should this pass nil instead if embed is meant to  use the simpler DNS server?

client/internal/connect_android_embed.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/netbirdio/netbird/client/internal/listener"
+	"github.com/netbirdio/netbird/client/internal/stdnet"
+)
+
+// RunOnAndroidEmbed is like RunOnAndroid but accepts a runningChan
+// so embed.Client.Start() can detect when the engine is ready.
+// This enables embed.Client to work on Android with netstack mode.
+func (c *ConnectClient) RunOnAndroidEmbed(
```
pappz: Why do you export this function? Is it used by the `func (c *ConnectClient) Run(runningChan chan struct{}, logPath string) error` only, or not?

client/internal/connect_android_default.go
```diff
@@ -0,0 +1,51 @@
...
+var _ stdnet.ExternalIFaceDiscover = noopIFaceDiscover{}
+var _ listener.NetworkChangeListener = noopNetworkChangeListener{}
+
+func init() {
+	// Wire up the default override so embed.Client.Start() works on Android
+	// with netstack mode. Provides no-op stubs for mobile dependencies.
+	// Applications that need P2P ICE should replace this by setting
+	// androidRunOverride before calling Start().
+	androidRunOverride = func(c *ConnectClient, runningChan chan struct{}, logPath string) error {
+		return c.RunOnAndroidEmbed(noopIFaceDiscover{}, noopNetworkChangeListener{}, runningChan)
```
pappz: noopNetworkChangeListener{} is non-nil, so engine.go:1770 `if e.mobileDep.NetworkChangeListener  != nil` will always take the full Android app path (NewDefaultServerPermanentUpstream), and the  `NewDefaultServer` fallback below is dead code. Should this pass nil instead if embed is meant to use the simpler DNS server?

pappz:
@tham-le Instead of adding nil-check if/else branches in engine.go, provide complete no-op stubs in connect_android_default.go including a noopDnsReadyListener.

This way, the engine’s existing Android code path (NewDefaultServerPermanentUpstream + DnsReadyListener.OnReady()) works unchanged with embed mode, and no modifications to engine.go
are needed.

Also, logPath is received by androidRunOverride but is not forwarded to RunOnAndroidEmbed — it gets silently dropped.

client/internal/engine.go
tham-le: so I made a complete set of stubs so the engine's existing Android path works unchanged and revert all engine.go modifications

client/internal/connect_android_embed.go
tham-le: my bad, I'll unexport it to runOnAndroidEmbed. It's only called from androidRunOverride within this package.

client/internal/connect_android_default.go
tham-le: I'll add a noopDnsReadyListener stub and pass it through, so the engine's Android DNS path executes normally without modifications. Thx

PR #5625, opened 2026-03-18
fix(dns): skip probe for NetBird overlay addresses at engine startup

sbeaudry-qms:
## Describe your changes

When a nameserver is configured with a NetBird overlay IP (in the `100.64.0.0/10` CGNAT range), `ProbeAvailability()` is called at engine startup before the WireGuard tunnel to that peer is established. The probe always fails at this point — not because the nameserver is down, but because the tunnel isn't up yet. This produces a spurious "Unable to reach one or more DNS servers" warning in `netbird status -d` even though DNS resolves correctly once the connection settles.

This change skips the startup probe for any nameserver whose IP falls within the NetBird CGNAT range. A debug log message is emitted instead. These nameservers continue to be exercised normally by real DNS queries, and the `waitUntilResponse` retry loop still handles genuine failures.

Non-overlay nameservers (LAN IPs, public DNS) are unaffected and continue to be probed as before.

N/A — no existing issue. This addresses a known annoyance discussed in the NetBird community forum: https://forum.netbird.io/t/unable-to-reach-one-or-more-dns-servers/204

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation

This is a bug fix for a false-positive warning with no user-facing behaviour change. DNS resolution is unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Skip CGNAT overlay (100.64.0.0/10) addresses during startup probes to avoid false-positive DNS failures; skipping is logged and treated as passed.
  * Improve shutdown coordination and concurrency safety for probing and disabling to ensure clean stop behavior.
  * Aggregate probe errors and strengthen context-aware probing/cancellation for more reliable failure reporting.

* **Tests**
  * Added comprehensive tests covering overlay detection and probing behavior across LAN, public, mixed, and overlay-only scenarios.

### DIFF
--- a/client/internal/dns/upstream.go
+++ b/client/internal/dns/upstream.go
@@ -45,6 +45,19 @@ const (
 
 const testRecord = "com."
 
+// netbirdCGNATRange is the CGNAT range (100.64.0.0/10) used by NetBird for
+// overlay peer IP addresses. Addresses in this range are only reachable once
+// the WireGuard tunnel to the peer is fully established.
+var netbirdCGNATRange = netip.MustParsePrefix("100.64.0.0/10")
+
+// isNetBirdOverlayAddr returns true if the given address is within the NetBird
+// CGNAT range (100.64.0.0/10). Such addresses are only reachable after the
+// WireGuard tunnel is established, so probing them at engine startup will
+// always produce a false-positive "unable to reach" warning.
+func isNetBirdOverlayAddr(addr netip.AddrPort) bool {
+	return netbirdCGNATRange.Contains(addr.Addr())
+}
+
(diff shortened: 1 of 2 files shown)


sbeaudry-qms:
Regarding the CodeRabbit nitpick of the externalCtx parameter, it was pre-existing, not introduced by this PR, so I'm not addressing it.

Cheers,
...Steve...

lixmal:
Superseded by #5971, which takes a different approach: rather than skipping probes for overlay IPs, it drops the synthetic probe entirely and derives group health from real query outcomes. Thanks for putting this together — the problem statement in your description informed the design.

Issue #5551, opened 2026-03-09
NetBird v0.64.1+ crashes on kernels without iptable_raw module due to unconditional initNoTrackChain() call


thvevirtue:
Environment:
•  Ubiquiti UXG-Enterprise gateway
•  Kernel: 5.15.72-ui-cn9670 (ARM64, Marvell OcteonTX CN9670)
•  Debian 11 (bullseye)
•  Available iptables tables: filter, nat, mangle — no raw table
•  The iptable_raw kernel module is not shipped with this kernel (not on disk, can't be loaded)

Problem:

NetBird v0.66.2 fails to start with netbird firewall enabled:
create firewall manager: init firewall: init notrack chain: create chain: 
running [/usr/sbin/iptables -t raw -N NETBIRD-RAW --wait]: exit status 3: 
iptables v1.8.7 (legacy): can't initialize iptables table 'raw': Table does not exist

Root cause:

PR #5166 (d4f7df2, merged Jan 27, 2026) added initNoTrackChain() to manager_linux.go, which is called unconditionally during Init(). This creates a NETBIRD-RAW chain in the raw iptables table on every startup, regardless of whether the eBPF proxy is actually in use.

On kernels that don't ship the iptable_raw module, this is a fatal error. Combined with #5208 ("Ensured NetBird stops on firewall initialization failure"), the client cannot start at all with DisableFirewall: false.

Workaround:

Downgrading to v0.64.0 (released Jan 20, 2026, before #5166) resolves the issue. The firewall works correctly using only the filter, nat, and mangle tables.

Suggested fix:

initNoTrackChain() should either:
1. Be called lazily — only when SetupEBPFProxyNoTrack() is actually invoked, or
2. Gracefully degrade when the raw table is unavailable (log a warning, skip notrack setup), since the NOTRACK rules are only needed to prevent conntrack interference with eBPF proxy loopback traffic — not for core firewall functionality.

karthikteleo:
I'm hitting a variant of this on NVIDIA Jetson Orin devices (arm64, Ubuntu 22.04, kernel 5.15.136-tegra). Same root cause — no `iptable_raw` module — but a different symptom: instead of crashing, the client falls back to userspace ACL processing, leaving the iptables `NETBIRD-ACL-INPUT` chain empty. Since the chain structure (conntrack → NETBIRD-ACL-INPUT → DROP) is still created, all new inbound connections on wt0 are silently dropped.

[...]

Issue #2512, opened 2024-09-01
OpenWRT not able to login


Gauss23:
**Describe the problem**

Clean installation of gl-inet AR300M with OpenWrt 22.03.4 with their firmware 4.3.17. Installed Netbird `https://downloads.openwrt.org/snapshots/packages/mips_24kc/packages/netbird_0.28.9-r1_mips_24kc.ipk
`
Netbird up with setup key says:
`Error: login backoff cycle failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded
`
In the logfile I see:
```
2024-09-01T20:36:00Z ERRO client/server/server.go:262: failed login: context deadline exceeded
2024-09-01T20:36:00Z ERRO util/grpc/dialer.go:38: Failed to dial: dial: dial tcp ####:443: connect: invalid argument

```
Seems like a Go problem. Version 0.17 works for OpenWrt but is missing important features. Would be great to get the current version running. Looks like something changed between 0.17 and 0.28.9 with the go implementation to connect to the management server.

**To Reproduce**

1. Fresh firmware on gl-inet AR300M
2. Download netbird package from the snapshot tree
3. install netbird
4. run netbird up with setup-key
5. wait until it times out and read the above error msg

**Expected behavior**

It should just connect, like it does with 0.17

**Are you using NetBird Cloud?**

self-hosted, but it doesn't matter, I tried it also with the the cloud version, same error

**NetBird version**

0.28.9

**NetBird status -dA output:**

Daemon status: LoginFailed
Run UP command to log in with SSO (interactive login):
 netbird up 
If you are running a self-hosted version and no SSO provider has been configured in your Management Server,
you can use a setup-key:
 netbird up --management-url <YOUR_MANAGEMENT_URL> --setup-key <YOUR_SETUP_KEY>
More info: https://docs.netbird.io/how-to/register-machines-using-setup-keys

**Do you face any (non-mobile) client issues?**

Please provide the file created by `netbird debug for 1m -AS`.
We advise reviewing the anonymized files for any remaining PII.
Output:
Error: failed to up: up already in progress: current status LoginFailed

**Screenshots**

none

**Additional context**

none

Gauss23:
Update: I exchanged the original gl-inet firmware with a vanilla OpenWRT installation. The newest Netbird package (atm 0.28.9) works perfectly.

nazarewk:
FYI: this might be caused by using legacy `iptables` instead of new `nftables` on the system running Netbird.

[...]

Issue #2116, opened 2024-06-11
Netbird not working on Unifi after Firmware 4.0


Llamrei80:
Since the update to beta firmware 4.0 on the Unifi Dream Machine Pro Netbird doesn't connect anymore, it worked fine on firmware 3.2.12.

**Are you using NetBird Cloud?**

Yes

**NetBird version**

0.27.10

**NetBird status -d output:**

OS: linux/arm64
Daemon version: 0.27.10
CLI version: 0.27.10
Management: Connected to https://api.netbird.io:443
Signal: Connected to https://signal.netbird.io:443
Relays: 
  [stun:stun.netbird.io:5555] is Unavailable, reason: dial: failed to listen: dial: dial udp 3.73.3.142:5555: connect: invalid argument
  [turns:turn.netbird.io:443?transport=tcp] is Unavailable, reason: dial: dial: dial tcp 18.157.58.205:443: connect: invalid argument

nicdercole:
I have the same identical problem, but my version is stuck at 25.5. Any version higher than that does not allow me to connect to the Netbird cloud.

nicdercole:
Daemon version: 0.25.5
CLI version: 0.25.5
Management: Connected to https://api.netbird.io:443
Signal: Connected to https://signal.netbird.io:443
Relays: 
  [stun:stun.netbird.io:5555] is Available
  [turns:turn.netbird.io:443?transport=tcp] is Available
FQDN: *******.netbird.cloud
NetBird IP: ***.**.***.*/16
Interface type: Kernel
Peers count: 4/8 Connected


____________________________________


2024-07-09T13:35:29+02:00 WARN client/cmd/root.go:234: retrying Login to the Management service in 1.299560714s due to error rpc error: code = DeadlineExceeded desc = context deadline exceeded
Error: login backoff cycle failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded
root@****:/persistent/dpkg/bullseye/packages# netbird status -d
Daemon status: LoginFailed

Run UP command to log in with SSO (interactive login):

 netbird up 

If you are running a self-hosted version and no SSO provider has been configured in your Management Server,
you can use a setup-key:

 netbird up --management-url <YOUR_MANAGEMENT_URL> --setup-key <YOUR_SETUP_KEY>

More info: https://docs.netbird.io/how-to/register-machines-using-setup-keys

Llamrei80:
Just tried downgrading and testet a few versions:

[...]

PR #5653, opened 2026-03-23
fix: add netlink.RouteGet() fallback for ICE/P2P on devices with policy-based routing

thvevirtue:
## Problem

On devices using policy-based routing (e.g., Ubiquiti UniFi gateways), the default route is not in the `main` routing table but in a separate policy table (e.g., `201.eth4`). NetBird uses `go-netroute` for route lookups, which only reads the main table via `syscall.NetlinkRIB(RTM_GETROUTE, AF_UNSPEC)`. This causes two failures:

1. **`SharedSocket.WriteTo()`** (`sharedsock/sock_linux.go`) — calls `s.router.Route()` to determine source IP for raw UDP packets. Fails with `Route referenced unknown interface`.
2. **`GetNextHop()`** (`systemops_generic.go`) — called via write hooks when ICE sends packets. Same failure.

**Result:** All STUN and ICE connectivity-check packets are dropped. Local candidates are gathered and remote candidates are received, but no packets can be sent. ICE always times out and falls back to relay. Normal TCP traffic (management/signal/relay) works fine because standard Go `net.Dial` lets the kernel handle routing.

### Example routing setup (UXG-Max)
\\\
# ip route show table main
192.168.1.0/24 dev br0 proto kernel scope link src 192.168.1.1
(... only directly connected subnets, NO default route)

# ip route show table 201.eth4
default via <gateway> dev eth4 proto dhcp metric 200

# ip rule show
0:     from all lookup local
32000: from all lookup main
32766: from all lookup 201.eth4
32767: from all lookup default
\\\

### Key log lines
\\\
DEBG systemops_generic.go:348: Failed to get route for <IP>: Route referenced unknown interface
ice WARNING: Failed get server reflexive address: failed to send STUN packet: got an error while checking route, err: Route referenced unknown interface
ice INFO: Failed to send packet: got an error while checking route, err: Route referenced unknown interface
(repeated hundreds of times for every peer)
\\\

## Fix

Adds a `netlink.RouteGet()` fallback when `go-netroute`'s `Route()` fails. `netlink.RouteGet()` asks the kernel to perform the actual routing decision, respecting policy routing rules (`ip rule`) and all tables — exactly matching what happens for real packets.

### Changes
- **`sharedsock/sock_linux.go`** — `WriteTo()`: when `s.router.Route()` fails, fall back to `netlink.RouteGet()` for source IP resolution
- **`systemops_generic.go`** — `GetNextHop()`: same fallback pattern
- **`route_fallback_linux.go`** (new) — Linux-specific `getNextHopViaNetlink()` using `netlink.RouteGet()`
- **`route_fallback_other.go`** (new) — no-op stub for non-Linux platforms

### Safety
- Fallback only activates when `go-netroute` fails — devices with standard routing are completely unaffected
- `vishvananda/netlink` is already a dependency
- `netlink.RouteGet()` is the standard kernel route query mechanism on Linux

## Testing

**Tested and confirmed working** on Ubiquiti UXG-Max (kernel `5.4.213-ui-ipq5322`, ARM64):
- Before: all 5 peers `Connection type: Relayed`, ICE candidates `-/-`
- After: peers switch to `Connection type: P2P`

## Affected platforms
- Ubiquiti UXG-Max, UXG-Enterprise, UDM Pro (firmware 4.0+)
- Likely all UniFi OS gateway devices
- OpenWRT devices with `mwan3` or custom policy routing
- Any Linux device where the default route is in a non-main routing table

## Related issues
- #5551 — Separate: `iptable_raw` crash on kernels without the module
- #2530 — Teltonika RutOS, possibly same underlying cause
- #2512 — OpenWRT, `connect: invalid argument`
- #2116 — UniFi UDM Pro firmware 4.0

Co-Authored-By: Oz <oz-agent@warp.dev>

## Release Notes

* **Improvements**
  * Enhanced routing system with improved fallback mechanisms for more reliable network connectivity on Linux
  * Improved source IP address determination when routing network packets
  * Better cross-platform compatibility with graceful error handling when standard route lookups fail
  * More robust network packet handling with enhanced source address selection

### DIFF
--- a/client/internal/routemanager/systemops/route_fallback_linux.go
+++ b/client/internal/routemanager/systemops/route_fallback_linux.go
@@ -1,0 +1,58 @@
+//go:build linux && !android
+
+package systemops
+
+import (
+	"fmt"
+	"net"
+	"net/netip"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/vishvananda/netlink"
+)
+
+// getNextHopViaNetlink uses netlink.RouteGet() to determine the next hop for a destination IP.
+// Unlike go-netroute which only reads the main routing table, netlink.RouteGet() asks the kernel
+// to perform an actual routing decision, respecting policy routing rules (ip rule) and all tables.
+// This is needed on devices like UniFi gateways where the default route lives in a separate
+// policy routing table (e.g., 201.eth4) rather than in the main table.
+func getNextHopViaNetlink(ip netip.Addr) (Nexthop, error) {
+	dst := ip.AsSlice()
+	routes, err := netlink.RouteGet(net.IP(dst))
+	if err != nil {
+		return Nexthop{}, fmt.Errorf("netlink.RouteGet(%s): %w", ip, err)
+	}
+	if len(routes) == 0 {
+		return Nexthop{}, fmt.Errorf("no route to %s via netlink", ip)
+	}
+
+	route := routes[0]
+
+	var intf *net.Interface
+	if route.LinkIndex > 0 {
+		intf, err = net.InterfaceByIndex(route.LinkIndex)
+		if err != nil {
+			log.Debugf("Failed to get interface for index %d: %v", route.LinkIndex, err)
+		}
+	}
+
+	if route.Gw != nil {
+		addr, err := ipToAddr(route.Gw, intf)
+		if err != nil {
+			return Nexthop{}, fmt.Errorf("convert gateway to address: %w", err)
+		}
+		log.Debugf("Policy routing fallback: route to %s via gw %s (table %d, iface %v)", ip, route.Gw, route.Table, intf)
+		return Nexthop{IP: addr, Intf: intf}, nil
+	}
+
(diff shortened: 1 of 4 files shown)


lixmal:
Thanks for digging into this and the detailed write-up!

I think this is superseded by #6062 (bump go-netroute to v0.4.0 and drop our fork). v0.2.1's Linux backend dumps `RTM_GETROUTE` against the main table only, which is the root cause you hit on UniFi/OpenWRT. v0.4.0 rewrote `netroute_linux.go` to send a per-destination `RTM_GETROUTE` with `Table: RT_TABLE_UNSPEC`, so the kernel performs the real routing decision and honors `ip rule` plus all policy tables: functionally equivalent to the `netlink.RouteGet()` fallback you added, but at the library level.

If you can, it would be great if you could test #6062 on the UXG-Max to confirm peers go P2P without the fallback. If that works, we can close this in favor of the bump.

Issue #4542, opened 2025-09-25
Disabling Masquerade on Routed network resources does not work.


pquan:
**Describe the problem**

When exposing network resources via routing peers (using the new Networks feature) AND disabling masquerading, the network resources will still see the routing peer IP on IP traffic and not the remote client IP in the private 100.64.x.x space.

**To Reproduce**

My setup: PfSense firewalls running Netbird 0.55.1 (this is currently the latest compatible version with the latest 2.8.1 production pfsense).
My clients are windows, running 0.58.1 (latest).

The Firewalls have DMZs and LANs behind them (different subnets, 4 networks total). 

Now connect from a client to the network and PING a device on the LAN or DMZ exposed by the firewalls. The device will see the IP coming from the firewall and not from the static IP of the peer.

Note on Natting: Natting not enabled on any but the public IP interfaces of the firewalls. However, PFsense is set up to `Manual Outbound NAT rule generation. (AON - Advanced Outbound NAT)`. In my understanding this setting should only install the NAT rules present on the Nat rules page. There I only have nat rules on the external interfaces and nothing on the internal DMZ and specially not on the LAN. As a proof , if I ping from any other internal resource on these firewalls (which are interconnected via another transport network) I see pure local IP addresses.

**Expected behavior**

With masquerading disabled, the firewalls should not masquerade the source IP address. The packed should be untouched.

**Are you using NetBird Cloud?**

No. I'm using self-hosted.

**NetBird version**

- Latest docker pulled on the admin.
- pfsense (freebsd) firewalls running 2.8.1 PfSense and 0.55.1 Netbird (latest compatible and available).

**Is any other VPN software installed?**

OpenVPN, irrelevant in this setup (no routes etc). 

**Debug output**
n/a
**Screenshots**
n/a
**Additional context**
n/a
**Have you tried these troubleshooting steps?**
- [x] Reviewed [client troubleshooting](https://docs.netbird.io/how-to/troubleshooting-client) (if applicable)
- [x] Checked for newer NetBird versions
- [x] Searched for similar issues on GitHub (including closed ones)
- [x] Restarted the NetBird client
- [x] Disabled other VPN software
- [x] Checked firewall settings

lixmal:
NetBird's userspace routing doesn't support routing without masquerading. You can tick the `Disable firewall` box to leave routing/filtering/NAT to the operating system

pquan:
[...]

PR #5682, opened 2026-03-24
[client] Fix exit node masquerade overrides per-route masquerade=false setting

### LINKED ISSUE #2751: Disabling Masquerding not working
**Describe the problem**

I run a node with multiple advertised routes that were masqueraded using iptabels with an client installed on Rocky 8.
Setting up a routing back into netbird for the 100.24.0.0 adresses works as expected as far as being able to ping Netbird clients from non Netbird clients.

This is not always true for Netbird clients, we can ping through the node with iptables but everything else like ssh or http is not working and stops at the routing node. Using nftables results in not being able to ping all together.
I reproduced this for Debian and a Docker client.

**To Reproduce**

1. Set a route for a node to advertise a subnet
2. Disable masquerading (its important to have no exit node on this node as this enables masqerading for all routes on this node, even when disabled)
3. Testing ping should not work
4. Enable iptabels with `echo 'NB_SKIP_NFTABLES_CHECK=true' | sudo tee -a /etc/sysconfig/netbird` and restarting netbird
5. Ping should now work but nothing else

We can see from iptables that currently no route advertises masquerading 

```
iptables -t nat -S

-P PREROUTING ACCEPT
-P INPUT ACCEPT
-P POSTROUTING ACCEPT
-P OUTPUT ACCEPT
-N NETBIRD-RT-NAT
-A POSTROUTING -j NETBIRD-RT-NAT

```
And ping is working but no ssh or http
```
ping 10.1.0.1

Ping wird ausgeführt für 10.1.0.1 mit 32 Bytes Daten:
Antwort von 10.1.0.1: Bytes=32 Zeit=38ms TTL=63
```

For the ease of configuring masquerading i enable an exit node route which results in following ouput
```
iptables -t nat -S

-P PREROUTING ACCEPT
-P INPUT ACCEPT
-P POSTROUTING ACCEPT
-P OUTPUT ACCEPT
-N NETBIRD-RT-NAT
-A POSTROUTING -j NETBIRD-RT-NAT
-A NETBIRD-RT-NAT -i wt0 -j MASQUERADE
-A NETBIRD-RT-NAT -o wt0 -j MASQUERADE
```

And fixing the issues by being able to connect again with the network clients.

Self-hosted NetBird

**NetBird version**

`0.30.1`

### LINKED ISSUE #5683: Exit node masquerade overrides per-route masquerade=false setting
## Description

When a peer is configured as both an **exit node** (`0.0.0.0/0`) and a **routing peer** for specific subnets with masquerade disabled, the exit node's postrouting masquerade fires for all traffic — the per-route `masquerade=false` setting is silently ignored.

## Steps to reproduce

1. Peer B is a routing peer for subnet `192.168.0.0/24` with **masquerade=OFF**
2. Peer A is also an exit node (`0.0.0.0/0`) — masquerade cannot be disabled for exit nodes
3. From a client routed through peer A, send traffic to a host in `192.168.0.0/24`

**Expected:** traffic arrives at destination with client's NetBird IP as source  
**Actual:** traffic arrives masqueraded — source is peer B's WireGuard interface IP

## Root cause

`addNatRule()` sets fwmark `PreroutingFwmarkMasquerade` (`0x1BD21`) on **all** `wt0` traffic when adding the exit node route. `addPostroutingRules()` masquerades everything carrying that mark. There is no mechanism for a per-route `masquerade=false` to opt out of this blanket rule — the chain order means the exit node's rule always wins.

## Impact

Any service relying on the original source IP is broken for clients behind an exit node:
- Firewall rules matching the client's NetBird IP
- IPsec policies with `src-address` matching
- Per-device ACLs on the destination network

## Environment

- NetBird client (Linux, nftables and iptables backends both affected)
- Exit node and per-subnet routing used simultaneously

## Related

- #2751 (closed, same root cause reported)
- #4542 (open, related)
- Fix proposed in PR #5682

tobsec:
## Problem

When a peer acts as both an **exit node** (`0.0.0.0/0`) and a **routing peer** for specific subnets with masquerade disabled, the exit node's postrouting masquerade rule fires for all traffic — silently ignoring the per-route `masquerade=false` setting.

**Root cause:** `addNatRule()` sets fwmark `PreroutingFwmarkMasquerade` on all `wt0` traffic for the exit node. `addPostroutingRules()` then masquerades everything carrying that mark, regardless of per-route settings. There is no mechanism for a per-route `masquerade=false` to opt out of this blanket rule.

**Practical effect:** A client routing through an exit node cannot reach services that rely on the original source IP being preserved (e.g. firewall rules matching the client's NetBird IP, IPsec policies, per-device ACLs). All traffic appears to arrive from the routing peer's WireGuard interface IP instead.

This was reported in #2751 and is related to #4542.

## Fix

When `masquerade=false` for a route pair, insert a `RETURN` verdict at the **head** of the postrouting NAT chain (`netbird-rt-postrouting`) matching the route's destination prefix. This short-circuits evaluation before the exit node's catch-all masquerade rule fires.

Both the **nftables** and **iptables** backends are updated:

- New constant `NoMasqPostRoutingFormat` in `firewall/manager/firewall.go`
- New `addNoMasqPostRoutingRule()` in both `nftables/router_linux.go` and `iptables/router_linux.go`
- `AddNatRule()` / `RemoveNatRule()` updated to manage the return rule when `masquerade=false`

The rule is keyed and tracked in `r.rules` the same way as existing NAT rules, so it is correctly removed on route deletion or peer disconnect.

## Test scenario

1. Configure peer A as exit node (`0.0.0.0/0`, masquerade cannot be disabled)
2. Configure peer B as routing peer for `192.168.0.0/24` with masquerade=OFF
3. From a client behind the exit node, send traffic to `192.168.0.1`
4. **Before fix:** traffic arrives at `192.168.0.1` with src = peer B's WireGuard IP (masqueraded)
5. **After fix:** traffic arrives with the client's original NetBird IP preserved

## Related issues

- Closes #5683
- Closes #2751
- Related: #4542

## Summary by CodeRabbit

* **New Features**
  * Added destination-scoped "no-masquerade" postrouting rules for both directions to provide finer NAT control across Linux routing backends.

* **Bug Fixes**
  * Improved insertion ordering, deduplication, rollback, and cleanup of no-masquerade postrouting rules.
  * Graceful handling and removal of missing or stale rule entries to prevent leftover state.

### DIFF
--- a/client/firewall/iptables/router_linux.go
+++ b/client/firewall/iptables/router_linux.go
@@ -529,6 +529,9 @@ func (r *router) addNoMasqPostRoutingRule(pair firewall.RouterPair) error {
 		if err := r.iptablesClient.DeleteIfExists(tableNat, chainRTNAT, rule...); err != nil {
 			return fmt.Errorf("remove existing no-masq rule before reinstall: %w", err)
 		}
+		if err := r.decrementSetCounter(rule); err != nil {
+			log.Warnf("decrement set counter for reinstalled rule %s: %v", ruleKey, err)
+		}
 		delete(r.rules, ruleKey)
 	}
 
--- a/client/firewall/iptables/router_linux.go
+++ b/client/firewall/iptables/router_linux.go
@@ -259,16 +259,21 @@ func (r *router) AddNatRule(pair firewall.RouterPair) error {
 		}
 	}
 
-	if !pair.Masquerade {
-		return nil
-	}
+	if pair.Masquerade {
+		if err := r.addNatRule(pair); err != nil {
+			return fmt.Errorf("add nat rule: %w", err)
+		}
 
-	if err := r.addNatRule(pair); err != nil {
-		return fmt.Errorf("add nat rule: %w", err)
-	}
-
-	if err := r.addNatRule(firewall.GetInversePair(pair)); err != nil {
-		return fmt.Errorf("add inverse nat rule: %w", err)
+		if err := r.addNatRule(firewall.GetInversePair(pair)); err != nil {
+			return fmt.Errorf("add inverse nat rule: %w", err)
+		}
+	} else {
+		// Insert a RETURN rule at the head of the postrouting NAT chain for this
+		// destination, preventing the exit node's catch-all masquerade from
+		// rewriting the source IP for routes with masquerade disabled.
+		if err := r.addNoMasqPostRoutingRule(pair); err != nil {
+			return fmt.Errorf("add no-masquerade postrouting rule: %w", err)
+		}
 	}
 
 	r.updateState()
@@ -285,6 +290,19 @@ func (r *router) RemoveNatRule(pair firewall.RouterPair) error {
 
 		if err := r.removeNatRule(firewall.GetInversePair(pair)); err != nil {
 			return fmt.Errorf("remove inverse nat rule: %w", err)
+		}
+	} else {
+		ruleKey := firewall.GenKey(firewall.NoMasqPostRoutingFormat, pair)
+		if rule, exists := r.rules[ruleKey]; exists {
+			if err := r.iptablesClient.DeleteIfExists(tableNat, chainRTNAT, rule...); err != nil {
+				return fmt.Errorf("remove no-masquerade return rule: %w", err)
+			}
+			delete(r.rules, ruleKey)
+			if err := r.decrementSetCounter(rule); err != nil {
+				return fmt.Errorf("decrement ipset counter: %w", err)
+			}
+		} else {
+			log.Debugf("no-masquerade postrouting rule %s not found", ruleKey)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


tobsec:
Both points addressed in d007a62:

- `rollbackRules` now includes `NoMasqPostRoutingFormat` in its key list so the no-masq RETURN rule is rolled back on flush failure
- `RemoveNatRule` no-masq path now checks `rule.Handle == 0` and handles stale entries the same way `removeNatRule` does

lixmal:
Good feature, but the PR has 2 flaws:

1. The reverse direction is not applied (routed network -> netbird network)
2. Nat exemption breaks more specific prefixes. Consider this example:

  | Route | Masquerade | Rule type |
  |-------|------------|-----------|
  | `10.0.0.0/24` | on | prerouting mark |
  | `10.0.0.0/16` | off | RETURN in postrouting |
  | `10.0.0.0/8` | on | prerouting mark |

The more specific 10/24 would be swallowed by the exemption of 10/16

tobsec:
Thanks for the review, @lixmal — both points are valid.

**Issue 1 (inverse direction):** Agreed. The masquerade=ON path calls `addNatRule` for both the pair and `GetInversePair`, but our masquerade=OFF path only adds the RETURN rule for the forward direction. Easy fix.

**Issue 2 (prefix overlap):** This is the trickier one. The root cause is that masquerade=ON routes currently only add a prerouting mark rule — there is no corresponding postrouting rule per destination. The blanket postrouting masquerade fires for anything with the mark. Because of that, a RETURN in postrouting for a less-specific masquerade=OFF prefix will shadow more-specific masquerade=ON prefixes, regardless of insertion order.

The correct fix seems to be: masquerade=ON routes should also get explicit per-destination postrouting rules (mark + destination → masquerade), and all postrouting rules should be ordered most-specific-first so the longest prefix match wins. This means changing the masquerade=ON path as well — a bigger refactor than the current PR.

Before going down that path, does the project have a preferred approach for this? For example:
- Explicit per-route postrouting rules sorted by prefix length
- A nftables verdict map with longest-prefix-match (cleaner but more invasive)
- Something else entirely

Happy to rework the implementation once there's consensus on the right direction.

tobsec:
Docs change: not needed

tobsec:
**Issue 1 (inverse direction):** Addressed in 4754f31. `addNoMasqPostRoutingRule` now selects the correct fwmark based on `pair.Inverse` (`0x1BD21` for forward, `0x1BD22` for inverse), and `AddNatRule`/`RemoveNatRule`/`rollbackRules` call the helper for both the forward and inverse pair — matching the pattern used by the masquerade=ON path.

**Issue 2 (prefix overlap):** Agreed this is a real concern. The root cause is that masquerade=ON only adds a prerouting mark but no per-destination postrouting rule, so there's nothing more-specific to override the less-specific RETURN. A proper fix likely requires explicit per-route postrouting masquerade rules ordered by prefix length (most-specific first). Happy to take direction on the preferred approach — should this be addressed in this PR or as a follow-up?

joachimtingvold:
Stumbled upon this today via #2751 and #5683.

Seems like this PR has stopped due to lack of feedback and/or path forward from developers?

Regarding "issue 2" described above (prefix overlap), I would prefer to have a working setup for non-overlapping prefixes, which is better than the "masquerade everything if exit-node enabled"-method we have currently... if "issue 2" requires a major rewrite, could we implement the other fixes while waiting for that clarification/rewrite?

PR #5686, opened 2026-03-25
[client] Add IPv6 reverse DNS and host configurator support

lixmal:
## Describe your changes

- Add IPv6 reverse DNS (PTR) support using RFC 3596 nibble format for `ip6.arpa` zones, alongside existing `in-addr.arpa` for v4
- Fix DNS host configurators on Linux (systemd, NetworkManager) and iOS to handle IPv6 server addresses and socket binding
- Add tests for PTR generation, reverse zone naming, and mixed v4/v6 record collection

## Stack

- #5631
  - #5675
    - #5686 :point_left:
      - #5687
        - #5688
          - #5706
            - #5707
              - #5708

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

https://github.com/netbirdio/docs/pull/667

### DIFF
--- a/client/internal/engine_ssh.go
+++ b/client/internal/engine_ssh.go
@@ -41,6 +41,14 @@ func (e *Engine) setupSSHPortRedirection() error {
 	}
 	log.Infof("SSH port redirection enabled: %s:22 -> %s:22022", localAddr, localAddr)
 
+	if v6 := e.wgInterface.Address().IPv6; v6.IsValid() {
+		if err := e.firewall.AddInboundDNAT(v6, firewallManager.ProtocolTCP, 22, 22022); err != nil {
+			log.Warnf("failed to add IPv6 SSH port redirection: %v", err)
+		} else {
+			log.Infof("SSH port redirection enabled: [%s]:22 -> [%s]:22022", v6, v6)
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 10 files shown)


lixmal:
@coderabbitai review

client/internal/dns/upstream_ios.go
```diff
@@ -66,12 +70,23 @@ func (u *upstreamResolverIOS) exchange(ctx context.Context, upstream string, r *
...
-			return nil, 0, fmt.Errorf("create private client: %s", err)
+		var bindIP netip.Addr
+		switch {
+		case upstreamIP.Is6() && u.lIPv6.IsValid():
+			bindIP = u.lIPv6
+		case upstreamIP.Is4() && u.lIP.IsValid():
+			bindIP = u.lIP
+		}
+
+		if bindIP.IsValid() {
```
pappz, line 84: If have validation issue with lIPv6 or lIP then bindIP will be invalid.
The old code always used the private client for matching upstreams. This is a behavior change. If this i the expected then fine.

lixmal: Yes, this is a subtle intentional change. If neither lIP nor lIPv6 is valid for the upstream's address family, we fall through to the default (non-private) client instead of binding to an invalid address. In practice, if needsPrivate is true but the corresponding local IP is invalid, the configuration is broken and there's nothing useful to bind to.

PR #5691, opened 2026-03-25
[client] Fix Exit Node submenu separator accumulation on Windows

### LINKED ISSUE #4702: Bug in exit node menu
Hello!
There is some bug in Exit node menu
If change Exit node in meny every time there come new line ...

Image

I have latest V 0.59.8

tobsec:
## Problem

Fixes #4702

On Windows, the tray app uses a 10-second background poller to refresh the Exit Node menu (because `TrayOpenedCh` is not supported on Windows). Every poll cycle that finds a selected exit node calls `s.mExitNode.AddSeparator()` before adding the "Deselect All" item.

`AddSeparator()` returns no handle, so the separator is never removed in the cleanup pass of `recreateExitNodeMenu()`. All other items — exit node checkboxes and "Deselect All" — are properly tracked and removed each cycle. The separator is not.

After the client has been running for a while with an exit node selected, hundreds of separator lines stack up in the submenu, filling the full screen height with blank entries.

On Linux/FreeBSD this doesn't manifest because the parent `mExitNode` item itself is removed and recreated each cycle (to work around a different systray limitation), which wipes all children including the orphaned separators.

## Fix

Replace the untracked `AddSeparator()` call with a regular disabled sub-menu item. This gives us a handle that can be stored in `mExitNodeSeparator` and removed at the start of each `recreateExitNodeMenu()` call, alongside the existing `mExitNodeDeselectAll` cleanup.

```go
// Before
s.mExitNode.AddSeparator()

// After
sep := s.mExitNode.AddSubMenuItem("───────────────", "")
sep.Disable()
s.mExitNodeSeparator = sep
```

The visual result is equivalent on Windows (a greyed-out, unclickable divider row), and the item is now properly cleaned up on every menu refresh.

## Test

Run the Windows client with an exit node selected and leave it running for several minutes. Open the Exit Node submenu — previously it would grow with each poll; with this fix the menu stays clean regardless of uptime.

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Refactor**
  * Improved exit-node menu behavior: the separator now persists correctly across menu updates and the "Deselect All" option has been reorganized for more reliable presentation and interaction, reducing visual flicker and improving menu stability.

### DIFF
--- a/client/ui/network.go
+++ b/client/ui/network.go
@@ -457,31 +457,35 @@ func (s *serviceClient) recreateExitNodeMenu(exitNodes []*proto.Network) {
 	}
 
 	if showDeselectAll {
-		sep := s.mExitNode.AddSubMenuItem("───────────────", "")
-		sep.Disable()
-		s.mExitNodeSeparator = sep
-		deselectAllItem := s.mExitNode.AddSubMenuItem("Deselect All", "Deselect All")
-		s.mExitNodeDeselectAll = deselectAllItem
-		go func() {
-			for {
-				_, ok := <-deselectAllItem.ClickedCh
-				if !ok {
-					// channel closed: exit the goroutine
-					return
-				}
-				exitNodes, err := s.handleExitNodeMenuDeselectAll()
-				if err != nil {
-					log.Warnf("failed to handle deselect all exit nodes: %v", err)
-				} else {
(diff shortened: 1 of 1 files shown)


mlsmaycon:
Thanks for the contribution @tobsec

PR #5694, opened 2026-03-25
[client] Replace iOS DNS IsPrivate heuristic with route checker

lixmal:
## Describe your changes

- Replace `IsPrivate()` heuristic in iOS upstream DNS resolver with a route checker function that queries actual client routes from the route manager
- Add `SetRouteChecker` to the DNS `Server` interface, wired up from the engine after route manager creation

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Implemented route-aware DNS resolution that enables intelligent routing decisions based on client-configured network routes. The system can now match upstream DNS resolvers against configured route networks, replacing purely network-type-based decisions with route-aware logic. This improves DNS query handling in split-DNS and complex multi-route network configurations.

### DIFF
--- a/client/internal/dns/server.go
+++ b/client/internal/dns/server.go
@@ -57,6 +57,7 @@ type Server interface {
 	ProbeAvailability()
 	UpdateServerConfig(domains dnsconfig.ServerDomains) error
 	PopulateManagementDomain(mgmtURL *url.URL) error
+	SetRouteChecker(func(netip.Addr) bool)
 }
 
 type nsGroupsByDomain struct {
@@ -104,6 +105,7 @@ type DefaultServer struct {
 
 	statusRecorder *peer.Status
 	stateManager   *statemanager.Manager
+	routeMatch     func(netip.Addr) bool
 
 	probeMu     sync.Mutex
 	probeCancel context.CancelFunc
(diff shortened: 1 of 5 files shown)


lixmal:
@coderabbitai review

client/internal/dns/server.go
```diff
@@ -743,6 +753,7 @@ func (s *DefaultServer) registerFallback(config HostDNSConfig) {
 		log.Errorf("failed to create upstream resolver for original nameservers: %v", err)
 		return
 	}
+	handler.routeMatch = s.routeMatch
```
mlsmaycon, line 756: Should be wrapped by a lock/unlock?

client/internal/dns/server.go
```diff
@@ -852,6 +863,7 @@ func (s *DefaultServer) createHandlersForDomainGroup(domainGroup nsGroupsByDomai
 		if err != nil {
 			return nil, fmt.Errorf("create upstream resolver: %v", err)
 		}
+		handler.routeMatch = s.routeMatch
```
mlsmaycon, line 866: Same here

client/internal/dns/server.go
```diff
@@ -1036,6 +1048,7 @@ func (s *DefaultServer) addHostRootZone() {
 		log.Errorf("unable to create a new upstream resolver, error: %v", err)
 		return
 	}
+	handler.routeMatch = s.routeMatch
```
mlsmaycon, line 1051: And here

client/internal/dns/server.go
lixmal: Not with the current code, it's set in the init phase long before any reads and it's never changed again

PR #5695, opened 2026-03-25
[client] Add Expose support to embed library

pappz:
## Describe your changes

Add Expose API to the embed library, allowing programmatic creation and
management of service exposures (HTTP, HTTPS, TCP, UDP, TLS) from embedded    
client code.
                                                                              
Introduces ExposeSession with a blocking Wait method that keeps the exposure  
alive until the context is cancelled. Extracts ProtocolType into a shared
expose package for reuse across CLI and embed layers.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Expose local services publicly with support for HTTP, HTTPS, TCP, UDP, and TLS.
  * Client API to create and manage exposures, returning sessions with domain, service name, and service URL.
  * Session "wait" to keep exposures alive until explicitly stopped.

* **Bug Fixes / Improvements**
  * Improved protocol parsing using a typed protocol enum and clearer validation errors for unsupported protocol values.

### DIFF
--- a/client/embed/expose.go
+++ b/client/embed/expose.go
@@ -2,6 +2,7 @@ package embed
 
 import (
 	"context"
+	"errors"
 
 	"github.com/netbirdio/netbird/client/internal/expose"
 )


client/embed/expose.go
```diff
@@ -0,0 +1,42 @@
...
+type ExposeProtocolType = expose.ProtocolType
+
+// ExposeSession represents an active expose session. Use Wait to block until the session ends.
+type ExposeSession struct {
+	Domain      string
+	ServiceName string
+	ServiceURL  string
+
+	mgr *expose.Manager
+	ctx context.Context
```
lixmal: `Wait` should take its own `ctx context.Context`  instead of storing it

client/internal/expose/manager.go
```diff
@@ -22,7 +25,7 @@ type Request struct {
 	NamePrefix string
 	Domain     string
 	Port       uint16
-	Protocol   int
+	Protocol   ProtocolType
```
lixmal, line 30: `Request` is exported and part of the embed API surface (via the type alias), but it has no doc comment. Same for `KeepAlive` further down.

client/cmd/expose.go
```diff
@@ -211,16 +212,21 @@ func exposeFn(cmd *cobra.Command, args []string) error {
...
-	case "tcp":
+	case expose.ProtocolTCP:
 		return proto.ExposeProtocol_EXPOSE_TCP, nil
-	case "udp":
+	case expose.ProtocolUDP:
 		return proto.ExposeProtocol_EXPOSE_UDP, nil
-	case "tls":
+	case expose.ProtocolTLS:
 		return proto.ExposeProtocol_EXPOSE_TLS, nil
 	default:
```
lixmal, line 231: This default is now unreachable since `ParseProtocolType` already rejects unknown protocols. We can change the message to something like `"unhandled protocol type: %d"` so it's useful when someone adds a new protocol to `ProtocolType` without updating this switch.

client/embed/expose.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Wait blocks while keeping the expose session alive.
+// It returns when ctx is cancelled or a keep-alive error occurs, then terminates the session.
+func (s *ExposeSession) Wait(ctx context.Context) error {
+	if s == nil || s.mgr == nil {
+		return errors.New("expose session is not initialized")
+	}
+	return s.mgr.KeepAlive(ctx, s.Domain)
+}
```
lixmal, line 45: Build breaks here


```suggestion
```

lixmal (approved):

PR #5697, opened 2026-03-25
[client] Guard against container DNAT bypass of ACL rules in iptables

lixmal:
## Describe your changes

When container runtimes (Docker, Podman) publish ports, they DNAT traffic in nat PREROUTING, causing it to traverse the FORWARD chain instead of INPUT. If the container runtime starts after NetBird, its ACCEPT rules in filter FORWARD are inserted above NetBird's rules, bypassing ACL port restrictions.

- Add two guard rules in mangle FORWARD that enforce the ACL mark check for DNAT'd traffic from the WireGuard interface
- Mangle table runs before filter, so container runtimes cannot override the verdict regardless of rule insertion order
- Uses `--ctstate DNAT` to scope the guard to only DNAT'd traffic, leaving regular route-forwarded traffic unaffected
- Rules are persisted in shutdown state for proper cleanup

Native nftables is not affected since it uses separate tables with independent chain priorities.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

No user-facing behavior change. The fix is internal to the iptables firewall manager.

## Release Notes

* **Bug Fixes**
  * Improved firewall forwarding chain management for better connection tracking and traffic handling.
  * Enhanced security enforcement to prevent external traffic from circumventing ACL checks.


pappz (approved):

SISheogorath:
Thanks... This allowed me to spend my whole evening debugging why I can't access an internal service.

lixmal:
Please open a GitHub issue. There's no point doing that in this PR

PR #5780, opened 2026-04-02
[client] Don't abort UI debug bundle when up/down fails

lixmal:
## Describe your changes

Mirror the CLI fix from #5657 for the UI debug bundle: when Up/Down/SetLogLevel/SetSyncResponsePersistence calls fail during debug collection, log a warning and continue instead of aborting the entire bundle.

- Log warnings instead of returning errors in `configureServiceForDebug`, so the bundle is still generated even if service state changes fail
- Downgrade `restoreServiceState` error logging from `Errorf` to `Warnf` for consistency

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Bug Fixes**
  * Debug data collection now continues even if initial setup calls fail; failures are logged without blocking the process.
  * Timing/waiting delays during debug setup now occur only after successful operations to avoid unnecessary waits.

* **Chores**
  * Reduced log severity for restore-related failures from error to warning.
  * Added conditional tracking to ensure the service is restored to its prior "up" state when appropriate.


client/ui/debug.go
```diff
@@ -371,46 +372,51 @@ func (s *serviceClient) configureServiceForDebug(
...
 	enablePersistence bool,
-) error {
+) {
 	if state.wasDown {
 		if _, err := conn.Up(s.ctx, &proto.UpRequest{}); err != nil {
-			return fmt.Errorf("bring service up: %v", err)
+			log.Warnf("failed to bring service up: %v", err)
+		} else {
+			log.Info("Service brought up for debug")
+			time.Sleep(time.Second * 10)
```
pappz, line 381: Is not it block the UI?

client/ui/debug.go
```diff
@@ -371,46 +372,51 @@ func (s *serviceClient) configureServiceForDebug(
...
 		}
-		log.Info("Log level set to TRACE for debug")
 	}
 
 	if _, err := conn.Down(s.ctx, &proto.DownRequest{}); err != nil {
-		return fmt.Errorf("bring service down: %v", err)
+		log.Warnf("failed to bring service down: %v", err)
+	} else {
+		state.needsRestoreUp = !state.wasDown
+		time.Sleep(time.Second)
```
pappz, line 397: Is not it block the UI?

client/ui/debug.go
```diff
@@ -371,46 +372,51 @@ func (s *serviceClient) configureServiceForDebug(
...
 		}
-		log.Info("Sync response persistence enabled for debug")
 	}
 
 	if _, err := conn.Up(s.ctx, &proto.UpRequest{}); err != nil {
-		return fmt.Errorf("bring service back up: %v", err)
+		log.Warnf("failed to bring service back up: %v", err)
+	} else {
+		state.needsRestoreUp = false
+		time.Sleep(time.Second * 3)
```
pappz, line 414: Is not it block the UI?

client/ui/debug.go
lixmal: It's running in a goroutine, and it's just moved. The sleep was already there

client/ui/debug.go
lixmal: It's running in a goroutine, and it's just moved. The sleep was already there

client/ui/debug.go
lixmal: It's running in a goroutine, and it's just moved. The sleep was already there

PR #5794, opened 2026-04-03
[client] Add TTL-based cache refresh for management DNS resolver

### LINKED ISSUE #5113: Management DNS cache not auto update
**Describe the problem**

My NetBird management server is running on a machine with a non-static public IP address. I use DDNS to keep a domain name synchronized with the server’s current IP.

However, I observed that the NetBird client appears to cache the DNS resolution result of the management domain and does not automatically refresh it. When the public IP of the management server changes and the DDNS record is updated accordingly, the NetBird client continues to use the old IP address.

As a result, the client fails to reconnect to the management server after the IP change, even though the domain name correctly resolves to the new IP when queried externally (e.g., via dig or nslookup).

The client only reconnects successfully after being restarted or after some manual intervention that forces DNS re-resolution.

The domain cache start from: https://github.com/netbirdio/netbird/pull/4128

**To Reproduce**

just dig the management dns will found that:

```
2026-01-16T03:36:59Z TRAC client/internal/dns/handler_chain.go:167: DNS request domain=netbird.xxx.com., handlers (8):
2026-01-16T03:36:59Z TRAC client/internal/dns/handler_chain.go:175: handler matched: domain=netbird.plusls.com. -> pattern=netbird.xxx.com. wildcard=false match_subdomain=false priority=150
2026-01-16T03:36:59Z DEBG client/internal/dns/mgmt/mgmt.go:78: serving 1 cached records for domain=netbird.xxx.com.
```

**Expected behavior**

```
The NetBird client should periodically re-resolve the management server domain name, or at least re-resolve it on reconnection attempts, so that DDNS-based deployments continue to work correctly when the management server’s IP changes.
```

self-host 

**NetBird version**

0.62.1

no

**Debug output**

To help us resolve the problem, please attach the following anonymized status output

  netbird status -dA

Create and upload a debug bundle, and share the returned file key:

  netbird debug for 1m -AS -U

*Uploaded files are automatically deleted after 30 days.*

Alternatively, create the file only and attach it here manually:

  netbird debug for 1m -AS

**Screenshots**

**Additional context**

plusls:
## Describe your changes

Add TTL expiration mechanism to MgmtCacheResolver to handle DDNS IP changes. When cache entries expire (300s TTL), refresh synchronously using net.DefaultResolver. On refresh failure, serve stale cache to maintain connectivity.

Fixes #5113

https://github.com/netbirdio/netbird/issues/5113

### Checklist
- [X] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNS resolver caching with per-entry timestamps, TTL-based expiration, and automatic refresh of stale records. Failed refreshes now trigger a backoff while continuing to serve stale records for reliability. New logic also records insertion times and applies consistent TTLs to cached responses.

### DIFF
--- a/client/internal/dns/mgmt/mgmt.go
+++ b/client/internal/dns/mgmt/mgmt.go
@@ -18,14 +18,16 @@ import (
 )
 
 const (
-	dnsTimeout = 5 * time.Second
-	defaultTTL = 300 * time.Second
+	dnsTimeout          = 5 * time.Second
+	defaultTTL          = 300 * time.Second
+	refreshBackoff      = 30 * time.Second // wait time after failed refresh attempt
 )
 
 // cachedRecord holds DNS records with their cache timestamp.
 type cachedRecord struct {
-	records  []dns.RR
-	cachedAt time.Time
+	records           []dns.RR
+	cachedAt          time.Time
+	lastFailedRefresh *time.Time // timestamp of last failed refresh attempt, nil if never failed
 }
 
 // Resolver caches critical NetBird infrastructure domains
@@ -192,6 +194,7 @@ func (m *Resolver) AddDomain(ctx context.Context, d domain.Domain) error {
 
 // refreshDomain refreshes a stale cached domain using DefaultResolver.
 // On failure, it returns the stale records to avoid breaking connectivity.
+// A backoff mechanism prevents repeated blocking refresh attempts after failures.
 func (m *Resolver) refreshDomain(question dns.Question, stale *cachedRecord) []dns.RR {
 	m.refreshMutex.Lock()
 	defer m.refreshMutex.Unlock()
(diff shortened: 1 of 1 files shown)


LufeiCheng:
I've also encountered this issue, and I hope that this commit, or a similar one, will be merged into the client as soon as possible.

plusls:
Hi @lixmal , just a gentle ping on this PR.
I understand you might be busy — whenever you have time, I’d appreciate your thoughts. Thanks!

lixmal:
Superseded by #5945. Thanks for starting this, the TTL + backoff direction is the right one.

Reasons for a rewrite rather than iteration:

- `net.DefaultResolver` loops back when NetBird is the system DNS resolver. The refresh needs to route through the internal handler chain (skipping the mgmt cache itself) and only fall back to the OS resolver when we provably aren't the system resolver.
- Refresh ran synchronously inside `ServeDNS`, blocking the caller for up to 5s on a stale hit. Moved to async stale-while-revalidate so queries stay non-blocking.
- The global `refreshMutex` serialized refreshes across every domain. Replaced with per-question dedup via `singleflight`.
- `lastFailedRefresh` was mutated without the write lock.

Issue #5088, opened 2026-01-11
Add MFA to Local Users


Discolor3182:
Hello,

Please add MFA to local users

Thank you.

Camis:
Yes, this is must have!

ashokjp2:
Yes +1, must have considering this opens access to the entire network. Also currently I cant find option to even change password to make it more secure. If we had atleast MFA, then that issue could have waited.

obale:
+1

My current workaround is to add an Identity Provider via the dashboard, transfer ownership status to an IdP account and block the internal admin user.

ashokjp2:
There is no way I see to block or disable the internal user. Please let me know if I am missing something. If i delete, what happens to the device / peers mapped to the initial owner user account.

obale:
In my case I connected an external IdP, logged in with an IdP account first and made it Owner. After that under the Users page there is a switch allowing to block the internal user.

The internal account does not have any peers or resources. It was a new setup. This may not work for most setups though.

ashokjp2:
Perfect, thankyou. anyway to move peers and devices to the new admin as well?
If so that solves my problem temporarily :)

obale:
netbird down and netbird up on the device or logout and login with new account!?! I am not aware of an automated migration approach.

LorenzoVaccher01:
+1!!!!

heymoe:
My guess is Netbird is waiting on Dex to add 2FA support before it will add that functionality to local user management without having to use a 3rd party / external IdP provider. I found this thread tracking the progress of adding 2FA to Dex if anyone is interested in following the progress:

https://github.com/dexidp/dex/pull/3712

proteus88:
For small deployments this is a must have. +1

adrianbaena:
If anybody is interested, here the team behind Dex is following the whole process of SSO and MFA: https://github.com/dexidp/dex/issues/4560

stelle007:
MFA was included when you used Zentinal for the quick deploymend. why did you make the product worse for self-hosted setup and remove MFA ?

This is a must have.

+1

hyllm:
+1 ... also using Google as IdP for Owner and no local admin for now

svenkispert:
It's merged! :) Will be included in the upcoming release.

https://github.com/dexidp/dex/pull/3712 

[...]

PR #4760, opened 2025-11-07
[client] Add ssh server status output

lixmal:
## Describe your changes

Example status output:

```
NetBird IP: 100.0.212.2/16
Interface type: Kernel
Quantum resistance: false
Lazy connection: false
SSH Server: Enabled (1 active session)
  [google-oauth2|123594234982349@100.0.169.249:57630 -> root] sleep [1 args]
Networks: -
Forwarding rules: 0
Peers count: 1/1 Connected
```

### Checklist
- [ ] Is it a bug fix
- [ ] Is a typo/documentation fix
- [x] Is a feature enhancement
- [ ] It is a refactor
- [ ] Created tests that fail without the change (if possible)

> By submitting this pull request, you confirm that you have read and agree to the terms of the [Contributor License Agreement](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT.md).

## Documentation
Select exactly one:

- [ ] I added/updated documentation for this change
- [x] Documentation is **not needed** for this change (explain why)

### Docs PR URL (required if "docs added" is checked)
Paste the PR link from https://github.com/netbirdio/docs here:

https://github.com/netbirdio/docs/pull/__

Issue #619, opened 2022-12-12
How to change `netbird.selfhosted` domain?


ykorzikowski:
![grafik](https://user-images.githubusercontent.com/20556547/207015661-fd7c65e0-2388-4ba3-9df6-4763fa27207f.png)

I could not find it anywhere in documentation, and https://netbird.io/docs/overview/dns is empty

mlsmaycon:
Hello @ykorzikowski, sorry for the lack of documentation for that. 

The domain name can be changed by passing the flag `--dns-domain` to the management service. Using our docker-compose.yml file this is done in the management service command definition.

e.g.:
```yml
command: ["--port", "443", "--log-file", "console", "--disable-anonymous-metrics=false", "--single-account-mode-domain=example.com","--dns-domain=example.com"]
```

ykorzikowski:
thx @mlsmaycon :) working

PR #5804, opened 2026-04-06
[management] Enable MFA for local users

jnfrati:
## Describe your changes

Added support to enable MFA (TOTP only) for the embedded IDP. 
Now the server can use the config `server.auth.enableLocalMFA` which will automatically configure TOTP for all local users. 

### Dependencies

Due to the way the dashboard is implemented and how dex behaves on authentication requests, this implementation will request MFA on each tab the user opens. 
This behavior will be addressed by any of the following:
- [Issue #4760](https://github.com/dexidp/dex/issues/4760) where we want to implement mfa tracking on auth sessions
- [Dashboard PR #619](https://github.com/netbirdio/dashboard/pull/619) where we implement some functionality to persist sessions through different tabs on a browser

https://github.com/netbirdio/netbird/issues/5088

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Dashboard

Related PR for dashboard:
https://github.com/netbirdio/dashboard/pull/615

## Documentation
Select exactly one:

https://github.com/netbirdio/docs/pull/704

## Summary by CodeRabbit

* **New Features**
  * Configurable local Multi‑Factor Authentication (TOTP & WebAuthn) with session behavior controls and an option to enable local MFA that is applied to the embedded identity provider.
  * New user-facing pages: MFA verification, WebAuthn redirect entry, logout confirmation, and a redirecting home entry.

* **Chores**
  * Updated many third‑party dependencies to newer versions.


idp/dex/web/templates/webauthn_verify.html
```diff
@@ -0,0 +1,12 @@
...
+<script>globalThis.location.replace("/");</script>
+<noscript>
+    <div class="nb-card">
+        <h1 class="nb-heading">Redirecting…</h1>
+        <p class="nb-subheading">You are being redirected to the NetBird dashboard.</p>
+        <a href="/" class="nb-btn" style="display:block;text-align:center;text-decoration:none">Go to Dashboard</a>
+    </div>
+</noscript>
+
+{{ template "footer.html" . }}
```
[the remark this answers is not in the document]
jnfrati: This is expected as we do not support webauthn for now

zcj6758:
When can the merger be approved? Two factor authentication is highly necessary.

management/internals/server/modules.go
```diff
@@ -113,30 +114,47 @@ func (s *BaseServer) AccountManager() account.Manager {
...
 		embeddedEnabled := s.Config.EmbeddedIdP != nil && s.Config.EmbeddedIdP.Enabled
 		if embeddedEnabled {
-			idpManager, err = idp.NewEmbeddedIdPManager(context.Background(), s.Config.EmbeddedIdP, s.Metrics())
+			embeddedMgr, err := idp.NewEmbeddedIdPManager(context.Background(), s.Config.EmbeddedIdP, s.Metrics())
 			if err != nil {
 				log.Fatalf("failed to create embedded IDP service: %v", err)
 			}
-			return idpManager
+
+			if val := isMFAEnabledForAccount(s.Store().GetAllAccounts(context.Background())); val {
```
bcmmbaga, line 137: `GetAllAccounts` currently does not populate the `LocalMfaEnabled` field from SQL. Since startup loads accounts to determine whether `SetMFAEnabled` should be called, MFA enabled via the API can be lost after restart because the value defaults to `false`.

We should include the `settings_local_mfa_enabled` support here:
https://github.com/netbirdio/netbird/blob/9168302a4d335a8d6deeb95ce7c92438ef090a08/management/server/store/sql_store.go#L1499 so the setting is properly hydrated when loading accounts from the database.

combined/cmd/config.go
```diff
@@ -133,13 +133,18 @@ type ManagementConfig struct {
 
 // AuthConfig contains authentication/identity provider settings
 type AuthConfig struct {
```
bcmmbaga, line 135: `server.auth.enableLocalMFA` is currently not wired up. `AuthConfig` doesn't define the `yaml:"enableLocalMFA"` field and startup only enables MFA based on the persisted account setting via `isMFAEnabledForAccount`.

As a result, setting `server.auth.enableLocalMFA: true` has no effect and the documented config path doesn't actually enable TOTP.

bcmmbaga (changes requested):

combined/cmd/config.go
jnfrati: Ah yeah, this was earlier behavior, we decided to change it so that it's easier for people to enable after updates as there's not really something that requires a full server restart to be enabled. 

Will remove this from the PR description!

management/internals/server/modules.go
jnfrati: Thanks for flagging this one up! Totally missed it

bcmmbaga (approved):

Issue #5670, opened 2026-03-24
[macOS] ICE agent never initializes on arm64 with Homebrew CLI daemon — all peers permanently relay-only


thatgrant:
**Describe the problem**
On macOS (darwin/arm64), the ICE agent never initializes when using the CLI daemon installed via Homebrew. All peer connections fall back to relay permanently. The log repeatedly shows `ICE Agent is not initialized yet` warnings for every peer, and ICE candidate fields always show `-/-` in status output. This means any brief relay disruption drops all connections, as there is no P2P fallback.

**To Reproduce**
1. Install NetBird via Homebrew on macOS arm64 (`brew install netbirdio/tap/netbird`)
2. Start the daemon with `netbird service start`
3. Connect to a self-hosted management server with `netbird up`
4. Run `netbird status --detail` — all peers show `Connection type: Relayed` and `ICE candidate (Local/Remote): -/-`
5. Check logs — `ICE Agent is not initialized yet` warnings appear repeatedly for all peers

**Expected behavior**
ICE should initialize and attempt P2P hole punching. Peers on the same network or with compatible NAT should connect via P2P, not relay.

**Are you using NetBird Cloud?**
Self-hosted control plane (management, signal, relay all self-hosted via Docker).

**NetBird version**
0.67.0 (CLI and daemon)

**Is any other VPN software installed?**
No.

**Debug output**

```
Peers detail:
 newsletter.netbird.anon-PzmaI.domain:
  NetBird IP: 100.88.227.221/32
  Status: Idle
  Connection type: -
  ICE candidate (Local/Remote): -/-

 unterwelt01.netbird.anon-PzmaI.domain:
  Status: Connected
  Connection type: Relayed
  ICE candidate (Local/Remote): -/-

[all 7 connected peers show identical Relayed / -/- ICE candidate pattern]

OS: darwin/arm64
Daemon version: 0.67.0
Interface type: Userspace
Relays: [stun] Available, [rels] Available
Peers count: 7/8 Connected
```

Debug bundle file key: `5a7b5dc2d6370d0dfa12cb23bbc8c84de0f2fac1057cf041c84d747b5ce35cf5/56c9e20d-1d58-4415-b43c-3fc390ceb110`

**Additional context**
- macOS interface type is `Userspace` (expected on macOS, not kernel)
- STUN is reachable; `stunclient` confirms independent/port-preserving NAT — P2P should be possible
- TURN credentials are empty (`"turns": []`) in the network map
[...]

Issue #4225, opened 2025-07-25
Netbird Self-Hosted - P2P is not working - Connections are Relayed


maxideus85:
I have a test environment with an Ubuntu VM running Docker CE and the Netbird server containers deployed following the advanced guide. I also deployed a container for the Netbird agent on the same host to act as an exit node that's using host networking.  As a test, I also deployed the Linux Netbird agent directly on the Ubuntu host itself and also got the same result.

In the prior version of 0.51.1 for all containers and agents, I was able to get P2P connections between my Windows 11 endpoint and the exit node. However, in the latest release of 0.51.2, I am now only able to get relayed connections. I have ensured and verified that all documented and required ports are accessible so I don't think it's a firewall or NAT issue on the server side.  

Windows 11 endpoint Upload key
e72bcca5e3d73634c00a7b0be17aa15a54a72dd723d9fe22619affbdd23db377/1bb5e298-06b6-4fd4-8f89-e310fd5c685e

Ubuntu Docker host Upload key
e72bcca5e3d73634c00a7b0be17aa15a54a72dd723d9fe22619affbdd23db377/08403185-8cb6-4b30-93e8-f1a5a42cf53f

lixmal:
I noticed two things in your logs:

1. You seem to run the exit node on the Netbird server, which makes the stun server return `127.0.0.1` as "public IP". 
    The machine also doesn't have a public IP on its interface. The result is that this peer cannot transmit its public IP as a candidate. I suggest either using an external stun server or setting up this `netbird up`  flag:
    >      --external-ip-map strings         Sets external IPs maps between local addresses and interfaces.You can specify a comma-separated list with a single IP and IP/IP or IP/Interface Name. An empty string "" clears the previous configuration. E.g. --external-ip-map 12.34.56.78/10.0.0.1 or --external-ip-map 12.34.56.200,12.34.56.78/10.0.0.1,12.34.56.80/eth1 or --external-ip-map ""

2. The client peer is missing the STUN server for some reason:
    ```
    Relays: 
      [rel://vpn.anon-QMHla.domain:33080] is Available
    ```
    This leads to this peer not being able to transmit its public IP either

Fixing either of these should restore P2P.

maxideus85:
Hi there,

Thank you very much for looking into this!

I originally had designated the Netbird agent Docker container as the exit node but ran into the relay issue I described.  To troubleshoot, I installed the Linux agent on the Docker host itself, but it had the same issues.  

[...]

Issue #4031, opened 2025-06-22
No P2P/direct connection, only relayed?


juju4:
**Describe the problem**

I'm testing different configurations when two devices are on the same physical network with same/different subnets. one of them could be a mobile. reviewing connectivity, speed and so on.
Setup1 works fine both in Relayed and P2P mode but on slower side likely limited by physical layer (mix cable, wifi)
```asciiflow
             ┌─────────────────┐                       
     ┌───────┤ Internet router ┼────┐ 
     │       └─────────────────┘    │ 
     │                              │ 
     │                              │ 
     │                              │ 
┌────┼────┐                 ┌───────┼───────┐
│ Device1 │                 │Device2/Gateway│
└─────────┘                 └───────┬───────┘
                                    │                  
                               ┌────┼───┐              
                               │Subnet2 │              
                               └────────┘              
```
Setup2 for now works only in Relayed mode (cable only, same switch). I reviewed and unblocked things in firewall at the point that there is no drop logs that seems related to netbird (outside of rotating endpoints that I'm still chasing per https://github.com/netbirdio/netbird/issues/3783). I also disabled dns on Device2 as was conflicting with my internal dns setup.
```asciiflow
             ┌─────────────────┐    
             │ Internet router ├───┐
             └─────────────────┘   │
                            ┌──────┼────────┐ 
                          ┌─┼Device2/Gateway│ 
                          │ └────────────┬──┘ 
                          │              │             
                     ┌────┼───┐      ┌───┼───┐         
                     │Device1 │      │Subnet2│         
                     └────────┘      └───────┘         
```
I don't see any "discovered local candidate" in my client log like seen in https://github.com/netbirdio/netbird/issues/1753
Debug logs were uploaded `netbird debug for 1m --system-info --anonymize --upload-bundle` (https://docs.netbird.io/how-to/troubleshooting-client)

Netbird 0.46.0-0.47.2 on linux for both devices currently tested.

**To Reproduce**

See above setup

**Expected behavior**

Have P2P connection for systems on same LAN

**Are you using NetBird Cloud?**

Yes

**NetBird version**

0.46.0-0.47.2

**Is any other VPN software installed?**

One endpoint also has openvpn.
The other, none

**Debug output**
(note upload option should follow web proxy...)

Upload file key:
f79e391890ab27fb37c88b3b4be7011e22aa2e5ca6f38ffa9c4481884941f726/9d81097b-39cc-4e35-9d3e-65243afa3f5d

[...]

Issue #3339, opened 2025-02-17
Why can't the same local area network successfully P2P?


moqiu365:
**Describe the problem**

![Image](https://github.com/user-attachments/assets/d1a63361-1237-47e7-b3cd-333787222e98)

The above picture shows the structure of my local area network,
Among them, Android phones, WIN10, and Ubuntu all have installed netbird, and WIN10 connects to Android WiFi hotspots
However, communication between WIN10 and Ubuntu cannot be successfully P2P，I can successfully P2P using other virtual LAN tools (such as VNT, Tailscale), but Netbird cannot

Android phones<--->Ubuntu uses P2P communication
WIN10<--->Ubuntu uses a transit server
WIN10<--->Android phones use P2P communication

When WIN10 is directly connected to the LAN, WIN10 and Ubuntu can successfully P2P, and the firewalls of all devices except Android phones have been turned off

Here are the detailed information of netbird status - d

![Image](https://github.com/user-attachments/assets/5fddc0a9-7782-444d-8b31-593e50a9b5ee)
![Image](https://github.com/user-attachments/assets/ba6ffd02-bdf0-460d-9914-35c04660e113)
![Image](https://github.com/user-attachments/assets/d8665855-f144-49c7-8398-410583d49a67)

The following is the configuration of turnserver. conf

![Image](https://github.com/user-attachments/assets/eec6a328-9b00-49c4-b2a7-8870eeb188c2)

Here are the netbird-coturn-1 logs for Docker

![Image](https://github.com/user-attachments/assets/6d6cd0d7-5a1c-40ea-99b4-887a83a22adf)

**To Reproduce**

not have

**Expected behavior**

P2P communication should be successful

**Are you using NetBird Cloud?**

Self deployed netbird

**NetBird version**

`The latest netbird`

**NetBird status -dA output:**

If applicable, add the `netbird status -dA' command output.

**Do you face any (non-mobile) client issues?**

Please provide the file created by `netbird debug for 1m -AS`.
We advise reviewing the anonymized files for any remaining PII.

**Screenshots**

**Additional context**

moqiu365:
> 你好
> 
> 可能与 [#3325](https://github.com/netbirdio/netbird/issues/3325) 有关
> 
> [#3325 （评论）](https://github.com/netbirdio/netbird/issues/3325#issuecomment-2660676927)
> 
> 我会尝试编辑 win10 的 config.json 并放入 Android 手机给出的 win10 的 IP

[...]

PR #5805, opened 2026-04-06
[client] Fix ICE reconnection loop and buffer candidates before agent initialization

### LINKED ISSUE #3669: Unable to connect to netbird peer on differnt LAN with the same public IP
Hello, 

Hugh fan of the Netbird project. 

I have all of my internal services running in a dedicated network, let's call it LAN. 
I have other Netbird peers running on a different separated network, let's call it IOT-LAN. Both networks share the same WAN. Running a OPNsence instance, and is behind CG-NAT.

Some of my NetBird agents located in the IOT-LAN are unable to connect the the Netbird peers located in the LAN. All peers from outside the network, such as on the internet are able to connect to the peers in the LAN with no problems. 

I suspect this is an issue with the relay service. " [rels://relay.netbird.io:443] is Unavailable, reason: relay client not connected"

**Expected behavior**

Are you using NetBird Cloud? - Yes

NetBird version - V0.41.0

Is any other VPN software installed? - No

**Debug output**

"Peers detail:
 iphone-nevellsb.netbird.cloud:
  NetBird IP: 100.72.5.10
  Public key: q4jsleN8mYGANXW+KP+7XDQoA+6Hz34ZaX70Dc/fzWI=
  Status: Connected
  -- detail --
  Connection type: Relayed
  ICE candidate (Local/Remote): host/relay
  ICE candidate endpoints (Local/Remote): 127.0.0.1:51820/198.51.100.0:12132
  Relay server address:
  Last connection update: 14 minutes, 16 seconds ago
  Last WireGuard handshake: 1 minute, 43 seconds ago
  Transfer status (received/sent) 900 B/2.3 KiB
  Quantum resistance: false
  Networks: -
  Latency: 242.2069ms

 birdcon-1.netbird.cloud:
  NetBird IP: 100.72.105.244
  Public key: LglatLZGN0L2YyrpwgvoCjiBSg0xojq4zBHfx4imzwU=
  Status: Connected
  -- detail --
  Connection type: Relayed
  ICE candidate (Local/Remote): srflx/relay
  ICE candidate endpoints (Local/Remote): 198.51.100.1:11666/198.51.100.2:32911
  Relay server address:
  Last connection update: 14 minutes, 16 seconds ago
  Last WireGuard handshake: 1 minute, 44 seconds ago
  Transfer status (received/sent) 644 B/2.3 KiB
  Quantum resistance: false
  Networks: 192.168.6.0/24
  Latency: 206.321ms

 web-03-clone.netbird.cloud:
  NetBird IP: 100.72.111.7
  Public key: IAJShlSwmqhb/Bg/P5gBqc5Q/YIZLTKrKcsfVRAYlBE=
  Status: Connected
  -- detail --
  Connection type: Relayed
  ICE candidate (Local/Remote): host/relay
  ICE candidate endpoints (Local/Remote): 127.0.0.1:51820/198.51.100.3:50323
  Relay server address:
  Last connection update: 14 minutes, 15 seconds ago
  Last WireGuard handshake: 1 minute, 43 seconds ago
  Transfer status (received/sent) 644 B/2.3 KiB
  Quantum resistance: false
  Networks: -
  Latency: 376.556ms

 lab-birdconnect.netbird.cloud:
  NetBird IP: 100.72.124.247
  Public key: 3ExFJeUc9Wxk9vNOmzIHr0YAR9+7CTDEqJirN8YORTk=
  Status: Disconnected
  -- detail --
  Connection type:
  ICE candidate (Local/Remote): -/-
  ICE candidate endpoints (Local/Remote): -/-
  Relay server address:
  Last connection update: -
  Last WireGuard handshake: -
  Transfer status (received/sent) 0 B/0 B
  Quantum resistance: false
  Networks: -
  Latency: 0s

Events:
  [INFO] SYSTEM (18e75b94-707d-4853-bec8-2b8da58497b9)
    Message: Network map updated
    Time: 2 hours, 8 minutes ago
  [INFO] SYSTEM (5e2a503a-10f8-4e9f-89b1-dc88924ba92b)
    Message: Network map updated
    Time: 1 hour, 59 minutes ago
  [INFO] SYSTEM (e078dde9-be79-44b5-9b02-cfa6bcac6696)
    Message: Network map updated
    Time: 1 hour, 32 minutes ago
  [INFO] SYSTEM (06c74158-794b-4477-87b9-84603a03d8cd)
    Message: Network map updated
    Time: 1 hour, 28 minutes ago
  [INFO] SYSTEM (634a310f-c235-494b-ab09-83cc5dfa6294)
    Message: Network map updated
    Time: 1 hour, 21 minutes ago
  [INFO] SYSTEM (b9996879-7ed7-41e0-8102-088c07a28e54)
    Message: Network map updated
    Time: 1 hour, 20 minutes ago
  [INFO] SYSTEM (2d465aa0-c2f9-48b6-b07f-ffa366bef611)
    Message: Network map updated
    Time: 1 hour, 18 minutes ago
  [INFO] SYSTEM (49e41599-0bb0-4c14-bacc-1059cfcb789f)
    Message: Network map updated
    Time: 1 hour, 18 minutes ago
  [INFO] SYSTEM (ca271077-f423-4ccd-be7c-012428b53ef7)
    Message: Network map updated
    Time: 52 minutes, 47 seconds ago
  [INFO] SYSTEM (7819bb89-1ef6-4736-8dd5-489018e66830)
    Message: Network map updated
    Time: 14 minutes, 19 seconds ago
OS: windows/amd64
Daemon version: 0.41.0
CLI version: 0.41.0
Management: Connected to https://api.netbird.io:443/
Signal: Connected to https://signal.netbird.io:443
Relays:
  [stun:stun.netbird.io:5555] is Available
  [turns:turn.netbird.io:443?transport=tcp] is Available
  [rels://relay.netbird.io:443] is Unavailable, reason: relay client not connected
Nameservers:
FQDN: laptop-hoeekb4b.netbird.cloud
NetBird IP: 100.72.155.197/16
Interface type: Userspace
Quantum resistance: false
Networks: -
Forwarding rules: 0
Peers count: 3/4 Connected"

 
[netbird.debug.3916181182 (2).zip](https://github.com/user-attachments/files/19729859/netbird.debug.3916181182.2.zip)

### LINKED ISSUE #5672: ICE Agent is not initialized yet on android app
**Describe the problem**

On the android app we have connectivity problems. The app can authenticate to our self-hosted peer. But the peer doesn't connect, can be seen by "0 out of 1 Peers connected". Digging further in the problem we found out that no ICE packets are sent. Which we saw by tcpdumping on the server peer.  ICE fails so the wireguard connection cannot be made. We found following logs:

```
03-23 16:13:12.043 10117 10193 E GoLog    : [INFO] [peer: <REDACTED_PEER_ID>] client/internal/peer/conn.go:280 OnRemoteAnswer, priority: None, status ICE: Disconnected, status relay: Disconnected
03-23 16:13:12.043 10117 10193 E GoLog    : [INFO] [peer: <REDACTED_PEER_ID>] client/internal/peer/handshaker.go:105 received answer, running version development, remote WireGuard listen port 11000, session id: <REDACTED_SESSION_ID>
03-23 16:13:12.043 10117 10193 E GoLog    : [INFO] [peer: <REDACTED_PEER_ID>] client/internal/peer/worker_relay.go:49 Relay is not supported by remote peer
03-23 16:13:12.043 10117 10193 E GoLog    : [INFO] [peer: <REDACTED_PEER_ID>] client/internal/peer/worker_relay.go:49 Relay is not supported by remote peer
03-23 16:13:12.045 10117 10499 E GoLog    : [DEBG] [peer: <REDACTED_PEER_ID>] client/internal/peer/worker_ice.go:154 OnRemoteCandidate from peer <REDACTED_PEER_ID> -> udp4 host <REDACTED_IP>:11000 (resolved: <REDACTED_IP>:11000) candidate:<REDACTED_CANDIDATE_HASH>
03-23 16:13:12.045 10117 10499 E GoLog    : [WARN] [peer: <REDACTED_PEER_ID>] client/internal/peer/worker_ice.go:156 ICE Agent is not initialized yet
03-23 16:13:12.046 10117 10499 E GoLog    : [DEBG] [peer: <REDACTED_PEER_ID>] client/internal/peer/worker_ice.go:154 OnRemoteCandidate from peer <REDACTED_PEER_ID> -> udp4 srflx <REDACTED_IP>:11000 related 0.0.0.0:11000 (resolved: <REDACTED_IP>:11000) candidate:<REDACTED_CANDIDATE_HASH>
03-23 16:13:12.046 10117 10499 E GoLog    : [WARN] [peer: <REDACTED_PEER_ID>] client/internal/peer/worker_ice.go:156 ICE Agent is not initialized yet
```
We get a ip from the range but cannot sent packages over the Wireguard tunnel. On the management server we see that we are connected.
Uninstalling and installing the app doesn't fix the problem.

**To Reproduce**

1. Install Netbird android app from play store
2. Change management url to self-hosted infra
3. Connect -> authenticate
4. Result => connected
5. Disconnect
6. Go to Android Settings -> Apps -> Apps management -> Netbird -> Storage usage -> Clear data
7.  Repeat steps 1-3
8. Result authenticated but "0 of 1 Peers connected"

**Expected behavior**

It should read "1 of 1 Peers connected" and we should be able to send packages over the Wireguard tunnel. 

We self-host management server, signal server, STUN, zitadel and OIDC.

**NetBird version**

We use following versions:

* Android app: v0.3.2 (core 0.64.1)
* Server-peer, management and signal (v0.64.5)

No

**Debug output**

**netbird status -dA for the broken peer**
<REDACTED_DEVICE_NAME>.domain:
  NetBird IP: <REDACTED_INTERNAL_IP>
  Public key: <REDACTED_PUBLIC_KEY>
  Status: Connecting
  -- detail --
  Connection type: -
  ICE candidate (Local/Remote): -/-
  ICE candidate endpoints (Local/Remote): -/-
  Relay server address: 
  Last connection update: 16 hours, 47 minutes ago
  Last WireGuard handshake: -
  Transfer status (received/sent) 0 B/0 B
  Quantum resistance: false 
  Networks: -
  Latency: 0s

**Screenshots**

**Additional context**
We could reproduce this error on two different androids, didn't find a way to fix this bug:

OS: Android 16
Kernel: 6.1.134

and

OS: Android 16
Kernel: 6.1.134

MichaelUray:
## Summary
- **Guard Loop Fix**: When both peers are behind the same NAT (same public IP), their connectivity guards fire simultaneously every ~12s, each sending a new offer with a new session ID. This cancels the other side's ICE agent before it can finish negotiating, creating an infinite reconnection loop. Fix: skip new offers while `agentConnecting == true` to let the current ICE attempt complete.
- **Candidate Buffering**: Remote ICE candidates that arrive before the local ICE agent is created were silently dropped. Add a `pendingCandidates` buffer that collects early candidates and flushes them once the agent is ready.

Both fixes are tightly coupled — without the buffer, the guard loop fix alone would still miss early candidates.

## Checklist
- [x] Bug fix
- [x] Create tests that fail without the change: ICE connections between peers behind the same NAT consistently failed before this fix
- [x] Documentation not needed — internal ICE behavior change, no user-facing API change

By submitting this pull request, I confirm that I have read and agree to the terms of the [Contributor License Agreement](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT.md).

## Related Issues

Fixes #3669 — Unable to connect to netbird peer on different LAN with the same public IP
Fixes #5672 — ICE Agent is not initialized yet on android app
Related #2703 — Multiple peers behind NAT getting relayed
Related #3339 — Same LAN cannot successfully P2P
Related #4031 — No P2P/direct connection, only relayed
Related #4225 — Netbird Self-Hosted P2P not working
Related #5670 — macOS ICE agent never initializes (arm64)

## Summary by CodeRabbit

* **Bug Fixes**
  * Buffer early remote connection candidates with a bounded cap to avoid lost candidates and limit memory use.
  * Expand and queue applicable server-reflexive candidates for later delivery instead of dropping them.
  * Prevent rapid reconnection churn by skipping replacement of a recently-started agent while it’s still connecting (adds a short protection window).
  * Flush buffered candidates into a newly created connection and clear the buffer on close to prevent stale data.

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -1625,7 +1625,7 @@ func (e *Engine) receiveSignalEvents() {
 					return err
 				}
 
-				go conn.OnRemoteCandidate(candidate, e.routeManager.GetClientRoutes())
+				conn.OnRemoteCandidate(candidate, e.routeManager.GetClientRoutes())
 			case sProto.Body_MODE:
 			case sProto.Body_GO_IDLE:
 				e.connMgr.DeactivatePeer(conn)
--- a/client/internal/peer/worker_ice.go
+++ b/client/internal/peer/worker_ice.go
@@ -30,6 +30,10 @@ import (
 // than typical STUN round-trips.
 const agentLoopProtectionWindow = 3 * time.Second
 
+// maxPendingCandidates is the maximum number of remote ICE candidates that
+// can be buffered before the local agent is initialized.
+const maxPendingCandidates = 50
+
 type ICEConnInfo struct {
 	RemoteConn                 net.Conn
 	RosenpassPubKey            []byte
@@ -72,6 +76,10 @@ type WorkerICE struct {
 
 	// we record the last known state of the ICE agent to avoid duplicate on disconnected events
 	lastKnownState ice.ConnectionState
+
+	// pendingCandidates buffers remote candidates that arrive before the ICE agent is initialized.
+	// They are flushed through the normal candidate pipeline once the agent is ready.
+	pendingCandidates []ice.Candidate
 
 	// portForwardAttempted tracks if we've already tried port forwarding this session
 	portForwardAttempted bool
@@ -160,8 +168,10 @@ func (w *WorkerICE) handleExistingAgent(remoteOfferAnswer *OfferAnswer) (skip bo
 	sessionID, err := NewICESessionID()
 	if err != nil {
 		w.log.Errorf("failed to create new session ID: %s", err)
+		w.sessionID = ""
+	} else {
+		w.sessionID = sessionID
 	}
-	w.sessionID = sessionID
 	w.agent = nil
 
 	return false
@@ -202,6 +212,8 @@ func (w *WorkerICE) OnNewOffer(remoteOfferAnswer *OfferAnswer) {
 		w.remoteSessionID = ""
 	}
 
+	w.flushPendingCandidates(agent)
+
 	go w.connect(dialerCtx, agent, remoteOfferAnswer)
 }
 
@@ -210,16 +222,31 @@ func (w *WorkerICE) OnRemoteCandidate(candidate ice.Candidate, haRoutes route.HA
 	w.muxAgent.Lock()
 	defer w.muxAgent.Unlock()
 	w.log.Debugf("OnRemoteCandidate from peer %s -> %s", w.config.Key, candidate.String())
-	if w.agent == nil {
-		w.log.Warnf("ICE Agent is not initialized yet")
-		return
-	}
 
+	// Filter routed candidates before buffering or adding, so both paths
+	// behave identically regardless of arrival timing.
 	if candidateViaRoutes(candidate, haRoutes) {
 		return
 	}
 
-	if err := w.agent.AddRemoteCandidate(candidate); err != nil {
+	if w.agent == nil {
+		if len(w.pendingCandidates) >= maxPendingCandidates {
+			w.log.Warnf("pending candidate buffer full (%d), dropping candidate: %s", maxPendingCandidates, candidate.Type())
+			return
+		}
+		w.log.Infof("ICE Agent not ready, buffering remote candidate: %s", candidate.Type())
+		w.pendingCandidates = append(w.pendingCandidates, candidate)
+		return
+	}
+
+	w.addRemoteCandidate(w.agent, candidate)
+}
+
+// addRemoteCandidate adds a remote candidate to the agent and, when applicable,
+// also generates and adds the synthetic extra server-reflexive candidate.
+// This is the single code path for both live and buffered candidates.
+func (w *WorkerICE) addRemoteCandidate(agent *icemaker.ThreadSafeAgent, candidate ice.Candidate) {
+	if err := agent.AddRemoteCandidate(candidate); err != nil {
 		w.log.Errorf("error while handling remote candidate")
 		return
 	}
@@ -233,11 +260,25 @@ func (w *WorkerICE) OnRemoteCandidate(candidate ice.Candidate, haRoutes route.HA
 			return
 		}
 
-		if err := w.agent.AddRemoteCandidate(extraSrflx); err != nil {
+		if err := agent.AddRemoteCandidate(extraSrflx); err != nil {
 			w.log.Errorf("error while handling remote candidate")
 			return
 		}
 	}
+}
+
+// flushPendingCandidates drains the buffered candidates through the normal
+// candidate pipeline once the ICE agent is ready. The caller must hold w.muxAgent.
+func (w *WorkerICE) flushPendingCandidates(agent *icemaker.ThreadSafeAgent) {
+	if len(w.pendingCandidates) == 0 {
+		return
+	}
+
+	w.log.Infof("flushing %d buffered remote candidates", len(w.pendingCandidates))
+	for _, c := range w.pendingCandidates {
+		w.addRemoteCandidate(agent, c)
+	}
+	w.pendingCandidates = nil
 }
 
 func (w *WorkerICE) GetLocalUserCredentials() (frag string, pwd string) {
@@ -381,11 +422,14 @@ func (w *WorkerICE) closeAgent(agent *icemaker.ThreadSafeAgent, cancel context.C
 		sessionID, err := NewICESessionID()
 		if err != nil {
 			w.log.Errorf("failed to create new session ID: %s", err)
+			w.sessionID = ""
+		} else {
+			w.sessionID = sessionID
 		}
-		w.sessionID = sessionID
 		w.agent = nil
 		w.agentConnecting = false
 		w.remoteSessionID = ""
+		w.pendingCandidates = nil
 	}
 	return sessionChanged
 }
--- a/client/internal/peer/worker_ice.go
+++ b/client/internal/peer/worker_ice.go
@@ -30,10 +30,6 @@ import (
 // than typical STUN round-trips.
 const agentLoopProtectionWindow = 3 * time.Second
 
-// maxPendingCandidates is the maximum number of remote ICE candidates that
-// can be buffered before the local agent is initialized.
-const maxPendingCandidates = 50
-
 type ICEConnInfo struct {
 	RemoteConn                 net.Conn
 	RosenpassPubKey            []byte
@@ -76,10 +72,6 @@ type WorkerICE struct {
 
 	// we record the last known state of the ICE agent to avoid duplicate on disconnected events
 	lastKnownState ice.ConnectionState
-
-	// pendingCandidates buffers remote candidates that arrive before the ICE agent is initialized.
-	// They are flushed through the normal candidate pipeline once the agent is ready.
-	pendingCandidates []ice.Candidate
 
 	// portForwardAttempted tracks if we've already tried port forwarding this session
 	portForwardAttempted bool
@@ -212,8 +204,6 @@ func (w *WorkerICE) OnNewOffer(remoteOfferAnswer *OfferAnswer) {
 		w.remoteSessionID = ""
 	}
 
-	w.flushPendingCandidates(agent)
-
 	go w.connect(dialerCtx, agent, remoteOfferAnswer)
 }
 
@@ -230,55 +220,27 @@ func (w *WorkerICE) OnRemoteCandidate(candidate ice.Candidate, haRoutes route.HA
 	}
 
 	if w.agent == nil {
-		if len(w.pendingCandidates) >= maxPendingCandidates {
-			w.log.Warnf("pending candidate buffer full (%d), dropping candidate: %s", maxPendingCandidates, candidate.Type())
-			return
-		}
-		w.log.Infof("ICE Agent not ready, buffering remote candidate: %s", candidate.Type())
-		w.pendingCandidates = append(w.pendingCandidates, candidate)
+		w.log.Warnf("ICE Agent is not initialized yet")
 		return
 	}
 
-	w.addRemoteCandidate(w.agent, candidate)
-}
-
-// addRemoteCandidate adds a remote candidate to the agent and, when applicable,
-// also generates and adds the synthetic extra server-reflexive candidate.
-// This is the single code path for both live and buffered candidates.
-func (w *WorkerICE) addRemoteCandidate(agent *icemaker.ThreadSafeAgent, candidate ice.Candidate) {
-	if err := agent.AddRemoteCandidate(candidate); err != nil {
+	if err := w.agent.AddRemoteCandidate(candidate); err != nil {
 		w.log.Errorf("error while handling remote candidate")
 		return
 	}
 
 	if shouldAddExtraCandidate(candidate) {
-		// sends an extra server reflexive candidate to the remote peer with our related port (usually the wireguard port)
-		// this is useful when network has an existing port forwarding rule for the wireguard port and this peer
 		extraSrflx, err := extraSrflxCandidate(candidate)
 		if err != nil {
 			w.log.Errorf("failed creating extra server reflexive candidate %s", err)
 			return
 		}
 
-		if err := agent.AddRemoteCandidate(extraSrflx); err != nil {
+		if err := w.agent.AddRemoteCandidate(extraSrflx); err != nil {
 			w.log.Errorf("error while handling remote candidate")
 			return
 		}
 	}
-}
-
-// flushPendingCandidates drains the buffered candidates through the normal
-// candidate pipeline once the ICE agent is ready. The caller must hold w.muxAgent.
-func (w *WorkerICE) flushPendingCandidates(agent *icemaker.ThreadSafeAgent) {
-	if len(w.pendingCandidates) == 0 {
-		return
-	}
-
-	w.log.Infof("flushing %d buffered remote candidates", len(w.pendingCandidates))
-	for _, c := range w.pendingCandidates {
-		w.addRemoteCandidate(agent, c)
-	}
-	w.pendingCandidates = nil
 }
 
 func (w *WorkerICE) GetLocalUserCredentials() (frag string, pwd string) {
@@ -429,7 +391,6 @@ func (w *WorkerICE) closeAgent(agent *icemaker.ThreadSafeAgent, cancel context.C
 		w.agent = nil
 		w.agentConnecting = false
 		w.remoteSessionID = ""
-		w.pendingCandidates = nil
 	}
 	return sessionChanged
 }
--- a/client/internal/peer/worker_ice.go
+++ b/client/internal/peer/worker_ice.go
@@ -21,6 +21,15 @@ import (
 	"github.com/netbirdio/netbird/route"
 )
 
+// agentLoopProtectionWindow is the maximum age of a freshly created ICE
+// agent during which OnNewOffer will refuse to replace it with a different
+// session ID. Must be short enough not to stall crash recovery (the ICE
+// FailedTimeout is 6s) and long enough to cover the millisecond-range race
+// between two guards stomping on each other when both peers are behind the
+// same NAT. 3s is well below the guard's 3s initial tick but still longer
+// than typical STUN round-trips.
+const agentLoopProtectionWindow = 3 * time.Second
+
 type ICEConnInfo struct {
 	RemoteConn                 net.Conn
 	RosenpassPubKey            []byte
@@ -46,6 +55,7 @@ type WorkerICE struct {
 	agent             *icemaker.ThreadSafeAgent
 	agentDialerCancel context.CancelFunc
 	agentConnecting   bool      // while it is true, drop all incoming offers
+	agentStartTime    time.Time // when the current agent was created; used to bound the loop-prevention window
 	lastSuccess       time.Time // with this avoid the too frequent ICE agent recreation
 	// remoteSessionID represents the peer's session identifier from the latest remote offer.
 	remoteSessionID ICESessionID
@@ -95,36 +105,75 @@ func NewWorkerICE(ctx context.Context, log *log.Entry, config ConnConfig, conn *
 	return w, nil
 }
 
+// handleExistingAgent checks whether an incoming offer should be skipped
+// (because the current agent is still valid) or accepted (tearing down the
+// existing agent first). It returns true when the offer must be skipped.
+// The caller must hold w.muxAgent.
+func (w *WorkerICE) handleExistingAgent(remoteOfferAnswer *OfferAnswer) (skip bool) {
+	if w.agent == nil && !w.agentConnecting {
+		return false
+	}
+
+	// backward compatibility with old clients that do not send session ID
+	if remoteOfferAnswer.SessionID == nil {
+		w.log.Debugf("agent already exists, skipping the offer")
+		return true
+	}
+
+	if w.remoteSessionID == *remoteOfferAnswer.SessionID {
+		w.log.Debugf("agent already exists and session ID matches, skipping the offer: %s", remoteOfferAnswer.SessionIDString())
+		return true
+	}
+
+	// Reconnection-loop protection.
+	//
+	// When two peers are behind the same NAT (same public IP) both of
+	// their guards can fire almost simultaneously, each sending an offer
+	// with a new session ID that tears down the other side's freshly
+	// created ICE agent before it has a chance to finish. We therefore
+	// ignore a replacement offer if the current agent was created only
+	// moments ago AND is still in the connecting phase.
+	//
+	// The window is intentionally small (agentLoopProtectionWindow):
+	//   * the guard's initial tick is 3s, so two guards can only stomp
+	//     on each other during the first few hundred ms after recreate;
+	//   * once the window has elapsed we MUST accept offers with a new
+	//     session ID, otherwise crash-recovery of the remote peer would
+	//     be stalled for the full ICE failed-timeout (6s) instead of
+	//     happening immediately. See pappz's review comment on PR #5805.
+	agentAge := time.Since(w.agentStartTime)
+	if w.agentConnecting && agentAge < agentLoopProtectionWindow {
+		w.log.Infof("ICE agent started %s ago, skipping new offer with different session to break reconnection loop: %s",
+			agentAge, remoteOfferAnswer.SessionIDString())
+		return true
+	}
+
+	w.log.Debugf("agent already exists, recreate the connection")
+	w.remoteSessionChanged = true
+	w.agentDialerCancel()
+	if w.agent != nil {
+		if err := w.agent.Close(); err != nil {
+			w.log.Warnf("failed to close ICE agent: %s", err)
+		}
+	}
+
+	sessionID, err := NewICESessionID()
+	if err != nil {
+		w.log.Errorf("failed to create new session ID: %s", err)
+	}
+	w.sessionID = sessionID
+	w.agent = nil
+
+	return false
+}
+
 func (w *WorkerICE) OnNewOffer(remoteOfferAnswer *OfferAnswer) {
 	w.log.Debugf("OnNewOffer for ICE, serial: %s", remoteOfferAnswer.SessionIDString())
 	w.muxAgent.Lock()
 	defer w.muxAgent.Unlock()
 
-	if w.agent != nil || w.agentConnecting {
-		// backward compatibility with old clients that do not send session ID
-		if remoteOfferAnswer.SessionID == nil {
-			w.log.Debugf("agent already exists, skipping the offer")
-			return
-		}
-		if w.remoteSessionID == *remoteOfferAnswer.SessionID {
-			w.log.Debugf("agent already exists and session ID matches, skipping the offer: %s", remoteOfferAnswer.SessionIDString())
-			return
-		}
-		w.log.Debugf("agent already exists, recreate the connection")
-		w.remoteSessionChanged = true
-		w.agentDialerCancel()
-		if w.agent != nil {
-			if err := w.agent.Close(); err != nil {
-				w.log.Warnf("failed to close ICE agent: %s", err)
-			}
-		}
-
-		sessionID, err := NewICESessionID()
-		if err != nil {
-			w.log.Errorf("failed to create new session ID: %s", err)
-		}
-		w.sessionID = sessionID
-		w.agent = nil
+	if w.handleExistingAgent(remoteOfferAnswer) {
+		return
 	}
 
 	var preferredCandidateTypes []ice.CandidateType
@@ -146,6 +195,7 @@ func (w *WorkerICE) OnNewOffer(remoteOfferAnswer *OfferAnswer) {
 	w.agent = agent
 	w.agentDialerCancel = dialerCancel
 	w.agentConnecting = true
+	w.agentStartTime = time.Now()
 	if remoteOfferAnswer.SessionID != nil {
 		w.remoteSessionID = *remoteOfferAnswer.SessionID
 	} else {


pappz:
@MichaelUray Thank you for the PR.
Regarding the candidate buffering: I think this can be dropped.                                          
                                                                                                                             
The race you're protecting against — OnRemoteCandidate arriving before the agent is created — requires a very specific timing: the remote peer would need to create its agent, start gathering, discover a candidate, and signal it back, all before our handshaker goroutine reads the offer from its channel and calls OnNewOffer. That's microseconds (channel read) vs tens-to-hundreds of milliseconds (STUN/TURN network I/O for candidate gathering), so in practice this almost never happens.                   
                                                                                                                             
If we ever wanted to truly fix this race, the correct approach would be to remove the go from engine.go:1612:              
 ``` 
  // current:                                                                                                                
  go conn.OnRemoteCandidate(candidate, e.routeManager.GetClientRoutes())
```
```
  // fix:                                                                                                                    
  conn.OnRemoteCandidate(candidate, e.routeManager.GetClientRoutes())
```
But of course it is a risky change.

**Comment by pappz** [comment]
The session ID mechanism exists for two specific reasons:

1. Force renegotiation after silent disconnect — when a peer crashes and reconnects with a new session ID, the remote side must tear down its agent and start fresh.
2. Duplicate offer suppression — same session ID on resend (ICE + relay paths) gets deduplicated.

By skipping all offers with a different session ID while `agentConnecting == true`, case 1 regresses. If the remote peer
crashed and came back, the local peer ignores its new session ID and keeps trying the current (now dead) ICE attempt until it times out. Recovery gets delayed by the full ICE timeout (30s+).

**Comment by MichaelUray** [comment]
Thanks @pappz for the thorough review!

I've reworked the PR based on your feedback:

**Candidate buffering — removed.** You're right that the race is practically impossible given the timing constraints (microsecond channel read vs hundreds of ms STUN/TURN I/O). All three buffering commits have been dropped.

**Session-ID skip during agentConnecting — replaced with a time-windowed guard.** Instead of unconditionally skipping offers while connecting, the new approach only skips offers with a different session ID during the first 3 seconds after agent creation. This addresses both concerns:

- **Loop prevention**: The guard-vs-guard race happens within milliseconds of agent creation (both guards fire on the same ~3s tick). The 3s window covers this with plenty of margin.
- **Crash recovery**: After 3s, offers with new session IDs are accepted normally. Since the ICE FailedTimeout is 6s, the worst-case crash-recovery delay is 3s (vs the full 6s+ that the unconditional approach would have caused).

The 3s window aligns with the guard's `InitialInterval` of 3s, so by the time the first regular guard retry fires, the protection window has just expired.

One factual note: the ICE FailedTimeout in this codebase is 6 seconds (see `iceFailedTimeoutDefault` in `client/internal/peer/ice/agent.go`), not 30s+.

Tested on a real Android device (Galaxy S26) with multiple peers in the same LAN (same public IP) — direct P2P connections establish reliably without the reconnection loop.

**Comment by MichaelUray** [comment]
Addressed the coderabbitai review on the candidate buffer path in commit 772014584 — buffered candidates now flow through the same `candidateViaRoutes()` and `shouldAddExtraCandidate()` pipeline as normal candidates, and the session ID error path no longer continues with a stale ID. Ready for another look @pappz.

**Comment by MichaelUray** [comment]
@pappz friendly ping — all review feedback has been addressed. Ready for another look whenever you have time. Thanks!

**Comment by MichaelUray** [comment]
@pappz Following up with detailed test results on both the candidate buffer and the synchronous OnRemoteCandidate change.

### Candidate buffer: confirmed unnecessary

I deployed the current PR branch (without buffer, with synchronous OnRemoteCandidate) to a Cudy TR3000 router (OpenWrt, ARM64) and tested against multiple peers:

| Peer | Version | Lost candidates | Connection |
|------|---------|----------------|------------|
| dk20 (same LAN) | dev (this PR) | 6 | **P2P** ✓ |
| uray-mic-dh (same LAN) | dev (this PR) | 13 | **P2P** ✓ |
| uray-mic-d4 (remote) | v0.67.1 (stock) | 1040 | Relayed |

The "ICE Agent is not initialized yet" warning still appears (even with synchronous OnRemoteCandidate — see below), but **P2P works regardless**. The lost candidates are re-signaled on subsequent guard rounds. You were right that the buffer has no practical impact.

### Synchronous OnRemoteCandidate: doesn't fix the race, but doesn't hurt

The race condition isn't at the goroutine level — it's at the **signaling message level**. The remote peer sends host candidates (instant, no STUN needed) *before* sending its answer. In the `receiveMSG` loop, CANDIDATE messages arrive before the ANSWER message:

```
10.598s - Router sends OFFER
10.765s - CANDIDATE message arrives (167ms later) - agent NOT yet created
10.786s - ANSWER message arrives (188ms later) - OnNewOffer creates agent
```

Host candidates don't need STUN/TURN — they're discovered instantly when the agent is created. So the remote peer signals them before signaling the answer. Making `OnRemoteCandidate` synchronous doesn't change this ordering since both messages are processed sequentially in the same `receiveMSG` goroutine.

That said, making it synchronous is still a reasonable cleanup (removes unnecessary goroutine spawn), so I've kept that change.

### The 1040 lost candidates from uray-mic-d4

These are from the guard loop on the stock v0.67.1 client — it sends new offers every ~12-30s, each with a new session ID. Each time, the remote peer's host candidates arrive before the answer. The time-windowed guard (3s) on our side prevents the reconnection loop stomping, but the old client keeps re-triggering. The lost candidates accumulate but don't prevent P2P once both sides have the fix.

uray-mic-d4 stays Relayed not because of lost candidates, but because it's behind a different NAT and its guard loop prevents ICE from completing (same root cause this PR fixes).

### Current PR state

- ✅ Time-windowed guard (`agentLoopProtectionWindow = 3s`) — prevents reconnection loop while allowing crash recovery
- ✅ Candidate buffer removed
- ✅ Synchronous OnRemoteCandidate
- ✅ Session ID error handling improved

Ready for another look whenever you have time.

**Comment by pappz** [comment]
@MichaelUray 
I checked your PR and I don't see evidence for the root cause of the issue. The offer/answer and candidate ordering is important, and as far as I can tell it's correct. The session ID mechanism was specifically designed to avoid
unnecessary agent recreation. Just because two peers try to connect to each other at the same time, it should not get stuck in a loop.

Could you send a debug bundle from both peers where you highlight the exact sequence that demonstrates the problem you want to fix? Ideally I'd like to see the OnNewOffer lines with session IDs across 3-4 guard cycles on both sides, so we can trace where the session-ID dedup fails.

**Comment by MichaelUray** [comment]
Hi @pappz, here is the debug evidence you requested. The analysis below was captured on **stock `upstream/main` at `3098f48b2`** with only additive `Infof` log lines – no behavioral changes – so that the OnNewOffer / closeAgent / send/recv-offer sequence is visible with session IDs on both sides.

<details><summary>Reproduction environment</summary>

Two peers, same LAN, same public IP. IPv6 + IPv4 peer-to-peer UDP blocked at netfilter/WFP to force ICE FailedTimeout (relay/signal/management paths untouched). Without that block ICE succeeds fast enough that `closeAgent` never fires; the loop is the post-failure steady state.
</details>

# PR #5805 – Session-ID Dedup Failure Evidence

## Setup

- Base code: `upstream/main` at `3098f48b2` **with zero fix-logic**, only
  `TRACE-PR5805` `Infof` log lines added in `OnNewOffer`, `closeAgent`,
  `sendOffer`/`sendAnswer`, and `RECV_OFFER`/`RECV_ANSWER`. No behavioral
  changes – the instrumentation is purely additive (see the debug branch).
- Two peers in the **same LAN** (same public IP, same /24):
  - `dk20` – 192.168.91.120 – linux/amd64
  - `w11-test1` – 192.168.91.193 – windows/amd64
- Both peers have `DisableIPv6Discovery: true` and – since the kernel/UDPMux
  still binds IPv6 regardless – additional netfilter/WFP rules that drop
  any IPv6 UDP **and** IPv4 UDP directly between the two peer addresses.
  This forces ICE host-candidate checks to fail so that the FailedTimeout
  path (which rotates sessions via `closeAgent`) actually fires. Relay
  paths are untouched; traffic to the NetBird management/signal/relay
  servers is unaffected.

The "force ICE failure" step is important: on a happy-path same-LAN pair
(e.g. both peers with IPv6 or with free IPv4 UDP) ICE succeeds fast enough
that `closeAgent` never runs and the loop never starts. The loop is the
post-failure steady state.

## Log fields

```
+NNNNms   peer        EVENT                  key=val ...
```

- `+NNNNms` – milliseconds since the first captured event (both peers merged
  on a common monotonic axis; dk20 is in UTC, w11-test1 in +02:00, both
  normalised to UTC before merging).
- `SEND_OFFER` / `RECV_OFFER` / `RECV_ANSWER` – handshaker signaling events.
- `OnNewOffer ENTER` – the state snapshot at the beginning of `OnNewOffer`
  (`incoming`, `local_before`, `stored_remote`, `agent_nil`, `connecting`).
- `DECISION=` – which branch was taken (`create_first_agent` / `recreate` /
  `skip_dedup`).
- `agent_created` – new ICE agent installed, new session stored.
- `closeAgent` – `connect()` goroutine finished (ICE Failed/Disconnected/
  Closed), fresh local session generated, `remoteSessionID` cleared.

## Full chronological trace

Captured on stock `upstream/main` at 3098f48b2. The timeline starts at
the first `SEND_OFFER`; the loop is fully visible within 110 seconds.

```
+     0ms  dk20       SEND_OFFER local_session=e7737d235f
+    43ms  dk20       RECV_ANSWER remote_session=5a85359990
+    43ms  dk20       OnNewOffer ENTER incoming=5a85359990 local_before=e7737d235f stored_remote= agent_nil=true connecting=false
+    43ms  dk20       DECISION=create_first_agent incoming=5a85359990 local=e7737d235f
+    44ms  dk20       agent_created local=e7737d235f stored_remote=5a85359990 connecting=true

# --- ~12s later: ICE FailedTimeout fires on dk20, closeAgent runs ---
+ 12095ms  dk20       closeAgent old_local=e7737d235f new_local=e21d611f3a cleared_remote=5a85359990 session_changed=false

# dk20 sends offer with its NEW post-closeAgent local session
+ 20554ms  dk20       SEND_OFFER local_session=e21d611f3a
+ 20596ms  dk20       RECV_ANSWER remote_session=5056316402
+ 20596ms  dk20       OnNewOffer ENTER incoming=5056316402 local_before=e21d611f3a stored_remote= agent_nil=true connecting=false
+ 20596ms  dk20       DECISION=create_first_agent incoming=5056316402 local=e21d611f3a
+ 20597ms  dk20       agent_created local=e21d611f3a stored_remote=5056316402 connecting=true

# Both sides' guards fire inside the same agent lifetime – the intra-agent
# dedup path is fine, these offers are ignored:
+ 22967ms  dk20       SEND_OFFER local_session=e21d611f3a
+ 24705ms  dk20       SEND_OFFER local_session=e21d611f3a
+ 28086ms  dk20       SEND_OFFER local_session=e21d611f3a

# Another ~12s, another closeAgent on dk20
+ 32625ms  dk20       closeAgent old_local=e21d611f3a new_local=bd2c5ce78d cleared_remote=5056316402 session_changed=false

+ 34926ms  dk20       SEND_OFFER local_session=bd2c5ce78d
+ 34927ms  dk20       RECV_OFFER remote_session=a8877b548b
+ 34927ms  dk20       OnNewOffer ENTER incoming=a8877b548b local_before=bd2c5ce78d stored_remote= agent_nil=true connecting=false
+ 34927ms  dk20       DECISION=create_first_agent incoming=a8877b548b local=bd2c5ce78d
+ 34927ms  dk20       SEND_ANSWER local_session=bd2c5ce78d
+ 34927ms  dk20       agent_created local=bd2c5ce78d stored_remote=a8877b548b connecting=true

# This is the dedup working as intended – same-agent retransmit:
+ 34968ms  dk20       OnNewOffer ENTER incoming=a8877b548b local_before=bd2c5ce78d stored_remote=a8877b548b agent_nil=false connecting=true
+ 34968ms  dk20       DECISION=skip_dedup reason=session_matches incoming=a8877b548b stored_remote=a8877b548b
+ 34968ms  dk20       RECV_ANSWER remote_session=a8877b548b

+ 45782ms  dk20       SEND_OFFER local_session=bd2c5ce78d
+ 47010ms  dk20       closeAgent old_local=bd2c5ce78d new_local=ca6e8d0d7b cleared_remote=a8877b548b session_changed=false
+ 47388ms  dk20       SEND_OFFER local_session=ca6e8d0d7b
+ 50813ms  dk20       SEND_OFFER local_session=ca6e8d0d7b

# --- w11-test1 side becomes visible (Windows netbird.exe buffers early lines) ---
+ 52914ms  w11-test1  SEND_OFFER local_session=3e3d2cf6c1
+ 52947ms  w11-test1  RECV_ANSWER remote_session=ca6e8d0d7b
+ 52947ms  w11-test1  OnNewOffer ENTER incoming=ca6e8d0d7b local_before=3e3d2cf6c1 stored_remote= agent_nil=true connecting=false
+ 52947ms  w11-test1  DECISION=create_first_agent incoming=ca6e8d0d7b local=3e3d2cf6c1
+ 52955ms  w11-test1  agent_created local=3e3d2cf6c1 stored_remote=ca6e8d0d7b connecting=true

+ 53714ms  dk20       RECV_OFFER remote_session=3e3d2cf6c1
+ 53714ms  dk20       OnNewOffer ENTER incoming=3e3d2cf6c1 local_before=ca6e8d0d7b stored_remote= agent_nil=true connecting=false
+ 53714ms  dk20       DECISION=create_first_agent incoming=3e3d2cf6c1 local=ca6e8d0d7b
+ 53714ms  dk20       SEND_ANSWER local_session=ca6e8d0d7b
+ 53714ms  dk20       agent_created local=ca6e8d0d7b stored_remote=3e3d2cf6c1 connecting=true

# Both peers now have lockstep ~12s lifetimes, rotating in sync:
+ 64998ms  w11-test1  closeAgent old_local=3e3d2cf6c1 new_local=22a0af9a10 cleared_remote=ca6e8d0d7b session_changed=false
+ 65798ms  dk20       closeAgent old_local=ca6e8d0d7b new_local=f5b6935722 cleared_remote=3e3d2cf6c1 session_changed=false

+ 68923ms  w11-test1  RECV_OFFER remote_session=f5b6935722
+ 68923ms  w11-test1  OnNewOffer ENTER incoming=f5b6935722 local_before=22a0af9a10 stored_remote= agent_nil=true connecting=false
+ 68923ms  w11-test1  DECISION=create_first_agent incoming=f5b6935722 local=22a0af9a10
+ 68932ms  w11-test1  SEND_ANSWER local_session=22a0af9a10
+ 68932ms  w11-test1  agent_created local=22a0af9a10 stored_remote=f5b6935722 connecting=true

+ 69689ms  dk20       SEND_OFFER local_session=f5b6935722
+ 69732ms  dk20       RECV_ANSWER remote_session=22a0af9a10
+ 69732ms  dk20       OnNewOffer ENTER incoming=22a0af9a10 local_before=f5b6935722 stored_remote= agent_nil=true connecting=false
+ 69732ms  dk20       DECISION=create_first_agent incoming=22a0af9a10 local=f5b6935722
+ 69733ms  dk20       agent_created local=f5b6935722 stored_remote=22a0af9a10 connecting=true

+ 81013ms  w11-test1  closeAgent old_local=22a0af9a10 new_local=838fdcd663 cleared_remote=f5b6935722 session_changed=false
+ 81776ms  dk20       closeAgent old_local=f5b6935722 new_local=c5df0189a9 cleared_remote=22a0af9a10 session_changed=false

+ 96180ms  w11-test1  RECV_OFFER remote_session=c5df0189a9
+ 96180ms  w11-test1  OnNewOffer ENTER incoming=c5df0189a9 local_before=838fdcd663 stored_remote= agent_nil=true connecting=false
+ 96180ms  w11-test1  DECISION=create_first_agent incoming=c5df0189a9 local=838fdcd663
+ 96189ms  w11-test1  SEND_ANSWER local_session=838fdcd663
+ 96189ms  w11-test1  agent_created local=838fdcd663 stored_remote=c5df0189a9 connecting=true

+ 96947ms  dk20       SEND_OFFER local_session=c5df0189a9
+ 96988ms  dk20       RECV_ANSWER remote_session=838fdcd663
+ 96988ms  dk20       OnNewOffer ENTER incoming=838fdcd663 local_before=c5df0189a9 stored_remote= agent_nil=true connecting=false
+ 96988ms  dk20       DECISION=create_first_agent incoming=838fdcd663 local=c5df0189a9
+ 96989ms  dk20       agent_created local=c5df0189a9 stored_remote=838fdcd663 connecting=true

+108266ms  w11-test1  closeAgent old_local=838fdcd663 new_local=2b6ed58608 cleared_remote=c5df0189a9 session_changed=false
+109038ms  dk20       closeAgent old_local=c5df0189a9 new_local=452cfe710b cleared_remote=838fdcd663 session_changed=false
```

## Session-ID timeline (7 cycles in 109s)

| Cycle | dk20 local session | w11-test1 local session |
|-------|--------------------|-------------------------|
| 0     | e7737d235f         | 5a85359990              |
| 1     | e21d611f3a         | 5056316402              |
| 2     | bd2c5ce78d         | a8877b548b              |
| 3     | ca6e8d0d7b         | 3e3d2cf6c1              |
| 4     | f5b6935722         | 22a0af9a10              |
| 5     | c5df0189a9         | 838fdcd663              |
| 6     | 452cfe710b         | 2b6ed58608              |

Every ~12s (= `iceFailedTimeoutDefault`) **both** peers rotate their session
IDs, and the next offer/answer the other side receives carries that rotated
session.

## Where session-ID dedup fails

Every `OnNewOffer ENTER` log in the steady-state loop shows

```
stored_remote=  agent_nil=true  connecting=false
```

i.e. the dedup check

```go
if w.remoteSessionID == *remoteOfferAnswer.SessionID {
    // skip
}
```

can never match: `w.remoteSessionID` has been reset to the empty string by
the previous `closeAgent` (`worker_ice.go:338`), while the incoming
`SessionID` is the remote's freshly rotated value. Dedup is *not* failing
because of a bug in its comparison – it is simply never reachable, because
both sides have `closeAgent`-cleared `remoteSessionID` **and** a rotated
local session at the exact moment the next offer arrives.

That is the core of the loop:

1. `closeAgent` on A – `remoteSessionID=""`, new `sessionID`.
2. `closeAgent` on B – `remoteSessionID=""`, new `sessionID`.
3. A's next `SEND_OFFER` uses the new local session; B receives it, sees
   empty stored-remote, has `agent_nil=true`, so it goes to
   `create_first_agent` and stores the new remote.
4. In parallel, B's `SEND_OFFER` does the same thing to A.
5. Both agents start the 12s ICE attempt, both fail, both rotate, back to
   step 1.

Because the rotation on both sides is triggered by the same global signal
(the ICE FailedTimeout of the previous round, which fires within a few
hundred milliseconds of each other), the cycles stay phase-locked for as
long as ICE keeps failing. In the captured run the two `closeAgent` events
are 763 ms, 800 ms, 772 ms, and 772 ms apart across four cycles.

## Does the session-ID mechanism fulfil its two documented purposes?

Yes, both still work inside a single agent lifetime. The merged trace at
`+34927ms/+34968ms` shows the "duplicate offer on ICE+relay paths"
dedup firing correctly (`DECISION=skip_dedup reason=session_matches`). And
the "crash-recovery after a silent remote restart" path is preserved – a
remote with a new session ID still tears down the old agent. The loop is
specifically the **simultaneous** case where *both* sides rotate inside the
same ICE failed-timeout window.

## Current PR #5805 behaviour

With `agentLoopProtectionWindow` applied, the first rotation on side A
will still happen (A cannot see B's session rotation in advance), but
within the protection window A refuses to tear its freshly-rebuilt agent
down a second time. That one-sided stability is enough to let the guard
settle, because the *other* side then does observe a stable remote
session on its next offer and its dedup path kicks in normally. After the
protection window expires, the "new session ID replaces current agent"
semantics are restored, so crash-recovery is delayed by at most
`agentLoopProtectionWindow` (currently 3 s), not by the ICE FailedTimeout.

## Files in this bundle

- `dk20.log` – raw `TRACE-PR5805` lines from dk20, UTC.
- `w11-test1.log` – raw `TRACE-PR5805` lines from w11-test1, +02:00.
- `merged.log` – both peers merged on a common UTC axis, offsets in ms
  from the first event.
- `analysis.md` – this document.

The instrumentation patch (against `upstream/main` at 3098f48b2) and the
reproduction scripts are on branch `debug/ice-session-trace-for-pr5805`.

pappz:
Thank you for your analysis. I tried to reproduce the issue to confirm the original report. In a lab environment, I introduced 1, 1.5, 3, etc. second delays in the connection. The P2P connection always succeeded and never got stuck in a loop. Could you please provide instructions on how I can repeat your tests?

MichaelUray:
@pappz Thank you for the thorough review and the back-and-forth on this PR. Your feedback shaped the implementation considerably — dropping the candidate buffer, replacing the unconditional session-ID skip with a time-windowed guard, and pushing for proper repro evidence all made the change much more defensible than what I originally opened.

After today's comprehensive re-testing on real hardware, I'm closing this PR. Here's the analysis.

## Test setup (2026-04-25)

Six Cudy TR3000 routers (OpenWrt 25.12.2-FWR, aarch64, kmod-wireguard / userspace WG) joined into one NetBird group with an `all-to-all` policy. UDP between same-NAT peer LAN IPs was blocked at `nft` to force the failure mode. `DisableIPv6Discovery=true`. Each round ran 90 s with `--log-level debug`.

| Peer | Site / WAN |
|---|---|
| 5723C, 572A2, 57262 | Graz home (same public NAT, 91.0/24 LAN) |
| 5731A | Graz home, LTE/CGNAT WAN |
| D95820, 572E2 | Remote site (same public NAT, 0.0/24 LAN) |

Three rounds, identical setup, only the binary swapped.

## Results

| Round | Binary | OnNewOffer (range across 6 routers) | ICE Failed | ICE Closed | recreate-agent | `skipping new offer` (#5805 marker) |
|---|---|---|---|---|---|---|
| 1 | vanilla 0.67.4 | **21 – 27** | **23 – 27** | **23 – 27** | 21 – 27 | 0 |
| 2 | this PR (`pr5805`-rebased on main, sha `90b59d944`) | 15 – 19 | 18 – 20 | 18 – 20 | 15 – 19 | 0 |
| 3 | upstream main `d6f08e484` (#5820 merged, **without** #5805) | 15 – 21 | 19 – 20 | 19 – 20 | 15 – 21 | 0 |

ICE Connected was 0 in all rounds (UDP fully blocked, all paths relayed — that's by design; the test isolates the reconnect-loop dynamics, not the relay/p2p selection).

## What this means

1. **Vanilla 0.67.4 reproduces the loop on every router**, including the remote site — so the bug isn't environment-specific to my home setup.
2. **PR #5820 alone** (your "Suppress ICE signaling" change merged on Apr 21) **already cuts the churn by ~25 %** and eliminates the lockstep character. This matches your hypothesis that the original report was driven by signaling-level amplification rather than the OnNewOffer-tear-down path.
3. **This PR on top of main shows no measurable additional reduction.** Round 2 vs Round 3 are statistically indistinguishable with this sample.
4. **The `agentLoopProtectionWindow` (3 s) never fires in practice** — the observed cycle period is consistently > 3 s, so the `skipping new offer` log line stays at 0 even with the fix in place. The defensive logic exists but doesn't engage in the scenarios I can reproduce.

## Closing rationale

Given that:

- the visible behavior (lockstep loop, session-ID rotation in stock 0.67.4) is already addressed by #5820,
- the additional guard in this PR adds machinery without an observable effect in the test scenarios I have,
- and you raised reasonable concerns about the session-ID mechanism semantics that I haven't fully resolved,

it doesn't make sense to keep this open as a candidate for merge. If the original lockstep pattern resurfaces on a future build, the right starting point will be a fresh investigation against then-current `main`, not a rebase of this branch.

## For the lab repro you asked about

The reason artificial latency injection didn't reproduce the loop in your lab is that the trigger isn't latency — it's complete UDP loss between the two peer LAN endpoints (so all ICE candidates fail FailedTimeout) combined with both peers being on the same public NAT (so neither side ever falls back into a "different routes" branch that would break symmetry). Reproduction recipe for posterity:

```sh
# On both peers in the same LAN, behind the same NAT:
nft add table inet ice_test
nft 'add chain inet ice_test input { type filter hook input priority -100; }'
nft 'add chain inet ice_test output { type filter hook output priority -100; }'
nft "add rule inet ice_test input  ip saddr <other_peer_lan_ip> meta l4proto udp drop"
nft "add rule inet ice_test output ip daddr <other_peer_lan_ip> meta l4proto udp drop"

# Disable IPv6 discovery so it doesn't paper over the v4 failure:
sed -i 's|"DisableIPv6Discovery": false|"DisableIPv6Discovery": true|' /etc/netbird/config.json
```

On stock 0.67.4 you'll see lockstep `OnNewOffer` → `ICE ConnectionState Failed` → `Closed` → `recreate ICE agent` cycles every 12-22 s, with session IDs rotating. On current `main` (with #5820) the cycle still happens but no longer in lockstep and at ~75 % of the rate.

Thanks again for sticking with this — both for the technical pushback that made me prove the claims properly, and for getting #5820 across the line.

— Michael

Issue #4657, opened 2025-10-17
Bug: iOS Client loses all routes when Posture Checks are enabled


Miki84:
Description
When using Posture Checks to prevent routing conflicts in home networks, the iOS client loses all routes and becomes non-functional, while the Windows client works as expected.

Current Behavior

Windows Client: Posture Checks work correctly - routes are properly managed based on network detection
iOS Client: All routes disappear when Posture Checks are enabled, regardless of which network the device is connected to

Steps to Reproduce

Set up self-hosted Netbird with "Connect on Startup" enabled
Configure Posture Check to block peer network range 192.168.1.0/24
Connect iOS client to Netbird
Observe that all routes disappear from the iOS client

Expected Behavior
The iOS client should handle Posture Checks the same way as the Windows client:

Routes should be dynamically managed based on the current network environment
When NOT in the blocked network range (e.g., at work), routes should be established normally
When IN the blocked network range (e.g., at home on 192.168.1.0/24), specific routes should be blocked to prevent routing conflicts

Context/Use Case
Problem being solved:
When at home, the Netbird connection takes priority over the local LAN/WiFi connection due to lower routing metrics. This causes traffic to home network devices (192.168.1.0/24) to be routed through the Netbird tunnel instead of using the direct local connection, which is inefficient and causes connectivity issues.
Workaround strategy:
Using Posture Checks to block the peer network range 192.168.1.0/24 prevents this routing conflict by detecting when the client is on the home network and adjusting routes accordingly.
Environment

Netbird Version Windows: 0.59.6
Netbird Version iOS: 0.55.1
Deployment: Self-hosted
Affected Platform: iOS
Working Platform: Windows
Network Configuration:

Home network: 192.168.1.0/24
Posture Check: Blocking peer network range 192.168.1.0/24

Impact
This makes the iOS client unusable when Posture Checks are required, forcing users to either:

Disable Posture Checks (resulting in routing conflicts at home)
Manually disable/enable VPN connection (defeats the purpose of "Connect on Startup")

Additional Information
The same Posture Check configuration works perfectly on Windows, suggesting this is an iOS-specific implementation issue with how routes are handled when Posture Checks are active.

levindecaro:
same here, any peer network range posture policy enforcement define in the network route will disable the route in iOS, however MacOS, Windows are working as expected.

Netbird Server version .0.59.6
Netbird iOS version: 0.0.11

robotkid450:
This is also the case on Android.
[...]

Issue #3968, opened 2025-06-12
Posture checks peer network range failed on iphone


mannguyen0107:
**Describe the problem**

I have setup policy for my personal group (which my iPhone peer is apart of) to connect to my homelab server group with posture checks peer network range for my home server subnets so if I'm connecting to my wifi the policy doesnt apply. However even if I'm on 5G or any wifi that the subnet is not apart of the subnets defined in the posture check the policy still block my iPhone because I do not see my home lab peers when I'm supposed to. The policy works if I'm on my MacBook connecting to a different wifi with subnet not in the posture check

**To Reproduce**

1. Assign 2 peers (1 iPhone and another peer) to 2 different groups A and B
2. Create Posture check peer network range (blocked) and put in the subnet of home wifi for example 192.168.1.0/24
3. Create a policy that allow group A (iPhone peer apart of) to connect to group B if posture check is satisfy 
4. Connect iPhone to 5G and connect to netbird
5. Do not see peer(s) in group B in the list of connected peers

**Expected behavior**

Expect to see peer(s) in group B in the list of connected peers on netbird iPhone App

**Are you using NetBird Cloud?**

I'm using self-host NetBird's control plane.

**NetBird version**

netbird server version: 0.46.0
iphone netbird agent version: 0.36.4-dev 

**Is any other VPN software installed?**

no

**Debug output**

To help us resolve the problem, please attach the following anonymized status output

  netbird status -dA

Create and upload a debug bundle, and share the returned file key:

  netbird debug for 1m -AS -U

*Uploaded files are automatically deleted after 30 days.*

Alternatively, create the file only and attach it here manually:

  netbird debug for 1m -AS

Not relevant with iPhone app?

**Screenshots**

**Additional context**

**Have you tried these troubleshooting steps?**
- [x] Reviewed [client troubleshooting](https://docs.netbird.io/how-to/troubleshooting-client) (if applicable)
- [x] Checked for newer NetBird versions
- [x] Searched for similar issues on GitHub (including closed ones)
- [x] Restarted the NetBird client
- [x] Disabled other VPN software
- [x] Checked firewall settings

citgot:
I have the same problem on Android, so you are not alone.

robotkid450:
Also having the same issue on Android. The peer network range posture check seems to always deny access.

MichaelUray:
Same issue on my Android phone.
Did that actually work before with older versions of the app on Android or iOS?

apollo13:
[...]

PR #5807, opened 2026-04-06
[client] Fix NetworkAddresses discovery on Android and re-sync on network change

MichaelUray:
## Summary
Two issues prevented posture checks from working on Android:

1. **`net.Interfaces()` broken on Android 11+**: SELinux blocks `NETLINK_ROUTE` sockets ([golang/go#40569](https://github.com/golang/go/issues/40569)). Use `wlynxg/anet` (already an indirect dependency via pion/ice) as drop-in replacement on Android.
2. **MAC address filter skipped all interfaces**: Android 10+ returns empty `HardwareAddr` for all interfaces, causing the MAC-based filter to skip everything. Removed the filter since loopback is already filtered by IP and MAC is only metadata not used for posture check matching.
3. **No re-sync on network change**: When a mobile device switches between WiFi and cellular, NetworkAddresses changed but were never re-synced with the management server. Added a 10s network address watcher goroutine with 30s debounce that detects changes and triggers `SyncMeta`.

Includes unit tests for `networkAddressesEqual()` and `networkAddresses()`.

Fixes #3614 #2962

## Checklist
- [x] Bug fix
- [x] Create tests that fail without the change: added unit tests for address comparison and discovery
- [x] Documentation is **not needed** for this change (internal client behavior, no user-facing API or configuration change)

By submitting this pull request, I confirm that I have read and agree to the terms of the [Contributor License Agreement](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT.md).

## Related Issues

Related #3968 — Posture checks peer network range failed on iPhone (Android side)
Related #4657 — iOS Client loses all routes when Posture Checks enabled (Android side)
Related #5810 — Feature Request: Expose peer LAN network addresses via API

## Summary by CodeRabbit

* **New Features**
  * Android can trigger immediate metadata re-syncs when the underlying network changes.
  * Periodic background detection of local network changes prompts resyncs.

* **Bug Fixes**
  * More reliable, debounced detection of network-address changes to reduce missed or duplicate updates.
  * Network discovery now respects caller context, captures interface MACs, and avoids loopback addresses or duplicates.

* **Tests**
  * Added tests for address comparison and uniqueness.

* **Chores**
  * Updated .gitignore to exclude build and generated docs directories.


client/system/network_addresses_android.go
```diff
@@ -0,0 +1,17 @@
...
+package system
+
+import (
+	"net"
+
+	"github.com/wlynxg/anet"
+)
+
+func getNetInterfaces() ([]net.Interface, error) {
+	return anet.Interfaces()
```
pappz: Did you consider to use `stdnet.ExternalIFaceDiscover` from mobile_dependencies.go?

MichaelUray: Done — switched to `stdnet.ExternalIFaceDiscover` from mobile_dependencies.go in commit 9fb609ef. The context propagation was a bit tricky (Android callbacks originate outside the engine context), so I injected the IFaceDiscover into the root context in 3dc24af0 to cover the login path as well.

MichaelUray:
Thanks @pappz for the review! I have reworked the PR based on your feedback and the coderabbitai suggestions:

- **Use `stdnet.ExternalIFaceDiscover`** (commit 9fb609ef): Switched to the existing ExternalIFaceDiscover API from mobile_dependencies.go instead of the custom Android approach.
- **Inject IFaceDiscover into root context** (commit 3dc24af0): Needed because Android network-change callbacks fire outside the engine context; this ensures the login path also has access.
- **Build resync payload atomically** (commit 3ae6931e): Rebuilt `GetInfoWithChecks()` snapshot commit semantics so we only update state after `SyncMeta` succeeds.
- **Removed debounce on explicit callbacks** (commit 3ae6931e): Platform callbacks now bypass the watch-interval debounce.
- **Redacted raw interface payloads** from parse warnings (commit 3ae6931e).
- **Extracted the watch interval** into a named constant for clarity (commit b4a69c2c).

Ready for another look.

MichaelUray:
@pappz friendly ping — all review feedback has been addressed. Ready for another look whenever you have time. Thanks!

pappz:
@MichaelUray Thank you for the PR!

Network address watcher may be unnecessary

The startNetworkAddressWatcher() (10s polling + 30s debounce) and ResyncNetworkAddresses() add significant complexity (new goroutine, two mutexes, debounce logic), but I believe the existing reconnection flow already covers this case. When the network changes (WiFi ↔ cellular), the network monitor triggers ErrResetConnection, which restarts the engine. The restart goes through login/sync, which sends fresh GetInfoWithChecks including up-to-date NetworkAddresses. Can you describe a scenario where the network addresses change but the network monitor doesn't fire? I prefer to remove this logic.

The system info part looks correct (I did a quick test). I don’t prefer the context value pattern, but I don’t have a better idea at the moment. I hope that sooner or later the fix for the Android-related network interface syscalls will be merged into the Go standard library.

MichaelUray:
@pappz You were right — the watcher is unnecessary. I've removed it entirely from the PR.

Your comment prompted me to dig deeper into *why* the network change wasn't triggering an engine restart on Android. The root cause turned out to be on the **Java side**, not the Go side:

### Root Cause: `ConcreteNetworkAvailabilityListener` bug

The `NetworkChangeDetector` in the Android app uses `ConnectivityManager.NetworkCallback` to detect network changes. But `ConcreteNetworkAvailabilityListener` only notified the `EngineRestarter` in two cases:
- `onNetworkAvailable`: only when **WiFi** became available
- `onNetworkLost`: only when **WiFi** was lost AND mobile was already registered

This missed a critical timing scenario during WiFi → Mobile switches:
1. User disables WiFi
2. `onLost(WIFI)` fires, but `isMobileAvailable` is still `false` (mobile not yet registered) → **no notification**
3. `onAvailable(MOBILE)` fires, but the code only checks for WiFi → **no notification**
4. Engine never restarts, stale NetworkAddresses remain on the management server
5. Posture check incorrectly blocks the peer

### Testing

Tested on Samsung Galaxy S26 (SM-S942B, Android 16) with a custom debug APK:

1. **Without fix:** WiFi → Mobile switch did not trigger engine restart. Management server kept old WiFi NetworkAddresses. Posture check `peer_network_range_check` (deny `192.168.91.0/24`) blocked the peer even though it was on mobile data.

2. **With fix (netbirdio/android-client#165):** WiFi → Mobile correctly triggers engine restart. Fresh NetworkAddresses sent. Posture check passes. Route to `192.168.91.0/24` via r1-pve5 becomes available. Switching back to WiFi also triggers correctly — posture check blocks the route (device is local), traffic goes directly via LAN.

### Go-side network monitor

The Go-side `NetworkMonitor` (`monitor_mobile.go`) is a no-op on Android — it returns `"network monitor not supported on mobile platforms"` immediately. All network change detection relies entirely on the Java-side `ConnectivityManager.NetworkCallback` → `EngineRestarter`. This is working as designed — the Java side is the right place for this on Android.

### Summary

- **Watcher removed** from this PR ✓
- **Root cause fixed** in netbirdio/android-client#165 (Java-side listener)
- **No Go-side changes needed** for network change detection on Android

Thank you for pushing back on the watcher — it led to finding and fixing the actual bug.

MichaelUray:
Addressed the three failing checks:

- **codespell**: fixed `unparseable` → `unparsable` in the Android warning log.
- **Client / Unit (FreeBSD)**: `info_freebsd.go` still called the old `networkAddresses()` signature; now passes `ctx` to match the other platforms.
- **Require docs PR URL or explicit "not needed"**: updated the PR body to use the exact template checkbox wording so the workflow recognises the selection.

The remaining `FAIL` in the Client (Docker) / Unit run was a flaky goroutine-leak in `flow/client_test.TestReceive_ProtocolErrorStreamReconnect` (unrelated to this PR). Should clear on the next run, but happy to investigate if it persists.

MichaelUray:
Follow-up after syncing the latest upstream `main`:

- Merged `upstream/main` into the branch (picks up #5900 which added `networkAddresses()` to `info_ios.go`).
- Adjusted `info_ios.go` to pass `ctx` to match the updated signature.

### CI status on the latest push

32 of 33 required checks are green (including codespell, all docs checks, Quality Gate, Client (Docker) / Unit, Client Build on iOS / Windows / macOS / Linux, etc.).

The one remaining red check is **Client / Unit** (macOS runner), which timed out after 5m00s inside `client/internal/dns.TestDNSPermanent_updateUpstream`. The stack dump shows the hang is in `resolver.LookupHost(ctx, "netbird.io")` — an external DNS lookup from the CI runner — not in any code path this PR touches (the PR is scoped to `client/system/*` and `client/internal/engine.go` network-address handling). This looks like a flaky pre-existing test; a re-run of just that one job should clear it. I don't have rerun rights on this repo — would one of the maintainers be able to trigger a retry?

MichaelUray:
Friendly re-ping @pappz — would you have a moment to take another look? The PR has been reworked per your feedback (the ICE-rebuild-on-network-change pattern was changed to use the in-place agent recreate approach you suggested), the CodeRabbit findings were addressed, and CI is currently green (26 checks passing). Happy to make further adjustments — just let me know.

apollo13:
@pappz any chance to get a final review here?

PR #5859, opened 2026-04-12
[management] fix resource and router changes not propagating to peers

### LINKED ISSUE #5813: Update to Routing Peers not applying on NetBird clients
**Describe the problem**

We updated the management service from 0.66.0 to 0.67.2 (also tested 0.67.4) and within this upgrade a bug was introduced. Previously, if we'd change any information here:

Image

it would trigger a `Network map updated`-event on all clients. Since 0.67.2 (perhaps earlier, haven't tested exactly when it was introduced) this is not the case anymore. This made it so changes to the routing peers were not properly applied on clients. We had to restart the management service to apply the changes to all clients.

We downgraded the management service back to 0.66.0 to fix the issue. Clients were not downgraded or updated and we only had to downgrade the management service to fix the issue.

The reason why we relied on this as I wanted to drain traffic from one peer, and I did not expect traffic to still flow through that peer when I was working on it - resulting in downtime due my changes.

**To Reproduce**

1. Go to a Network's routing peers
2. If you have 2 routing peers, disable one of them.

**Expected behavior**

I'd expect the following to appear on all clients:
```
  [INFO] SYSTEM (59d5313b-8b60-42e9-830a-2c0dc7b698db)
    Message: Network map updated
    Time: Now
```

But it does not happen. When rolling back to 0.66.0 it works again.

Self-hosted.

**NetBird version**

```
❯ netbird version
0.67.4
```

No.

**Debug output**

```
Events:
  [INFO] SYSTEM (05aa35a5-db43-4d2f-8b74-221da1c0617f)
    Message: Network map updated
    Time: 15 minutes, 4 seconds ago
  [INFO] SYSTEM (beba6a56-07f8-4048-a348-b77962887afe)
    Message: Network map updated
    Time: 7 minutes, 44 seconds ago
  [INFO] SYSTEM (c534e80d-d65e-4f60-8e08-e3b953f3028f)
    Message: Network map updated
    Time: 7 minutes, 37 seconds ago
  [INFO] SYSTEM (69adb3ff-edff-41b4-b525-a7c666d1255d)
    Message: Network map updated
    Time: 7 minutes, 35 seconds ago
  [INFO] SYSTEM (c74bad56-2acd-4fbe-ac1f-ffa641df3202)
    Message: Network map updated
    Time: 5 minutes, 52 seconds ago
  [INFO] SYSTEM (66ece6f7-49bf-4821-bbaf-f01bbd6d5f23)
    Message: Network map updated
    Time: 3 minutes, 46 seconds ago
  [INFO] SYSTEM (30bed1be-647d-4901-a98a-2030e3545bfc)
    Message: Network map updated
    Time: 3 minutes, 36 seconds ago
  [INFO] SYSTEM (b86719ec-9b1a-4112-b9a1-a5ea17d2f4a2)
    Message: Network map updated
    Time: 3 minutes, 32 seconds ago
  [INFO] SYSTEM (59d5313b-8b60-42e9-830a-2c0dc7b698db)
    Message: Network map updated
    Time: 1 minute, 3 seconds ago
  [INFO] SYSTEM (d2a54878-7849-4821-854c-77ed13abc940)
    Message: Network map updated
    Time: 1 minute ago
OS: linux/amd64
Daemon version: 0.67.4
CLI version: 0.67.4
Profile: cloudbear
Management: Connected to https://mesh.anon-jdwc0.domain:443
Signal: Connected to https://mesh.anon-jdwc0.domain:443
Relays: 
  [stun:mesh.anon-jdwc0.domain:3478] is Available
  [stun:relay-ams.anon-jdwc0.domain:3478] is Available
  [turn:mesh.anon-jdwc0.domain:3478?transport=udp] is Available
  [rels://relay-ams.anon-jdwc0.domain:443] is Available
Nameservers: 
FQDN: wvanos-framework.anon-UwtoQ.domain
NetBird IP: 100.19.155.24/16
Interface type: Userspace
Quantum resistance: false
Lazy connection: false
SSH Server: Disabled
Networks: -
Peers count: 14/14 Connected
```

Wouter0100:
## Describe your changes

Goroutines in the networks package called `UpdateAccountPeers` with the HTTP request context. After the response was sent, the request context was cancelled, causing `GetAccountWithBackpressure` to return `ctx.Err()` and silently drop the peer notification.

Fix by passing `context.WithoutCancel(ctx)` to these goroutines so they survive the request lifecycle while still inheriting trace/log values.

Group handlers were unaffected because they call `UpdateAccountPeers` synchronously before the response is returned.

Curious to hear if this is as desired. I've verified the changes on my NetBird setup.

Fixes #5813

### Checklist
- [X] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Bugfix

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where peer updates could be interrupted when creating, updating, or deleting networks, resources, and routers.

### DIFF
--- a/management/server/networks/resources/manager.go
+++ b/management/server/networks/resources/manager.go
@@ -162,7 +162,7 @@ func (m *managerImpl) CreateResource(ctx context.Context, userID string, resourc
 		event()
 	}
 
-	go m.accountManager.UpdateAccountPeers(ctx, resource.AccountID)
+	go m.accountManager.UpdateAccountPeers(context.WithoutCancel(ctx), resource.AccountID)
 
 	return resource, nil
 }
@@ -270,7 +270,7 @@ func (m *managerImpl) UpdateResource(ctx context.Context, userID string, resourc
 		}
 	}()
 
-	go m.accountManager.UpdateAccountPeers(ctx, resource.AccountID)
+	go m.accountManager.UpdateAccountPeers(context.WithoutCancel(ctx), resource.AccountID)
 
 	return resource, nil
 }
@@ -352,7 +352,7 @@ func (m *managerImpl) DeleteResource(ctx context.Context, accountID, userID, net
 		event()
 	}
 
-	go m.accountManager.UpdateAccountPeers(ctx, accountID)
+	go m.accountManager.UpdateAccountPeers(context.WithoutCancel(ctx), accountID)
 
 	return nil
 }
--- a/management/server/networks/manager.go
+++ b/management/server/networks/manager.go
@@ -177,7 +177,7 @@ func (m *managerImpl) DeleteNetwork(ctx context.Context, accountID, userID, netw
 		event()
 	}
 
-	go m.accountManager.UpdateAccountPeers(ctx, accountID)
+	go m.accountManager.UpdateAccountPeers(context.WithoutCancel(ctx), accountID)
 
 	return nil
 }
--- a/management/server/networks/routers/manager.go
+++ b/management/server/networks/routers/manager.go
@@ -119,7 +119,7 @@ func (m *managerImpl) CreateRouter(ctx context.Context, userID string, router *t
 
 	m.accountManager.StoreEvent(ctx, userID, router.ID, router.AccountID, activity.NetworkRouterCreated, router.EventMeta(network))
 
-	go m.accountManager.UpdateAccountPeers(ctx, router.AccountID)
+	go m.accountManager.UpdateAccountPeers(context.WithoutCancel(ctx), router.AccountID)
 
 	return router, nil
 }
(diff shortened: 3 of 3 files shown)


Wouter0100:
I have rather limited experience in writing tests, so if this is a correct fix for the bug - I'll dive into that :)

management/server/networks/resources/manager.go
```diff
@@ -162,7 +162,7 @@ func (m *managerImpl) CreateResource(ctx context.Context, userID string, resourc
 		event()
 	}
 
-	go m.accountManager.UpdateAccountPeers(ctx, resource.AccountID)
+	go m.accountManager.UpdateAccountPeers(context.WithoutCancel(ctx), resource.AccountID)
```
jnfrati, line 165: Hey @Wouter0100, thanks for the PR!
Chatted with the team about this, and it feels like it would be better to move the "context.WithoutCancel" inside UpdateAccountPeers, so that we avoid running into this scenario on the future 👍

pascal-fischer: Also lets keep a timeout on this one of like 5 min (which is really high but gives enough buffer for slow environments under pressure) to not pile up requests forever

Wouter0100:
I see https://github.com/netbirdio/netbird/pull/5878/changes reverted the code that was causing it. Closing this.

PR #5878, opened 2026-04-14
[management] revert ctx dependency in get account with backpressure

crn4:
## Describe your changes
the pr revert ctx cancel dependency on GetAccountWithBackpressure 

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

bug fix 

## Summary by CodeRabbit

* **Bug Fixes**
  * Modified account request handling to change cancellation behavior during request transmission and result retrieval.

### DIFF
--- a/management/server/account_request_buffer.go
+++ b/management/server/account_request_buffer.go
@@ -63,20 +63,11 @@ func (ac *AccountRequestBuffer) GetAccountWithBackpressure(ctx context.Context, 
 
 	log.WithContext(ctx).Tracef("requesting account %s with backpressure", accountID)
 	startTime := time.Now()
+	ac.getAccountRequestCh <- req
 
-	select {
-	case <-ctx.Done():
-		return nil, ctx.Err()
-	case ac.getAccountRequestCh <- req:
-	}
-
-	select {
-	case <-ctx.Done():
-		return nil, ctx.Err()
-	case result := <-req.ResultChan:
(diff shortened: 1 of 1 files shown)


pascal-fischer (approved):

management/server/account_request_buffer.go
```diff
@@ -63,20 +63,11 @@ func (ac *AccountRequestBuffer) GetAccountWithBackpressure(ctx context.Context,
...
-	}
-
-	select {
-	case <-ctx.Done():
-		return nil, ctx.Err()
-	case result := <-req.ResultChan:
-		log.WithContext(ctx).Tracef("got account with backpressure after %s", time.Since(startTime))
-		return result.Account, result.Err
-	}
+	result := <-req.ResultChan
```
mlsmaycon, line 68: What happens if there is a database failure? Will we block indefinitely?

mlsmaycon (changes requested):

management/server/account_request_buffer.go
crn4: no, we have a condition for account == nil case https://github.com/netbirdio/netbird/pull/5878/changes#diff-64fdebe6d3ae051b1c1f2634ed6bd4ce69268aff565f57b55ed438eb80ba3892L103

mlsmaycon (approved):

management/server/account_request_buffer.go
mlsmaycon: ok, the account result in the same function would kick in

PR #5886, opened 2026-04-14
[management, client] Add management-controlled client metrics push

pappz:
## Describe your changes

Allow enabling/disabling client metrics push from the dashboard via account settings instead of requiring env vars on every client.

- Add MetricsConfig proto message to NetbirdConfig
- Add MetricsPushEnabled to account Settings (DB-persisted)
- Expose metrics_push_enabled in OpenAPI and dashboard API handler
- Populate MetricsConfig in sync and login responses
- Client dynamically starts/stops push based on management config
- NB_METRICS_PUSH_ENABLED env var overrides management when explicitly set
- Add activity events for metrics push enable/disable

https://github.com/netbirdio/dashboard/pull/613
https://github.com/netbirdio/docs/pull/701

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Release Notes

* **New Features**
  * Added an account-level setting (`metrics_push_enabled`) to enable or disable client metrics push for all account peers.
  * Management now sends metrics push configuration during login; clients apply it after login and when configuration updates are received.
  * Added activity events for metrics push enabled/disabled account changes.

* **Behavior Updates**
  * Clients still honor an explicit environment override (`NB_METRICS_PUSH_ENABLED`): when set, management changes are ignored.


bcmmbaga (approved):

lixmal (approved):

lixmal (dismissed):
Approving the client portion

bcmmbaga (dismissed):

management/server/types/settings.go
```diff
@@ -73,6 +73,9 @@ type Settings struct {
 	// For new accounts this defaults to the All group.
 	IPv6EnabledGroups []string `gorm:"serializer:json"`
 
+	// MetricsPushEnabled globally enables or disables client metrics push for the account
+	MetricsPushEnabled bool `gorm:"default:false"`
```
lixmal, line 77: This new setting won't take effect on Postgres deployments.

The Postgres read path in `management/server/store/sql_store.go` (`getAccountPgx`) loads account settings with a hand-written column list and manual row scan, not via gorm. `settings_metrics_push_enabled` is missing from that `SELECT` and `Scan`, so on Postgres this field is always read back as `false` and the toggle never reaches clients, even though it works in local SQLite dev.

Fix: add `settings_metrics_push_enabled` to the SELECT, add a `sql.NullBool` var, add it to the `Scan(...)`, and assign it the same way the other settings fields are (e.g. next to `sLocalMFAEnabled`).

client/internal/engine.go
```diff
@@ -1066,6 +1068,14 @@ func (e *Engine) handleFlowUpdate(config *mgmProto.FlowConfig) error {
 	return e.flowManager.Update(flowConfig)
 }
 
+func (e *Engine) handleMetricsUpdate(config *mgmProto.MetricsConfig) {
+	if config == nil {
+		return
+	}
+	log.Infof("received metrics configuration from management: enabled=%v", config.GetEnabled())
+	e.clientMetrics.UpdatePushFromMgm(e.ctx, config.GetEnabled())
```
lixmal: This starts the push under the per-connection engine context (`e.ctx`), but the login path in `connect.go` deliberately uses the daemon context so the push survives engine restarts.

`clientMetrics` lives for the whole daemon lifetime. `startPushLocked` sets `c.push` but only `stopPushLocked` clears it. So if the push is started from here and the engine later restarts, `e.ctx` is cancelled and the push goroutine exits, but `c.push` stays non-nil. On the next login, `UpdatePushFromMgm(c.ctx, true)` sees `c.push != nil` and returns early, so the push is never restarted, it stays dead until the setting is toggled off and back on.

Better to run the push under a daemon-scoped context, e.g. have `ClientMetrics` hold its own base context and derive the push context from that, so a per-cycle context can't leak into the push lifetime.

lixmal (approved):

mlsmaycon (approved):

PR #5905, opened 2026-04-16
DNS-aware dial for HTTP reverse proxy — use NetBird nameservers for hostname resolution

aturkenov:
### Problem

The HTTP reverse proxy transport (`dialWithDNSResolution` in dns.go) was a stub.  As a result, target URLs that contained **hostnames** (e.g. `https://myapp.netbird.cloud/`) could not be dialled — only literal IP addresses worked.  Users wanting to forward traffic to a domain couldn't do so through a resource group or peer proxy target.

### Solution

Route hostname resolution through NetBird's **own DNS infrastructure** (custom zones, nameserver groups configured in the management UI) instead of the host-OS resolver or a hardcoded list of DNS servers.

The embedded NetBird client already runs an internal DNS server (bound to the last IP of the WireGuard network on port 53 in netstack/userspace mode).  This server already knows about all NetBird-managed zones and upstream nameserver groups.  We now query it directly during each dial that involves a hostname.

current state:
image

expected result:
image

## Release Notes

* **New Features**
  * Added DNS server address and port retrieval capabilities to the client.
  * Enhanced network operations with automatic hostname resolution that uses the NetBird DNS server when available and falls back to the system default resolver.

### DIFF
--- a/client/embed/embed.go
+++ b/client/embed/embed.go
@@ -507,3 +507,14 @@ func (c *Client) getNet() (*wgnetstack.Net, netip.Addr, error) {
 
 	return nsnet, addr, nil
 }
+
+// GetDNSAddrPort returns the address of the NetBird DNS resolver for this client.
+// Returns the zero AddrPort and false if the client is not started or the DNS
+// server is not yet initialized.
+func (c *Client) GetDNSAddrPort() (netip.AddrPort, bool) {
+	engine, err := c.getEngine()
+	if err != nil {
+		return netip.AddrPort{}, false
+	}
+	return engine.GetDNSAddrPort()
+}
--- a/client/internal/dns/mock_server.go
+++ b/client/internal/dns/mock_server.go
@@ -53,6 +53,10 @@ func (m *MockServer) DnsIP() netip.Addr {
 	return netip.MustParseAddr("100.10.254.255")
 }
 
+func (m *MockServer) DnsAddrPort() netip.AddrPort {
+	return netip.AddrPortFrom(netip.MustParseAddr("100.10.254.255"), 53)
+}
+
 func (m *MockServer) OnUpdatedHostDNSServer(addrs []netip.AddrPort) {
 	// TODO implement me
 	panic("implement me")
--- a/client/internal/dns/server.go
+++ b/client/internal/dns/server.go
@@ -51,6 +51,8 @@ type Server interface {
 	Initialize() error
 	Stop()
 	DnsIP() netip.Addr
+	// DnsAddrPort returns the full address (IP + port) of the DNS resolver.
+	DnsAddrPort() netip.AddrPort
 	UpdateDNSServer(serial uint64, update nbdns.Config) error
 	OnUpdatedHostDNSServer(addrs []netip.AddrPort)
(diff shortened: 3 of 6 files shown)


lixmal:
What is this trying to solve? DNS already goes through netstack when dialing a domain

aturkenov:
I just need reverse proxy service that forwards all traffic to custom http domain target.
That domain can be resolved using existing dns nameservers and zones.
Thank you for responding! @lixmal

lixmal:
Can you elaborate on what's not working? The dial already resolves DNS internally with the same nameserver.

aturkenov:
@lixmal, its not a bug, its a feature, everything is working fine.
thank you for that!
current state:
image
expected result:
image
I just need reverse proxy service that forwards all traffic to custom http domain target.
thank you for answer!

aturkenov:
@lixmal, what to do next?

lixmal:
What are you trying to solve? From my understanding you just move the resolution out of the dialer (via netstack)

aturkenov:
I have a **NetBird client deployed in Docker Swarm**, and I now need to deploy **Traefik** as a reverse proxy.

The goal is to use **NetBird’s reverse proxy** with an **HTTP target inside the Docker Swarm overlay network**.
The problem is that I **cannot rely on static IP addresses** in the overlay network, so I need **DNS-based service discovery**.

### My idea:

1. Deploy a **custom DNS server** inside Docker Swarm.
2. Configure it to use Docker’s internal DNS (`127.0.0.11`) as the upstream resolver, so it can resolve service names in the overlay network.
3. Add this custom DNS server as a **nameserver in NetBird**.
4. Then create **NetBird reverse proxy services** that point to custom internal domains instead of IP addresses.

This would allow me to expose Docker services through NetBird reverse proxy using **stable domain names**, even when container IPs change.

Another possible improvement would be to use **NetBird DNS zones** for creating custom aliases as reverse proxy targets.
This would be useful when:

* multiple services share the same IP address,
* or when the target IP changes and you only need to update the DNS record instead of reconfiguring the reverse proxy.

I think this might solve an issue many users face when trying to use **custom DNS names as reverse proxy targets** in NetBird.

If this sounds useful, I can also search for similar discussions or feature requests from other users.

@lixmal, Thank you for responding!

HunterNyan:
Can this please be merged, I currently have a few different k8s clusters running with nginx-ingress that I'd like to use netbird reverse proxy to point at and this looks to fix the bug with reverse proxy's not using my local DNS to resolve the ingress IP

lixmal:
@HunterNyan what bug? Have you tried running the build from the PR?

aturkenov:
@lixmal, its not a bug! its feature request

lixmal:
So what feature is it implementing? Have you tested it and what is the difference? Like I said all I'm seeing is that you move DNS resolution outside the dial while the dial already consults the internal nameserver just fine.

aturkenov:
@lixmal, I haven't had a chance to test these changes yet. If needed, I can work on an alternative implementation that forwards the proxy target's peer IP address after it has been resolved internally by NetBird. Let me know if you'd like me to go that route.

PR #5915, opened 2026-04-18
[proxy] set session cookie path to root

### LINKED ISSUE #5626: Reverse Proxy doesn't set path of auth cookie
**Describe the problem**

The cookie set by the auth control of the reverse proxy is missing the `Path` options ([src](https://github.com/netbirdio/netbird/blob/80a8816b1dbb46d9dd3525f54abdb948ce04da66/proxy/internal/auth/middleware.go#L413)). According to MDN if the path is omitted, the browser will fill in the path component of the request URL ([docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#pathpath-value)). 

This causes issues with services such as [Immich](https://github.com/immich-app/immich)s share feature, as the share URL will look like `example.com/share/aaabbcccc`, but when opening it, it will also try to load resources from `example.com/_app/xyz`. Since the cookie is set with path `/share`, the requests to `/_app` will fail with `401`.

**To Reproduce**

1. Host some kind of webserver (for example this dummy one using `ncat`: `while true ; do ncat -l -p 1500 -c 'echo "HTTP/1.1 200 OK\n\n $(date)"'; done`)
2. Expose it as a reverse proxy service protected by authentication
3. Navigate to `/hello/world` and authenticate
4. Navigate to `/foo/bar` and observe that you need to authenticate again

**Expected behavior**

Authentication should only happen once.

NetBird Cloud

**NetBird version**

`0.65.3`

No

- [ ] Reviewed [client troubleshooting](https://docs.netbird.io/how-to/troubleshooting-client) (not applicable)

alsruf36:
## Describe your changes
Explicitly set `Path: "/"` on the session cookie written by the reverse proxy's auth middleware.

Without an explicit `Path`, browsers default to the path of the request URI that set the cookie ([MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#pathpath-value)). This means a user authenticating at `example.com/app/page` gets a cookie scoped to `/app`, and subsequent requests to sibling paths like `/api` or `/_app/*` are treated as unauthenticated — breaking SPAs, share links, and any service that fetches resources from multiple top-level paths.

This was originally flagged by CodeRabbit during review of #5587 as an outside-diff comment, but wasn't picked up before merge.

Closes #5626

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

- [x] Documentation is **not needed** for this change (one-line behavioral fix to existing helper; no user-facing API change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed session cookie scope to apply across all application paths, ensuring consistent session persistence for users navigating throughout the application.

### DIFF
--- a/proxy/internal/auth/middleware.go
+++ b/proxy/internal/auth/middleware.go
@@ -433,6 +433,7 @@ func setSessionCookie(w http.ResponseWriter, token string, expiration time.Durat
 	http.SetCookie(w, &http.Cookie{
 		Name:     auth.SessionCookieName,
 		Value:    token,
+		Path:     "/",
 		HttpOnly: true,
 		Secure:   true,
 		SameSite: http.SameSiteLaxMode,
--- a/proxy/internal/auth/middleware_test.go
+++ b/proxy/internal/auth/middleware_test.go
@@ -391,6 +391,15 @@ func TestProtect_SchemeAuthRedirectsWithCookie(t *testing.T) {
 	assert.Equal(t, http.SameSiteLaxMode, sessionCookie.SameSite)
 }
 
+func TestSetSessionCookieHasRootPath(t *testing.T) {
+	w := httptest.NewRecorder()
+	setSessionCookie(w, "test-token", time.Hour)
+
+	cookies := w.Result().Cookies()
+	require.Len(t, cookies, 1)
+	assert.Equal(t, "/", cookies[0].Path, "session cookie must be scoped to root so it applies to all paths")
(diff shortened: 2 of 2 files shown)


lixmal:
Thanks, it has been there from the beginning and needed more evaluation

lixmal (approved):

PR #5926, opened 2026-04-20
[client] Move macOS sleep detection into the daemon (purego)

lixmal:
## Describe your changes

Run macOS sleep/wake detection from the daemon instead of the UI so it keeps working when the UI isn't running (headless installs, UI quit/crashed). Also replaces the CGO implementation with purego.

- Move sleep detection from `client/ui` into the daemon; drop the `NotifyOSLifecycle` gRPC and regenerate proto
- Rewrite `client/internal/sleep/detector_darwin.go` using purego (IOKit + CoreFoundation via dlopen), with a single process-lifetime callback thunk and a dedicated OS-locked goroutine for the CFRunLoop
- Make Register synchronous: surface IOKit registration failures to the caller, and have Deregister await runloop setup so teardown can't skip cleanup on a fast register/deregister
- Add `NB_DISABLE_SLEEP_DETECTOR` env var to opt out of sleep detection
- Promote `ebitengine/purego` to a direct dependency

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

Internal refactor; user-facing behavior is unchanged (sleep/wake handling is daemon-internal).

## Summary by CodeRabbit

* **New Features**
  * Server now starts OS sleep/wake detection automatically; UI no longer initializes it and events are forwarded internally.

* **Refactor**
  * Sleep detection rewritten to use runtime dynamic bindings, serialized lifecycle transitions, safer callback dispatch, and robust startup/shutdown.

* **Chores**
  * Made native runtime binding dependency explicit in module configuration.

* **Breaking Change**
  * Removed legacy OS-lifecycle RPC — external lifecycle notifications must stop.

* **Config**
  * Detector can be disabled via NB_DISABLE_SLEEP_DETECTOR.


client/internal/sleep/detector_darwin.go
```diff
@@ -2,217 +2,320 @@
...
-	if !exists {
+	if _, exists := serviceRegistry[d]; !exists {
+		serviceRegistryMu.Unlock()
 		return nil
 	}
 
-	// cancel and remove this detector
-	d.cancel()
+	close(d.done)
 	delete(serviceRegistry, d)
```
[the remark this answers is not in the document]
lixmal: Fixed the re-register panic by moving the `done` channel creation into `Register`. Skipping the callback mutex — `SleepHandler` already serializes these. Added recovers in a few places while I was in there.

client/internal/sleep/detector_darwin.go
```diff
@@ -2,217 +2,352 @@
...
+	err := runLoopErr
+	if err != nil {
+		delete(serviceRegistry, d)
+		runLoopReady = nil
+	}
+	serviceRegistryMu.Unlock()
+
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
lixmal: fixed

client/internal/sleep/detector_darwin.go
```diff
@@ -2,217 +2,366 @@
...
 )
 
-var (
-	serviceRegistry   = make(map[*Detector]struct{})
-	serviceRegistryMu sync.Mutex
+// IOKit message types from IOKit/IOMessage.h.
+const (
+	kIOMessageSystemWillSleep    uintptr = 0xe0000280
+	kIOMessageSystemHasPoweredOn uintptr = 0xe0000300
 )
```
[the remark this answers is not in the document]
lixmal: Added `kIOMessageCanSystemSleep`; it now acknowledges via `IOAllowPowerChange` without dispatching an event. Factored the port lookup + ack into an `allowPowerChange` helper since there are two call sites now.

client/internal/sleep/detector_darwin.go
```diff
@@ -2,217 +2,366 @@
...
+	go runRunLoop()
+	<-ready
 
-		C.registerNotifications()
-	}()
+	if err := d.rollbackIfRunLoopFailed(); err != nil {
+		serviceRegistryMu.Lock()
+		runLoopReady = nil
+		serviceRegistryMu.Unlock()
+		return err
```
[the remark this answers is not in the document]
lixmal: Added a `lifecycleMu` around `Register`/`Deregister` so the two can't interleave. Not reachable with our current single-detector usage, but the API is sound now.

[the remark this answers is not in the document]
lixmal: Fixed

client/internal/sleep/detector_darwin.go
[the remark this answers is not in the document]
lixmal: Fixed

client/internal/sleep/detector_darwin.go
```diff
@@ -2,217 +2,386 @@
...
-	log.Info("sleepCallbackBridge event triggered")
-
-	serviceRegistryMu.Lock()
-	defer serviceRegistryMu.Unlock()
+	// lifecycleMu serializes Register and Deregister so concurrent lifecycle
+	// transitions can't interleave (e.g. a new registration starting a second
+	// runloop while the previous teardown is still pending).
+	lifecycleMu sync.Mutex
+
+	// runtime state, protected by serviceRegistryMu
```
pappz, line 60: Create the protected vars a common struct. Do not need to copy all arguments any more.

```
type runLoopSession struct {
	rl       uintptr
	port     uintptr
	notifier uintptr
	rp       uintptr
}

session *runLoopSession
```

If session is nil, means need to setup.

client/internal/sleep/detector_darwin.go
```diff
@@ -2,217 +2,386 @@
...
+	runLoopRef   uintptr
+	notifyPort   uintptr
+	notifierObj  uintptr
+	rootPort     uintptr
+	runLoopReady chan struct{}
```
pappz, line 65: runLoopErr / runLoopReady don't need to be package-level (pass the channel to the go routine). Hard to keep in synced (check is it nil + mutex lock all the time).

The error only flows from the setup goroutine to the single Register call that started it. A local errCh :=
make(chan error, 1) passed into runRunLoop achieves the same thing, confines the failure mode to the goroutine's scope, and removes the cross-function state. Same for runLoopReady, it's a one-shot handshake between Register and its goroutine, not a long-lived package signal.

client/internal/sleep/detector_darwin.go
```diff
@@ -2,217 +2,386 @@
...
+		log.Warn("sleep callback timed out")
 	}
 }
+
+// runRunLoop registers IOKit notifications and blocks on CFRunLoopRun.
+// Must own a locked OS thread because CFRunLoop is thread-affine. Publishes
+// runloop state to the package globals, then signals runLoopReady. On setup
+// failure runLoopErr is set and the goroutine exits without entering the
+// runloop.
+func runRunLoop() {
```
pappz, line 337: runRunLoop is doing too much and the `readyClosed` bookkeeping is a symptom of it.

If we split setup out into its own function that returns  `error`, the "report once" invariant becomes a one-liner:       
```             
func runRunLoop(errCh chan<- error) {
	runtime.LockOSThread()
	defer runtime.UnlockOSThread()

	err := initSession()
	errCh <- err
	if err != nil {
		return
	}

	defer func() {
		if r := recover(); r != nil {
			log.Errorf("panic in sleep runloop: %v", r)
		}
	}()
	cf.CFRunLoopRun()
}
```

pappz (changes requested):

client/internal/sleep/detector_darwin.go
```diff
@@ -2,217 +2,386 @@
...
 	d.callback = callback
+	d.done = make(chan struct{})
+	serviceRegistry[d] = struct{}{}
 
-	d.ctx, d.cancel = context.WithCancel(context.Background())
-
-	if len(serviceRegistry) > 0 {
-		serviceRegistry[d] = struct{}{}
-		return nil
+	if len(serviceRegistry) > 1 {
```
pappz, line 191: lifecycleMu serializes Register calls, so Register(B) can only start after Register(A) has fully returned. That means runLoopReady is already closed and any failure
already rolled itself out. Waiting on the channel and calling rollbackIfRunLoopFailed is work that's already been done. B just needs if session == nil { start }.

pappz (changes requested):

client/internal/sleep/detector_darwin.go
lixmal: Done in fe4fe4008: introduced runLoopSession; nil means setup required.

client/internal/sleep/detector_darwin.go
lixmal: Done in fe4fe4008: errCh is now local to Register and passed into runRunLoop. Dropped the package-level runLoopReady/runLoopErr.

client/internal/sleep/detector_darwin.go
lixmal: Done in fe4fe4008: setup extracted into setupSession, runRunLoop is now a handshake + CFRunLoopRun.

client/internal/sleep/detector_darwin.go
lixmal: Done in fe4fe4008: dropped the concurrent rollback path; Register just checks session == nil and starts one if so.

pappz (approved):

pappz (changes requested):
Wait for changes

pappz (approved):

PR #6003, opened 2026-04-27
[management] Enable PAT creation during setup

jnfrati:
## Describe your changes

Automation requires a way for the user to generate some kind of token they can use to control and populate the instance, this feature introduces the option to enable PAT creation through the `/api/setup` endpoint. 

Now the user will be able to enable PAT creation on setup by:
- Setting the env var `NB_SETUP_PAT_ENABLED=true`
- Sending in the /setup request the property `"create_pat": true`
  - Optionally the admin will be able to use `"pat_expire_in": 30` to specify the expiration time in days (defaults to 1)

As answer they'll receive the same response with an extra field `personal_access_token` with the actual token they can use to control the instance. 

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

  OpenAPI schema/response documentation is updated in this PR; no separate docs-site PR is needed.

https://github.com/netbirdio/docs/pull/714

## Summary by CodeRabbit

* **New Features**
  * Optional Personal Access Token (PAT) creation during instance setup, controlled by a server feature flag; PAT returned in the setup response when created and allowed. Response includes Cache-Control: no-store.

* **Validation**
  * PAT expiration validated and constrained to configurable bounds (default 1 day, min/max enforced); invalid values return 422.

* **Reliability**
  * Setup flow includes automated rollback of created resources on failures, with clearer error and rollback behaviors.

* **Documentation**
  * API docs and request/response schemas updated to reflect PAT controls and response behavior.

* **Tests**
  * Expanded test coverage for PAT flows, defaults, error cases, and rollback.


management/server/instance/setup_service.go
```diff
@@ -0,0 +1,206 @@
...
+			log.WithContext(ctx).Errorf("failed to roll back setup account %s for user %s after %s: original error: %v, rollback error: %v", accountID, userID, reason, origErr, err)
+		} else {
+			log.WithContext(ctx).Warnf("rolled back setup account %s for user %s after %s: %v", accountID, userID, reason, origErr)
+		}
+	}
+
+	if err := m.instanceManager.RollbackSetup(ctx, userID); err != nil {
+		log.WithContext(ctx).Errorf("failed to roll back setup user %s after %s: original error: %v, rollback error: %v", userID, reason, origErr, err)
+		return
+	}
```
bcmmbaga, line 160: I think we should stop rollback if `rollbackSetupAccount` fails. 

If account deletion fails but we continue to `RollbackSetup`, we can end up with an existing account whose owner has been deleted from the embedded IdP. That leaves `/api/setup` non-retryable because setup is blocked whenever any account exists, even if there is no local admin user left

bcmmbaga (changes requested):

management/server/instance/setup_service.go
jnfrati: Done!

bcmmbaga (approved):

PR #6004, opened 2026-04-27
[client, relay] Advertise relay server IP via signal for foreign-relay fallback dial

lixmal:
### Problem

When two peers are on different relays and P2P (ICE) cannot be established, each learns the other peer's relay address through signal as an FQDN. If that FQDN cannot be resolved on the receiving side, the foreign-relay dial fails and the peers never reach each other.

This deadlocks on first connect when the only configured exit-node peer is the one whose relay needs to be resolved: the overlay default route is installed before the exit-node WireGuard handshake completes, so DNS goes through `wt0` and fails. Management's relay cache doesn't help here, the foreign peer's home relay is only learned through signal, not from management.

### Fix

Peer B already knows the IP it is connected to on its relay. We add that IP to the signal `Body` so peer A can use it as a dial-time fallback when the FQDN cannot be resolved. SNI/cert validation still uses the FQDN, so TLS is unaffected.

- Add `relayServerIP` (bytes, tag 11) to signal `Body`; reserve tag 9 (previously int64 sessionId) to avoid wire-type collisions
- Surface the underlying TCP RemoteAddr from the websocket relay conn so `ConnectedIP` returns the resolved IP after an FQDN dial
- Plumb the IP as `netip.Addr` through `Manager.OpenConn` and `Client` with a fallback dial path that substitutes the IP into the URL while keeping the original host as SNI
- Add unit tests for the fallback dial, `ConnectedIP` after an FQDN dial, and `substituteHost`; add an end-to-end manager test exercising the foreign-relay path

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal protocol field; no user-visible behavior change beyond reachability of foreign relays when DNS for the FQDN is unavailable.

## New Features
* Added IP-based fallback mechanism for relay server connections when DNS resolution fails, improving resilience in connectivity scenarios
* Relay servers now exchange their connected IP addresses to enable direct IP dialing as a fallback option

## Tests
* Added integration tests validating relay server IP recovery and fallback dialing behavior


shared/relay/client/client_fallback_test.go
```diff
@@ -0,0 +1,258 @@
...
+)
+
+// TestClient_FallbackIPRecoversFromUnresolvableFQDN verifies that when the
+// primary FQDN-based dial fails (unresolvable .invalid host), Connect
+// recovers via the fallback IP and SNI still uses the FQDN.
+func TestClient_FallbackIPRecoversFromUnresolvableFQDN(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+
+	listenAddr := "127.0.0.1:50301"
```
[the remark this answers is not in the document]
lixmal: Done

shared/signal/proto/signalexchange.proto
```diff
@@ -65,7 +65,15 @@ message Body {
...
 
+  reserved 9;
+
   optional bytes sessionId = 10;
+
+  // relayServerIP is the IP the sender is connected to on its relay server,
+  // encoded as 4 bytes (IPv4) or 16 bytes (IPv6). Receivers may use it as a
+  // fallback dial target when DNS resolution of relayServerAddress fails.
+  // SNI/TLS verification still uses relayServerAddress.
+  bytes relayServerIP = 11;
```
pappz: Why not optional? (realayServerAddress also should be optional)

shared/relay/client/manager.go
```diff
@@ -202,6 +206,19 @@ func (m *Manager) RelayInstanceAddress() (string, error) {
 	return m.relayClient.ServerInstanceURL()
 }
 
+// RelayInstanceIP returns the IP address of the live home relay connection.
+// Zero value if not connected. Sent alongside RelayInstanceAddress so remote
+// peers can dial directly without their own DNS lookup.
+func (m *Manager) RelayInstanceIP() netip.Addr {
```
pappz: Return the `RelayInstanceAddress() `along with this variable next to the domain. We can avoid a rare inconsistent state during a relay reconnection.

pappz (changes requested):

shared/relay/client/client.go
```diff
@@ -170,13 +175,16 @@ type Client struct {
 }
 
 // NewClient creates a new client for the relay server. The client is not connected to the server until the Connect
-func NewClient(serverURL string, authTokenStore *auth.TokenStore, peerID string, mtu uint16) *Client {
+// is called. fallbackIP, when valid, is used as a dial-time fallback if the FQDN-based dial fails. TLS
+// verification still uses the FQDN from serverURL via SNI.
+func NewClient(serverURL string, fallbackIP netip.Addr, authTokenStore *auth.TokenStore, peerID string, mtu uint16) *Client {
```
pappz: Please create a new constructor with the extended parameter because the new parameter is not mandatory for all cases. NewClientForForeginServer or NewClientWithFallbackIp.

pappz (changes requested):

shared/signal/proto/signalexchange.proto
lixmal: Doesn't really make a difference for `bytes`, but I'll add it for consistency

lixmal: Fixed

shared/relay/client/manager.go
lixmal: Fixed

shared/relay/client/client.go
lixmal: Fixed

shared/relay/client/client_serverip_test.go
```diff
@@ -0,0 +1,280 @@
...
+// the caller to bind. Avoids hardcoded ports that can collide.
+func freeAddr(t *testing.T) (string, int) {
+	t.Helper()
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("get free port: %s", err)
+	}
+	addr := l.Addr().(*net.TCPAddr)
+	_ = l.Close()
+	return addr.String(), addr.Port
```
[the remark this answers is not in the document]
lixmal: Deferring this. The relay server API (`Server.Listen(ListenerConfig)` and the underlying `ws.Listener` / `quic.Listener`) only accepts an `Address` string and binds itself, so there is no `Serve(net.Listener)` path to feed a pre-bound listener into. Doing it properly means widening that API across both listeners just for tests, which is out of scope here. The close-then-bind window on 127.0.0.1 is microseconds and hasn't flaked in CI; can revisit in a follow-up if it ever does.

shared/relay/client/client.go
```diff
@@ -304,6 +318,41 @@ func (c *Client) ServerInstanceURL() (string, error) {
...
+	addr := conn.RemoteAddr()
+	if addr == nil {
+		return netip.Addr{}
+	}
+	return extractIPLiteral(addr.String())
+}
+
+// extractIPLiteral returns the IP from address forms produced by the relay
+// dialers (URL or host:port). Zero value if the host is not an IP.
+func extractIPLiteral(s string) netip.Addr {
```
pappz: Move the static function to the end of function list

shared/relay/client/client.go
```diff
@@ -332,10 +381,17 @@ func (c *Client) Close() error {
 func (c *Client) connect(ctx context.Context) (*RelayAddr, error) {
 	dialers := c.getDialers()
 
-	rd := dialer.NewRaceDial(c.log, dialer.DefaultConnectionTimeout, c.connectionURL, dialers...)
-	conn, err := rd.Dial(ctx)
+	conn, err := c.dialDirect(ctx, dialers)
 	if err != nil {
-		return nil, err
+		if c.serverIP.IsValid() {
```
pappz: We never connect to the home server using an IP address. From this source, it’s hard to determine that fact. This validation check is already performed in c.dialDirect(ctx, dialers), so it would be better to do it only once and will be clear the logic of the connection.
```
if c.serverIP.IsValid() {
	var err error

	conn, err = c.dialDirect(ctx, dialers)
	if err != nil {
		c.log.Infof(
			"dial via server IP %s failed, falling back to FQDN: %v",
			c.serverIP,
			err,
		)
		conn = nil
	}
}

if conn == nil {
	rd := dialer.NewRaceDial(
		c.log,
		dialer.DefaultConnectionTimeout,
		c.connectionURL,
		dialers...,
	)

	var err error
	conn, err = rd.Dial(ctx)
	if err != nil {
		return nil, fmt.Errorf("dial via FQDN: %w", err)
	}
}

c.relayConn = conn
```

shared/relay/client/client.go
```diff
@@ -332,10 +381,17 @@ func (c *Client) Close() error {
 func (c *Client) connect(ctx context.Context) (*RelayAddr, error) {
 	dialers := c.getDialers()
 
-	rd := dialer.NewRaceDial(c.log, dialer.DefaultConnectionTimeout, c.connectionURL, dialers...)
-	conn, err := rd.Dial(ctx)
+	conn, err := c.dialDirect(ctx, dialers)
```
pappz: Please rename it to `dialRaceDirect(ctx, dialers)` becasue it is also a race dialer.

shared/relay/client/dialer/quic/quic.go
```diff
@@ -32,11 +32,14 @@ func (d Dialer) Dial(ctx context.Context, address string) (net.Conn, error) {
...
 
-	// Set ServerName to hostname if not an IP address
-	host, _, splitErr := net.SplitHostPort(quicURL)
-	if splitErr == nil && net.ParseIP(host) == nil {
-		// It's a hostname, not an IP - modify directly
-		tlsClientConfig.ServerName = host
+	switch {
+	case serverName != "" && net.ParseIP(serverName) == nil:
+		tlsClientConfig.ServerName = serverName
+	case serverName == "":
```
pappz: I think this `case` should be replaced to `default`

pappz (changes requested):

shared/relay/client/dialer/quic/quic.go
lixmal: Fixed

shared/relay/client/client.go
lixmal: Fixed

shared/relay/client/client.go
lixmal: Fixed

shared/relay/client/client.go
lixmal: Fixed

pappz (approved):

PR #6030, opened 2026-04-29
[client] Add SteamOS install script for rootless deployment

shuuri-labs:
## Describe your changes

Adds a standalone installer script (`release_files/install-steamos.sh`) for deploying NetBird on SteamOS/Bazzite as a user-level service optimized for game streaming.

SteamOS (or other SteamOS-like distros) uses an immutable A/B partition scheme with a read-only rootfs, which breaks traditional installation methods. This script takes a different approach from the `systemd-sysext` pattern (used by e.g. Tailscale) by keeping everything in `/home` and using Linux file capabilities instead of running as root:

- **User-level service** — runs as the `deck` user with `systemctl --user`, binary and config live under `/home`
- **Real kernel TUN interface** — uses userspace WireGuard (wireguard-go) with a kernel TUN device for proper network performance (important for Moonlight/Sunshine game streaming)
- **File capabilities instead of root** — `sudo setcap cap_net_admin,cap_net_raw+eip` grants the binary the network capabilities it needs without running as root (one-time sudo at install, reapplied per update)
- **No sysext / overlay** — no version pinning, no extension-release files to break on OS updates
- **Survives all OS updates** — nothing on the read-only rootfs
- **SHA-256 checksum verification** — validates downloads against the release checksums.txt before installing
- **Atomic updates** — downloads and verifies the new binary to a staging directory before stopping the running service, so failed downloads leave the existing installation untouched
- **Supports install, update, and uninstall** via `--update` and `--uninstall` flags
- **Auto-connect** with `NB_SETUP_KEY` environment variable for headless auth
- **`FORCE=true` escape hatch** for other immutable Linux distros (e.g. Bazzite)

### Why not netstack mode?

An earlier iteration used `NB_USE_NETSTACK_MODE=true` for a fully rootless install. However, netstack routes all traffic through a userspace gVisor network stack, which adds latency and jitter unsuitable for real-time game streaming. The TUN + file capabilities approach gives kernel-level routing performance while still keeping all files in `/home`.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Single install script with inline `--help` usage. Docs PR can follow once the approach is validated on hardware.

## Summary by CodeRabbit

* **New Features**
  * Added SteamOS installer script for automated setup, updates, and removal
  * Automatic service configuration with optional provisioning setup
  * Built-in checksum verification for secure installation

### DIFF
--- a/release_files/install-steamos.sh
+++ b/release_files/install-steamos.sh
@@ -71,7 +71,7 @@ check_arch() {
 
 check_dependencies() {
     local missing=""
-    for cmd in curl tar systemctl sudo; do
+    for cmd in curl tar systemctl sudo sha256sum; do
         if ! command -v "$cmd" >/dev/null 2>&1; then
             missing="$missing $cmd"
         fi
@@ -334,6 +334,7 @@ do_update() {
     info "Updating ${installed_version} -> ${latest_version}"
 
     check_arch
+    check_dependencies
 
     # Download and verify new binary to a staging directory before touching the running service
     local staging_dir
--- a/release_files/install-steamos.sh
+++ b/release_files/install-steamos.sh
@@ -1,0 +1,352 @@
+#!/bin/bash
+# NetBird installer for SteamOS (Steam Deck)
+#
+# Installs NetBird as a rootless, user-level service running entirely from /home.
+# Uses netstack mode (userspace WireGuard) — no root, no sysext, no TUN device needed.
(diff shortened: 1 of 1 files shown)


shuuri-labs:
@coderabbitai review

shuuri-labs:
@coderabbitai review

Issue #5989, opened 2026-04-25
Consolidate connection-mode flags; add p2p-dynamic


MichaelUray:
> **Edit 2026-05-08:** Removed `p2p-dynamic-lazy` as a separate mode; its two-tier teardown is now part of `p2p-dynamic` itself, based on implementation feedback. The proposal is now four enum values, not five.

## Summary

cc @pappz — would value your input given the engine-side context from #5807 / netbirdio/android-client#152.

This is an RFC-style proposal to replace the two independent peer-connection flags (`NB_FORCE_RELAY` and `NB_ENABLE_EXPERIMENTAL_LAZY_CONN`) with a single `connection-mode` enum that has four explicit values, including one new mode (`p2p-dynamic`) that combines an activity-triggered relay-first wake-up with a two-tier teardown (ICE first, then relay, then full idle) for bursty mobile/LTE access. Inactivity thresholds become explicit, configurable settings rather than a single env var.

The companion proposal #5990 extends this with per-peer/per-group server-side override of both the mode and the thresholds.

This addresses the same use case as #5589 (mobile default flip) and #4103 (UI/CLI exposure of relay-only mode) with a broader mechanism. If maintainers agree this is the right direction, the original authors of #5589 / #4103 may want to consider whether their issues are still independently needed or can be closed in favor of this work.

## Background

The peer-connection state machine in `client/internal/peer/conn.go` is currently controlled by two independent settings whose effects overlap on the same code path:

- `NB_FORCE_RELAY` / `EnvKeyNBForceRelay` ([peer/env.go](https://github.com/netbirdio/netbird/blob/main/client/internal/peer/env.go))
- `NB_ENABLE_EXPERIMENTAL_LAZY_CONN` / `LazyConnectionEnabled` ([lazyconn/env.go](https://github.com/netbirdio/netbird/blob/main/client/internal/lazyconn/env.go))

[...]

PR #6081, opened 2026-05-05
[client, management] Phase 1+2+3 of #5989: ConnectionMode foundation, two-timer lifecycle, iceBackoff (stack 1/4)


MichaelUray:
# [client+management] Phase 1+2+3 of #5989: ConnectionMode foundation, two-timer lifecycle, iceBackoff

**Branch:** `pr/a-p2p-dynamic-foundation` → base `main`
**Compare:** https://github.com/netbirdio/netbird/compare/main...MichaelUray:netbird:pr/a-p2p-dynamic-foundation?expand=1

---

## Summary

This PR is the foundation for the work proposed in issue #5989 ("Consolidate connection-mode flags; add p2p-dynamic and p2p-dynamic-lazy modes"). It introduces a single first-class `ConnectionMode` enum on the wire and across Settings, replaces the asymmetric `LazyConnectionEnabled` / `IsForceRelayed` precedence with that enum, adds the per-peer two-timer lifecycle (`P2pTimeoutSeconds` for ICE-inactivity detach, `RelayTimeoutSeconds` for full close), and lands the per-peer `iceBackoffState` so failed pair-checks don't loop forever (`P2pRetryMaxSeconds` controls the cap).

It supersedes the older PR #6047 (which only landed Phase 1 and is no longer mergeable).

## Why

NetBird currently has two flags fighting for the same job:
- `LazyConnectionEnabled` (account-wide bool) — eager vs. lazy initial connect.
- `IsForceRelayed` (per-peer hint) — skip ICE entirely.

A user who wants "all my peers should start lazy, but on demand try P2P, and if P2P never comes up, fall back to relay" can't express that with two booleans. The proposed `ConnectionMode` is one of `p2p`, `p2p-lazy`, `p2p-dynamic`, `p2p-dynamic-lazy`, `relay-forced`. Each mode unambiguously selects one combination of (eager-vs-lazy, skip-ICE-yes-no, ICE-fallback-to-relay-yes-no).

The `p2p-dynamic` variants additionally need two timers (one for ICE-only inactivity, one for full conn close) and an exponential ICE-failure backoff so a chronically-broken NAT path can't pin a peer to relay-with-stalled-retries forever.

### Phase 1 — Wire + Settings + CLI plumbing
- New proto field `PeerConfig.ConnectionMode` (enum) + `P2pTimeoutSeconds`, `RelayTimeoutSeconds`.
[...]

PR #6082, opened 2026-05-05
[client] Phase 3.5 + 3.7d-h of #5989: network-change ICE recovery + GUI mode tab (stack 2/4)


MichaelUray:
# [client] Phase 3.5 + 3.7d-h of #5989: network-change ICE recovery + GUI mode tab

**Branch:** `pr/b-phase3.5-network-change` → base **`pr/a-p2p-dynamic-foundation`** (stacked PR — depends on PR-A landing first)
**Compare:** https://github.com/netbirdio/netbird/compare/MichaelUray:netbird:pr/a-p2p-dynamic-foundation...MichaelUray:netbird:pr/b-phase3.5-network-change?expand=1

---

## Summary

This PR adds network-change handling, signal-trigger ICE re-attach, a short-grace post-network-change ICE-failure delay (so we don't burn the exponential backoff on the well-known "first ICE pair-check after WiFi/cellular handover fails on stale NAT mappings"), the Guard-Loop fix from the closed PR #5805, and the GUI surface for Connection Mode + timeouts in the client UI.

This is the second of four stacked PRs implementing #5989. It builds on PR-A.

## Why

The Phase-1+2+3 foundation (PR-A) handles steady-state operation well, but real devices do not stay in steady state:
- Laptops roam between WiFi and cellular.
- Mobile devices wake from Doze/StandBy with stale NAT mappings.
- LTE modems briefly drop and re-attach.

After such an event the existing ICE agent is bound to a departing local interface and pion's pair-checks fail. The exponential backoff (PR-A) then kicks in and the peer stays on relay for minutes.

Two related fixes are needed:
1. **Re-attach ICE after a network change** — recreate the ICE worker on a fresh socket, send a new offer. (Phases 3.5, 3.7d, 3.7e.)
2. **Short retry delay during the post-network-change grace window** — the first pair-check often fails on stale mappings even when subsequent ones succeed. Falling back to the normal 1-minute initial backoff after that single failure leaves the peer on relay far longer than the underlying connectivity warrants. A 60-second grace window with 2-second retries lets ~4-5 attempts run while the new mapping primes. (Phases 3.7f, 3.7h.)

Plus: previously closed PR #5805 ("Fix ICE reconnection loop and buffer candidates before agent initialization") had a known reconnect-loop where ICE worker rebuild on every signal-state change racked up offers faster than pion could complete pair-checks. The clean fix is to skip new offers while the ICE agent is still in `Connecting`.

### Phase 3.5 — Reset on network change
- `client/peer/conn`: reset ICE backoff + recreate `workerICE` when the signal/relay layer reconnects after a network change.
- `client/peer/conn`: refactor `onNetworkChange` to use the in-place agent recreate path; drop `SendOffer` from `onNetworkChange` to fix offer-storm; revert and replace an interim `worker_ice` candidate-buffering attempt with a clean approach.

### Phase 3.7d — Signal-trigger re-attach
- Re-attach ICE on every signal trigger so a peer that lost ICE while the daemon was Disconnected from signaling resumes correctly when signaling is back.

### Phase 3.7e — Listener inside onNetworkChange
- Re-attach the ICE listener inside `onNetworkChange` itself so a network change fires a fresh listener registration without waiting for the next activity event.

### Phase 3.7f + 3.7h — Grace window
- `iceBackoff.markFailure`: while inside `networkChangeGracePeriod` (60s) of the most recent `Reset()`, cap the suspend delay at `networkChangeRetryDelay` (2s) and do NOT advance the long-term exponential schedule. Outside the window the normal 1-minute initial backoff applies.
- Window widened from initial 30s to 60s and delay reduced from 5s to 2s after observing real-world LTE-bounce behaviour: cold NAT mappings often need 3-4 ICE attempts to prime, and the previous 30s window only fit ~2 attempts.

### Phase 3.7g — Skip ICE-Close on NC when still connected
- If ICE is still in `Connected` state when the network-change fires, don't call `workerICE.Close()` — the existing pair is fine, just refresh the listener.

### #5805 cleanup — Guard-Loop fix
- `worker_ice`: skip new offers while the ICE agent is in `Connecting`. Drops the candidate-buffer approach the closed PR proposed in favour of a much simpler offer-skip during connecting.

### GUI surface (Phase 3.7h GUI)
- `client/cmd/service`: persist connection-mode + timeouts on install/reconfigure.
- `client/ui`: Connection Mode + timeouts in Network tab.
- `client`: surface server-pushed connection-mode/timeouts via daemon-RPC.
- `client/ui`: "Follow-Server (currently: …)" display + remove the now-redundant Lazy menu.
- `client/android`: gomobile getters for ConnectionMode + ServerPushed values.

## Tests

- `go test ./client/internal/peer ./client/internal/peer/guard` — pass (full ICE-backoff suite including the new grace-window cases).
- `go build ./client/...` — pass on linux/amd64, linux/arm64, windows/amd64.
- Hardware-validated against three OpenWrt routers (LAN host candidate, internet srflx, LTE srflx) plus a Windows 11 daemon. After a forced LTE-modem reset, the LTE peer recovers to P2P srflx within ~50 s instead of the 2-3 minutes the pre-3.7f code took.

## Test plan
- [x] Grace window: failure inside 60s → 2s retry; failure after 60s → exponential restarts at ~1m.
- [x] No regression for the non-network-change failure path (still uses the exponential schedule).
- [x] Signal/relay reconnect drives `Reset()` → grace window engages.
- [ ] Maintainer to verify: GUI changes on macOS / Linux ARM (I tested Windows + Linux/amd64 only).

## Use case

A laptop that switches between docked Ethernet, office WiFi, and tethered LTE during the day. Without this PR every transition burns the exponential backoff and the user sees "stuck on relay" for 5-15 minutes. With this PR the post-handover recovery is bounded at ~50s.

## Linked work

- Stacks on PR-A (Phase 1+2+3 foundation).
- Picks up the offer-skip-during-connecting fix from closed #5805 (the original PR's candidate-buffer approach is replaced with the simpler pattern).
- Closes part of #5989.

**Maintainers are welcome to push directly to this branch.**

## Summary by CodeRabbit

* **New Features**
  * Configurable Connection Modes (relay-forced, p2p, p2p-lazy, p2p-dynamic) with per-mode timeouts, local overrides, and server-pushed defaults surfaced to clients.
  * ICE failure backoff with visible status (suspended, failures, next-retry) and per-peer retry cap; settings persist via profile/install.

* **UI**
  * Network/Settings UI updated to manage connection mode, timeout fields, and show server-provided hints.

* **Tests**
  * Expanded coverage for mode resolution, inactivity timers, ICE backoff, and related flows.

## Documentation

These changes are internal lifecycle / behavioural improvements; no user-visible API or CLI flag added that warrants new public docs. Existing flags/Settings already documented at netbirdio/docs cover the surface area.


MichaelUray:
Stack: #6081 → **#6082 (this PR)** → #6083 → #6084. Depends on #6081 landing first.

Issue #3258, opened 2025-01-31
Client does not set DNS Server in systemd-resolved on Linux


NiclasPe:
**Describe the problem**

At first, thanks for that awesome Piece of Software!

I have a Ubuntu 24.04.1 LTS Desktop PC with GNOME. For DNS-Resolution it uses systemd-resolved as Local DNS "Server" as the default is.
Because systemd-resolved have Problems with .local domains, i disabled it and just used network manager. 1 day after that, netbird dns stopped working. The error keeps after reverting the changes and reinstalling netbird client. 
After some investiation i learned that netbird-client spins up its own dns server on the netbird internal ip. 
The Problem is, it dose not add it to systemd-resolved automatically. When i add it manually to the netbird interface via `resolvectl` everything works well.

**Expected behavior**

Netbird Client should set the DNS Server for its interface by its own.

Is it possible without systemd-resolved?

**Are you using NetBird Cloud?**

self-hosted -> other clients work like a charm

**NetBird version**

`0.36.5`

**NetBird status -dA output:**

```bash
~ netbird status -dA
Peers detail:
 i-netbird-gw-01.anon-3AphZ.domain:
  NetBird IP: 100.98.94.23
  Public key: .
  Status: Connected
  -- detail --
  Connection type: Relayed
  ICE candidate (Local/Remote): relay/prflx
  ICE candidate endpoints (Local/Remote): 198.51.100.0:65335/198.51.100.1:49935
  Relay server address: 
  Last connection update: 13 minutes, 15 seconds ago
  Last WireGuard handshake: 1 minute, 15 seconds ago
  Transfer status (received/sent) 1.7 KiB/2.5 KiB
  Quantum resistance: false
  Routes: 10.10.1.20/32, 10.10.3.20/32
  Networks: 10.10.1.20/32, 10.10.3.20/32
  Latency: 307.392541ms

 abc-pve02.anon-3AphZ.domain:
  NetBird IP: 100.98.109.170
  Public key: .
  Status: Connected
  -- detail --
  Connection type: P2P
  ICE candidate (Local/Remote): srflx/srflx
  ICE candidate endpoints (Local/Remote): 198.51.100.2:20572/198.51.100.3:61463
  Relay server address: rel://entry.anon-WnKDa.domain:33080
  Last connection update: 13 minutes, 16 seconds ago
  Last WireGuard handshake: 1 minute, 58 seconds ago
  Transfer status (received/sent) 1.4 KiB/1.8 KiB
[...]

PR #6087, opened 2026-05-06
dns: prefer loopback over WireGuard IP when binding DNS listener

### LINKED ISSUE #4110: timeout towards netbird domain DNS forwarder on arch linux
**Describe the problem**

I cannot get any connectivity to the netbird DNS mechanism on an arch linux system.

While investigating the issue, it looks like the eBPF maps aren't populated with the DNS entries. At least, if I understood the mechanism correctly.

My question is now, how can I debug this further.

**To Reproduce**

Having a recent (recently updated) arch linux installation connected to a self-hosted netbird instance. A custom netbird DNS domain is used, but I don't know if this affects the issue.

When I try to look up any peer by DNS, it is timing out:

```plaintext
$ cat /etc/resolv.conf
# Generated by NetBird
# If needed you can restore the original file by copying back /etc/resolv.conf.original.netbird

options timeout:4 attempts:1
search access.anon-L7cGS.domain --
nameserver 100.110.148.82
nameserver 192.168.50.155
nameserver 192.168.119.217

$ dig @100.110.148.82 netrelay.access.anon-L7cGS.domain
;; communications error to 100.110.148.82#53: timed out

; <<>> DiG 9.20.10 <<>> @100.110.148.82 netrelay.access.anon-L7cGS.domain
; (1 server found)
;; global options: +cmd
;; no servers could be reached
```

I also saw that the `nb_map_dns_ip` is empty:
```plaintext
$ sudo bpftool map dump name nb_map_dns_ip
key: 00 00 00 00  value: 00 00 00 00
key: 01 00 00 00  value: 00 00 00 00
key: 02 00 00 00  value: 00 00 00 00
key: 03 00 00 00  value: 00 00 00 00
key: 04 00 00 00  value: 00 00 00 00
key: 05 00 00 00  value: 00 00 00 00
key: 06 00 00 00  value: 00 00 00 00
key: 07 00 00 00  value: 00 00 00 00
key: 08 00 00 00  value: 00 00 00 00
key: 09 00 00 00  value: 00 00 00 00
Found 10 elements
```

Running the agent in foreground (so without systemd hardenings) also doesn't produce a different result. `sudo netbird up -F --config /etc/netbird/main.json --log-file console`.

Connectivity to a service running on a netbird IP (on a different peer) works.

**Expected behavior**

I would expect DNS resolution of netbird DNS domains to work.

self-hosted instance

**NetBird version**

`0.49.0` and also tried it with `0.50.1`.

- Wireguard, manually configured connections. But this is also the case when it is not running after a fresh reboot.

**Debug output**

To help us resolve the problem, please attach the following anonymized status output

  netbird status -dA
```plaintext
Peers detail:
 boris-wintower.access.anon-L7cGS.domain:
  NetBird IP: 100.110.181.146/32
  Public key: pFUC+DmaquTK+81S12HfAPD0CjTLsAXDEXh6arHOZEo=
  Status: Idle
  -- detail --
  Connection type: 
  ICE candidate (Local/Remote): -/-
  ICE candidate endpoints (Local/Remote): -/-
  Relay server address: 
  Last connection update: 5 minutes, 50 seconds ago
  Last WireGuard handshake: -
  Transfer status (received/sent) 0 B/0 B
  Quantum resistance: false
  Networks: -
  Latency: 0s

 cph2449eea.access.anon-L7cGS.domain:
  NetBird IP: 100.110.100.50
  Public key: ObxaqdHPK5rVcjZrph6nQ/7yO5dkLyRkzmt1pvV9hC0=
  Status: Connected
  -- detail --
  Connection type: P2P
  ICE candidate (Local/Remote): host/host
  ICE candidate endpoints (Local/Remote): 192.168.119.158:51820/192.168.119.217:51820
  Relay server address: rel://birdy.anon-07nyF.domain:33080
  Last connection update: 5 minutes, 48 seconds ago
  Last WireGuard handshake: 2 minutes, 47 seconds ago
  Transfer status (received/sent) 2.6 KiB/2.5 KiB
  Quantum resistance: false
  Networks: -
  Latency: 7.971143ms

 netrelay.access.anon-L7cGS.domain:
  NetBird IP: 100.110.236.28
  Public key: 3iHnhFhvs656rtSeD4RKJVTldj3h4/L1WJpKbFn5yUU=
  Status: Connected
  -- detail --
  Connection type: Relayed
  ICE candidate (Local/Remote): -/-
  ICE candidate endpoints (Local/Remote): -/-
  Relay server address: rel://birdy.anon-07nyF.domain:33080
  Last connection update: 38 minutes, 37 seconds ago
  Last WireGuard handshake: 45 seconds ago
  Transfer status (received/sent) 2.8 KiB/4.2 KiB
  Quantum resistance: false
  Networks: 172.16.20.0/24, 172.16.22.0/24, 172.16.22.137/32
  Latency: 0s

Events:
  [INFO] SYSTEM (64e85152-a6e8-43dd-ad8d-196d4ef2e56b)
    Message: Network map updated
    Time: 38 minutes, 37 seconds ago
  [INFO] SYSTEM (3948db80-a0e4-4464-aefe-a9f714cf46fd)
    Message: Network map updated
    Time: 5 minutes, 50 seconds ago
OS: linux/amd64
Daemon version: 0.49.0
CLI version: 0.49.0
Management: Connected to https://birdy.anon-07nyF.domain:33073
Signal: Connected to http://birdy.anon-07nyF.domain:10000
Relays: 
  [stun:birdy.anon-07nyF.domain:3478] is Unavailable, reason: stun request: context deadline exceeded
  [turn:birdy.anon-07nyF.domain:3478?transport=udp] is Available
  [rel://birdy.anon-07nyF.domain:33080] is Available
Nameservers: 
  [9.9.9.9:53, 149.112.112.112:53] for [.] is Available
FQDN: arch.access.anon-L7cGS.domain
NetBird IP: 100.110.148.82/16
Interface type: Kernel
Quantum resistance: false
Lazy connection: false
Networks: -
Forwarding rules: 0
Peers count: 2/3 Connected
```

Create and upload a debug bundle, and share the returned file key:

  netbird debug for 1m -AS -U: `fc1a3b80da002537c8f61ade127c2f5a6d47035e640f710f6ea29a21a478965c/52e3c97c-4fb0-4978-9e41-643a74e5042f`

*Uploaded files are automatically deleted after 30 days.*

**Screenshots**

N/A

**Additional context**

- [X] Reviewed [client troubleshooting](https://docs.netbird.io/how-to/troubleshooting-client) (if applicable)
- [X] Checked for newer NetBird versions
- [X] Searched for similar issues on GitHub (including closed ones)
- [X] Restarted the NetBird client
- [X] Disabled other VPN software
- [X] Checked firewall settings

logan71f100:
## Describe your changes

\`testFreePort\` in \`client/internal/dns/service_listener.go\` tries the WireGuard interface IP first when scanning for a free port. If the DNS server binds to that IP (e.g. \`100.71.x.x:53\`), queries routed to it are silently dropped by the WireGuard kernel module — the host's own WireGuard IP is not in any peer's \`AllowedIPs\`, so packets never reach the listener. DNS resolution breaks entirely.

This fix reorders the candidate IP list so loopback (\`127.0.0.1\`) is tried first, then \`127.0.0.153\`, with the WireGuard IP as a last resort. The DNS proxy then binds to a loopback address where packets are always delivered locally.

\`\`\`go
// Before
ips = []netip.Addr{s.wgInterface.Address().IP, defaultIP, customIP}

// After
ips = []netip.Addr{defaultIP, customIP, s.wgInterface.Address().IP}
\`\`\`

Verified on Ubuntu 25.10 with systemd-resolved: after the fix \`netbird\` binds to \`127.0.0.1:53\` and DNS resolves correctly. Before the fix it bound to the WireGuard IP and all DNS queries timed out.

Fixes #4110 (timeout towards netbird domain DNS forwarder on arch linux)
Related to #3258 (client does not set DNS server in systemd-resolved on Linux)

Both issues stem from the DNS listener binding to the WireGuard interface IP, which the WireGuard kernel module drops packets for since the local IP is not in any peer's AllowedIPs.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

- [x] Documentation is **not needed** for this change (internal bind address selection, no user-facing behavior change)

## Summary by CodeRabbit

* **Refactor**
  * Adjusted DNS port probing order on non-Darwin systems to prefer loopback addresses when determining a free port.
* **Tests**
  * Added a test ensuring loopback addresses are preferred over other interfaces when selecting a free DNS port.

### DIFF
--- a/client/internal/dns/service_listener.go
+++ b/client/internal/dns/service_listener.go
@@ -188,11 +188,10 @@ func (s *serviceViaListener) RuntimeIP() netip.Addr {
 	return s.listenIP
 }
 
-
-// evalListenAddress figure out the listen address for the DNS server
-// first check the 53 port availability on lo or WG interface, if not success
-// pick a random port on WG interface for eBPF, if not success
-// check the 5053 port availability on lo or WG interface without eBPF usage,
+// evalListenAddress figure out the listen address for the DNS server.
+// First tries port 53 on loopback (127.0.0.1, then 127.0.0.153), then the WireGuard IP.
+// If port 53 is unavailable everywhere, tries eBPF on a random port bound to the WireGuard IP.
+// As a final fallback, retries port 5053 in the same loopback-first order.
 func (s *serviceViaListener) evalListenAddress() (netip.Addr, uint16, error) {
 	if s.customAddr != nil {
 		return s.customAddr.Addr(), s.customAddr.Port(), nil
--- a/client/internal/dns/service_listener.go
+++ b/client/internal/dns/service_listener.go
@@ -190,9 +190,9 @@ func (s *serviceViaListener) RuntimeIP() netip.Addr {
 
 
 // evalListenAddress figure out the listen address for the DNS server
-// first check the 53 port availability on WG interface or lo, if not success
+// first check the 53 port availability on lo or WG interface, if not success
 // pick a random port on WG interface for eBPF, if not success
-// check the 5053 port availability on WG interface or lo without eBPF usage,
+// check the 5053 port availability on lo or WG interface without eBPF usage,
 func (s *serviceViaListener) evalListenAddress() (netip.Addr, uint16, error) {
 	if s.customAddr != nil {
 		return s.customAddr.Addr(), s.customAddr.Port(), nil
--- a/client/internal/dns/service_listener.go
+++ b/client/internal/dns/service_listener.go
@@ -220,7 +220,7 @@ func (s *serviceViaListener) evalListenAddress() (netip.Addr, uint16, error) {
 func (s *serviceViaListener) testFreePort(port int) (netip.Addr, bool) {
 	var ips []netip.Addr
 	if runtime.GOOS != "darwin" {
-		ips = []netip.Addr{s.wgInterface.Address().IP, defaultIP, customIP}
+		ips = []netip.Addr{defaultIP, customIP, s.wgInterface.Address().IP}
 	} else {
 		ips = []netip.Addr{defaultIP, customIP}
 	}
--- a/client/internal/dns/service_listener_test.go
+++ b/client/internal/dns/service_listener_test.go
@@ -12,6 +12,23 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestTestFreePort_PrefersLoopback verifies that testFreePort returns a loopback
+// address rather than the WireGuard interface IP when both are available.
+// Before the fix, the WireGuard IP was tried first; binding there caused DNS
+// queries to be silently dropped by the WireGuard kernel module.
+func TestTestFreePort_PrefersLoopback(t *testing.T) {
+	// Find a free port by letting the OS pick one, then releasing it.
+	udpLn, err := net.ListenUDP("udp", net.UDPAddrFromAddrPort(netip.MustParseAddrPort("0.0.0.0:0")))
+	require.NoError(t, err)
+	port := uint16(udpLn.LocalAddr().(*net.UDPAddr).Port)
+	require.NoError(t, udpLn.Close())
+
+	svc := newServiceViaListener(&mocWGIface{}, nil, nil)
+	ip, ok := svc.testFreePort(int(port))
+	require.True(t, ok, "expected to find a free port on loopback")
+	assert.True(t, ip.IsLoopback(), "expected loopback address, got %s — WireGuard IP must not be preferred", ip)
+}
+
 func TestServiceViaListener_TCPAndUDP(t *testing.T) {
 	handler := dns.HandlerFunc(func(w dns.ResponseWriter, r *dns.Msg) {
 		m := new(dns.Msg)


lixmal:
Thanks for the PR.

What exactly is this solving? On Linux, packets destined for the WireGuard interface's own IP are delivered locally via the loopback path, they never traverse the WireGuard module's AllowedIPs check. That only applies to traffic entering or leaving the tunnel.

This seems like an XY problem and I want to avoid we're just masking it instead of fixing it.

logan71f100:
On a fresh Ubuntu install, `systemd-resolved` listens on `127.0.0.53`, leaving `127.0.0.1:53` free. With the old ordering (`testFreePort` tried the WG IP first), binding to e.g. `100.71.234.215:53` succeeded, and NetBird then registered that IP with systemd-resolved as the link-scoped nameserver.

The problem is what happens next: systemd-resolved sends DNS queries *to* `100.71.234.215:53`. Whether those are delivered locally or go out through the tunnel depends on NetBird's ip rules. If a routing rule causes lookup in a table where the NetBird WG subnet is routed through `wg0`, the packet exits the host through WireGuard — and gets silently dropped because the host's own WG IP isn't in any peer's AllowedIPs.

Binding to `127.0.0.1` sidesteps this entirely: systemd-resolved queries loopback, which can never exit the host.

That said, I recognize the fix is minimal and CodeRabbit's suggestion is worth considering: remove the WG IP from `testFreePort` entirely (matching Darwin), since eBPF is already the only legitimate consumer of the WG IP for DNS. That would make a silent failure (bad bind) into an explicit one (no free port → error). Happy to update the PR that way if you prefer.

lixmal:
>Whether those are delivered locally or go out through the tunnel depends on NetBird's ip rules. If a routing rule causes lookup in a table where the NetBird WG subnet is routed through wg0, the packet exits the host through WireGuard — and gets silently dropped because the host's own WG IP isn't in any peer's AllowedIPs.

This is not something netbird sets up, so there is no need to defend against it.

>That said, I recognize the fix is minimal and CodeRabbit's suggestion is worth considering: remove the WG IP from testFreePort entirely (matching Darwin), since eBPF is already the only legitimate consumer of the WG IP for DNS. That would make a silent failure (bad bind) into an explicit one (no free port → error). Happy to update the PR that way if you prefer.

We cannot skip binding to the netbird IP. If `127.0.0.1:53` is already occupied, this fails without any fallback. That's the reason we try the netbird IP first. Only `0.0.0.0:53` or `[::]:53` listeners break this.

There's most likely something else going on on your machine and I suggest to open a new issue instead, with debug info attached.

PR #6108, opened 2026-05-08
[client] util/log: ensure log directory exists on first write (fixes #4392)

### LINKED ISSUE #4392: Directory /var/log/netbird should get created on each start if does not exist
> To check what happened I removed all the directories and re-installed everything:
> ```
> root@r1:~# ls -l /var/lib/netbird /var/log/netbird
> ls: /var/lib/netbird: No such file or directory
> ls: /var/log/netbird: No such file or directory
> ```
> 
> ```
> root@r1:~# netbird service install --service-env="NB_CONFIG=/etc/netbird/config.json"
> NetBird service has been installed
> ```
> 
> It looks the directory `/var/log/netbird` gets only created when the service gets installed:
> ```
> root@r1:~# ls -l /var/lib/netbird /var/log/netbird
> ls: /var/lib/netbird: No such file or directory
> /var/log/netbird:
> ```
> 
> ```
> root@r1:~# netbird service start
> NetBird service has been started
> ```
> 
> `/var/lib/netbird` gets created on each start of the Netbird process and I think the same should happen for `/var/log/netbird`.
> ```
> root@r1:~# ls -l /var/lib/netbird /var/log/netbird
> /var/lib/netbird:
> -rw-------    1 root     root            45 Aug 22 09:37 active_profile.json
> -rw-r--r--    1 root     root            58 Aug 22 09:37 resolv.conf
> -rw-------    1 root     root           236 Aug 22 09:37 state.json
> 
> /var/log/netbird:
> -rw-r--r--    1 root     root          5322 Aug 22 09:37 client.log
> -rw-r--r--    1 root     root             0 Aug 22 09:37 netbird.err
> -rw-r--r--    1 root     root             0 Aug 22 09:37 netbird.log
> ```

 _Originally posted by @MichaelUray in [#4322](https://github.com/netbirdio/netbird/issues/4322#issuecomment-3213456500)_

The directory `/var/log/netbird` only gets created during the installation of the service, but not later during the start of the service process, which could cause problems on systems with volatile `/var` directories like OpenWRT.
I think `/var/log/netbird` should get created on each start if it does not exist.

MichaelUray:
## Summary

Closes #4392.

`lumberjack.Logger` creates the log file lazily on first write, but does **not** create the parent directory. If the directory is missing, the first write fails with `no such file or directory` and the daemon never produces logs. The user-reported failure mode is on a fresh install where `netbird service install` was never run (or where `/var/log/netbird/` was removed manually): `service start` and direct daemon runs then produce no log output, while `/var/lib/netbird/` is created automatically by the profile manager (asymmetry between data-dir and log-dir handling).

## Change

[util/log.go](util/log.go) `newRotatedOutput()`: pre-create the parent directory before constructing the `lumberjack.Logger`. `MkdirAll` is a no-op when the directory already exists, so installs that already went through `service install` are unaffected.

If the daemon lacks permission to create the directory (e.g. running as a less-privileged user against a system-owned `/var/log` path), we log a warning and continue — `lumberjack` will then surface a concrete open-error on the first write rather than silently failing. Permissions match the existing `0o750` used by [`service_installer.go:84`](client/cmd/service_installer.go#L84) so behavior is symmetric across both creation paths.

## Tests

New file `util/log_test.go` with three cases:

- `TestNewRotatedOutputCreatesMissingParentDir` — the bug-reproducer: nested log path under a tempdir whose intermediate directories don't exist; the test asserts that the parent dir is created and the subsequent write succeeds.
- `TestNewRotatedOutputWithExistingParentDir` — sanity: function is a no-op when the parent already exists.
- `TestNewRotatedOutputWithRootOrCwdParentDoesNotPanic` — guards against spurious `MkdirAll` calls for paths whose `Dir()` resolves to `.` or `/`.

```
$ go test ./util/ -run TestNewRotatedOutput -v
=== RUN   TestNewRotatedOutputCreatesMissingParentDir
--- PASS: TestNewRotatedOutputCreatesMissingParentDir (0.00s)
=== RUN   TestNewRotatedOutputWithExistingParentDir
--- PASS: TestNewRotatedOutputWithExistingParentDir (0.00s)
=== RUN   TestNewRotatedOutputWithRootOrCwdParentDoesNotPanic
--- PASS: TestNewRotatedOutputWithRootOrCwdParentDoesNotPanic (0.00s)
PASS
ok      github.com/netbirdio/netbird/util       0.004s
```

Hardware-validated end-to-end:

```
$ rm -rf /tmp/nb-test/
$ ./netbird service status --log-file /tmp/nb-test/netbird/client.log --daemon-addr unix:///nonexistent.sock
NetBird service status: Running
$ ls -la /tmp/nb-test/netbird/
drwxr-x---  2 ai-agent ai-agent  40 May  8 08:32 .
drwxr-x---  3 ai-agent ai-agent  60 May  8 08:32 ..
```

Without this PR, the same invocation leaves `/tmp/nb-test/` non-existent because lumberjack never reaches the file-open step on a write-only logger that didn't get any writes — and worse, on the daemon `run` path the missing directory leads to silent log loss until the directory is manually created.

## Test plan
- [x] Unit tests cover the create / already-exists / `.`-and-`/` cases.
- [x] No regression for the path that already goes through `service install` (existing dir → `MkdirAll` is no-op).
- [x] Permission failure path returns a writable lumberjack logger and a clear warning rather than panicking.

## Use case

Real-world: OpenWrt routers using a packaged `init.d` wrapper that calls `/usr/sbin/netbird` directly without going through `netbird service install`, and any host where `/var/log/netbird/` was removed (e.g. log-rotation cleanup gone wrong). With this fix the daemon recreates the directory automatically on next start instead of failing silently.

- [ ] I added/updated documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced log file handling to automatically create missing parent directories before writing logs, ensuring logs are reliably written even if directories have been removed or never created, preventing silent logging failures.

* **Tests**
  * Added tests to verify proper log file creation and handling of various directory scenarios.

### DIFF
--- a/util/log.go
+++ b/util/log.go
@@ -152,6 +152,28 @@ func setupLogFile(logPath string, disableRotation bool) (io.Writer, error) {
 }
 
 func newRotatedOutput(logPath string) io.Writer {
+	// Ensure the parent directory exists. lumberjack.Logger creates the
+	// log file lazily on first write, but does NOT create the parent
+	// directory; if the directory is missing, the first write fails with
+	// "no such file or directory" and the daemon never produces logs.
+	// On installs that go through `netbird service install`, the directory
+	// is created at install time (see client/cmd/service_installer.go),
+	// but on installs where the directory is removed or never existed
+	// (e.g. fresh OpenWrt setups using the packaged init.d wrapper, or
+	// after a manual `rm -rf /var/log/netbird`), `service start` and
+	// direct daemon runs produce no log output. Fixes #4392.
+	if dir := filepath.Dir(logPath); dir != "" && dir != "." && dir != string(filepath.Separator) {
+		if err := os.MkdirAll(dir, 0o750); err != nil {
+			// Don't fail startup just because we couldn't pre-create the
+			// log directory: let lumberjack attempt the open and surface a
+			// concrete error path. Common reasons MkdirAll fails here are
+			// permission denied (daemon running as a less-privileged user
+			// against a system-owned /var/log path) — in those setups the
+			// directory should already exist via packaging.
+			log.Warnf("could not create log directory %q (continuing, lumberjack will retry on first write): %v", dir, err)
+		}
+	}
+
 	maxLogSize := getLogMaxSize()
 	timberjackLogger := &timberjack.Logger{
 		// Log file absolute path, os agnostic
--- a/util/log_test.go
+++ b/util/log_test.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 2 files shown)


cousinit99:
https://forum.openwrt.org/t/25-12-x-bpi-r4-netbird-service-fails-to-start-upon-boot-due-to-missing-var-log-netbird-folder-tmpfs/251559

PR #6134, opened 2026-05-13
[client] Bracket IPv6 in embed listeners, expand debug bundle

lixmal:
## Describe your changes

Fixes an IPv6 handling bug at the embed listener boundary and expands the debug bundle so v6 firewall state and the sysctls netbird touches are captured.

- Build embed `ListenTCP`/`ListenUDP` bind addresses with `net.JoinHostPort` so IPv6 overlay addresses are bracketed.
- Add `sysctls.txt` to the debug bundle: forwarding (v4 + v6, global + per-interface), reverse-path filter, source-validation mark, and conntrack accounting.
- Collect `ip6tables.txt` (in addition to `iptables.txt`), split `ipset` into its own file, and update the bundle README.

The proxy-side IPv6 target-URL bracketing originally included here was dropped: the right layer is management URL construction (RFC 3986 §3.2.2 makes brackets mandatory whenever the host is an IPv6 literal, port or no port), not defensive bracketing on the proxy after a non-conforming URL has already been emitted. That fix lands in a separate PR.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal debug bundle README updated in-tree; no user-facing docs affected.

### DIFF
--- a/proxy/server_test.go
+++ b/proxy/server_test.go
@@ -4,113 +4,11 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
 func TestDebugEndpointDisabledByDefault(t *testing.T) {
 	s := &Server{}
 	assert.False(t, s.DebugEndpointEnabled, "debug endpoint should be disabled by default")
-}
-
-func TestParseTargetURL(t *testing.T) {
-	tests := []struct {
-		name         string
-		input        string
-		wantHost     string
-		wantHostname string
-		wantPort     string
-		wantErr      bool
-	}{
-		{
-			name:         "unbracketed ipv6 gets bracketed",
-			input:        "http://fb00:cafe:1::3/",
-			wantHost:     "[fb00:cafe:1::3]",
-			wantHostname: "fb00:cafe:1::3",
-			wantPort:     "",
-		},
-		{
-			name:         "bracketed ipv6 is preserved",
-			input:        "http://[fb00:cafe:1::3]/",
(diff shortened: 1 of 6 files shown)


proxy/server_test.go
```diff
@@ -4,13 +4,108 @@ import (
...
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestDebugEndpointDisabledByDefault(t *testing.T) {
 	s := &Server{}
 	assert.False(t, s.DebugEndpointEnabled, "debug endpoint should be disabled by default")
 }
 
+func TestParseTargetURL(t *testing.T) {
```
pappz: Could you extend the test cases with this:
http://fb00:cafe::1:8080/

pappz (changes requested):

proxy/server_test.go
lixmal: Done, it's parsed an IPv6 addr: `http://[fb00:cafe::1:8080]`

pappz (approved):

proxy/server_test.go
riccardomanfrin: but what was the actual intended meaning of http://fb00:cafe::1:8080/ ?
I mean, is it right to allow for this ambiguity?

lixmal: It cannot mean port, that's why we have brackets. I moved the wrapping to the management layer where it belongs, so there is no ambiguity anymore.

riccardomanfrin: same thinking

Issue #4341, opened 2025-08-14
Android app crashes when Rosenpass is enabled


guedressel:
**Describe the problem**

In the Android app I enable Rosenpass in the advance settings, then I hit "connect". After connecting and showing the number of peers the app crashes.  
When I disable Rosenpass again, the connection gets established without crashing.

**Are you using NetBird Cloud?**

Yes - NetBird Cloud.

**NetBird version**

Android app version 0.1.7

**Is any other VPN software installed?**

Yes. Tailscale and Proton VPN, but they are disabled.

**Debug output**

```
08-14 13:44:58.548 16782 16814 I GoLog   : time=2025-08-14T11:44:58.548Z level=DEBUG msg="Sending message" pid="YlGGu+cD21f3rjjIbZetFv48emNOVz1AmoTGnjagqpg=" type=InitHello
08-14 13:44:58.550 16782 16814 I GoLog   : time=2025-08-14T11:44:58.550Z level=DEBUG msg="Started new handshake" pid="YlGGu+cD21f3rjjIbZetFv48emNOVz1AmoTGnjagqpg=" sidi=6aa98752
08-14 13:44:58.551 16782 16814 I GoLog   : time=2025-08-14T11:44:58.551Z level=DEBUG msg="Added peer" pid="YlGGu+cD21f3rjjIbZetFv48emNOVz1AmoTGnjagqpg="
08-14 13:44:58.554 16782     0 E Go      : panic: runtime error: invalid memory address or nil pointer dereference
08-14 13:44:58.554 16782     0 E Go      : [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x6c606e327c]
08-14 13:44:58.554 16782     0 E Go      : 
08-14 13:44:58.554 16782     0 E Go      : goroutine 494 [running]:
08-14 13:44:58.554 16782     0 E Go      : github.com/netbirdio/netbird/client/internal/rosenpass.(*NetbirdHandler).AddPeer(...)
08-14 13:44:58.554 16782     0 E Go      : 	/Users/pzoli/netbird/android-client/netbird/client/internal/rosenpass/netbird_handler.go:44
[...]

PR #6136, opened 2026-05-13
[client] Improve rosenpass support

riccardomanfrin:
## Describe your changes

This PR bundles several related changes to the Rosenpass support in the
client agent:

1. **Bump `cunicu.li/go-rosenpass` from v0.4.0 to v0.5.42.** The upstream
   library was pinned at a release from September 2023, ~40 releases behind
   the current upstream. The bump pulls in protocol fixes from upstream
   (resolves the recurring `failed to load biscuit (ICR1): detected
   replay` slog errors observed in production). Transitive bumps:
   - `cilium/ebpf` 0.15 → 0.19, 
   - `gopacket/gopacket` 1.1.1 → 1.4.0, 
   - `wireguard/wgctrl` updated. Wire-protocol interop between v0.4.0 and
   v0.5.42 verified end-to-end (cross-version handshake works), so no
   flag-day rollout is required.

2. **Fix nil-receiver crash on Android (#4341).** When `Engine.Run()` had
   not completed before `OnConnected` fired, `m.rpWgHandler` was nil and
   `m.rpWgHandler.AddPeer(...)` in `addPeer` panicked. `NewManager` now
   pre-initializes `rpWgHandler`, and `addPeer` has explicit nil guards
   for both `m.server` and `m.rpWgHandler` as defense in depth.
   `generateConfig` keeps replacing the handler on each `Run()` so peer
   state from previous engine sessions is not retained.

3. **Normalize peer endpoint to IPv4-mapped IPv6.** The Rosenpass UDP
   server binds on the IPv6 wildcard `[::]` and `net.ResolveUDPAddr` returns
   a 4-byte IPv4 for IPv4 hosts, which the kernel rejected with `EDESTADDRREQ`
   on Linux when the lib called `sendto` from the AF_INET6 socket. The
   initiator path in `addPeer` now converts the remote endpoint to
   IPv4-mapped IPv6 so its family matches the listening socket.

4. **Move `rosenpassDetermKey` into the `rosenpass` package** as the
   exported `DeterministicSeedKey`, closing the `// todo: move this logic
   into Rosenpass package` left at `conn.go:911`. While moving:
   - Replace `append(lk[:16], rk[:16]...)` with `append` onto an empty
     slice so the `//nolint:gocritic` suppression is no longer needed.
   - Validate input length (`>= 16` bytes) so the function returns an
     error instead of panicking on out-of-range slicing.
   - Wrap returned errors with package-scoped context.
   - Doc comment makes explicit that the seed is deterministic and **not**
     post-quantum safe; Rosenpass replaces it with a real PQ PSK on
     first rotation.

5. **Decouple `rp.Server` behind an `rpServer` interface** so the manager
   can be exercised without spinning up a real UDP server, and add unit
   tests covering the initiator/responder tie-break, peer lifecycle,
   PSK rotation `updateOnly` semantics, error propagation, and the new
   nil-guard branches.

GitHub: <https://github.com/netbirdio/netbird/issues/4341>

### Checklist
- [x] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

Internal change only. The public NetBird CLI/API surface and the Rosenpass
feature behavior are unchanged: the dependency bump is transparent to users
(wire-compatible with v0.4.0), the nil-guard prevents a crash without
altering normal flow, the endpoint normalization restores the intended send
behavior on Linux, and the function move/refactor preserves inputs, outputs,
and call semantics.

N/A

## Summary by CodeRabbit

* **New Features**
  * Deterministic preshared-key generation added for peer handshakes.

* **Bug Fixes**
  * Prevented panics by initializing internal handlers earlier and returning explicit errors when components are unready.
  * Normalized IPv4 endpoints to IPv6-mapped form for consistent connectivity.

* **Tests**
  * Expanded test suite covering key derivation, manager/handler behaviors, endpoint selection, and edge cases.

* **Chores**
  * Updated several dependencies (Rosenpass, WireGuard, eBPF, packet libraries).

[Review Change Stack](https://app.coderabbit.ai/change-stack/netbirdio/netbird/pull/6136)

### DIFF
--- a/client/internal/rosenpass/manager_test.go
+++ b/client/internal/rosenpass/manager_test.go
@@ -1,14 +1,381 @@
 package rosenpass
 
 import (
+	"errors"
+	"os"
+	"sync"
 	"testing"
 
+	rp "cunicu.li/go-rosenpass"
 	"github.com/stretchr/testify/require"
+	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
 )
+
+// --- test doubles -----------------------------------------------------------
+
+type addPeerCall struct {
+	cfg rp.PeerConfig
+}
+
+type removePeerCall struct {
+	id rp.PeerID
+}
+
+type mockServer struct {
+	mu         sync.Mutex
+	addCalls   []addPeerCall
+	removed    []removePeerCall
+	nextID     rp.PeerID
+	addErr     error
+	removeErr  error
+	closed     bool
+	ran        bool
+}
+
+func (m *mockServer) AddPeer(cfg rp.PeerConfig) (rp.PeerID, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.addCalls = append(m.addCalls, addPeerCall{cfg: cfg})
+	if m.addErr != nil {
+		return rp.PeerID{}, m.addErr
+	}
+	// Increment a byte in nextID so distinct peers get distinct IDs.
+	m.nextID[0]++
+	return m.nextID, nil
+}
+
+func (m *mockServer) RemovePeer(id rp.PeerID) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.removed = append(m.removed, removePeerCall{id: id})
+	return m.removeErr
+}
+
+func (m *mockServer) Run() error  { m.ran = true; return nil }
+func (m *mockServer) Close() error { m.closed = true; return nil }
+
+type setPSKCall struct {
+	peerKey    string
+	psk        wgtypes.Key
+	updateOnly bool
+}
+
+type mockIface struct {
+	mu    sync.Mutex
+	calls []setPSKCall
+	err   error
+}
+
+func (m *mockIface) SetPresharedKey(peerKey string, psk wgtypes.Key, updateOnly bool) error {
+	m.mu.Lock()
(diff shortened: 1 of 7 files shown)


client/internal/rosenpass/manager_test.go
```diff
@@ -1,14 +1,412 @@
...
+}
+
+func TestHashRosenpassKey_DifferentInputsDifferOutputs(t *testing.T) {
+	require.NotEqual(t, hashRosenpassKey([]byte("a")), hashRosenpassKey([]byte("b")))
+}
+
+func TestGetLogLevel_DefaultWhenUnset(t *testing.T) {
+	// Snapshot + unset to exercise the LookupEnv ok=false branch. t.Setenv
+	// can only set, not delete, so do it manually with restore via t.Cleanup.
+	prev, hadPrev := os.LookupEnv(defaultLogLevelVar)
```
lixmal, line 116: Let's use something like this here:

```
  t.Setenv(defaultLogLevelVar, "")
```

riccardomanfrin: The manager doesn't check for empty value, just for its presence or not, and there isn't an explicit `t.Unsetenv`

pappz (approved):

PR #6197, opened 2026-05-18
[management] Add SSO session extend flow (management)

pappz:
Adds the management-server half of the SSO session-extension feature:

- New ExtendAuthSession gRPC RPC that refreshes a peer's session expiry using a fresh JWT, validated through the same pipeline as Login but without tearing down the tunnel or redoing the NetworkMap sync.
- Per-peer SessionExpiresAt timestamp on every LoginResponse and SyncResponse so connected clients learn the deadline on the existing long-lived stream, and admin-side changes (toggling expiration, changing the expiration window) reach every peer within seconds.
- SessionExpiresAt(...) helper on Peer that derives the absolute UTC deadline from LastLogin + the account-level PeerLoginExpiration setting, returning zero when the peer is not SSO-tracked or expiration is disabled.

The matching client-side consumer of these fields lands separately.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * Peers can extend SSO sessions using a fresh JWT without performing a full sync.
  * Login and sync responses now include session expiration timestamps when applicable.
  * Account settings changes now trigger timely peer login-expiration updates.
  * Client libraries support the new session-extension call.

* **Tests**
  * Added unit test to validate session-expiration encoding and round-trip.

* **Other**
  * Activity logging added for user-initiated session extensions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/netbirdio/netbird/pull/6197?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


management/internals/shared/grpc/server.go
```diff
@@ -821,6 +822,70 @@ func (s *Server) Login(ctx context.Context, req *proto.EncryptedMessage) (*proto
...
+	deadline, err := s.accountManager.ExtendPeerSession(ctx, peerKey.String(), userID)
+	if err != nil {
+		log.WithContext(ctx).Warnf("failed extending session for peer %s: %v", peerKey.String(), err)
+		return nil, mapError(ctx, err)
+	}
+
+	resp := &proto.ExtendAuthSessionResponse{}
+	if !deadline.IsZero() {
+		resp.SessionExpiresAt = timestamppb.New(deadline)
+	}
```
pascal-fischer: why do we have the condition here? even if the new deadline in zero (which I guess can only happen if someone updates the settings while we are in a session extension flow) we should still send the new updated timestamp to the client? I assume the client will ignore nil values and disable on Zero?

pascal-fischer (approved):

PR #6199, opened 2026-05-19
[misc, infrastructure] proto: unify toolchain & gate CI on drift

riccardomanfrin:
## Describe your changes

Unifies the protobuf code-generation toolchain across the repo and adds a CI
gate that fails on drift between `.proto` sources and committed `*.pb.go`.

### Why

The five proto `generate.sh` scripts pinned different (and stale) tool
versions, and none pinned `protoc` itself. The committed `*.pb.go` files were
generated with at least four different `protoc` versions (`v3.12.4`,
`v3.21.9`, `v6.33.1`, `v7.34.1`) and a `protoc-gen-go-grpc` newer than the
pinned `v1.1`.

### What changed

- New `proto-tools.env` at repo root — single source of truth for
  `PROTOC_VERSION`, `PROTOC_GEN_GO_VERSION`, `PROTOC_GEN_GO_GRPC_VERSION`.
- All five proto `generate.sh` scripts (`client/proto`,
  `shared/signal/proto`, `shared/management/proto`, `flow/proto`,
  `encryption/testprotos`) now `source` `proto-tools.env`, install pinned
  plugin versions, and refuse to run if the local `protoc` does not match
  the pinned release.
- All `*.pb.go` regenerated with the pinned toolchain
  (`protoc v6.33.1`, `protoc-gen-go v1.36.6`, `protoc-gen-go-grpc v1.6.1`).
  Headers now match across every generated file.
- `.github/workflows/proto-version-check.yml` rewritten: installs the
  pinned `protoc` and plugins, runs every `generate.sh`, and fails via
  `git diff --exit-code` on any drift. Trigger paths extended to
  `**/*.proto`, `**/*.pb.go`, `**/generate.sh`, `proto-tools.env`, and the
  workflow itself.

### Holes in the previous CI check

The old `proto-version-check.yml` parsed the PR diff with a regex on
header comments. It missed drift in several ways:

- **Regex did not match `protoc-gen-go-grpc`.** The pattern
  `protoc(?:-gen-go)?` accepts `protoc` or `protoc-gen-go`, never the
  `-grpc` plugin. Version changes there passed silently.
- **`protoc-gen-go-grpc` v1.5+ stopped emitting the version header.**
  Nothing left to match — total bypass.
- **Trigger restricted to `**/*.pb.go`.** A PR that edits `.proto` but
  forgets to regenerate never runs the check.
- **No canonical pin.** Check compared old vs new in the same PR, not
  against a repo-wide pinned toolchain. Two PRs introducing two
  different tool versions both pass.
- **Patch truncation on large PRs.** API omits `patch` data, check
  explicitly bails with "verify manually".

The new workflow regenerates with the pinned toolchain and runs
`git diff --exit-code`. Byte-for-byte comparison against canonical
output. None of the above gaps remain.

### Compatibility

- Wire format: unchanged (`.proto` sources untouched). Cross-version peers
  interoperate.
- In-repo Go API: regenerated together, no internal call sites broken.
- External consumers importing these proto packages must embed
  `UnimplementedXxxServer` (already required by `protoc-gen-go-grpc` v1.5+);
  the in-repo implementations already do.

### Verification

- `proxy/web` UI build green (`tsc && vite build`, 1602 modules,
  no TS errors).
- `go build` on every proto package green (`client/proto`,
  `shared/signal/proto`, `shared/management/proto`, `flow/proto`,
  `encryption/testprotos`).

N/A

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

The new CI workflow itself is the regression test: if a future PR commits
`*.pb.go` generated with a different toolchain, the job fails.

## Documentation
Select exactly one:

Internal build/CI infrastructure only. No user-facing API or behavior change.

N/A

## Summary by CodeRabbit

* **Chores**
  * Centralized pinning of protobuf compiler and Go code-gen plugin versions for proto generation.
  * CI now regenerates proto artifacts and fails PRs if generated outputs drift from pinned toolchain or sources.
  * Hardened proto generation scripts with robust path handling, environment sourcing, stricter error checks, and runtime version verification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/netbirdio/netbird/pull/6199?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


client/proto/generate.sh
```diff
@@ -9,9 +9,21 @@ then
...
+. "$repo_root/proto-tools.env"
+
+actual_protoc=$(protoc --version | awk '{print $2}')
+if [[ "$actual_protoc" != "$PROTOC_VERSION" ]]; then
+  echo "ERROR: protoc version $actual_protoc differs from pinned $PROTOC_VERSION" >&2
+  echo "Install protoc $PROTOC_VERSION from https://github.com/protocolbuffers/protobuf/releases" >&2
+  exit 1
+fi
+
+go install "google.golang.org/protobuf/cmd/protoc-gen-go@${PROTOC_GEN_GO_VERSION}"
```
dmitri-netbird, line 26: we could rely on golang's built-in tool management instead (just for plugins though)?
go.mod would list an indirect dependency on each plugin and a tool section:
google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.6.1 // indirect
...
tool (
        google.golang.org/grpc/cmd/protoc-gen-go-grpc
        google.golang.org/protobuf/cmd/protoc-gen-go
)

Tool updates/additions would be managed via go get -tool some-tool-here@version.
To install all tools: go install tool
A potential issue is that each tool is considered project's dependency now, and will affect versions of shared dependencies.

dmitri-netbird: We could also handle tool installation from the Makefile, this way we won't need to duplicate tool installation in multiple shell scripts?

proto-tools.env
```diff
@@ -0,0 +1,14 @@
...
+# Sourced by every proto generate.sh and the proto-generation-check CI workflow.
+# When bumping a version, regenerate all *.pb.go files in the same PR.
+
+# protoc release tag from https://github.com/protocolbuffers/protobuf/releases
+# `protoc --version` reports `libprotoc ${PROTOC_VERSION}`.
+# Generated pb.go headers embed `protoc v6.${PROTOC_VERSION}`.
+PROTOC_VERSION="33.1"
+
+# google.golang.org/protobuf/cmd/protoc-gen-go
+PROTOC_GEN_GO_VERSION="v1.36.6"
```
dmitri-netbird, line 11: I just discovered that protoc_gen_go version is expected to match that of the protobuf-go module, see here for details: https://github.com/protocolbuffers/protobuf-go/blob/v1.36.11/README.md#compatibility, which could be an argument to manage the plugin through go.mod

.github/workflows/proto-version-check.yml
```diff
@@ -3,60 +3,74 @@ name: Proto Version Check
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Load pinned proto toolchain versions
+        run: |
+          # shellcheck source=/dev/null
+          . ./proto-tools.env
+          {
+            echo "PROTOC_VERSION=${PROTOC_VERSION}"
+            echo "PROTOC_GEN_GO_VERSION=${PROTOC_GEN_GO_VERSION}"
```
dmitri-netbird, line 26: if I understand this correctly, protobuf module version is what determines the version of protoc-gen-go that should be used; perhaps we should check just for that?

client/proto/generate.sh
riccardomanfrin: Ciao Dmitri,
I think it's cool if you can move the versions in the go.mod. Maybe I'd keep the install in the `generate.sh` at least for now since you have to call that only  and it's documented already (moving to Makefile would require ppl to familiarize with new way..)

Do you want to add the change yourself?

proto-tools.env
riccardomanfrin: yep see above ^

dmitri-netbird:
we use a rather old (deprecated) version of golang protobuf bindings (see here for example: https://github.com/netbirdio/netbird/blob/main/encryption/message.go#L4).

I created a test to verify marshalling-unmarshalling of protobuf messages between the old and the new versions of the library, see here: https://gist.github.com/dmitri-netbird/c857b34f7c00e2ade4c1d5b27ab2c0ce

It's currently failing in old/marshal - new/unmarshal direction, will investigate further

riccardomanfrin:
> we use a rather old (deprecated) version of golang protobuf bindings (see here for example: https://github.com/netbirdio/netbird/blob/main/encryption/message.go#L4).
> 
> I created a test to verify marshalling-unmarshalling of protobuf messages between the old and the new versions of the library, see here: https://gist.github.com/dmitri-netbird/c857b34f7c00e2ade4c1d5b27ab2c0ce
> 
> It's currently failing in old/marshal - new/unmarshal direction, will investigate further

@dmitri-netbird  I took the time to check the gist since I'm compiling a list of breaking changes and I noticed that your gist is testing against "gogo" which is an old 2021 dropped project.

The correct testing should be against  google.golang.org/protobuf I think right?

dmitri-netbird:
@riccardomanfrin here's a gist using a correct v2 protobuf module: https://gist.github.com/dmitri-netbird/747c37cf50e38d315eba33781288e9cd. The test no longer fails, supposedly v1 should be mostly? compatible with v2 on the wire, so we could approach this thus:

* add a fuzz-like test to verify serialization/deserialization of v2-v1 sync and login messages
* extract serialization so it can be configured on startup, keep v1 as default
* switch to v2 by default in cloud, run it for awhile
* switch to v2 by default everywhere after a couple of releases
* remove v1 after a few releases

PR #6226, opened 2026-05-20
[management, client, proxy] add expose NetBird-only services over tunnel peers

mlsmaycon:
## Describe your changes

Adds a new "private" service mode for the reverse proxy: services reachable exclusively over the embedded WireGuard tunnel, gated by per-peer group membership instead of operator auth schemes.

Wire contract
- ProxyMapping.private (field 13): the proxy MUST call ValidateTunnelPeer and fail closed; operator schemes are bypassed.
- ProxyCapabilities.private (4) + supports_private_service (5): capability gate. Management never streams private mappings to proxies that don't claim the capability; the broadcast path applies the same filter via filterMappingsForProxy.
- ValidateTunnelPeer RPC: resolves an inbound tunnel IP to a peer, checks the peer's groups against service.AccessGroups, and mints a session JWT on success. checkPeerGroupAccess fails closed when a private service has empty AccessGroups.
- ValidateSession/ValidateTunnelPeer responses now carry peer_group_ids + peer_group_names so the proxy can authorise policy-aware middlewares without an extra management round-trip.
- ProxyInboundListener + SendStatusUpdate.inbound_listener: per-account inbound listener state surfaced to dashboards.
- PathTargetOptions.direct_upstream (11): bypass the embedded NetBird client and dial the target via the proxy host's network stack for upstreams reachable without WireGuard.

Data model
- Service.Private (bool) + Service.AccessGroups ([]string, JSON- serialised). Validate() rejects bearer auth on private services. Copy() deep-copies AccessGroups. pgx getServices loads the columns.
- DomainConfig.Private threaded into the proxy auth middleware. Request handler routes private services through forwardWithTunnelPeer and returns 403 on validation failure.
- Account-level SynthesizePrivateServiceZones (synthetic DNS) and injectPrivateServicePolicies (synthetic ACL) gate on len(svc.AccessGroups) > 0.

Proxy
- /netbird proxy --private (embedded mode) flag; Config.Private in proxy/lifecycle.go.
- Per-account inbound listener (proxy/inbound.go) binding HTTP/HTTPS on the embedded NetBird client's WireGuard tunnel netstack.
- proxy/internal/auth/tunnel_cache: ValidateTunnelPeer response cache with single-flight de-duplication and per-account eviction.
- Local peerstore short-circuit: when the inbound IP isn't in the account roster, deny fast without an RPC.
- proxy/server.go reports SupportsPrivateService=true and redacts the full ProxyMapping JSON from info logs (auth_token + header-auth hashed values now only at debug level).

Identity forwarding
- ValidateSessionJWT returns user_id, email, method, groups, group_names. sessionkey.Claims carries Email + Groups + GroupNames so the proxy can stamp identity onto upstream requests without an extra management round-trip on every cookie-bearing request.
- CapturedData carries userEmail / userGroups / userGroupNames; the proxy stamps X-NetBird-User and X-NetBird-Groups on r.Out from the authenticated identity (strips client-supplied values first to prevent spoofing).
- AccessLog.UserGroups: access-log enrichment captures the user's group memberships at write time so the dashboard can render group context without reverse-resolving stale memberships.

OpenAPI/dashboard surface
- ReverseProxyService gains private + access_groups; ReverseProxyCluster gains private + supports_private. ReverseProxyTarget target_type enum gains "cluster". ServiceTargetOptions gains direct_upstream. ProxyAccessLog gains user_groups.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

https://github.com/netbirdio/docs/pull/767

## Summary by CodeRabbit

* **New Features**
  * Private reverse-proxy services (NetBird-only) with group-based access control, synthesized private DNS records, and per-account inbound HTTP(S) listeners
  * Direct-upstream target mode to route via host network
  * Tunnel-based peer validation RPC and session tokens carrying email/groups
  * Client/local lookup helpers to resolve peer identity by tunnel IP

* **Enhancements**
  * Proxy cluster capability reporting (private/custom-port)
  * Identity propagation to upstream requests (email/groups)
  * DNS answers filter disconnected peers; positive tunnel-validation caching
* **Tests**
  * Extensive unit/integration coverage for proxy, auth, routing, DNS, and storage behaviors

[Review Change Stack](https://app.coderabbit.ai/change-stack/netbirdio/netbird/pull/6226?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


management/server/store/sql_store.go
```diff
@@ -5908,7 +5929,7 @@ func (s *SqlStore) getClusterCapability(ctx context.Context, clusterAddr, column
 		AnyTrue       bool
 	}
 
-	err := s.db.
+	err := s.db.WithContext(ctx).
```
pascal-fischer: no withContext

proxy/cmd/proxy/cmd/root.go
```diff
@@ -105,6 +106,12 @@ func init() {
 	rootCmd.Flags().StringVar(&preSharedKey, "preshared-key", envStringOrDefault("NB_PROXY_PRESHARED_KEY", ""), "Define a pre-shared key for the tunnel between proxy and peers")
 	rootCmd.Flags().BoolVar(&supportsCustomPorts, "supports-custom-ports", envBoolOrDefault("NB_PROXY_SUPPORTS_CUSTOM_PORTS", true), "Whether the proxy can bind arbitrary ports for UDP/TCP passthrough")
 	rootCmd.Flags().BoolVar(&requireSubdomain, "require-subdomain", envBoolOrDefault("NB_PROXY_REQUIRE_SUBDOMAIN", false), "Require a subdomain label in front of the cluster domain")
+	// --private is internal: set by the embedded `netbird proxy` subcommand
+	// via NB_PROXY_PRIVATE so management can distinguish per-peer / private
+	// clusters from centralised ones. Hidden so the standalone CLI doesn't
+	// surface it as an operator-facing toggle.
+	rootCmd.Flags().BoolVar(&private, "private", envBoolOrDefault("NB_PROXY_PRIVATE", false), "Mark this proxy as embedded/private (internal flag)")
```
pascal-fischer: I think this description is misleading. It sound like marking private sets
it into a private mode. We should be more clear that this additionally supports private services

proxy/internal/auth/middleware.go
```diff
@@ -246,18 +311,117 @@ func (mw *Middleware) forwardWithSessionCookie(w http.ResponseWriter, r *http.Re
 	if err != nil {
 		return false
 	}
-	userID, method, err := auth.ValidateSessionJWT(cookie.Value, host, config.SessionPublicKey)
+	if requestIsPlainHTTP(r) {
+		mw.logger.WithFields(log.Fields{
+			"host":   host,
+			"remote": r.RemoteAddr,
+		}).Warn("session cookie on plain HTTP path; cookie auth requires TLS — use port 443")
+	}
```
pascal-fischer: not sure how usefull this is as it will be in our logs but not propagated to the user

proxy/internal/auth/middleware.go
```diff
@@ -518,18 +693,27 @@ func (mw *Middleware) validateSessionToken(ctx context.Context, host, token stri
 			}).Debug("Session validation denied")
 			return &validationResult{
 				UserID:       resp.UserId,
+				UserEmail:    resp.GetUserEmail(),
 				Valid:        false,
 				DeniedReason: resp.DeniedReason,
+				Groups:       resp.GetPeerGroupIds(),
+				GroupNames:   resp.GetPeerGroupNames(),
```
pascal-fischer: should we expose this if the session is invalid?

proxy/lifecycle_test.go
```diff
@@ -0,0 +1,134 @@
...
+	defer cancel()
+
+	err := srv.Start(ctx)
+	require.Error(t, err, "Start must surface management dial failures")
+
+	assert.True(t, srv.started, "started flag is set before any dial attempt so a second Start fails fast")
+
+	err = srv.Start(ctx)
+	require.Error(t, err, "second Start must reject")
+	assert.Contains(t, err.Error(), "already started", "error must explain why the call was rejected")
```
pascal-fischer: The code that is tested in lifecycle_test.go does not life in lifecycle.go but server.go

pascal-fischer (changes requested):

pascal-fischer (dismissed):

management/internals/shared/grpc/proxy.go
```diff
@@ -882,24 +682,41 @@ func (s *ProxyServiceServer) SendServiceUpdateToCluster(ctx context.Context, upd
...
+			return false
+		}
+	}
 	if mapping.ListenPort == 0 || mapping.Mode == "tls" {
 		return true
 	}
-	// Old proxies that never reported capabilities don't understand
-	// custom port mappings.
 	return conn.capabilities != nil && conn.capabilities.SupportsCustomPorts != nil
 }
```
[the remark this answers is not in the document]
mlsmaycon: not from this PR

client/internal/peer/status.go
```diff
@@ -305,6 +305,25 @@ func (d *Status) PeerByIP(ip string) (string, bool) {
...
+// PeerStateByIP returns the full peer State for the given tunnel IP.
+// Matches against either the IPv4 (State.IP) or IPv6 (State.IPv6) tunnel
+// address so dual-stack peers are reachable on either family. Returns the
+// zero State and false when no peer matches or the input is empty.
+func (d *Status) PeerStateByIP(ip string) (State, bool) {
+	if ip == "" {
+		return State{}, false
+	}
+	d.mux.Lock()
+	defer d.mux.Unlock()
```
lixmal: Ideally this should be a read lock now since we run it for every request

proxy/internal/roundtrip/multi.go
```diff
@@ -0,0 +1,68 @@
...
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}
+	direct := &http.Transport{
+		DialContext:           dialWithTimeout(dialer.DialContext),
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
```
lixmal: These bypass the env overrides defined in proxy/internal/roundtrip/transport.go

proxy/internal/roundtrip/multi.go
```diff
@@ -0,0 +1,68 @@
...
+		insecure: insecure,
+	}
+}
+
+// RoundTrip dispatches by reading the direct-upstream flag from the request
+// context. When set, the request is forwarded via the stdlib transport,
+// honouring the existing per-request skip-TLS-verify flag. Otherwise it
+// goes through the embedded NetBird roundtripper.
+func (m *MultiTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if DirectUpstreamFromContext(req.Context()) || m.embedded == nil {
```
lixmal: nil m.embedded shouldn't be a silent fallback. This needs to be explicit

proxy/internal/proxy/reverseproxy.go
```diff
@@ -426,3 +431,46 @@ func opErrorContains(err error, substr string) bool {
...
+	groupNames := cd.GetUserGroupNames()
+	labels := make([]string, len(groupIDs))
+	for i, id := range groupIDs {
+		if i < len(groupNames) && groupNames[i] != "" {
+			labels[i] = groupNames[i]
+			continue
+		}
+		labels[i] = id
+	}
+	r.Out.Header.Set(headerNetBirdGroups, strings.Join(labels, ","))
```
lixmal: SThis is not safe if the groups contain ",". Better repeat the same header for multiple values.

We also shouldn't accept arbitrary values here since it can break the request (e.g. illegal characters end up in headers). Same goes for email above

proxy/inbound.go
```diff
@@ -0,0 +1,537 @@
...
+
+// privateInboundPortHTTP is the WG-side plain-HTTP port.
+const privateInboundPortHTTP = 80
+
+// inboundManager wires per-account inbound listeners into the proxy
+// pipeline when --private-inbound is enabled. When disabled the manager
+// is nil and every method on *Server that touches it short-circuits.
+type inboundManager struct {
+	logger    *log.Logger
+	handler   http.Handler
```
lixmal, line 49: Worth checking if we should exclude geo and crowdsec middleware here since it's not very useful for overlay traffic

lixmal (changes requested):

proxy/internal/proxy/reverseproxy.go
mlsmaycon: will sanitize it and drop the headers

proxy/internal/proxy/reverseproxy.go
```diff
@@ -426,3 +431,70 @@ func opErrorContains(err error, substr string) bool {
...
+	labels := make([]string, 0, len(groupIDs))
+	for i, id := range groupIDs {
+		label := id
+		if i < len(groupNames) && groupNames[i] != "" {
+			label = groupNames[i]
+		}
+		if !isHeaderValueSafe(label) || strings.ContainsRune(label, ',') {
+			continue
+		}
+		labels = append(labels, label)
```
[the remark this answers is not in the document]
mlsmaycon: don't want to go for partial here or fallbacks.

lixmal (approved):

PR #6260, opened 2026-05-25
[client] Add commands to discover and write Kubernetes configuration

phillebaba:
## Describe your changes

This change adds a new Kubernetes subcommand. It allows listing of Kubernetes clusters that are exposed through the api proxy peers. It allows has a command that will write to the local kubeconfig so that the cluster can be accessed.

https://github.com/netbirdio/kubernetes-operator/issues/274

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

https://github.com/netbirdio/docs/pull/776

## Summary by CodeRabbit

* **New Features**
  * Added a `kubernetes` CLI to discover reachable Kubernetes clusters and show cluster name, FQDN and Kubernetes version; prints “No Kubernetes clusters available.” when none found.
  * `write-kubeconfig` creates/updates kubeconfig entries (server URL, user `netbird` token placeholder, context), sets current-context, and supports cluster-name filtering.

* **Tests**
  * Added tests for cluster fingerprinting, kubeconfig path resolution, and kubeconfig writing.

* **Chores**
  * Updated project module dependencies.

### DIFF
--- a/client/cmd/kubernetes.go
+++ b/client/cmd/kubernetes.go
@@ -1,0 +1,301 @@
+package cmd
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/goccy/go-yaml"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/netbirdio/netbird/client/proto"
+)
+
+const (
+	KubernetesDNSSuffix = "netbird-kubeapi-proxy"
+)
+
+var kubernetesCmd = &cobra.Command{
+	Use:   "kubernetes",
+	Short: "Kubernetes cluster commands.",
+	Long:  "Kubernetes cluster commands.",
+}
+
+var kubernetesListCmd = &cobra.Command{
+	Use:   "list",
+	RunE:  kubernetesList,
+	Short: "List Kubernetes clusters.",
+	Long:  "List Kubernetes clusters by discovering NetBird peers running netbird-kubeapi-proxy.",
+}
+
+var kubernetesWriteKubeconfigCmd = &cobra.Command{
+	Use:   "write-kubeconfig",
+	RunE:  kubernetesWriteKubeconfig,
+	Args:  cobra.ExactArgs(1),
+	Short: "Write kubeconfig for a Kubernetes cluster.",
+	Long:  "Updates kubeconfig in place to allow token-less access to the Kubernetes cluster through NetBird.",
+}
+
+func init() {
+	kubernetesWriteKubeconfigCmd.Flags().String("kubeconfig", "", "path to kubeconfig file")
+}
+
+func kubernetesList(cmd *cobra.Command, _ []string) error {
+	conn, err := getClient(cmd)
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+	client := proto.NewDaemonServiceClient(conn)
+	statusResp, err := client.Status(cmd.Context(), &proto.StatusRequest{GetFullPeerStatus: true})
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/netbirdio/netbird
 
-go 1.25.5
+go 1.26.0
```
phillebaba: Kubernetes dependency requires 1.26 as minimum version. Should we change to setting toolchain version instead?

client/cmd/kubernetes.go
```diff
@@ -0,0 +1,300 @@
...
+		return nil, "", err
+	}
+	versionData := map[string]string{}
+	err = json.Unmarshal(b, &versionData)
+	if err != nil {
+		return nil, "", err
+	}
+	version, ok := versionData["gitVersion"]
+	if !ok {
+		return nil, "", err
```
lixmal: This can return a nil error

```suggestion
		return nil, "", errors.New("gitVersion not found in version response"
```

client/cmd/kubernetes.go
```diff
@@ -0,0 +1,300 @@
...
+		// https://github.com/golang/go/issues/17093
+		PreferGo: true,
+	}
+
+	kcs := []kubernetesCluster{}
+	attempted := map[string]any{}
+	for _, peer := range peers {
+		fqdns, err := resolver.LookupAddr(ctx, peer.IP)
+		if err != nil {
+			return nil, err
```
lixmal, line 141: This will abort the whole list command when a single peer doesn't have a RR

client/cmd/kubernetes.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("could not determine home directory: %w", err)
+	}
+	return filepath.Join(home, ".kube", "config"), nil
+}
+
+func writeKubeconfig(kubeconfigPath string, kc kubernetesCluster) error {
+	b, err := os.ReadFile(kubeconfigPath)
```
lixmal, line 233: This will fail for users that never had a kubecfg before. Side note: I think this will also drop all comments from the kube cfg if there were any

client/cmd/kubernetes.go
```diff
@@ -0,0 +1,300 @@
...
+		Transport: transport,
+	}
+	resolver := net.Resolver{
+		// Required so both DNS records are returned.
+		// https://github.com/golang/go/issues/17093
+		PreferGo: true,
+	}
+
+	kcs := []kubernetesCluster{}
+	attempted := map[string]any{}
```
lixmal: This should be a `map[string]struct{}{}`

lixmal (changes requested):

client/cmd/kubernetes.go
phillebaba: Kubectl already drops comments from kubeconfig when it does modifications. I was debating if we should create a file or not if it does not exist. I can add a check if that is better. Think this is a matter of taste.

client/cmd/kubernetes.go
phillebaba: Would there be a case where a peer does not have a record and things are working properly? I assumed a failed lookup is an indication that things are not working with DNS. My thinking was that silently logging the error could cause people chasing errors.

If this happens a lot then I am fine with logging the error instead.

client/cmd/kubernetes.go
phillebaba: I have always just done sets with any but I can change it if this is the standard here.

client/cmd/kubernetes.go
lixmal: In general, this shouldn't happen, but management sends a partial list this will error out. We can leave it for now

client/cmd/kubernetes.go
lixmal: Then it's alright. leave it

client/cmd/kubernetes.go
lixmal: It saves a few bytes so it's good practice

lixmal (approved):

PR #6296, opened 2026-05-31
[client] propagate exit-node deselect to synthesized v6 (::/0) route

pappz:
When a client deselects an IPv4 exit node, the auto-generated IPv6 default route (::/0) was still selected and pushed onto the tunnel interface, even though the user disabled the exit node. On an exit node without a real IPv6 egress this blackholes IPv6 traffic, and because clients prefer IPv6 (happy eyeballs) it can break general connectivity.

Root cause: the synthesized v6 route gets a different NetID than its v4 base (base + "-v6"). The route selector keys deselects by NetID and defaults unknown NetIDs to selected, so the "-v6" entry was never matched by the v4 deselect. The effectiveNetID() mirror that solves exactly this is used by HasUserSelectionForRoute and FilterSelectedExitNodes, but categorizeUserSelection called the raw IsSelected(), bypassing it and mis-categorizing the v6 pair as user-selected.

Add RouteSelector.IsSelectedForExitNode(), which applies effectiveNetID before the selection check, and use it in categorizeUserSelection. IsSelected() is left untouched so non-exit code paths don't make unrelated "*-v6" routes inherit v4 state. Adds regression tests for the v4/v6 deselect mirror and explicit-v6 override.

  1. The exit node advertises both the 0.0.0.0/0 and ::/0 routes.
  2. UI: you enable the exit node → ExpandV6ExitPairs → selected={v4, v6}. (The -v6 key is persisted at this point.)
  3. On the management side, disable IPv6.
  4. UI: you disable the exit node → ExpandV6ExitPairs can't find the v6 pair (it's no longer in the routesMap) → only the v4 gets added to deselected.
  5. On the management side, enable IPv6 → ::/0 reappears → on the next route update it gets applied to the tunnel, even though the exit node is OFF.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Bug Fixes**
  * v6 exit-route selections now mirror their v4 counterparts so deselected v4 routes no longer leave orphaned v6 selections or cause routing leaks.

* **New Features**
  * iOS: per-route connection status added to route selection details (Connected / Idle).
  * Route selector: explicit v4↔v6 pairing sync to keep selection state consistent.

* **Tests**
  * Added tests covering v4/v6 pairing, sync semantics, orphaned v6 regression, and filtering behavior.

### DIFF
--- a/client/internal/routemanager/manager.go
+++ b/client/internal/routemanager/manager.go
@@ -428,10 +428,16 @@ func (m *DefaultManager) UpdateRoutes(
 	var merr *multierror.Error
 	if !m.disableClientRoutes {
 
+		log.Debugf("DIAG UpdateRoutes: incoming %d client networks: %v", len(clientRoutes), haKeysList(clientRoutes))
+		log.Debugf("DIAG UpdateRoutes: selector BEFORE management update: %s", m.routeSelector.MarshalSummary())
+
 		// Update route selector based on management server's isSelected status
 		m.updateRouteSelectorFromManagement(clientRoutes)
 
+		log.Debugf("DIAG UpdateRoutes: selector AFTER management update: %s", m.routeSelector.MarshalSummary())
+
 		filteredClientRoutes := m.routeSelector.FilterSelectedExitNodes(clientRoutes)
+		log.Debugf("DIAG UpdateRoutes: AFTER filter, %d networks remain: %v", len(filteredClientRoutes), haKeysList(filteredClientRoutes))
 
 		if err := m.updateSystemRoutes(filteredClientRoutes); err != nil {
 			merr = multierror.Append(merr, fmt.Errorf("update system routes: %w", err))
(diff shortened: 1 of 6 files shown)


riccardomanfrin (approved):
Maybe it could be worth to separate the iOS routes reachability status feature from the bugfix itself (they seem to be orthogonal).

Future hardening (out of scope for this PR): the pair detection currently keys off the user-influenced -v6 NetID suffix, so a route a user literally names foo-v6 with ::/0 (alongside foo+0.0.0.0/0) is indistinguishable from a synthesized exit pair and gets force-mirrored. A more robust signal is the system-generated ID: the server stamps the synthesized v6 exit with ID = r.ID + "-v6-default" (networkmap_components.go:480), which a user cannot influence. We could detect the synthesized pairs by iterating clientRoutes and selecting routes whose r.ID ends in -v6-default, then still derive r.NetID to call SyncPairedSelection — detection by ID, write by NetID (the selector is and must stay NetID-keyed). The server ID is delivered to the client in the NetworkMap, so it's available at this layer. Cleanest of all would be a dedicated typed flag on route.Route set at synthesis time, rather than string-suffix matching on either identifier.

Apart for this potentially cleaner but more invasive approaches the PR is ok to me

Issue #5449, opened 2026-02-25
netbird ssh JWT auth fails - client discards fresh SSH token, sends stale token from netbird up


chris-prosser:
**Describe the problem**

  After upgrading our self-hosted NetBird management server to v0.65.2, netbird ssh with JWT authentication always fails. The SSH server's JWT validator rejects the token with age=5m0.Xs, max=5m0s
  every time, regardless of how quickly the SSH command is run after netbird up. The OIDC flow for SSH shows "Authentication successful!" but the token sent to the target peer is always the one
  obtained during netbird up, not a fresh token. This was working before the management server upgrade.

**To Reproduce**

  1. Self-hosted NetBird with Entra ID (Azure AD) OIDC
  2. Run netbird up - OIDC completes successfully
  3. Run netbird ssh username@<peer-ip> 'hostname' - OIDC completes, shows "Authentication successful!"
  4. SSH handshake fails: ssh: unable to authenticate, attempted methods [none password], no supported methods remain
  5. Target peer log shows: JWT authentication failed: token expired for user=...: age=5m0.884820059s, max=5m0s
  6. Repeating immediately gives the same result - token age is always ~5m0s

**Expected behavior**

  The client should obtain a fresh JWT token for the netbird ssh session rather than reusing the cached token from netbird up. Alternatively, the max age for JWT validation should be configurable on
  the SSH server side. The current hardcoded max=5m0s in client/ssh/server/server.go makes JWT SSH auth unusable since the token from netbird up will always be expired by the time an SSH session is
  attempted.

**Are you using NetBird Cloud?**

  No. Self-hosted management server, signal, relay, and dashboard - all on v0.65.2. Identity provider is Microsoft Entra ID (Azure AD).

**NetBird version**

  Tested with both client versions - identical behavior:
  - netbird version: 0.65.3
  - netbird version: 0.66.0

**Is any other VPN software installed?**

No

**Debug output**

  Target peer SSH server log (/var/log/netbird/client.log):
  INFO client/ssh/server/server.go:779: SSH connection from NetBird peer 100.x.x.x:51482 allowed
  INFO client/ssh/server/server.go:463: JWT validator initialized successfully
  WARN [session: hydra-admin@100.x.x.x:51482] client/ssh/server/server.go:610: JWT authentication failed: token expired for user=...: age=5m0.884820059s, max=5m0s

  Client output:
  SSH authentication required.
  Please do the SSO login in your browser.
  Waiting for authentication...
[...]

PR #6325, opened 2026-06-02
[management] Make ssh JWT timeout configurable

allomanta:
## Describe your changes
Currently the JWToken's max age for netbird ssh is always set to be 10 minutes on the server, which was updated from 5 minutes in #5471. When used with an SSO like Authentik this can lead to validation issues if a user calls netbird up with an --ssh-jwt-cache-ttl larger than 600.
This PR adds routing for the tokens MaxAge config value such that it can be configured by the management UI. I can make a related PR for the UI if this is indeed desired. Currently we've tested locally with a fork where I just added a field to the authentication tab. 

I also think that #6197 might be related to this.

#5449, #5471, maybe #6197
### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

I believe that documentation only needs to be changed if the management UI also updates as this doesn't change anything on its own. But I'm happy to update any documentation if that's an incorrect assumption

## Summary by CodeRabbit

* **New Features**
  * SSH JWT settings can be updated at runtime without restarting the SSH service; configurable SSH JWT max token age respected.
  * JWT cache now validates tokens using iat/exp claims and honors the current TTL when serving cached tokens.

* **Documentation**
  * API schema and client models expose SSH JWT max token age.

* **Tests**
  * Added tests for runtime JWT reconfiguration, JWT cache claim/TTL behavior, and max-token-age validation/persistence.

### DIFF
--- a/client/internal/engine_ssh.go
+++ b/client/internal/engine_ssh.go
@@ -24,6 +24,7 @@ type sshServer interface {
 	Stop() error
 	GetStatus() (bool, []sshserver.SessionInfo)
 	UpdateSSHAuth(config *sshauth.Config)
+	UpdateJWTConfig(config *sshserver.JWTConfig)
 }
 
 func (e *Engine) setupSSHPortRedirection() error {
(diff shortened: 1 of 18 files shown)


allomanta:
Maybe this should be a separate PR, but I noticed that the expiration times could get out of sync, resulting in the error:
```
SSH connection to NetBird server failed: SSH handshake: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none password], no supported methods remain
```
I believe this is due to the client only checking the time since it stored the JWToken, and not the actual exp or iat times. Changing to that setup seems to work for me so far.

allomanta:
I could move some functions from client/ssh/server/server.go to shared/auth/jwt/extractor.go and shared/auth/jwt/validator.go in order to make the length smaller but that refactor would be orthogonal to the changes for this PR. I also don't understand how the length is calculated as the file has 978 lines, are empty lines skipped?

jarrah31:
I really want this fix too please. Having to authenticate new SSH connections after the 10 min timeout is extremely annoying.

duartemvix:
Bump. Please any updates on this? This really sets back Netbird compared to Tailscale. I couldn't believe this wasn't looked upon sooner.

PR #6332, opened 2026-06-03
[client] iptables: fall back to per-IP ACL rules when ipset is unavailable

Lamera:
## Problem

Since v0.68, every peer ACL rule on the **iptables** backend depends on an
ipset. On kernels that do not provide the ipset hash module
(`ip_set_hash_net`) — common on embedded/IoT devices and stripped-down custom
kernels — ipset creation fails with `invalid argument`.

`AddPeerFiltering` returns that error, but `applyPeerACLs` only logs it and
continues. As a result the `NETBIRD-ACL-INPUT` chain stays empty, and the
trailing catch-all `DROP` blocks **all** policy-permitted inbound traffic.

The failure is silent and fail-closed: the peer becomes unreachable even though
the policy explicitly allows the traffic. Last working release was v0.67.4; the
regression was introduced in v0.68.0.

Reported in discussion #6125.

## Fix

- Probe ipset support **once** during `aclManager.init()` by attempting to
  create (and destroy) a small `hash:net` probe set.
- When ipset is unavailable, fall back to writing **per-IP iptables rules**
  (`-s <ip>`), matching the pre-0.68 behavior, so ACLs keep working. The
  existing `filterRuleSpecs` already supports the empty-ipset path.
- Log a clear `WARN` pointing at the missing kernel module so the degradation
  is visible to operators (instead of silently dropping traffic).

This keeps the ipset fast path for kernels that support it, while restoring
correct ACL behavior on kernels that don't.

## Testing

- Added `TestIptablesACLIPSetFallback`, which forces the fallback path and
  asserts that the resulting rule uses `-s <ip>` (no `--match-set`) and is
  actually present in the ACL chain.
- Verified on a real affected device (aarch64 embedded kernel without
  `ip_set_hash_net`): inbound connectivity to policy-permitted peers is
  restored.

## Scope

iptables backend only. The nftables backend uses native sets and is not
affected.

Fixes #6125

## Summary by CodeRabbit

* **Bug Fixes**
  * Firewall ACLs now detect when kernel ipset support is unavailable and automatically fall back to effective per-IP iptables rules, keeping access controls enforced.

* **Tests**
  * Added coverage verifying fallback rules are created and installed correctly when ipset support is unavailable.

## Documentation


Lamera:
@coderabbitai full review

Lamera:
@mlsmaycon can you support here with the failed checks? I also signed the CLA, but it doesn't finish the pending re-check.

Lamera:
Hey @lixmal,
Is there any chance this gets merged soon to main? Currently we lose access to many of our hosts when upgrading NetBird to 0.68.0 or higher. This fix is pretty critical to us.

xcstatus:
I also encountered the same problem.

xcstatus:
Is this blocked by #6322?  Should the fallback probe cover the exact ipset type used by ACL rules, including hash:net,port if applicable?

lixmal (approved):

PR #6367, opened 2026-06-08
[client] Migrate profile identity from display name to ID and allow renaming of profiles

theodorsm:
## Describe your changes

This PR decouples a profile's identity from its display name. Previously the profile name was used for its on-disk config filename, forcing unique, character-restricted names. Now each profile has a stable ID (the filename) while the name is free-form and stored inside the profile JSON. New rpc and cli cmd are added for renaming of profiles.

Changes
- IDs: new profiles get a generated 32-char hex ID, legacy profiles keep their name as their ID, so existing files keep working with no migration.
- Resolution: handles resolve server-side by exact ID > unique ID prefix > unique name. Ambiguous handles return a typed error with candidates.
- Validation: filename stems are guarded against path traversal and restricted charset (id.go). Display names are UTF-8-validated, control-char stripped, and length-capped.
- gRPC: `Profile`, `GetActiveProfileResponse`, and the `Add`/`Switch`/`Remove` responses gain an additive id field (name retained for older clients). New `RenameProfile` rpc. `profileNames` in profile requests acts as handles instead of only profile names (backward compat).
- CLI: `profile list --show-id adds an ID column: select/remove accept name, ID, or unique prefix, with disambiguation help on ambiguity. `profile rename` allows users to rename profiles, even the default one.
image

- UI: the tray client now keys profile operations (switch/remove/logout, config/login requests) on profile ID instead of name, logout confirmation still shows the display name. If there are multiple profiles with the same name, the id is added to the name to differentiate.
- Android: profile manager updated to the ID model.

Compatibility: fully backward compatible with legacy profiles, existing state files, and older clients all keep working.

Tests: new service_test.go plus updates across profilemanager/server/cmd tests.
OS test: have done manual testing on linux, macos, windows and android as they all use the go implementation of profiles (ios uses a native swift implementation).

Related PR for android-client: https://github.com/netbirdio/android-client/pull/189

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

https://github.com/netbirdio/docs/pull/781

## Summary

* **New Features**
  * Profiles now use stable, shortable IDs end-to-end; the CLI can optionally display IDs with `profile list --show-id`.
  * Added `netbird profile rename` command.
  * Profile switch/add/remove operations now show the resolved (short) profile ID.

* **Bug Fixes**
  * Improved profile resolution and disambiguation when handles/names are ambiguous.
  * Updated login, logout, and config flows to reliably target the active profile by ID.
  * Enhanced Android/UI profile selection to operate on IDs.

* **Tests**
  * Updated and expanded coverage for ID-based profile lifecycle and validation.


theodorsm:
This is the based on https://github.com/netbirdio/netbird/pull/6326, but a new PR due to rebase diff issues with main.

client/android/profile_manager.go
```diff
@@ -174,53 +185,56 @@ func (pm *ProfileManager) LogoutProfile(profileName string) error {
...
 	return nil
 }
 
 // getProfileConfigPath returns the config file path for a profile
 // This is needed for Android-specific path handling (netbird.cfg for default profile)
-func (pm *ProfileManager) getProfileConfigPath(profileName string) (string, error) {
-	if profileName == "" || profileName == profilemanager.DefaultProfileName {
+func (pm *ProfileManager) getProfileConfigPath(id string) (string, error) {
+	if id == "" || id == profilemanager.DefaultProfileName {
+		if !profilemanager.IsValidProfileFilenameStem(profilemanager.ID(id)) {
```
riccardomanfrin: `IsValidProfileFilenameStem` is checkin the default (`""` or `"default"`) while it's not checking the non default id path

client/internal/profilemanager/id.go
```diff
@@ -0,0 +1,118 @@
...
+		return "", fmt.Errorf("read random bytes: %w", err)
+	}
+	return ID(hex.EncodeToString(buf)), nil
+}
+
+// IsValidProfileFilenameStem reports whether id is safe to use as the stem
+// of a profile JSON filename.
+func IsValidProfileFilenameStem(id ID) bool {
+	s := id.String()
+	if s == "" || len(s) > maxProfileIDLen {
```
riccardomanfrin, line 49: The legacy behavior was for `s == ""` to be defaulted to the default config file (see `getProfileConfigPath`)
We are now disallowing this (maybe it's wanted)

riccardomanfrin (commented):
I made some comments around and there is the ui binary to remove

client/android/profile_manager.go
theodorsm: Moved the check before the default path.

client/internal/profilemanager/id.go
theodorsm: An empty id should not be valid. `getProfileConfigPath` or `GetConfigPath` are not called directly by the Android bindings, so this should not affect legacy cases.

riccardomanfrin (approved):
Did few tests also removing default.json and renaming it ant it works fine

PR #6373, opened 2026-06-08
[client] Preserve posture checks on config-only sync updates

pappz:
When management sends a MessageTypeControlConfig update (e.g. relay token rotation), the SyncResponse carries no NetworkMap and no Checks. Moving the updateChecksIfNew call after the nm == nil guard ensures posture checks are only updated when a full network map is present, preventing relay token rotation from silently clearing the previously applied posture check state.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed sync handling so configuration-only updates no longer trigger posture-check changes.
  * Ensured network configuration (TURN/STUN, relay, flow, DNS) and full-sync state are applied and persisted only when a complete network map is present, preventing partial updates from altering runtime checks.

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -914,13 +914,15 @@ func (e *Engine) handleSync(update *mgmProto.SyncResponse) error {
 		// todo update signal
 	}
 
-	if err := e.updateChecksIfNew(update.Checks); err != nil {
-		return err
-	}
-
 	nm := update.GetNetworkMap()
 	if nm == nil {
+		// config-only update (e.g. relay token rotation): posture checks and network map are intentionally absent,
+		// preserving the previously applied state
 		return nil
+	}
+
+	if err := e.updateChecksIfNew(update.Checks); err != nil {
+		return err
 	}
 
 	// Persist sync response under the dedicated lock (syncRespMux), not under syncMsgMux.
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -880,38 +880,8 @@ func (e *Engine) handleSync(update *mgmProto.SyncResponse) error {
 		e.handleAutoUpdateVersion(update.NetworkMap.PeerConfig.AutoUpdate)
 	}
 
-	if update.GetNetbirdConfig() != nil {
-		wCfg := update.GetNetbirdConfig()
-		err := e.updateTURNs(wCfg.GetTurns())
-		if err != nil {
-			return fmt.Errorf("update TURNs: %w", err)
-		}
-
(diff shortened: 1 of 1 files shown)


client/internal/engine.go
theodorsm, line 906: Is there a reason why this needs to be after the sync lock and store? If not, we should move it up to the `nm==nil` check to group the logic.

client/internal/engine.go
```diff
@@ -914,15 +914,17 @@ func (e *Engine) handleSync(update *mgmProto.SyncResponse) error {
...
+		// config-only update (e.g. relay token rotation): posture checks and network map are intentionally absent,
+		// preserving the previously applied state
 		return nil
 	}
 
+	if err := e.updateChecksIfNew(update.Checks); err != nil {
+		return err
+	}
+
 	// Persist sync response under the dedicated lock (syncRespMux), not under syncMsgMux.
```
theodorsm: We should make this logic even more explicit by a comment: 

```
NetworkMap != nil + checks present = set
NetworkMap != nil + checks nil = posture removed
NetworkMap == nil =  checks untouched (config-only)
```

theodorsm (commented):
I think it would improve the readability of the logic here if we move all the update logic to functions similar to `updateCheckIfNew`, so: `updateNetworkMapIfNew`and `updateNetbirdConfig`, `updateSignal` etc. This would let us improve unit testing in the future also.

client/internal/engine.go
```diff
@@ -947,6 +910,64 @@ func (e *Engine) handleSync(update *mgmProto.SyncResponse) error {
...
+		return
+	}
+
+	if err := e.syncStore.Set(update); err != nil {
+		log.Errorf("failed to persist sync response: %v", err)
+		return
+	}
+
+	log.Debugf("sync response persisted with serial %d", update.GetNetworkMap().GetSerial())
+}
```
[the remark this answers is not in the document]
pappz: On the caller side, we already check whether the update is nil.

pappz:
> I think it would improve the readability of the logic here if we move all the update logic to functions similar to `updateCheckIfNew`, so: `updateNetworkMapIfNew`and `updateNetbirdConfig`, `updateSignal` etc. This would let us improve unit testing in the future also.

Probably right, but in this PR I would like to be sure that I only changed the logic requested by the backend team. Anything else risks introducing a bug.

theodorsm (approved):

PR #5986, opened 2026-04-24
feat: add MDM managed app configuration support for Android and iOS

dbrieck:
## Summary

Add MDM (Mobile Device Management) managed app configuration support to the Go SDK, enabling Android and iOS native apps to receive and apply configuration from enterprise MDM solutions (Intune, Google Workspace, Jamf, etc.).

Fixes #1918

### Go SDK - Android (`client/android/`)
- **`managed_config.go`** — New `ManagedConfig` struct with gomobile-compatible setter API for 7 MDM keys: `managementUrl`, `setupKey`, `adminUrl`, `preSharedKey`, `rosenpassEnabled`, `rosenpassPermissive`, `disableAutoConnect`
- **`managed_config_test.go`** — Unit tests covering setters, Apply(), config override, key constants
- **`login.go`** — Added `LoginWithSetupKeySync()` for silent setup key registration from MDM

### Go SDK - iOS (`client/ios/NetBirdSDK/`)
- **`managed_config.go`** — iOS variant using `DirectUpdateOrCreateConfig()` for tvOS sandbox compatibility
- **`login.go`** — Added `LoginWithSetupKeySync()` matching Android API

## How It Works

1. Native app reads MDM config (Android: `RestrictionsManager`, iOS: `UserDefaults` managed domain)
2. Populates `ManagedConfig` via setter methods (gomobile-compatible)
3. Calls `Apply(configPath)` to write values to NetBird client config
4. Optionally calls `LoginWithSetupKeySync()` for zero-touch enrollment

MDM values are authoritative — they override any existing user configuration on every app launch.

## Companion PRs

- **android-client**: netbirdio/android-client — Android Enterprise integration (app_restrictions.xml, ManagedConfigReader, EngineRunner integration)
- **ios-client**: netbirdio/ios-client — Apple Managed App Configuration integration (ManagedConfigReader.swift, NetworkExtensionAdapter/PacketTunnelProvider integration)

## Summary by CodeRabbit

* **New Features**
  * Added synchronous login entry points for Android and iOS.
  * Added managed configuration support for Android Enterprise and iOS AppConfig: management URL, setup key, admin URL, pre-shared key, Rosenpass flags, and disable-auto-connect, with Apply persistence behavior.

* **Tests**
  * Added comprehensive tests validating managed-config state, persistence behavior, and non-persistence of setup keys.

PR #6374, opened 2026-06-09
[client] add MDM configuration profile support (Windows registry + macOS plist)

riccardomanfrin:
## Describe your changes

First MDM integration milestone for NetBird. Adds a `client/mdm/` package that
reads OS-native managed-configuration sources (Windows registry under
`HKLM\Software\Policies\NetBird` and macOS plist at
`/Library/Managed Preferences/io.netbird.client.plist`) and applies them to
the daemon's resolved configuration as the highest-priority layer. Scope of
this PR is Windows + macOS only; iOS and Android coverage is deferred to
follow-up PRs.

Highlights:

1. New package `client/mdm/` with per-OS loaders behind build tags. Shared
   `Policy` struct with typed accessors (`GetString` / `GetBool` / `GetInt` /
   `GetStringSlice`), case-insensitive `canonicalKey` lookup so admins can
   use either PascalCase ADMX-style names or camelCase plist-style names,
   1-minute reload ticker with `reflect.DeepEqual` diff + change callback,
   and a `SecretKeys` redaction set for log safety. 16 keys are surfaced:
   `managementURL`, `disableUpdateSettings`, `disableProfiles`,
   `disableNetworks`, `disableClientRoutes`, `disableServerRoutes`,
   `blockInbound`, `disableMetricsCollection`, `allowServerSSH`,
   `disableAutoConnect`, `preSharedKey`, `rosenpassEnabled`,
   `rosenpassPermissive`, `wireguardPort`, `splitTunnelMode`,
   `splitTunnelApps`.

2. Daemon integration in `client/internal/profilemanager/config.go`:
   `Config.apply()` now runs `applyMDMPolicy(LoadPolicy())` as the last
   layer (defaults → on-disk JSON → env → CLI/UI → MDM). MDM-derived
   values are never persisted back to `default.json` / `<profile>.json`,
   so un-enrollment takes effect on the next reload with no stale policy
   on disk.

3. `Server.Start` arms the MDM ticker; any diff triggers
   `restartEngineForMDM` which re-resolves config + re-spawns
   `connectWithRetryRuns` + emits `SystemEvent{type=config_changed,
   source=mdm}` so the GUI tray / CLI subscribers can refresh without
   polling. Same event is now also emitted from `Server.Start` (source
   `startup`) and `Server.Up` (source `up_rpc`).

4. Per-field MDM gate on `SetConfig` (`client/server/server.go`) and on
   `Login` (same file) — value-aware: a request that sets a field to the
   same value the policy already enforces is a no-op echo, not a
   conflict. Conflicts return `codes.FailedPrecondition` with the new
   `MDMManagedFieldsViolation { repeated string fields }` proto detail
   listing the offending key names. The `Login` gate is required because
   the CLI `netbird up --flag=...` path falls through to `Login` when
   `SetConfig` is rejected (legacy backward-compat behaviour); without
   the matching `Login` gate, MDM policy would be bypassable from the
   CLI.

5. `checkProfilesDisabled` / `checkUpdateSettingsDisabled` /
   `checkNetworksDisabled` now resolve MDM-first: when the MDM key is
   present in the policy, that value wins (true OR false), and the CLI
   `--disable-*` flag is the fallback only when the MDM source is silent
   on the key. This honours the "MDM is authoritative" semantic: an
   admin pushing `disableNetworks=false` via MDM re-enables the feature
   even on a host installed with `--disable-networks`.

6. gRPC surface: `GetConfigResponse.mDMManagedFields[]` lists the
   currently MDM-enforced key names so the UI can grey out the
   corresponding inputs. New `MDMManagedFieldsViolation` message used as
   gRPC error detail (see point 4).

7. Legacy Fyne tray UI (`client/ui/client_ui.go`): submenu checkboxes
   (`Allow SSH`, `Connect on Startup`, `Enable Quantum-Resistance`,
   `Block Inbound Connections`) get a ` (MDM)` title suffix and
   `Disable()` when MDM-managed; the value itself is sourced from the
   daemon-resolved config so the checkbox always reflects ground truth.
   Advanced Settings form: `iMngURL` and `iInterfacePort` get an inline
   ` (MDM)` suffix in the text; password entry `iPreSharedKey` uses an
   empty Text + placeholder (`MDM-managed` / `configured`) so the eye
   toggle never leaks the redaction sentinel. `mProfile.setEnabled(false)`
   now disables parent + children (Manage Profiles / Deregister /
   per-profile sub-items) instead of `Hide()` so the user sees a
   greyed-out menu rather than a missing menu.

8. UI 2-second GetFeatures poll replaced with event-driven refresh
   hooked to `config_changed`. One initial sync at boot; the periodic
   loop continues to handle status/connection updates (no event for
   those yet). Transition caches (`profilesEnabled`,
   `networksMenuEnabled`) avoid systray `ClickedCh` subscription drop
   from repeated Enable/Disable calls.

9. New `netbird debug config` subcommand
   (`client/cmd/debug.go`) dumps the daemon's `GetConfigResponse` as
   JSON via protojson with `EmitUnpopulated:true` — used for verifying
   MDM application end-to-end (the `mDMManagedFields` array is the
   single source of truth for "which keys is the daemon actually
   enforcing right now").

Admin manifests + deployment helpers (`docs/`):

- `netbird.admx` + `netbird.adml` — Group Policy / Intune ADMX
  ingestion. All 16 keys.
- `netbird-macos.mobileconfig` — Apple Configuration Profile template
  with `com.apple.ManagedClient.preferences` payload targeting bundle
  `io.netbird.client`.
- `netbird-policy.reg` — Windows `.reg` sample (full set of 16 keys)
  for `reg import` or sidecar attachment.
- `netbird-policy.reg.ps1` — JumpCloud Windows command companion that
  imports an attached `netbird-policy.reg` after wiping the existing
  policy key (single source of truth).
- `macos-netbird-mdm.sh` — JumpCloud macOS command counterpart;
  inline-variable script that emits the plist.

Internal scoping task. Anchor for reviewers: the previous
community attempt at MDM support is at
https://github.com/netbirdio/netbird/pull/5986 — this PR is a from-scratch
rewrite with the gaps identified in that PR's review fixed
(daemon-side enforcement, gRPC `mDMManagedFields` surface, per-field
`FailedPrecondition`, no `mdm.json` intermediate file, 1-minute reload
ticker). 
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

This PR ships the admin-facing manifests inline under `docs/`:

- `io.netbird.client.plist`
- `netbird-macos.mobileconfig`
- `netbird-macos.sh`
- `netbird-policy.reg`
- `netbird-policy.reg.ps1`
- `netbird.adml`
- `netbird.admx`

Additionally the PR is paired by a `netbirdio/docs` PR for the end-user-facing admin
guide (linked below). The in-tree manifests target tooling that
expects them at known paths (ADMX/ADML in `PolicyDefinitions`,
`.mobileconfig` for `profiles install`); the `netbirdio/docs` PR
covers the narrative ("how do I push the NetBird policy from Intune /
Jamf / JumpCloud") that doesn't fit inside a manifest.

https://github.com/netbirdio/docs/pull/783

## Summary by CodeRabbit

* **New Features**
  * Full MDM support: daemon reports managed fields, enforces policy on config changes (rejects conflicts with violation details), auto-applies policy updates with live restarts, and exposes MDM-driven feature flags and UI locks.
  * CLI: new debug subcommand to dump resolved client config as indented JSON.

* **UI**
  * Tray and Settings reflect MDM locks and show PSK as configured/managed without leaking secrets.

* **Tests**
  * Added comprehensive tests for MDM loading, enforcement, ticker behavior, and client/server flows.

* **Documentation**
  * Added macOS/Windows MDM templates, deployment scripts, and policy guidance.


client/mdm/ticker.go
```diff
@@ -0,0 +1,140 @@
...
+// key set with identical values. Nil and empty policies compare equal.
+func PoliciesEqual(a, b *Policy) bool {
+	if a.IsEmpty() && b.IsEmpty() {
+		return true
+	}
+	if a == nil || b == nil {
+		return false
+	}
+	return reflect.DeepEqual(a.values, b.values)
+}
```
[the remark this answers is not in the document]
riccardomanfrin: False positive: LoadPolicy always return a policy even when there is an error or there is a nil value map

riccardomanfrin:
@coderabbitai generate docstrings

riccardomanfrin:
@coderabbitai generate docstrings

client/mdm/ticker.go
```diff
@@ -0,0 +1,159 @@
...
+// minutes. It has no effect on production builds, where testing.Testing()
+// always returns false.
+const testReloadInterval = 1 * time.Second
+
+// reloadInterval returns the production cadence, or the accelerated test
+// cadence when running under `go test`. Centralising the choice here keeps
+// reloadInterval selects the polling interval used to re-read the OS-native MDM policy.
+// reloadInterval selects the polling interval used for policy reloads.
+// It returns testReloadInterval when running under `go test` (testing.Testing() == true), and defaultReloadInterval otherwise.
+func reloadInterval() time.Duration {
```
pappz: Please keep the Go recommendation in the order of the source code like
const->var->struct->constructor->exported func->unexported-> static->

client/mdm/ticker.go
```diff
@@ -0,0 +1,159 @@
+package mdm
+
+import (
+	"context"
+	"reflect"
+	"sort"
+	"testing"
```
pappz: Do not import testing in prod code. Pass it as parameter to the constructor. The default prod parameter can be still const.

pappz (changes requested):

client/server/server.go
```diff
@@ -237,11 +243,20 @@ func (s *Server) Start() error {
...
+// Server.connectionGoroutineRunning() which non-blockingly checks the close state
+// of clientGiveUpChan. The defer does NOT touch s.mutex; the daemon's
+// "intent" (clientRunning) is maintained by the RPC handlers, not by this
+// goroutine.
 func (s *Server) connectWithRetryRuns(ctx context.Context, profileConfig *profilemanager.Config, statusRecorder *peer.Status, runningChan chan struct{}, giveUpChan chan struct{}) {
 	defer func() {
-		s.mutex.Lock()
-		s.clientRunning = false
-		s.mutex.Unlock()
+		if giveUpChan != nil {
```
pappz, line 257: if you do not set `s.clientRunning to = false` then what will be the side effect? In case of backoff exhaust the state will be invalid and the `Start()` will not work any more. Maybe same with the `ExposeService()`

The close in defer is fine.

client/server/mdm.go
```diff
@@ -36,34 +36,45 @@ var loadMDMPolicy = mdm.LoadPolicy
...
 	// Wait for previous connectWithRetryRuns to exit so we don't end up
 	// with two goroutines fighting over the same status recorder + engine.
-	if giveUpChan != nil {
+	// The teardown engages a fan-out of engine goroutines (peer workers,
+	// signal handler, route manager, ...). close(clientGiveUpChan)
+	// happens in the function-scope defer of connectWithRetryRuns, on
+	// every exit path (ctx cancel, DisableAutoConnect single-shot,
+	// backoff exhausted, panic) — see the defer in server.go.
+	if s.clientGiveUpChan != nil {
+		giveUpChan := s.clientGiveUpChan
```
pappz: Why do you copy?

client/server/mdm.go
```diff
@@ -36,34 +36,45 @@ var loadMDMPolicy = mdm.LoadPolicy
...
 		select {
 		case <-giveUpChan:
 		case <-time.After(5 * time.Second):
```
pappz: After the timeout, the old engine may still be running, and you start another one. Avoid this because it can cause an inconsistent state in the OS involving the WireGuard interface, routes, and DNS settings.

pappz (changes requested):

client/server/server.go
riccardomanfrin: My understanding is that `Start` is never executed when the process is already running. It's only called at program start. In this sense also the 
```
if s.clientRunning {
     return nil
}
```
in server.go:150 would not be needed.

So I only protected the "Up" where I check both conditions: that the client is running (`clientRunning == true`) which will *always* be true after the Start/Up and until a Down or Logout occurs (places from which `cleanupConnection()` is called) or if it was started/upped (`clientRunning == true`) but go routine has terminated in which case the `connectionGoroutineRunning()` will return `false`

client/server/mdm.go
riccardomanfrin: I copied this from the  `Down()` pattern, after which we can run the Up, and end up in the same situation. I believed it was a sufficient amount of time. Is increasing it a solution?

client/server/mdm.go
riccardomanfrin: leftover of the refactor fixed :)

client/internal/profilemanager/config.go
```diff
@@ -174,6 +175,23 @@ type Config struct {
 	LazyConnectionEnabled bool
 
 	MTU uint16
+
+	// policy is the MDM policy that produced the currently-set values for
+	// any MDM-enforced fields. Set by applyMDMPolicy at the tail of apply()
+	// and reset on every apply() invocation. Never persisted to disk.
+	// Callers query enforcement state via Policy() and the mdm.Policy API
+	// (HasKey, ManagedKeys, IsEmpty).
+	policy *mdm.Policy `json:"-"`
```
pappz, line 188: Why need to json:"-" if it is not exported

client/internal/profilemanager/config.go
```diff
@@ -612,10 +630,97 @@ func (config *Config) apply(input ConfigInput) (updated bool, err error) {
...
+	// supersedes defaults, on-disk config, env vars and CLI input.
+	config.applyMDMPolicy(loadMDMPolicy())
+
 	return updated, nil
 }
 
-// parseURL parses and validates a service URL
+// loadMDMPolicy is the package-level indirection used by apply() to read the
+// active MDM policy. Tests override this to inject a fake policy.
+var loadMDMPolicy = mdm.LoadPolicy
```
pappz: Move vars to the beggining to the file

client/mdm/ticker.go
```diff
@@ -0,0 +1,129 @@
...
+			t.prev = curr
+		}
+	}
+}
+
+// PoliciesEqual reports whether two Policy instances carry the same
+// managed key set with identical values. Nil and empty policies
+// compare equal; one-nil/one-non-empty compare not equal; otherwise
+// the underlying values maps are compared with reflect.DeepEqual.
+func PoliciesEqual(a, b *Policy) bool {
```
pappz: why is it exported?

pappz (changes requested):

client/internal/profilemanager/config.go
riccardomanfrin: Lower-case keeps it private to the package; the json:"-" tag forbids serialization to disk — now and in the future — to keep a single source of truth (the live MDM read). True, the lower case already implicitly prevents serialization, but I preferred an explicit forbid rule in case the field is ever made public

pappz (approved):

PR #6394, opened 2026-06-10
[client] introduce client-side event aggregation

dmitri-netbird:
## Describe your changes
To reduce the frequency at which flow events are emitted, a client-aggregation of flow events is implemented in this PR.

Events are aggregated over a time window by destination address, destination port, protocol, and icmp code (for icmp). Total number of start-, end-, and drop-events is tracked, as well as the number of transmitted and received bytes and packets.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

https://github.com/netbirdio/docs/pull/800

## Summary by CodeRabbit

* **New Features**
  * Network flow events now include an aggregation window with start/end timestamps.
  * Events now report per-flow counters for starts, ends, and drops.
  * Flow event delivery is more reliable: events are aggregated and unacknowledged events are retried automatically until ACKed.
* **Documentation**
  * Updated API schema and generated models to include the new aggregation window fields and start/end/drop counters.


client/internal/netflow/manager.go
```diff
@@ -236,6 +248,41 @@ func (m *Manager) receiveACKs(ctx context.Context, client *client.GRPCClient) {
...
+					// do not retry if the event is less than 1 sec old
+					continue
+				}
+				if err := m.send(e); err != nil {
+					ticker = time.NewTimer(retryBackoff.NextBackOff()) //nolint:staticcheck,wastedassign
+					break
+				}
+			}
+			retryBackoff.Reset()
+			ticker = time.NewTimer(time.Second)
```
riccardomanfrin: I think this is retrying every 1 second without backoff

client/internal/netflow/store/memory.go
```diff
@@ -48,3 +55,78 @@ func (m *Memory) DeleteEvents(ids []uuid.UUID) {
...
+func (am *AggregatingMemory) GetAggregatedEvents() []*types.Event {
+	aggregated := make(map[aggregationKey]*types.Event)
+	for _, v := range am.events {
+		lookupKey := aggregationKey{destAddr: v.DestIP, destPort: v.DestPort, protocol: uint8(v.Protocol), icmpType: v.ICMPType}
+		if _, ok := aggregated[lookupKey]; !ok {
+			aggregated[lookupKey] = v.Clone()
+			event := aggregated[lookupKey]
+
+			if event.Protocol != types.ICMP && event.Protocol != types.ICMPv6 && event.Protocol != types.UDP && event.Protocol != types.TCP {
+				lookupKey.unique = time.Now().UnixNano() // to make the lookup key unique so we don't aggregate on it
```
riccardomanfrin: I'm not sure about setting the unique here

client/internal/netflow/store/memory.go
```diff
@@ -48,3 +55,78 @@ func (m *Memory) DeleteEvents(ids []uuid.UUID) {
...
+	destAddr netip.Addr
+	destPort uint16
+	protocol uint8
+	icmpType uint8
+	unique   int64 // used to prevent aggregation on non icmp/udp/tcp events
+}
+
+func (am *AggregatingMemory) GetAggregatedEvents() []*types.Event {
+	aggregated := make(map[aggregationKey]*types.Event)
+	for _, v := range am.events {
```
riccardomanfrin, line 96: can it race on am.events?

client/internal/netflow/manager.go
```diff
@@ -236,6 +248,41 @@ func (m *Manager) receiveACKs(ctx context.Context, client *client.GRPCClient) {
 	}
 }
 
+func (m *Manager) startRetries(ctx context.Context) {
+	ticker := time.NewTimer(m.retryInterval)
```
riccardomanfrin: `ticker` variable can be overwritten below. `defer ticker.Stop()` would prevent timer leaks (is it ticker or timer?)

riccardomanfrin (commented):
Compared to what was there from before the aggregation key does not consider direction and source IP and port (credits to @lixmal ).

I think the source port as being ephemeral is ok to drop, but dir and source IP are probably to keep.

I think there might be one more little problem in general: all the fields that are part of the Event, originate as information from the traffic (e.g. NOT ID / Timestamp), but are not part of the Aggregation Key should ideally be removed from the sent event - which I suppose is not possible for  backward compat-. If not possible they should at least be cleared before sending the event. Otherwise you are populating them with the first collected event of the time window which is not statistically meaningful.

client/internal/netflow/manager.go
dmitri-netbird: the logic here is that all unacked events are retried once a second; if there's a an error on send, backoff kicks in. I didn't want to track retries per event, as those are not useful most of the time (when we see a failure on send); for the same reason I stop sending when an error occurs.

Does this make sense? Did you have something different in mind?

dmitri-netbird:
> Compared to what was there from before the aggregation key does not consider direction and source IP and port (credits to @lixmal ).

Nods, fixed this.


> I think there might be one more little problem in general: all the fields that are part of the Event, originate as information from the traffic (e.g. NOT ID / Timestamp), but are not part of the Aggregation Key should ideally be removed from the sent event - which I suppose is not possible for backward compat-. If not possible they should at least be cleared before sending the event. Otherwise you are populating them with the first collected event of the time window which is not statistically meaningful.

Nods, can't drop the fields, as it would be a breaking change. The fields that we are not explicitly aggregating on:
 * flowId -- I think it's used in quite a few places, but I think it's ok to keep it w/e value we end up with; an alternative is to create a new uuid, but I don't see much of a difference (I don't think we can end up with the same flowId for multiple distinct aggregated flows)
 * ruleId -- if I'm reading the code correctly it should be the same for a given dstAddr/port combination? Not sure if this is used for icmp traffic?
 *  SourceResourceID -- I think it's unique for a given src IP/port tuple, not sure what should be used instead? Possibly ok to reuse one of the resouceIds used in the aggregation?
 * DestResourceID -- should be the same for a given aggregated flow
 * icmp code -- not sure what to do with this, could reset it to zero?
 * the rest of the fields we aggregate on
 
 Got clarification re: SourceResourceID -- for a given src ip, dst ip, dst port, and direction, resouce id should stay the same.

client/internal/netflow/store/memory.go
dmitri-netbird: I could use a real random int here instead, is that would you meant?

client/internal/netflow/store/memory.go
dmitri-netbird: technically yes, but not how it's currently used: when `ResetAggregationWindow()` is called, current AggregatingMemory is returned, and a new one is created. The only user of the returned memory is `client/internal/netflow/manager.go`. I'll acquire a lock here to avoid issues in future.

client/internal/netflow/manager.go
dmitri-netbird: haha, sry about the confusing name, it's a timer (renamed). Since golang 1.23 there's no requirement to stop the timer/ticker to prevent leaks, as the channel used is no longer buffered, see here (https://tip.golang.org/doc/go1.23#timer-changes) for the description of the change.

dmitri-netbird:
proto version check is failing b/c flow grpc bindings (`flow/proto/flow_grpc.pb.go`) do not have protoc metadata header in main/HEAD.

client/internal/netflow/manager.go
```diff
@@ -236,6 +248,43 @@ func (m *Manager) receiveACKs(ctx context.Context, client *client.GRPCClient) {
...
 
+// We effectively never drop events (see MaxInterval), which makes eventsWithoutAcks unbounded.
+// We may want to limit the max size of the store, and start dropping oldest events when the threshold is reached.
+func (m *Manager) startRetries(ctx context.Context) {
+	timer := time.NewTimer(m.retryInterval)
+	retryBackoff := backoff.WithContext(&backoff.ExponentialBackOff{
+		InitialInterval:     1 * time.Second,
+		RandomizationFactor: 0.5,
+		Multiplier:          1.7,
+		MaxInterval:         m.flowConfig.Interval / 2,
```
lixmal: `.Interval` should be accesed under a lock

flow/proto/generate.sh
```diff
@@ -10,6 +10,7 @@ fi
 
 old_pwd=$(pwd)
 script_path=$(dirname $(realpath "$0"))
+echo "$script_path"
```
lixmal: debug leftover?

lixmal (changes requested):

client/internal/netflow/manager.go
dmitri-netbird: this is an interesting one, actually; golang race detector doesn't report it, as it would require a rapid succession of FlowConfig updates, which we don't test. Probably not a common case, but I fixed all accesses to FlowConfig.Interval (I don't see anything else being accessed from that struct outside of the 
lock) by passing a copy of the value to go routines...

client/internal/netflow/manager.go
```diff
@@ -236,6 +248,43 @@ func (m *Manager) receiveACKs(ctx context.Context, client *client.GRPCClient) {
 	}
 }
 
+// We effectively never drop events (see MaxInterval), which makes eventsWithoutAcks unbounded.
+// We may want to limit the max size of the store, and start dropping oldest events when the threshold is reached.
+func (m *Manager) startRetries(ctx context.Context) {
```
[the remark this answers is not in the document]
dmitri-netbird: this behaviour's been re-used from the previous implementation.

lixmal (dismissed):

riccardomanfrin (dismissed):

lixmal (dismissed):

dmitri-netbird:
closing in favour of https://github.com/netbirdio/netbird/pull/6627

PR #6409, opened 2026-06-11
[client] Fixes DisableAutoConnect semantics

riccardomanfrin:
## Describe your changes

Clarify the semantics of the `DisableAutoConnect` flag in the client
daemon. The flag previously short-circuited every code path that
considered spawning the connection goroutine — service Start, the Up
RPC handler, and future coming MDM-driven engine restart.

The flag's intent is narrower: it governs whether the daemon should
automatically connect **at service start**. An explicit user request,
or a policy refresh applied to an already-running engine, should not be
subordinated to the flag.

New semantics:

1. **Service Start + `DisableAutoConnect=true`** → daemon stays
   idle, no connection goroutine spawned, `clientRunning` stays
   `false`. The state RPC reports `StatusIdle`.
2. **Service Start + `DisableAutoConnect=false`** → daemon spawns
   `connectWithRetryRuns`, retry loop runs until ctx cancelled.
3. **Up RPC (any value of `DisableAutoConnect`)** → flag ignored.
   The user explicitly asked to connect — by definition not "auto".
4. **MDM-triggered restart (any value of `DisableAutoConnect`)** →
   flag ignored. An MDM policy change applies new config to an
   already-running engine; treated as an implicit Up.
5. **Down / Stop / Logout** → cancels ctx; `cleanupConnection`
   clears `clientRunning`. `DisableAutoConnect` is not involved.

Code changes:

- `client/server/server.go`: removed the `DisableAutoConnect` branch
  inside `connectWithRetryRuns` so the retry loop no longer
  collapses to a single-shot when the flag is set. The flag is now
  only consulted at the top of `Start()` (the early-return path that
  sets `StatusIdle`).
  
No public issue. Semantic divergence between the flag name and its
behaviour surfaced during the MDM integration review
(https://github.com/netbirdio/netbird/pull/6374) — `DisableAutoConnect`
was being read as a runtime kill switch in addition to a startup
gate. This PR confines it to its documented purpose so the MDM PR
does not have to choose between two reasonable but incompatible
interpretations.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

The `DisableAutoConnect` flag is already documented as a
startup-only auto-connect gate (CLI help text and ConfigInput
field comment). The previous implementation diverged from the
documented intent; this PR aligns the implementation with the
existing docs, so no doc update is required. The behavioural
contract becomes: setting the flag prevents the boot-time
connection only — explicit Up requests and MDM-driven restarts
remain unaffected.

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated connection retry behavior so connection attempts consistently follow the backoff-based retry flow.
  * Added improved debug logging when a client run exits with an error, making connection termination issues easier to diagnose.


client/server/server.go
```diff
@@ -259,14 +259,6 @@ func (s *Server) connectWithRetryRuns(ctx context.Context, profileConfig *profil
 		}
 	}()
 
-	if s.config.DisableAutoConnect {
```
mlsmaycon, line 262: The historical reason for this handler is that when the auto-connect is disabled, we should not use the backoff retry in case of a response from management. 

e.g., for any reason, management will respond with a permission denied, because a peer has expired, we are still trying for a few days just to confirm there wasn't an issue that should not cause a client shutdown.

mlsmaycon (commented):
we can probably leave it as is with more docs. But I don't see any harm in keeping it here. If we keep it, let's add some docs for it.

client/server/server.go
riccardomanfrin: I see the point, the problem is that we are abusing the autoconnect with the current approach: if the peer has expired either we want to wait some external intervention - if any- to fix it, or we want to exit, regardless ofthe autoconnect.

In any case I understand what is the desired expectation, so I will close the PR and think of a strategy that makes more sense in the other refactor that I'm attempting

PR #6431, opened 2026-06-14
[misc] Installation fallback for when host is missing openssl dependency

adikavemen:
## Describe your changes
Added checks for openssl dependency being met in the installation script. When the dependency is not met the script spins up temporary alpine openssl containers to generate keys. Tries to use pre-existing openssl image if present. Handles clean up if images are retrieved.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

This PR does not change how the script already functions. It provides a fallback for when openssl is not available on the host.

Paste the PR link from https://github.com/netbirdio/docs here: N/A

## Summary by CodeRabbit

* **Chores**
  * Improved the getting-started setup by initializing default configuration values before secret generation.
  * Updated secret generation to adapt to the environment: uses locally available cryptographic tooling when present, otherwise generates secrets via a containerized fallback.
  * Maintains consistent secret formatting across scenarios and streamlines related cleanup.


adikavemen:
Hi @pappz & @riccardomanfrin sorry for the ping. Just wanted to get some visibility on this PR.

champtar:
@adikavemen why not use `head -c 32 /dev/urandom | base64` ?

adikavemen:
@champtar I was unaware of /dev/urandom and how openssl often uses it. Openssl was what was previously used in the script so I was using the same key generation method. Openssl remains the better option as it would use available hardware in keygen where possible. It also uses more bits in encryption which makes openssl more secure than using /dev/urandom this way

adikavemen:
@jnfrati Is this going to be part of the single installation script consolidation?

PR #6455, opened 2026-06-17
[client] Forward non-address DNS record types through route forwarders

lixmal:
## Describe your changes

DNS queries for a domain handled by a NetBird DNS route are now fully owned by that route's forwarder. Previously only A/AAAA were served; any other record type fell through to the client's system resolver, which is not authoritative for the routed (often internal, split-horizon) zone and answers NXDOMAIN. Because NXDOMAIN is name-scoped, that poisoned the whole name and broke the A/AAAA records the route does serve.

- Forward all query types from the DNS route interceptor to the peer forwarder instead of passing non-A/AAAA queries down the handler chain
- Resolve MX, TXT, NS, SRV, CNAME and PTR through the host resolver and return real records
- Answer NODATA (NOERROR, empty) for record types the host resolver cannot serve, never NXDOMAIN or NOTIMP, so a routed name is not poisoned
- Distinguish NXDOMAIN from NODATA on a missing record by probing whether the name has any address, treating an inconclusive probe as existing
- Attach an Extended DNS Error ("Not Supported") on the NODATA returned for unsupported types when the client uses EDNS0

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Behavioral fix to existing DNS route forwarding; no user-facing configuration or API changes.

## Summary by CodeRabbit

* **New Features**
  * Added DNS forwarding support for MX, TXT, NS, SRV, CNAME, and PTR (including reverse-DNS PTR handling).
  * When EDNS0 is enabled, unsupported query types now return an EDNS0 EDE “not supported” response.

* **Bug Fixes**
  * Improved DNS response correctness for non-address lookups to return **NODATA** (empty answers) instead of **NXDOMAIN**.
  * For intercepted domains, queries are forwarded to the peer resolver to avoid system-resolver NXDOMAIN behavior.

* **Tests**
  * Expanded coverage for PTR name parsing, TXT chunking, record routing, and EDNS0 EDE behavior.


riccardomanfrin (approved):
Looks ok to me, it's improving the DNS support. Routing peer not updated would just reply with NotImplemented

PR #6462, opened 2026-06-18
prep for lneto use

soypat:
WIP


soypat:
abandon in favor of https://github.com/netbirdio/netbird/pull/6516

PR #6479, opened 2026-06-19
[client] Fix profile regressions in `up --profile` and `status`

riccardomanfrin:
## Describe your changes

PR #6367 (migrating profile identity from display name to ID) introduced two regressions in the CLI profile flow. This PR fixes both.

1. **`netbird up --profile <name>` no longer auto-created a missing profile.** Before #6367, switching to a non-existent profile name simply recorded it as the active profile and the config file was created lazily. After the migration, the daemon resolves the supplied handle against existing profiles and returns `codes.NotFound`, which broke ephemeral CI workflows relying on `netbird up --profile <name> --setup-key <key>` to provision the profile on the fly (reported in discussion #6476).

   `upFunc` in `client/cmd/up.go` now detects `codes.NotFound` from `switchProfile`, creates the profile via the daemon's existing `AddProfile` RPC (`createProfile` helper in `client/cmd/login.go`), and retries the switch. To make the status code inspectable by the caller, `switchProfile` now wraps the daemon error with `%w` instead of `%v`.

2. **`netbird status` showed an empty `Profile:` field** for both the default and custom profiles. After #6367, `ProfileManager.GetActiveProfile` returns only the profile ID (not the display name), and the CLI — running as the invoking user — cannot read the per-user profile files under `/var/lib/netbird/<user>/` (root-owned, `0700`) to resolve the name itself.

   `client/cmd/status.go` now resolves the display name through the daemon's existing `GetActiveProfile` RPC, which runs as root and maps the active ID to its display name. Added the `getActiveProfileName` helper and removed the now-unused `profilemanager` import.

Manually verified: `netbird up --profile <new-name> --setup-key <key>` provisions and connects, and `netbird status` shows the correct profile name for both default and custom profiles.

https://github.com/netbirdio/netbird/discussions/6476 — regression report for `netbird up --profile` + setup-key failing on v0.73.0 after #6367.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

These are regression fixes that restore the documented pre-0.73.0 behavior of `netbird up --profile` and the `netbird status` output. No public CLI flags, API surface, or configuration options change, so existing documentation remains accurate.

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RPC error handling for profile-related operations, including clearer error wrapping during daemon failures.
* **New Features**
  * The `up` command now switches to a specified profile, and if it doesn’t exist, creates it and then retries activation.
* **Improvements**
  * The `status` command now pulls the active profile’s display name from the running daemon.
  * Profile creation output is more reliable by using the daemon-returned profile ID after creation.

### DIFF
--- a/client/cmd/login.go
+++ b/client/cmd/login.go
@@ -248,7 +248,7 @@ func switchOrCreateProfile(ctx context.Context, pm *profilemanager.ProfileManage
 		// have created the profile between the NotFound above and this
 		// call, in which case the retried switch still succeeds. Only
 		// surface the create error if the switch also fails.
-		createErr := createProfile(ctx, handle, username)
+		_, createErr := createProfile(ctx, handle, username)
 		if resolvedID, err = switchProfile(ctx, handle, username); err != nil {
 			if createErr != nil {
 				return fmt.Errorf("create profile: %w", createErr)
@@ -264,13 +264,14 @@ func switchOrCreateProfile(ctx context.Context, pm *profilemanager.ProfileManage
 }
 
 // createProfile asks the daemon to create a new profile with the given
-// display name. Helpful for `netbird up --profile <name>` which
-// auto-creates a missing profile.
-func createProfile(ctx context.Context, profileName, username string) error {
+// display name and returns its generated ID. It is the single entry point
+// for profile creation, shared by `netbird profile add` and the
+// `netbird up --profile <name>` auto-create path.
+func createProfile(ctx context.Context, profileName, username string) (profilemanager.ID, error) {
 	conn, err := DialClientGRPCServer(ctx, daemonAddr)
 	if err != nil {
 		//nolint
-		return fmt.Errorf("failed to connect to daemon error: %v\n"+
+		return "", fmt.Errorf("failed to connect to daemon error: %v\n"+
 			"If the daemon is not running please run: "+
 			"\nnetbird service install \nnetbird service start\n", err)
 	}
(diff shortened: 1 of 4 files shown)


pappz (dismissed):

theodorsm (dismissed):

client/cmd/login.go
```diff
@@ -227,7 +227,69 @@ func switchProfile(ctx context.Context, handle string, username string) (profile
...
+// and returns the new profile's ID. It is the single entry point for profile
+// creation, shared by `netbird profile add` and the `netbird up --profile
+// <name>` auto-create path.
+func addProfileOnDaemon(ctx context.Context, client proto.DaemonServiceClient, profileName, username string) (profilemanager.ID, error) {
+	resp, err := client.AddProfile(ctx, &proto.AddProfileRequest{
+		ProfileName: profileName,
+		Username:    username,
+	})
+	if err != nil {
+		return "", fmt.Errorf("add profile failed: %w", err)
```
theodorsm: Let's move these helper functions to `client/cmd/profile.go`.

client/cmd/login.go
```diff
@@ -227,7 +227,69 @@ func switchProfile(ctx context.Context, handle string, username string) (profile
...
+			}
+			return err
+		}
+	}
+
+	if err := pm.SwitchProfile(resolvedID); err != nil {
+		return err
+	}
+	return nil
+}
```
theodorsm: Let's move this helper function to `client/cmd/up.go`.

theodorsm (changes requested):

theodorsm (dismissed):

theodorsm (approved):
Tested and works :shipit:

PR #6508, opened 2026-06-22
[client] Skip firewall ruleset rebuild when config is unchanged

pappz:
ApplyFiltering rebuilt every peer and route ACL and flushed the firewall on every sync, with no guard for an unchanged configuration. Management re-sends the same network map far more often than it actually changes (account-wide updates, peer meta churn), so on busy accounts this is the dominant client-side cost of redundant syncs — especially with a large route set and a userspace firewall.

Hash the inputs ApplyFiltering consumes (peer rules, route rules, the empty flag and the dns-route feature flag) and skip the rebuild + flush when the hash matches the last successfully applied update. Mirrors the guard the DNS server already uses (previousConfigHash). The hash is only recorded after apply and flush both succeed, so a failed update is not skipped on the next (possibly identical) sync and gets a chance to reconcile the firewall state.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved firewall/route ACL updates to avoid unnecessary rebuilds and flushes when the effective configuration hasn’t changed since the last successful update.

* **Tests**
  * Added regression coverage to ensure identical inputs don’t trigger redundant re-application.
  * Added hash determinism tests (order-independent) and benchmarks across multiple input sizes.

### DIFF
--- a/client/internal/acl/manager.go
+++ b/client/internal/acl/manager.go
@@ -116,14 +116,16 @@ func (d *DefaultManager) ApplyFiltering(networkMap *mgmProto.NetworkMap, dnsRout
 // firewall state, so an identical hash means an identical resulting ruleset.
 func (d *DefaultManager) firewallConfigHash(networkMap *mgmProto.NetworkMap, dnsRouteFeatureFlag bool) (uint64, error) {
 	return hashstructure.Hash(struct {
-		PeerRules           []*mgmProto.FirewallRule
-		PeerRulesIsEmpty    bool
-		RouteRules          []*mgmProto.RouteFirewallRule
-		DNSRouteFeatureFlag bool
+		PeerRules            []*mgmProto.FirewallRule
+		PeerRulesIsEmpty     bool
+		RouteRules           []*mgmProto.RouteFirewallRule
+		RouteRulesIsEmpty    bool
+		DNSRouteFeatureFlag  bool
 	}{
 		PeerRules:           networkMap.GetFirewallRules(),
 		PeerRulesIsEmpty:    networkMap.GetFirewallRulesIsEmpty(),
 		RouteRules:          networkMap.GetRoutesFirewallRules(),
+		RouteRulesIsEmpty:   networkMap.GetRoutesFirewallRulesIsEmpty(),
 		DNSRouteFeatureFlag: dnsRouteFeatureFlag,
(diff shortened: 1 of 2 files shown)


client/internal/acl/manager.go
```diff
@@ -57,6 +60,23 @@ func (d *DefaultManager) ApplyFiltering(networkMap *mgmProto.NetworkMap, dnsRout
...
+	// every peer/route ACL and flushing the firewall on every such sync is the
+	// dominant client-side cost when nothing changed. Mirrors the same guard the
+	// DNS server already uses (previousConfigHash). Only the fields ApplyFiltering
+	// consumes participate in the hash, so an unrelated map change cannot mask a
+	// real ACL change.
+	hash, err := d.firewallConfigHash(networkMap, dnsRouteFeatureFlag)
+	if err != nil {
+		log.Errorf("unable to hash firewall configuration, applying unconditionally: %v", err)
+	} else if d.hasAppliedConfig && d.previousConfigHash == hash {
+		log.Debugf("not applying the firewall configuration update as there is nothing new")
```
lixmal: Let's add the hash here

client/internal/acl/manager.go
```diff
@@ -70,13 +90,47 @@ func (d *DefaultManager) ApplyFiltering(networkMap *mgmProto.NetworkMap, dnsRout
...
+func (d *DefaultManager) firewallConfigHash(networkMap *mgmProto.NetworkMap, dnsRouteFeatureFlag bool) (uint64, error) {
+	return hashstructure.Hash(struct {
+		PeerRules           []*mgmProto.FirewallRule
+		PeerRulesIsEmpty    bool
+		RouteRules          []*mgmProto.RouteFirewallRule
+		DNSRouteFeatureFlag bool
+	}{
+		PeerRules:           networkMap.GetFirewallRules(),
+		PeerRulesIsEmpty:    networkMap.GetFirewallRulesIsEmpty(),
+		RouteRules:          networkMap.GetRoutesFirewallRules(),
```
lixmal, line 127: This is missing GetRoutesFirewallRulesIsEmpty

lixmal (dismissed):

lixmal (approved):

PR #6512, opened 2026-06-22
[client] bound system info / posture-check gathering with a timeout to prevent sync-loop freeze

riccardomanfrin:
## Describe your changes

The engine's management sync loop (`handleSync`) and the signal handler both run under a single mutex (`syncMsgMux`). On every checks change (for posture checks), `handleSync` → `updateChecksIfNew` calls `system.GetInfoWithChecks`, which performs **uncancellable** system calls — process enumeration (`getRunningProcesses` → gopsutil `Pids()`/`Exe()` over every PID), `os.Stat` on each posture-check path, and `exec.Command` (e.g. `sw_vers`). None of these were bounded by a timeout or context. When one of them hangs (observed on macOS clients of an account that uses **process posture checks**), the whole engine freezes: it stops consuming management *and* signal updates (shared mutex) until the call eventually returns. In two debug bundles this showed up as a single `handleSync` holding the lock for **7m03s** and **2m35s**, ending only when the user terminated the process — the client could not even receive the management update that would have fixed it. The same unbounded `GetInfoWithChecks` call also sits on the management-stream setup path in `receiveManagementEvents`.

This PR bounds that gathering. Three commits:

1. **Wrap system-info / posture-check gathering in a timeout.** New `system.GetInfoWithChecksTimeout(ctx, timeout, checks)` in `client/system/info.go` runs `GetInfoWithChecks` in a goroutine and returns `(nil, false)` if it does not finish within the timeout, so the caller proceeds instead of blocking. A buffered channel lets the abandoned goroutine exit once its blocking call returns (no leak beyond that call). Both call sites use it: `updateChecksIfNew` (skips the meta sync this cycle on timeout) and `receiveManagementEvents` (falls back to base `GetInfo` so the stream still connects). The duplicated `GetInfoWithChecks`-with-fallback logic and the 16-arg `SetFlags` block are factored into the shared helper and a new `Engine.applyInfoFlags`. `systemInfoTimeout` is 15s.

2. **Propagate context into the OS calls that can honor it.** `getRunningProcesses`/`checkFileAndProcess` (`client/system/process.go`) now take a `ctx`: `process.PidsWithContext`/`ExeWithContext`, plus a `ctx.Err()` bail in the per-PID and per-path loops; `sw_vers` uses `exec.CommandContext` (`info_darwin.go`). This makes the abandoned goroutine terminate promptly on timeout instead of running to completion. The mobile/wasm stubs (`info_android.go`, `info_ios.go`, `info_js.go`) get the matching signature.

3. **Distinguish timeout from cancellation in the log.** `ctx.Done()` fires both on the 15s deadline and on parent-context cancellation (e.g. shutdown); the log now reports `DeadlineExceeded` as a timeout warning and other cancellation separately, so a normal shutdown no longer prints a misleading "timed out" message.

Functional posture-check behavior is unchanged: when gathering succeeds, the same `Info.Files` is produced and synced. The only new behavior is that, on a hang, the sync loop is released after 15s instead of freezing indefinitely.

Tested: `go test ./client/system/` (new tests below), plus `go build`/`go vet` and a `GOOS=linux/darwin/windows` build sweep of the changed packages.

No public GitHub issue — internal reliability task (tracked privately; Linear/Plain links omitted per repo convention). The unbounded call being fixed exists on `main`: [`client/system/process.go` `getRunningProcesses`](https://github.com/netbirdio/netbird/blob/main/client/system/process.go) (no ctx/timeout), reached from `client/internal/engine.go` `updateChecksIfNew` → `system.GetInfoWithChecks` while holding `syncMsgMux`.

### Checklist
- [x] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

Internal reliability fix with no user-facing surface change: the CLI, the management/API protocol, configuration options, and posture-check semantics are all unchanged. The change only adds an internal 15s timeout around system-info gathering so a stuck OS call can no longer freeze the client's sync loop; on the normal (fast) path behavior is identical.

N/A

## Summary by CodeRabbit

* **New Features**
  * Added a 15-second timeout for system info collection that includes checks to prevent sync from stalling.
* **Improvements**
  * If check-based collection times out, the app falls back to basic system info for that sync cycle.
  * Engine feature settings are now applied to gathered system info before syncing.
  * Improved responsiveness by honoring context cancellations/deadlines during process and file checking.
* **Tests**
  * Added unit tests for the new timeout and context-cancellation behavior.


client/internal/engine.go
```diff
@@ -1066,11 +1072,23 @@ func (e *Engine) updateChecksIfNew(checks []*mgmProto.Checks) error {
...
+	if !ok {
+		// Gathering timed out; skip the meta sync this cycle rather than blocking the
+		// sync loop (and syncMsgMux) on a stuck system call. A later sync will retry.
+		return nil
 	}
+	e.applyInfoFlags(info)
+
+	if err := e.mgmClient.SyncMeta(info); err != nil {
+		log.Errorf("could not sync meta: error %s", err)
+		return err
```
lixmal: Let's return an error with `fmt.Errorf()` instead of logging + returning

riccardomanfrin: No problem removing the log it should be logged here already at grpc.go:477

lixmal (dismissed):

lixmal (approved):

PR #6571, opened 2026-06-28
[management] Enable lazy connections by default on new accounts

mlsmaycon:
## Describe your changes
With improvements in userspace lazy connection handling, we should be able to enable it for new accounts with less impact on users. 

These connections are cheaper and only target traffic that should go through the tunnels, leaving all other tunnels in an idle state. 

> Linux clients might lose the first byte. We are working on a fix for it in the next few days

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **New Features**
  * New accounts now start with Lazy Connection enabled by default, improving connection behavior out of the box.

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -2045,6 +2045,7 @@ func newAccountWithId(ctx context.Context, accountID, userID, domain, email, nam
 			Extra: &types.ExtraSettings{
 				UserApprovalRequired: true,
 			},
+			LazyConnectionEnabled: true,
 		},
 		Onboarding: types.AccountOnboarding{
 			OnboardingFlowPending: true,


lixmal (approved):

Silex:
I just want to make sure you guys are aware that the local inactivity teardown only runs in userspace/bind mode, not kernel-WireGuard mode — which matters more now that lazy is becoming the default.

It's gated here:

```go
// client/internal/lazyconn/manager/manager.go
if wgIface.IsUserspaceBind() {
    m.inactivityManager = inactivity.NewManager(...)
} else {
    log.Warnf("inactivity manager not supported for kernel mode, ...")
}
```

(`KernelConfigurer.LastActivities()` returns `nil`.) On-demand activation works in both modes — it's the local idle teardown that's userspace-only, so a kernel peer depends on the remote's `GO_IDLE`. That means two kernel-mode peers (e.g. two Linux hosts) connect on demand and then stay up indefinitely, since neither side ever idles out.

The reason kernel mode can't reuse the userspace signal: ICEBind filters keepalives in the datapath (`isTransportPkg`), but in kernel mode userspace only sees aggregate `wgctrl` counters, which keep ticking from our 25s NAT keepalive — so nothing ever looks idle. I assume that's why the gate exists.

pappz:
@mlsmaycon The client-side env variable has been designt originally to be live until this change only. When we change this, we should remove that env variable and create another one for opting out, because right now we can't disable this feature on the client side with env vars.

mlsmaycon:
> I just want to make sure you guys are aware that the local inactivity teardown only runs in userspace/bind mode, not kernel-WireGuard mode — which matters more now that lazy is becoming the default.
> 
> It's gated here:
> 
> ```go
> // client/internal/lazyconn/manager/manager.go
> if wgIface.IsUserspaceBind() {
>     m.inactivityManager = inactivity.NewManager(...)
> } else {
>     log.Warnf("inactivity manager not supported for kernel mode, ...")
> }
> ```
> 
> (`KernelConfigurer.LastActivities()` returns `nil`.) On-demand activation works in both modes — it's the local idle teardown that's userspace-only, so a kernel peer depends on the remote's `GO_IDLE`. That means two kernel-mode peers (e.g. two Linux hosts) connect on demand and then stay up indefinitely, since neither side ever idles out.
> 
> The reason kernel mode can't reuse the userspace signal: ICEBind filters keepalives in the datapath (`isTransportPkg`), but in kernel mode userspace only sees aggregate `wgctrl` counters, which keep ticking from our 25s NAT keepalive — so nothing ever looks idle. I assume that's why the gate exists.

@Silex we are aware of that, this won't disable on kernel mode, ending up in the same case as we have without lazy connection enabled. 

We have a few options we will evaluate in the next days, but impact is small on this.

PR #6572, opened 2026-06-28
[client] Reinject captured first packet on lazy connection activation

lixmal:
## Describe your changes

When a lazy connection activates, the packet that triggered activation was dropped, so WireGuard had to wait for its handshake retransmit timer (~5s) before the connection became usable. This captures that first packet and replays it through the real transport as soon as ICE or relay comes up, removing the activation stall.

- Capture the first packet that triggers activation in both the bind (userspace) and UDP (kernel) activity listeners
- Carry the captured packet through the activity event to the peer connection and reinject it once the real transport is established, via the proxy or directly over the ICE connection
- Stop removing the peer on activation in the kernel-mode listener, matching the userspace path, so kernel WireGuard's staged queue and the triggering packet are not wiped
- Widen the UDP listener read buffer to the interface MTU so the triggering packet can be captured
- Add an `InjectPacket` method to the proxy implementations to replay a raw packet over the underlying transport

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal client behavior change with no user-facing surface.

## Summary by CodeRabbit

* **New Features**
  * Added capture and replay of the first WireGuard handshake-init packet triggered by peer activity during lazy activation.
  * Activity notifications now include both the peer connection ID and the triggering packet bytes.
  * Added packet-injection capability across proxy implementations to support handshake replay.
* **Bug Fixes**
  * Improved handshake reinjection by replaying the captured packet once the transport/proxy path is ready.
  * Updated UDP packet capture to use MTU-sized buffering to better handle larger packets.
* **Tests**
  * Refreshed activity and UDP capture tests to assert the new event payload (including packet contents).

### DIFF
--- a/client/iface/wgproxy/bind/proxy.go
+++ b/client/iface/wgproxy/bind/proxy.go
@@ -136,6 +136,17 @@ func (p *ProxyBind) CloseConn() error {
 	return p.close()
 }
 
+// InjectPacket writes b to the remote peer over the underlying transport.
+func (p *ProxyBind) InjectPacket(b []byte) error {
+	if p.remoteConn == nil {
+		return errors.New("proxy not started")
+	}
+	if _, err := p.remoteConn.Write(b); err != nil {
+		return err
+	}
+	return nil
+}
+
 func (p *ProxyBind) close() error {
 	if p.remoteConn == nil {
 		return nil
--- a/client/iface/wgproxy/bind/proxy.go
+++ b/client/iface/wgproxy/bind/proxy.go
@@ -136,14 +136,8 @@ func (p *ProxyBind) CloseConn() error {
 	return p.close()
 }
 
-// InjectPacket writes b to the remote peer over the underlying transport.
-func (p *ProxyBind) InjectPacket(b []byte) error {
-	if p.remoteConn == nil {
-		return errors.New("proxy not started")
-	}
-	if _, err := p.remoteConn.Write(b); err != nil {
-		return err
-	}
+// InjectPacket is a no-op for the userspace proxy: first-packet reinjection is kernel-only.
+func (p *ProxyBind) InjectPacket(_ []byte) error {
 	return nil
 }
 
--- a/client/internal/peer/conn.go
+++ b/client/internal/peer/conn.go
@@ -143,13 +143,6 @@ type Conn struct {
 	pendingFirstPacket []byte
 }
 
-// SetPendingFirstPacket stashes a packet to be replayed once the connection is established.
-func (conn *Conn) SetPendingFirstPacket(b []byte) {
-	conn.mu.Lock()
-	conn.pendingFirstPacket = slices.Clone(b)
-	conn.mu.Unlock()
-}
-
 // injectPendingFirstPacket replays the captured handshake through the proxy if present, else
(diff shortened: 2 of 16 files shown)


client/iface/wgproxy/bind/proxy.go
```diff
@@ -136,6 +136,17 @@ func (p *ProxyBind) CloseConn() error {
 	return p.close()
 }
 
+// InjectPacket writes b to the remote peer over the underlying transport.
+func (p *ProxyBind) InjectPacket(b []byte) error {
```
pappz: Is it used in this implementation?

client/internal/peer/conn.go
```diff
@@ -136,6 +137,46 @@ type Conn struct {
...
 	metricsRecorder MetricsRecorder
 	metricsStages   *MetricsStages
+
+	// pendingFirstPacket is the lazyconn-captured handshake init, replayed once the real
+	// transport is up.
+	pendingFirstPacket []byte
+}
+
+// SetPendingFirstPacket stashes a packet to be replayed once the connection is established.
+func (conn *Conn) SetPendingFirstPacket(b []byte) {
```
pappz: Instead of setter please use alternative Open function.

pappz (approved):

pappz (changes requested):
request changes

client/internal/peer/conn.go
lixmal: Done in e01c828f3. Removed the `SetPendingFirstPacket` setter and added `OpenWithFirstPacket(ctx, firstPacket)` instead. The packet is stashed inside the open path and retained only on a successful open, so a failed open no longer leaves a stale handshake for a later retry to replay.

lixmal: Fixed

client/internal/peerstore/store.go
pappz: Call  p.Open
```
	s.peerConnsMu.RLock()
	defer s.peerConnsMu.RUnlock()

	p, ok := s.peerConns[pubKey]
	if !ok {
		return
	}

	if err := p.Open(ctx); err != nil {
		p.Log.Errorf("failed to open peer connection: %v", err)
	}
```

client/internal/peerstore/store.go
```diff
@@ -94,14 +94,9 @@ func (s *Store) PeerConnOpenWithFirstPacket(ctx context.Context, pubKey string,
 	if !ok {
 		return
 	}
-	if len(firstPacket) > 0 {
-		p.SetPendingFirstPacket(firstPacket)
-	}
 	// this can be blocked because of the connect open limiter semaphore
```
pappz: this comment is not valid any more

pappz (changes requested):

client/internal/peerstore/store.go
lixmal: Done in 57170cdac. PeerConnOpen now does its own lookup and calls p.Open(ctx) directly.

client/internal/peerstore/store.go
lixmal: Removed in 57170cdac.

pappz (approved):

PR #6594, opened 2026-06-29
[signal] Refresh receive liveness when worker handoff drains

lixmal:
## Describe your changes

Follow-up to #6530. The signal receive watchdog could briefly misread a healthy stream as dead: after the receive loop finished a worker handoff that blocked past the inactivity threshold, the handoff flag cleared while the last-received timestamp still pointed before the block, so the window before the next `Recv` looked stale.

- Refresh the receive liveness timestamp when the worker handoff drains, before clearing the handoff flag, so the gap until the next read is not seen as a dead stream
- Add a regression test that drives the receive loop through a blocked handoff and checks liveness in the post-handoff window

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal signal client fix with no user-facing surface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved receive-stream liveness handling so brief handoff delays no longer make the connection appear inactive.
  * The stream now stays marked healthy during worker handoff and refreshes its activity status before resuming the next receive cycle.
* **Tests**
  * Added coverage for blocked handoff behavior to verify the stream remains alive until the handoff completes and liveness is refreshed afterward.

### DIFF
--- a/shared/signal/client/grpc.go
+++ b/shared/signal/client/grpc.go
@@ -541,6 +541,9 @@ func (c *GrpcClient) receive(stream proto.SignalExchange_ConnectStreamClient) er
 		if err := c.decryptionWorker.AddMsg(c.ctx, msg); err != nil {
 			log.Errorf("failed to add message to decryption worker: %v", err)
 		}
+		// Refresh liveness before clearing the flag so the window between here and
+		// the next Recv does not read a stale timestamp as a dead stream.
+		c.markReceived()
 		c.receiveHandoffBlocked.Store(false)
 	}
 }
--- a/shared/signal/client/watchdog_test.go
+++ b/shared/signal/client/watchdog_test.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"io"
 	"net"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


shared/signal/client/grpc.go
```diff
@@ -541,6 +541,9 @@ func (c *GrpcClient) receive(stream proto.SignalExchange_ConnectStreamClient) er
 		if err := c.decryptionWorker.AddMsg(c.ctx, msg); err != nil {
 			log.Errorf("failed to add message to decryption worker: %v", err)
 		}
+		// Refresh liveness before clearing the flag so the window between here and
+		// the next Recv does not read a stale timestamp as a dead stream.
+		c.markReceived()
```
dmitri-netbird, line 546: this works, I think a slightly clearer way would be to clear `receiveHandoffBlocked` in `markReceived` (and probably rename it?) so it's called once within the receive loop, and acts as a barrier...

dmitri-netbird: ah, nm, we don't always reach `markReceived`

dmitri-netbird (approved):

Issue #5990, opened 2026-04-25
Per-peer and per-group connection-mode and inactivity-threshold override on the management server


MichaelUray:
## Summary

cc @pappz — same context as the companion proposal #5989.

RFC-style proposal: extend the connection-mode and inactivity-threshold settings (proposed in companion issue #5989) from single account-wide values to a hierarchical per-peer / per-group / account resolution on the management server. Each client receives the resolved values via the existing `NetworkMap.PeerConfig` channel; client-side override (covered in the companion proposal) remains available for power users.

This addresses the same use case as [#4758](https://github.com/netbirdio/netbird/issues/4758) (per-peer "server-only mode" request) with a more general mechanism. If maintainers agree this is the right direction, the original author of #4758 may want to consider whether their issue is still independently needed or can be closed in favor of this work.

## Background

The management server already computes `PeerConfig.LazyConnectionEnabled` per-peer ([engine.go:1385](https://github.com/netbirdio/netbird/blob/main/client/internal/engine.go#L1385) consumes the per-peer field). Today the value sent is uniformly the account-wide setting `Settings.LazyConnectionEnabled` ([management/server/types/settings.go](https://github.com/netbirdio/netbird/blob/main/management/server/types/settings.go)) — there is no way for an admin to express "all my mobiles should be `p2p-dynamic-lazy`, all my servers `p2p`, my dev-laptop `p2p-lazy`, with custom inactivity thresholds for the lazy ones" without touching every peer individually via env/CLI.

NetBird already has a mature group system used for Access Policies, Routes, Setup Keys, and Posture Checks. Group-based connection-mode and threshold configuration is the natural extensibility point.

## Proposed solution

Server-side resolution per peer in NetworkMap, applied independently to each setting (the connection-mode itself and each of the inactivity thresholds):

```
1. Explicit per-peer setting   (admin set on a specific peer in the dashboard)
   ↓ falls back to
2. Per-group setting           (peer is member of N groups; conflict resolved per the rule below)
   ↓ falls back to
3. Account default              (as today)
```

The resolved values are sent via the existing `PeerConfig` channel — `connection_mode`, `ice_idle_threshold`, `relay_idle_threshold` (newly added in the companion proposal; today only `LazyConnectionEnabled` exists).

[...]

PR #6616, opened 2026-06-30
[client] Tear down idle lazy connections in kernel mode

Silex:
## Describe your changes

Lazy connections have two halves: **activation** (establish a peer on first traffic) and **inactivity teardown** (drop a peer once it goes idle). Activation already works in both userspace and kernel-WireGuard mode, but teardown was gated to userspace only:

- `lazyconn/manager.NewManager` created the inactivity monitor only `if wgIface.IsUserspaceBind()`, otherwise logging `"inactivity manager not supported for kernel mode, wait for remote peer to close the connection"`.
- `KernelConfigurer.LastActivities()` returned `nil`, so the monitor never saw a kernel-mode peer as idle.

As a result a kernel-mode peer relied entirely on the *remote* peer's `GO_IDLE`. Two kernel-mode peers (e.g. two Linux hosts) would connect on demand and then stay up indefinitely, since neither side ever idled out. This becomes more visible as lazy connections move to default-on (#6571).

**This PR** gives kernel mode a real per-peer activity signal and removes the gate:

- `KernelConfigurer.LastActivities()` now infers activity from the aggregate Tx/Rx byte counters already exposed via `wgctrl` (reusing `GetStats()`). A peer counts as active when its combined counter grows past a small threshold between polls. The baseline advances every poll so the 25s persistent keepalive can't accumulate into a false positive; a counter reset (peer re-added) counts as activity; newly seen peers are seeded active, mirroring the userspace `ActivityRecorder`.
- The inactivity manager is now created unconditionally, so kernel mode reaches parity with userspace.

**Why byte counters and not packet capture:** in the linked discussion I initially floated tapping packets via AF_PACKET (before deleting it). On closer look that's the wrong tool for an always-on signal — it runs a continuous userspace read loop copying every decrypted overlay packet (negating kernel mode's performance benefit), is Linux-only (kernel mode also runs on FreeBSD), and shares the single debug-capture slot. The `wgctrl` counter dump is a once-a-minute netlink read the inactivity monitor already polls at, works on Linux + FreeBSD, and needs no extra privileges.

The change is split into two bisect-friendly commits: (1) add the kernel activity signal (no behavior change on its own), (2) remove the gate (activates teardown).

### Threshold, validated on real hardware
Kernel mode only exposes aggregate counters, so the idle noise floor is handled with a fixed `activityByteThreshold`. I measured it on idle kernel-WireGuard peers (Teltonika TRB500, ARMv7, OpenWrt 21.02, kernel 4.14) by sampling per-peer `wg show transfer` every 60s: a keepalive-only poll grows the combined Tx+Rx counter by ~64 bytes, and a poll that contains a WireGuard rekey handshake (which recurs ~every 2 min, kept alive by the persistent keepalive itself) grows it by ~400 bytes. The rekey overhead is the non-obvious part — handshake messages land in the per-peer counters too, so the floor is higher than keepalives alone. The threshold is set to **1024 bytes/poll**, comfortably above the measured ~400-byte ceiling and far below any real traffic. The per-poll baseline advances every poll so the floor can't accumulate across intervals. Happy to make it configurable if maintainers prefer.

Discussed on the default-on PR: https://github.com/netbirdio/netbird/pull/6571#issuecomment-4841303542
Related connection-mode RFCs: #5989, #5990

## Stack

N/A

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor
- [x] This change does **not** modify the public API, gRPC protocols, functionality behavior, CLI / service flags, or introduce a new feature — **OR** I have discussed it with the NetBird team beforehand (link the issue / Slack thread in the description). — *This changes kernel-mode behavior (idle lazy connections now tear down); raised on #6571 (link above) before submitting.*

## Documentation
Select exactly one:

- [x] Documentation is **not needed** for this change (explain why) — *No code-repo docs assert "userspace only" for inactivity. If the netbirdio/docs site notes that lazy/idle teardown is userspace-only, that note should be dropped — flag if so and I'll open a docs PR.*

N/A

---

## Test plan (not part of template — for our review)

Verified in a `golang:1.25.5` container:
- `gofmt -l` clean on all three changed files
- `go vet ./client/iface/configurer/... ./client/internal/lazyconn/...` — OK
- `go build` of both package trees — OK
- `go test ./client/iface/configurer/... ./client/internal/lazyconn/...` — all pass, including new table-driven tests for the tracker (new peer seeded active, growth > threshold → active, sub-threshold growth across polls → stays idle, counter reset → active, removed-peer pruning)

Validated against live hardware (kernel WireGuard, lazy enabled):
- Sampled per-peer `wg show transfer` on idle TRB500 peers over 6 min → idle floor ~64 B/poll (keepalive) rising to ~400 B/poll on rekey polls; drove the `activityByteThreshold = 1024` choice (see above). Real traffic on the same interface was orders of magnitude higher, cleanly separable.

Not yet done (needs deploying a patched build to two peers):
- End-to-end: two kernel-WireGuard peers with lazy enabled + low `NB_LAZY_CONN_INACTIVITY_THRESHOLD`, confirming connect-on-traffic → teardown-on-idle → re-activate. (The current released agent in kernel mode keeps idle peers connected indefinitely — reproduced live: a P2P peer idle for 6+ min with only keepalive/rekey traffic stayed `Connected`.)

### DIFF
--- a/client/iface/configurer/kernel_activity_test.go
+++ b/client/iface/configurer/kernel_activity_test.go
@@ -1,0 +1,99 @@
+//go:build (linux && !android) || freebsd
+
+package configurer
+
+import (
+	"testing"
+
+	"github.com/netbirdio/netbird/monotime"
+)
+
+func TestUpdateActivity_NewPeerSeededActive(t *testing.T) {
+	c := NewKernelConfigurer("")
+
+	now := monotime.Time(1000)
+	got := c.updateActivity(map[string]WGStats{
+		"peerA": {TxBytes: 500, RxBytes: 500},
+	}, now)
+
+	if got["peerA"] != now {
+		t.Fatalf("new peer should be seeded active at now=%d, got %d", now, got["peerA"])
+	}
+}
+
+func TestUpdateActivity_GrowthPastThresholdIsActive(t *testing.T) {
+	c := NewKernelConfigurer("")
+
+	t0 := monotime.Time(1000)
+	c.updateActivity(map[string]WGStats{"peerA": {TxBytes: 1000}}, t0)
+
+	// Grow well past the threshold.
+	t1 := monotime.Time(2000)
+	got := c.updateActivity(map[string]WGStats{
+		"peerA": {TxBytes: 1000 + activityByteThreshold + 1},
+	}, t1)
+
+	if got["peerA"] != t1 {
+		t.Fatalf("peer with >threshold growth should be active at t1=%d, got %d", t1, got["peerA"])
+	}
+}
+
+func TestUpdateActivity_SubThresholdStaysIdleNoAccumulation(t *testing.T) {
+	c := NewKernelConfigurer("")
+
+	t0 := monotime.Time(1000)
+	c.updateActivity(map[string]WGStats{"peerA": {TxBytes: 0}}, t0)
+
+	// Several polls, each growing by less than the threshold (keepalive noise).
+	// Even though the cumulative growth far exceeds the threshold, the per-poll
+	// delta never does, so the peer must keep reporting its original activity.
+	bytes := int64(0)
+	step := int64(activityByteThreshold / 2)
+	for i, ts := range []monotime.Time{2000, 3000, 4000, 5000, 6000} {
+		bytes += step
+		got := c.updateActivity(map[string]WGStats{"peerA": {TxBytes: bytes}}, ts)
+		if got["peerA"] != t0 {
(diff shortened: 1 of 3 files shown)


pappz:
@Silex  Thanks for your effort in this topic! We already tried this approach. https://github.com/netbirdio/netbird/commit/9307e7e0e. It contains unit tests that prove there are edge cases causing issues in environments where flaky connections cause more problems than they solve. We dropped it because we'd like something bulletproof. We could track the WireGuard-type messages in eBPF. Next time, please open a discussion before jumping into a PR.

Silex:
@pappz Alright. Closing this.

PR #6617, opened 2026-06-30
[client] Make lazy connections opt-out via NB_LAZY_CONN

lixmal:
## Describe your changes
Lazy connections are becoming the default (enabled by management for new accounts in #6571). This reworks the client-side control so lazy is opt-out, driven by management with explicit local overrides, and removes the old experimental opt-in.

Resolution order: `NB_LAZY_CONN` env var (if set) > MDM policy override > management feature flag > off.

- Add the tri-state `NB_LAZY_CONN` env var (`on`/`off`/unset); set, it overrides the management feature flag in both directions; unset defers to management
- Add an MDM policy override (`lazyConnection` key, `on`/`off`) with the same semantics, below the env var
- Remove the experimental `NB_ENABLE_EXPERIMENTAL_LAZY_CONN` variable
- Retire the `--enable-lazy-connection` CLI flag: inert and hidden (kept so existing invocations don't error)
- Remove the local `LazyConnectionEnabled` setting end to end (config, daemon apply, desktop UI checkbox, MDM bool, and login telemetry); lazy is no longer a persisted local toggle
- Point the Android/iOS env exports at the new variable name

The per-peer lazy flag previously reported to management via login metadata is no longer sent (it reflected the removed local setting). Runtime lazy status in `netbird status` is unaffected.

Complements #6571.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

https://github.com/netbirdio/docs/pull/821

## Summary by CodeRabbit

* **New Features**
  * Lazy connections are now controlled via `NB_LAZY_CONN` (`on`/`off`/unset), with unset deferring to management using consistent precedence.
  * Added MDM policy support for `lazyConnection`, including normalization of `on/off/yes/no` values.
* **Bug Fixes**
  * Local overrides reliably take priority; remote changes and related status updates are ignored when overridden.
* **UI / CLI**
  * Removed the lazy-connections toggle from the tray UI and settings sync.
  * Deprecated `--enable-lazy-connection`; it’s no longer used by `up`.
* **Tests / Other**
  * Expanded unit test coverage for env parsing, override resolution, and MDM mapping; updated debug output.

### DIFF
--- a/client/internal/lazyconn/env.go
+++ b/client/internal/lazyconn/env.go
@@ -49,7 +49,7 @@ func ParseState(raw string) State {
 
 	enabled, err := strconv.ParseBool(normalized)
 	if err != nil {
-		log.Warnf("failed to parse %s value %q: %v", EnvLazyConn, raw, err)
+		log.Warnf("failed to parse lazy connection value %q (from %s env or MDM policy): %v", raw, EnvLazyConn, err)
 		return StateUnset
 	}
 	if enabled {
--- a/client/internal/lazyconn/env.go
+++ b/client/internal/lazyconn/env.go
@@ -3,24 +3,57 @@ package lazyconn
 import (
 	"os"
 	"strconv"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 )
 
 const (
-	EnvEnableLazyConn      = "NB_ENABLE_EXPERIMENTAL_LAZY_CONN"
+	EnvLazyConn            = "NB_LAZY_CONN"
 	EnvInactivityThreshold = "NB_LAZY_CONN_INACTIVITY_THRESHOLD"
 )
 
-func IsLazyConnEnabledByEnv() bool {
-	val := os.Getenv(EnvEnableLazyConn)
-	if val == "" {
-		return false
+// State is the tri-state local override for lazy connections read from the environment.
+type State int
+
+const (
+	// StateUnset means no local override; defer to the management feature flag.
+	StateUnset State = iota
+	// StateOn forces lazy connections on, overriding management.
+	StateOn
+	// StateOff forces lazy connections off, overriding management.
+	StateOff
+)
+
+// EnvState reads NB_LAZY_CONN and returns the local override state.
+func EnvState() State {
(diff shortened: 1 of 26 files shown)


riccardomanfrin (commented):
I think you need to add the new key in `canonical_loaders.go:13`

Otherwise it will be discarded as unknown for darwin and windows

client/internal/profilemanager/config.go
```diff
@@ -728,6 +722,11 @@ func (config *Config) applyMDMPolicy(policy *mdm.Policy) {
 			log.Warnf("MDM wireguard port %d out of range [1,65535]; keeping previous value", v)
 		}
 	}
+
+	if v, ok := policy.GetString(mdm.KeyLazyConnection); ok {
```
riccardomanfrin: if we use `GetBool` we behave like the other bools (e.g. we allow also for things like "yes"/"no")

client/internal/lazyconn/env.go
```diff
@@ -3,24 +3,57 @@ package lazyconn
...
+	case "off":
+		return StateOff
 	}
-	enabled, err := strconv.ParseBool(val)
+
+	enabled, err := strconv.ParseBool(normalized)
 	if err != nil {
-		log.Warnf("failed to parse %s: %v", EnvEnableLazyConn, err)
-		return false
+		log.Warnf("failed to parse %s value %q: %v", EnvLazyConn, raw, err)
```
riccardomanfrin: this was a previous mistake but here we might log that the value might come either from NB_ env OR from MDM... might allow faster recon of why a flag is there

client/internal/profilemanager/config.go
lixmal: Fixed in a007c4588

client/internal/lazyconn/env.go
lixmal: Fixed in a007c4588

riccardomanfrin (dismissed):

riccardomanfrin (approved):

PR #6621, opened 2026-06-30
[management] detach JWT group sync write from request cancellation

### LINKED ISSUE #6620: JWT group sync silently never persists when the dashboard cancels the request
### Describe the problem

With JWT group sync enabled, a user's JWT-derived groups (`users.auto_groups`) are never persisted. The dashboard shows no groups for the user, and the management log repeats, on essentially every dashboard request:

```
management/server/http/middleware/auth_middleware.go:157: HTTP server failed to sync user JWT groups: context canceled
```

while `users.auto_groups` stays `[]`.

### Root cause

`checkJWTFromRequest` (`management/server/http/middleware/auth_middleware.go`) calls `syncUserJWTGroups(ctx, userAuth)` with `ctx = r.Context()` — the HTTP **request** context. `SyncUserJWTGroups` persists the groups in a DB transaction bound to that context (`ExecuteInTransaction`). The dashboard SPA routinely aborts in-flight `fetch()` calls on re-render/navigation, which cancels the request context mid-write; the transaction is rolled back and the groups never persist. The error is logged but swallowed (no `return`), so it repeats silently.

This is distinct from #2696 (the older lock/deadlock that hung the dashboard, fixed in v0.31.0) — here the dashboard is responsive, the write is simply rolled back.

### Steps to reproduce

1. Self-host `management` with an external IdP that emits a flat `groups` claim (e.g. Keycloak).
2. Enable JWT group sync (Settings → Groups), claim name `groups`.
3. Log into the dashboard as a user whose token carries `"groups": ["..."]`.
4. Observe `users.auto_groups` stays `[]` and the log repeats `failed to sync user JWT groups: context canceled`.

### Expected behaviour

The user's JWT groups are persisted to `auto_groups` regardless of whether the originating dashboard request is cancelled.

### Are you using NetBird Cloud?

No — self-hosted.

### NetBird version

management v0.73.2 (also reproduces on current `main`), PostgreSQL store.

### Notes

Claim extraction itself works (`ToGroups` returns the groups); the failure is purely the cancelled write. A fix (detaching the sync write from request cancellation) is in the PR linked below.

den-dw:
## Describe your changes

The HTTP auth middleware persists a user's JWT-derived groups (`users.auto_groups`) by calling `syncUserJWTGroups` inside the **request context**. The dashboard SPA routinely aborts in-flight `fetch()` calls on re-render/navigation, which cancels the request context mid-write. The group-sync write runs in a DB transaction bound to that context, so it is rolled back and the synced groups never persist. The error is logged but swallowed (no `return`), so it repeats silently on every request and `auto_groups` stays empty.

**Root cause:** a server-side persistence side effect was tied to the client request lifecycle.

**Fix:** detach the group-sync write with `context.WithoutCancel(ctx)` so it commits regardless of the client connection, while preserving context values (request id, logger). This matches the existing `context.WithoutCancel` usage in the management server (`peer.go`, the gRPC server, the network-map controller). The transaction stays bounded by the store's own timeout (`ExecuteInTransaction`, default 5 min).

Scoped to the HTTP middleware call-site where the SPA-abort symptom reproduces. The gRPC login path (`internals/shared/grpc/server.go`) calls the same method but isn't driven by an aborting SPA and is intentionally left out of this change.

A regression test (`TestAuthMiddleware_SyncUserJWTGroupsDetachedFromRequestCancellation`) dispatches an already-cancelled request and asserts the injected `syncUserJWTGroups` receives a non-cancelled context — it fails without the change (`context canceled`) and passes with it.

Fixes #6620

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor
- [x] This change does **not** modify the public API, gRPC protocols, functionality behavior, CLI / service flags, or introduce a new feature

## Documentation
Select exactly one:

- [x] Documentation is **not needed** for this change (explain why): internal server bug fix in the auth middleware, no user-facing surface.

### DIFF
--- a/management/server/http/middleware/auth_middleware.go
+++ b/management/server/http/middleware/auth_middleware.go
@@ -152,7 +152,11 @@ func (m *AuthMiddleware) checkJWTFromRequest(r *http.Request, authHeaderParts []
 		return err
 	}
 
-	err = m.syncUserJWTGroups(ctx, userAuth)
+	// Detach the group-sync write from the request's cancellation: the dashboard
+	// SPA aborts in-flight requests on re-render, which would otherwise cancel the
+	// DB transaction mid-write and silently drop the synced groups. Context values
+	// (request id, logger) are preserved; the store bounds the tx with its own timeout.
+	err = m.syncUserJWTGroups(context.WithoutCancel(ctx), userAuth)
 	if err != nil {
 		log.WithContext(ctx).Errorf("HTTP server failed to sync user JWT groups: %s", err)
 	}
--- a/management/server/http/middleware/auth_middleware_test.go
+++ b/management/server/http/middleware/auth_middleware_test.go
@@ -241,6 +241,66 @@ func TestAuthMiddleware_Handler(t *testing.T) {
 	}
 }
 
+// TestAuthMiddleware_SyncUserJWTGroupsDetachedFromRequestCancellation ensures the
+// JWT group sync write is not bound to the request context. The dashboard SPA
+// routinely aborts in-flight requests on re-render/navigation; if the sync ran in
+// the request context, the cancellation would roll back the DB transaction and the
+// synced groups would silently never persist. The sync must receive a context that
+// is not cancelled even when the originating request is.
+func TestAuthMiddleware_SyncUserJWTGroupsDetachedFromRequestCancellation(t *testing.T) {
+	var (
+		syncCalled bool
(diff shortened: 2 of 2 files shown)


pascal-fischer:
Hi @den-dw, Thanks for contributing. This looks correct, I will merge it

pascal-fischer (approved):

PR #6630, opened 2026-07-01
[management,proxy] Add per-provider skip_tls_verification for agent-network

mlsmaycon:
## Describe your changes

Adds per-provider **`skip_tls_verification`** to agent-network providers, so an operator can front a self-hosted / internal AI gateway that presents a private or self-signed certificate.

The proxy already supports skipping upstream TLS verification per target (`roundtrip.WithSkipTLSVerify`), but agent-network dials **per provider inside the `llm_router` middleware** (the reverse-proxy target is a single `noop` `DirectUpstream`), so a per-target flag alone wouldn't cover it. This wires the flag end to end:

> **Why not just set the existing target-level `skip_tls_verify`?** For a normal reverse-proxy service that would be enough — each target is a real upstream. But the agent-network service has a *single* target that all of an account's providers route through, with the real per-provider upstreams living in the `llm_router` config. Setting skip-tls on that one target is all-or-nothing for the account: enabling it for an internal self-signed gateway would also stop verifying a public provider like `api.openai.com` in the same account. Since the field is per-provider, the flag rides the router's per-route `UpstreamRewrite` so it applies only to the selected provider's upstream.

- **Provider** (`types/provider.go`): new `SkipTLSVerification` field, persisted via GORM `AutoMigrate` (new column, no manual migration). Request→provider→response mapping mirrors `enabled`: on update a `nil` value preserves the stored setting, an explicit `false` clears it. Never surfaced as a secret.
- **API** (`openapi.yml`): `skip_tls_verification` on both `AgentNetworkProviderRequest` (optional) and `AgentNetworkProvider` (required in the response); `types.gen.go` regenerated via `generate.sh`.
- **Synthesizer**: carries the flag into the `llm_router` route config (`skip_tls_verify`) sent to proxies.
- **Proxy**: `llm_router` sets `SkipTLSVerify` on the `UpstreamRewrite` mutation for the selected route, and `ReverseProxy.forwardUpstream` applies `roundtrip.WithSkipTLSVerify` when forwarding — so only that provider's upstream is dialed insecurely, not the whole service.

**E2E validation:** `TestProviderSkipTLSVerification` (container harness, build tag `e2e`) stands up a self-signed HTTPS upstream and provisions two providers pointing at it — `skip_tls_verification=true` and `false` — behind one proxy + client. The skip=true provider's chat returns 200 (reaches the self-signed upstream); the skip=false provider's returns 5xx (TLS handshake rejected). Same upstream, opposite outcome — which also demonstrates why the flag must be per-provider rather than on the shared synthesised target. Passing locally (images built from source):

```
--- PASS: TestProviderSkipTLSVerification (58.30s)
```

N/A — agent-network provider configuration enhancement.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Adds an optional field to an existing agent-network API surface, described inline in the OpenAPI spec.

## Summary by CodeRabbit

* **New Features**
  * Added a provider option to skip upstream TLS certificate verification.
  * Exposed the setting in the API (including update/omit behavior) and applied it in routing/proxy upstream dials.

* **Bug Fixes**
  * Requests to self-signed/private-certificate upstreams can now succeed when TLS verification is disabled.
  * Provider updates now correctly preserve the existing value when the setting is omitted.

* **Tests**
  * Added unit, middleware, and end-to-end coverage to verify TLS-skip propagation and real request behavior.

### DIFF
--- a/e2e/agentnetwork/skiptls_test.go
+++ b/e2e/agentnetwork/skiptls_test.go
@@ -1,0 +1,140 @@
+//go:build e2e
+
+package agentnetwork
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/netbirdio/netbird/e2e/harness"
+	"github.com/netbirdio/netbird/shared/management/http/api"
+)
+
+// TestProviderSkipTLSVerification proves skip_tls_verification is per-provider:
+// two providers share one self-signed upstream, one skipping TLS verification
+// and one not. The skip=true provider's chat reaches the upstream and returns
+// 200; the skip=false provider's chat fails at the TLS handshake — same
+// upstream, opposite outcome. This is the behaviour a target-level flag could
+// not give, since all of an account's providers share one synthesised target.
+func TestProviderSkipTLSVerification(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Minute)
+	defer cancel()
+
+	up, err := harness.StartFakeUpstream(ctx, srv)
+	require.NoError(t, err, "start self-signed upstream")
+	t.Cleanup(func() { _ = up.Terminate(context.Background()) })
+
+	grp, err := srv.API().Groups.Create(ctx, api.PostApiGroupsJSONRequestBody{Name: "e2e-skiptls"})
+	require.NoError(t, err, "create group")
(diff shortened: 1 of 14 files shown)


mlsmaycon:
https://github.com/netbirdio/netbird/actions/runs/28532546229

braginini (approved):

PR #6659, opened 2026-07-03
[client] Set routeselector atomic exit node

pappz:
## Describe your changes

Make exit-node reconciliation atomic by moving the full deselect/reselect logic into a single locked RouteSelector.SetExclusiveExitNode method, preventing a concurrent DeselectAllRoutes from being silently undone by an interleaved re-selection of the preferred exit node.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved exit-node selection so only the preferred exit node remains active when reconciling routes.
  * Kept a global “deselect all” state intact, avoiding unintended re-selection.
  * Made route updates more reliable by applying selection changes atomically.


client/internal/routeselector/routeselector.go
```diff
@@ -115,6 +115,35 @@ func (rs *RouteSelector) DeselectAllRoutes() {
...
+		if id == preferred {
+			continue
+		}
+		rs.deselectedRoutes[id] = struct{}{}
+		delete(rs.selectedRoutes, id)
+	}
+
+	if preferred != "" {
+		delete(rs.deselectedRoutes, preferred)
+		rs.selectedRoutes[preferred] = struct{}{}
```
theodorsm, line 143: Same comment as line 137

client/internal/routeselector/routeselector.go
```diff
@@ -115,6 +115,35 @@ func (rs *RouteSelector) DeselectAllRoutes() {
...
+
+	if rs.deselectAll {
+		return
+	}
+
+	for _, id := range exitIDs {
+		if id == preferred {
+			continue
+		}
+		rs.deselectedRoutes[id] = struct{}{}
```
theodorsm, line 137: We could check if this map is not nil to avoid potential panic. Seems to me that mostly NewRouteSelector is used, which is initializing the map, so this might not be a problem + it's an internal package.

theodorsm (dismissed):

client/internal/routeselector/routeselector.go
pappz: I think we instantiate this resource only through the constructor. It is safe.

client/internal/routeselector/routeselector.go
pappz: I think we instantiate this resource only through the constructor. It is safe.

Issue #2073, opened 2024-05-29
OIDC compatible mode for IdP data extraction and synchronization


oddlama:
**Is your feature request related to a problem? Please describe.**

When using self-hosted netbird there are a lot of choices for which IdP to use. Only a select few of them can currently be used with netbird's user synchronization, because it requires talking to the IdP through proprietary REST APIs. Not all IdP provide such an API and sometimes users might not want to synchronize all users from the IdP into netbird. See #2057 for the issue where this idea came up.

Currently this means that account data such as email or username is never synchronized and it is impossible to identify users:

![image](https://github.com/netbirdio/netbird/assets/31919558/cb2cc4d7-b421-4078-b304-985da3d24d1d)

**Describe the solution you'd like**

I would like to propose adding a OIDC-compatible mode, which can work with any IdP that supports OIDC and without requiring it to support any additional API and by extension no explicit support from netbird. By default the user list just stays unpopulated and whenever a user successfully authenticates, their information (mostly username & email) is automatically synchronized via the IdP by just updating it to whatever the JWT authentication token contains.

This is also how most other OIDC applications (grafana, forgejo/gitea, ...) work. The downside is of course that users need to login at least once to show up in netbird, but the upsides are that it will work with any IdP and only users that are actually allowed to use netbird will be synchronized.

In OIDC compatible mode, the user account should always be identified by the `sub` claim, and their username and email would automatically reflect the `preferred_username` and `email` claims (possibly configurable). When these values get updated in the IdP, netbird can just automatically update them each time the user authenticates.

By using the JWT group sync feature, one can also still pre-configure stuff for users that have never authenticated, so the downside is really just that users are not immediately displayed which is probably the expectation by self-hosters anyway.

**Describe alternatives you've considered**

The alternative is to proxy a supported IdP between the actual IdP I'd like to use and netbird, but that causes a lot of friction and introduces numerous points of failure. Without being familiar with netbird's internals I would still assume the feature to be quite simple to implement since the required data is already available in the JWT token netbird receives and processes.

ds-sebastian:
@oddlama  I also was trying to migrate to KanIDM, were you able to find a workaround?

oddlama:
No unfortunately i didn't find a workaround. I just live with the UUID names for the time being and will wait for this feature before onboarding more users.

ds-sebastian:
[...]

PR #6671, opened 2026-07-05
fix(management): refresh user name/email from ID-token claims on login

### LINKED ISSUE #6652: Refresh user name/email from the ID-token claims on login (parity with JWT group sync)
> ⚠️ This is **not** a directory-sync or SCIM request.
> It does **not** require enabling the server-side IdP directory manager (`IdpManagerConfig`), and it does **not** need the paid Cloud/SCIM sync. It only re-reads the `name` and `email` claims that are **already present in the login ID-token** — exactly like the existing JWT **group** sync (`JWTGroupsClaimName`) already does for groups. No extra IdP API call, no background job, no license.

### Observation
NetBird already persists `name` / `email` into the local user record **at JIT creation** (`addNewUserToDomainAccount` → `types.NewRegularUser(userId, email, name)` → `SaveUser`), sourced from the token claims. But on **subsequent logins**, `GetUserFromUserAuth` only updates `last_login` — it never refreshes `name` / `email` from the ID-token claims (confirmed in `v0.73.2`).

### Consequences
- Users created by an older version (before name/email were persisted at JIT) keep **empty** `name` / `email` forever: there is no login-time backfill, and the API does not allow editing those fields for IdP-managed users.
- Anyone running `IdpManagerConfig.ManagerType: "none"` (a common workaround to stop the directory refresh from resetting `is_blocked` / renaming users to their IdP ID) then sees blank name/email, since display falls back to the stored (empty) fields.

### Proposal
On login, refresh the stored `name` / `email` from the ID-token claims — mirroring the existing JWT **group** sync (`JWTGroupsClaimName`). The data is already in the token and already persisted at creation; this just keeps it consistent over time.

### Related
- #2073 (broader "OIDC-compatible sync"; this is a narrow, JWT/login-only subset of it).

Amilliox:
### Problem
On subsequent logins, `GetUserFromUserAuth` only updates `last_login` but never refreshes `name`/`email` from the ID-token claims. Users created by older versions keep empty name/email forever, and those running with `IdpManagerConfig.ManagerType: "none"` also see blank fields.

### Fix
After retrieving the user, check if `userAuth.Name` or `userAuth.Email` differ from stored values (and are non-empty). If so, update and save — mirroring the existing JWT group sync (`JWTGroupsClaimName`) pattern.

### Scope
- 1 file: `management/server/user.go`
- +13 lines
- No new dependencies
- No config changes
- No IdP API calls
- No SCIM/license needed

Fixes #6652
Related #2073

## Summary by CodeRabbit

* **Bug Fixes**
  * During login, the app now refreshes the stored user profile details (name and/or email) from the latest authentication token claims when available.
  * If updating those saved details fails, login still succeeds and the sign-in flow remains reliable.

### DIFF
--- a/management/server/user.go
+++ b/management/server/user.go
@@ -234,6 +234,19 @@ func (am *DefaultAccountManager) GetUserFromUserAuth(ctx context.Context, userAu
 		return nil, err
 	}
 
+	// Refresh name/email from ID-token claims on each login (parity with JWT group sync)
+	if (userAuth.Name != "" && user.Name != userAuth.Name) || (userAuth.Email != "" && user.Email != userAuth.Email) {
+		if userAuth.Name != "" {
+			user.Name = userAuth.Name
+		}
+		if userAuth.Email != "" {
+			user.Email = userAuth.Email
+		}
+		if err := am.Store.SaveUser(ctx, user); err != nil {
+			log.WithContext(ctx).Debugf("failed to update user name/email: %v", err)
+		}
+	}
+
 	// this code should be outside of the am.GetAccountIDFromToken(claims) because this method is called also by the gRPC
 	// server when user authenticates a device. And we need to separate the Dashboard login event from the Device login event.
 	newLogin := user.LastDashboardLoginChanged(userAuth.LastLogin)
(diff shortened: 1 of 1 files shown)


raku-cat:
Here to report a success with this patch, had a user registered via authelia OIDC, after an LDAP email update the change didn't make it all the way over. After applying this PR after the user logged in the email was instantly updated in all parts of the app. Thumbs up from me :+1:

PR #6674, opened 2026-07-06
[client] Fix forwarder peers never excluded from lazy connections

riccardomanfrin:
## Describe your changes

`toExcludedLazyPeers` compared each peer's `AllowedIPs` (CIDR, e.g. an overlay IP as `/32`) against `ForwardRule.TranslatedAddress.String()` (unmasked), so the match never fired and forward-target peers were **never** excluded from lazy connections — inbound forwarded traffic can't wake a lazy peer, so those peers must stay permanent. Extracted a `peerRoutesAddr` helper and switched to prefix containment.

Internal bug found while investigating lazy-connection DNS/forwarding breakage. Broken line on main: https://github.com/netbirdio/netbird/blob/main/client/internal/engine.go#L2568

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal lazy-connection exclusion fix. No public API, config, or CLI change.

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inbound forwarding peer exclusion by matching forwarded targets against peers’ routed address ranges (prefix containment) instead of relying on exact string matches, reducing missed exclusions for IPv4 and IPv6.
* **Tests**
  * Added unit tests covering address/prefix membership behavior (IPv4/IPv6, matching and non-matching cases) and validating forward-target exclusion.
  * Added coverage for handling when no forward rules are provided.

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -2624,7 +2624,11 @@ func (e *Engine) toExcludedLazyPeers(rules []firewallManager.ForwardRule, peers 
 // peerRoutesAddr verifies if the peer is a router for a given address.
 func peerRoutesAddr(p *mgmProto.RemotePeerConfig, addr netip.Addr) bool {
 	for _, allowedIP := range p.GetAllowedIps() {
-		if allowedIP == addr.String() {
+		prefix, err := netip.ParsePrefix(allowedIP)
+		if err != nil {
+			continue
+		}
+		if prefix.Contains(addr) {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


CoderSufiyan (approved):
Good catch. The string comparison between `allowedIP` (CIDR, e.g. `100.110.8.145/32`) and `ip.String()` (bare address, `100.110.8.145`) was the root cause - they can never match. Switching to `netip.Prefix.Contains` is the right fix.

Also a smart design choice to source AllowedIPs from the peer store rather than re-parsing the network map strings, since the lazy manager already consumes those same typed prefixes.

Tests look solid. Covered IPv4, IPv6, no-rules, and non-matching cases.

LGTM.

lixmal (approved):

PR #6689, opened 2026-07-07
[client] Add local Prometheus metrics endpoint

lixmal:
Adds an opt-in Prometheus `/metrics` endpoint to the client daemon so deployments can scrape connection health locally (requested for a RedHat evaluation: latency, connected peers, P2P vs relay).

- Add `netbird up --enable-local-metrics` and `--local-metrics-address` (default `127.0.0.1:9191`); the setting persists in the profile config and travels through SetConfig/Login like other config flags
- Expose the existing client metrics (peer connection stage durations, sync and login durations) by mirroring them into a Prometheus registry behind the same recording interface
- Add scrape-time gauges for current state: management/signal connectivity, known peers, connected peers by connection type, and per-peer latency for direct connections
- Add a Grafana dashboard for the client next to the existing management/signal/relay dashboards
- Gate the new settings behind `--disable-update-settings` and MDM policy (`enableLocalMetrics`, `localMetricsAddress` keys)
- Regenerate the daemon JSON gateway, which was missing routes for recently added RPCs

All labels are bounded enums except the per-peer latency series, which carries the peer FQDN only.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

- [ ] Documentation is **not needed** for this change

https://github.com/netbirdio/docs/pull/952

## Summary by CodeRabbit

* **New Features**
  * Added an optional local Prometheus `/metrics` endpoint for client connectivity, peer states, latency, and connection performance.
  * Added CLI and configuration options to enable local metrics and set its listening address.
  * Added MDM policy support for managing local metrics settings.
  * Added a Grafana dashboard for visualizing client metrics.
* **Documentation**
  * Debug bundles now include local metrics configuration details.
* **Bug Fixes**
  * Improved connection-duration metric calculations and reporting consistency.
  * Added safeguards and privilege checks when exposing metrics beyond the local device.


client/internal/localmetrics/localmetrics.go
```diff
@@ -0,0 +1,243 @@
...
+	ch <- c.peerLatency
+}
+
+// Collect implements prometheus.Collector.
+func (c *collector) Collect(ch chan<- prometheus.Metric) {
+	ch <- prometheus.MustNewConstMetric(c.managementConnected, prometheus.GaugeValue, boolToFloat(c.status.GetManagementState().Connected))
+	ch <- prometheus.MustNewConstMetric(c.signalConnected, prometheus.GaugeValue, boolToFloat(c.status.GetSignalState().Connected))
+
+	peers := c.status.GetPeerStates()
+	ch <- prometheus.MustNewConstMetric(c.peersTotal, prometheus.GaugeValue, float64(len(peers)))
```
theodorsm, line 215: This does not account for offline peers, which is also known peers. We should export `Status.numOfPeers` and use it here (or add offline peers to GetPeerStatus) if the intention of this metric is to mirror status "peers count: x/**y** connected".

Currently this is what status would show: 

Image

And this is what grafana would show: 
Image

client/internal/peer/status.go
```diff
@@ -1167,6 +1167,18 @@ func (d *Status) GetResolvedDomainsStates() map[domain.Domain]ResolvedDomainInfo
 	return maps.Clone(d.resolvedDomainsStates)
 }
 
+// GetPeerStates returns a snapshot of all known peer states.
+func (d *Status) GetPeerStates() []State {
+	d.mux.RLock()
+	defer d.mux.RUnlock()
+
+	states := make([]State, 0, len(d.peers))
```
theodorsm: Should we include offline peers too, like `GetFullStatus` ? Ref: https://github.com/netbirdio/netbird/pull/6689/changes#r3863137373

client/mdm/policy.go
```diff
@@ -47,6 +47,8 @@ const (
 	KeyRosenpassEnabled    = "rosenpassEnabled"
 	KeyRosenpassPermissive = "rosenpassPermissive"
 	KeyWireguardPort       = "wireguardPort"
+	KeyEnableLocalMetrics  = "enableLocalMetrics"
+	KeyLocalMetricsAddress = "localMetricsAddress"
```
theodorsm, line 51: KeyEnableLocalMetrics and KeyLocalMetricsAddress are never added to allKeys (`client/mdm/canonical_loaders.go`), so the Windows registry and macOS plist loaders discard them as unknown.

client/server/server.go
theodorsm, line 1222: Missing localMetrics.Reconcile. The profile we switch to might have different metrics settings.

theodorsm, line 690: Missing localMetrics.Reconcile. The profile might have changed.

client/internal/localmetrics/localmetrics.go
```diff
@@ -0,0 +1,243 @@
...
+		WriteTimeout:      writeTimeout,
+		IdleTimeout:       idleTimeout,
+	}
+	m.srv = srv
+	m.addr = addr
+
+	log.Infof("serving local metrics on http://%s/metrics", addr)
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			log.Errorf("failed to serve local metrics on %s: %v", addr, err)
```
theodorsm, line 117: We should reset the `m.srv` and `m.addr` fields like we do in `stop()`.

client/internal/localmetrics/localmetrics.go
```diff
@@ -0,0 +1,243 @@
...
+	return m
+}
+
+// Reconcile starts, stops, or restarts the metrics endpoint to match the
+// desired state. An empty addr falls back to DefaultListenAddress.
+func (m *Manager) Reconcile(enabled bool, addr string) {
+	if addr == "" {
+		addr = DefaultListenAddress
+	}
+	warnIfNotLoopback(addr)
```
theodorsm, line 72: We should gate it behind a privilege check like we do for ssh rather than just logging.

theodorsm (changes requested):
Foreground mode accepts the local metrics flags, but does not serve the endpoint. Do we want to support this feature for foreground mode?

client/internal/localmetrics/localmetrics.go
lixmal: Fixed

client/internal/peer/status.go
lixmal: Fixed

client/mdm/policy.go
lixmal: Fixed

client/server/server.go
lixmal: Fixed

lixmal: Fixed

client/internal/localmetrics/localmetrics.go
lixmal: Fixed

client/internal/localmetrics/localmetrics.go
lixmal: Fixed

theodorsm (approved):

PR #6707, opened 2026-07-09
[client] Restore residual state in foreground mode before login

dfry:
## Describe your changes

`netbird up --foreground-mode` goes straight to `foregroundLogin` without the residual-state recovery the daemon performs — `restoreResidualState` is called at three points in `client/server` before touching management, with a comment describing exactly why: *"Otherwise, we might not be able to connect to the management server to retrieve new config."* The foreground path never calls it.

If a previous run did not shut down cleanly (crash/kill — deferred cleanup never runs), leftover routing policy from `SetupRouting` (e.g. the `not fwmark → NetbirdVPNTableID` rule) can shunt the management connection into a stale routing table whose default route points at a dead `wt0`, and every subsequent login times out.

**Foreground mode is particularly exposed in containers** (its primary use): a crashed container restarts inside the same pod network namespace, so the stale rules survive while the process state does not. Observed in the field: a netbird router container crash-looped ~99 times over 8 hours on

```
Error: foreground login failed: ... dial tcp <mgmt-ip>:443: i/o timeout
```

while other pods in the same namespace reached the same management endpoint fine — only deleting the pod (fresh netns) recovered it. Daemon-mode installs never hit this, which is likely why it hasn't been reported.

**Fix:** export `RestoreResidualState` and call it in `runInForegroundMode` before `foregroundLogin`, mirroring the daemon's behaviour (warn and continue on error).

`go build ./client/...` and `go vet` clean; `client/server` tests pass (`client/cmd`'s `TestRemoveStaleUnixSocket*` failures are pre-existing on `main` in my environment, unchanged by this PR).

None — field incident described above; happy to open an issue if preferred.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Documentation is not needed: this fixes an internal crash-recovery omission — foreground mode now performs the same residual-state cleanup the daemon already does. No user-facing behaviour, flags, API, or configuration change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved residual state restoration during foreground startup and foreground login, ensuring consistent routing behavior.
  * Residual-state restore errors now warn without interrupting the ongoing start or login flow.
  * Residual-state cleanup now consistently performs restore/persist steps to keep residual handling accurate.
  * Foreground flows now initialize networking earlier so routing is available when login proceeds.

### DIFF
--- a/client/cmd/up.go
+++ b/client/cmd/up.go
@@ -22,6 +22,7 @@ import (
 	"github.com/netbirdio/netbird/client/internal/peer"
 	"github.com/netbirdio/netbird/client/internal/profilemanager"
 	"github.com/netbirdio/netbird/client/proto"
+	nbnet "github.com/netbirdio/netbird/client/net"
 	"github.com/netbirdio/netbird/client/server"
 	"github.com/netbirdio/netbird/client/system"
 	"github.com/netbirdio/netbird/shared/management/domain"
@@ -240,6 +241,11 @@ func runInForegroundMode(ctx context.Context, cmd *cobra.Command, activeProf *pr
 	if err := server.RestoreResidualState(ctx, profilemanager.NewServiceManager(configPath).GetStatePath()); err != nil {
 		log.Warnf("failed to restore residual state: %v", err)
 	}
+
+	// Enable advanced routing (as the daemon does on startup) so the
+	// management dial bypasses a leftover fwmark rule instead of being
+	// shunted into a stale routing table.
+	nbnet.Init()
 
 	err = foregroundLogin(ctx, cmd, config, providedSetupKey, activeProf.ID)
 	if err != nil {
--- a/client/cmd/up.go
+++ b/client/cmd/up.go
@@ -22,6 +22,7 @@ import (
 	"github.com/netbirdio/netbird/client/internal/peer"
 	"github.com/netbirdio/netbird/client/internal/profilemanager"
 	"github.com/netbirdio/netbird/client/proto"
+	"github.com/netbirdio/netbird/client/server"
 	"github.com/netbirdio/netbird/client/system"
 	"github.com/netbirdio/netbird/shared/management/domain"
 	"github.com/netbirdio/netbird/util"
@@ -228,6 +229,17 @@ func runInForegroundMode(ctx context.Context, cmd *cobra.Command, activeProf *pr
 	}
 
 	_, _ = profilemanager.UpdateOldManagementURL(ctx, config, configFilePath)
+
+	// Restore residual state left by a previous run that did not shut down
+	// cleanly, mirroring what the daemon does before connecting. Without this,
+	// leftover routing rules can shunt the management connection into a stale
+	// routing table and every subsequent login times out. Foreground mode is
+	// particularly exposed in containers: a crashed container restarts inside
+	// the same (pod) network namespace, so the stale rules survive while the
+	// process state does not.
+	if err := server.RestoreResidualState(ctx, profilemanager.NewServiceManager(configPath).GetStatePath()); err != nil {
+		log.Warnf("failed to restore residual state: %v", err)
+	}
 
 	err = foregroundLogin(ctx, cmd, config, providedSetupKey, activeProf.ID)
 	if err != nil {
--- a/client/cmd/up.go
+++ b/client/cmd/up.go
@@ -232,12 +232,14 @@ func runInForegroundMode(ctx context.Context, cmd *cobra.Command, activeProf *pr
 	_, _ = profilemanager.UpdateOldManagementURL(ctx, config, configFilePath)
 
 	// Restore residual state left by a previous run that did not shut down
-	// cleanly, mirroring what the daemon does before connecting. Without this,
-	// leftover routing rules can shunt the management connection into a stale
-	// routing table and every subsequent login times out. Foreground mode is
-	// particularly exposed in containers: a crashed container restarts inside
-	// the same (pod) network namespace, so the stale rules survive while the
-	// process state does not.
+	// cleanly, mirroring what the daemon does before connecting: it recovers
+	// DNS config (a stale resolv.conf takeover can make the management
+	// hostname unresolvable), firewall rules, ssh config and legacy routing.
+	// Route cleanup itself happens at engine start; nbnet.Init() below lets
+	// the management dial bypass a leftover fwmark rule until then.
+	// Foreground mode is particularly exposed in containers: a crashed
+	// container restarts inside the same (pod) network namespace, so stale
+	// state survives while the process does not.
 	if err := server.RestoreResidualState(ctx, profilemanager.NewServiceManager(configPath).GetStatePath()); err != nil {
 		log.Warnf("failed to restore residual state: %v", err)
 	}
--- a/client/cmd/login.go
+++ b/client/cmd/login.go
@@ -333,9 +333,9 @@ func doForegroundLogin(ctx context.Context, cmd *cobra.Command, setupKey string,
 		return fmt.Errorf("read config file %s: %v", configFilePath, err)
 	}
 
-	// Mirror runInForegroundMode: recover residual state from a previous
-	// unclean shutdown and enable advanced routing before dialing
-	// management, so leftover routing rules can't break the login.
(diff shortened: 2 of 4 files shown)


dfry:
CI note: the one failing job (`Client / Unit`, Windows) is `TestResetAggregationWindow` in `client/internal/netflow/store` — a time-comparison flake in a package this PR doesn't touch (diff is limited to `client/cmd/up.go` and `client/server/{server,state}.go`). The Linux matrix variants (amd64, 386, Docker) all pass.

On codecov/patch: the change is a call-site mirror of the existing daemon behaviour (`RestoreResidualState` itself is exercised by the `client/server` tests); happy to add a dedicated foreground-path test if maintainers prefer.

CoderSufiyan (approved):
Clean fix. The asymmetry between daemon mode and foreground mode was the root cause -- daemon cleaned up residual state before connecting, foreground did not. In containers where the network namespace persists across container restarts, this meant stale routing rules survived and blocked the next login.

Exporting `RestoreResidualState` and calling it from `runInForegroundMode` is the minimal correct fix. Error handling (warn, continue) matches the daemon pattern. Container justification in the comment is clear.

LGTM.

dfry:
Rebased onto main — `Client / Unit` now passes on all platforms including Windows, confirming the earlier `TestResetAggregationWindow` failure was the pre-existing netflow/store flake (this PR doesn't touch that package).

The one remaining red job, `Relay / Unit (386)`, is `TestForeginConnClose` failing on `listen tcp 127.0.0.1:52202: bind: address already in use` — a port-collision flake in `shared/relay/client`, also untouched by this diff. Codecov reds are the no-new-tests thresholds; simulating leftover routing state from a crashed process in a unit test isn't practical, but happy to add one if someone has a pattern for it.

theodorsm (changes requested):
Hi @dfry, thanks for reporting the issue and suggesting a fix. 

I have tried to recreate the issue in the following way: 

Running a debian vm with netbird as a regular peer in foreground mode, and another ubuntu vm as an exit node. 
1. Run both vms and bring up netbird. When connected, the debian vm has a default route through `wt0` via the ubuntu vm. `172.17.0.1` is the management IP, `100.107.172.93` is the ubuntu vm peer IP.
```
root@debian13:~# ip rule 
0:      from all lookup local
105:    from all lookup main suppress_prefixlength 0
110:    not from all fwmark 0x1bd00 lookup 7120
32766:  from all lookup main
32767:  from all lookup default

root@debian13:~# ip route show table 7120
default dev wt0

root@debian13:~# ip route get 172.17.0.1
172.17.0.1 dev wt0 table 7120 src 100.107.172.93 uid 0 
    cache
```

2. Stop netbird process with `SIGKILL`
3. Debian vm has dangling routes and interface
4. Bring netbird down for ubuntu vm to remove the routing path to management.
5. Try to bring up debian in foreground mode again. This fails with `no route to host` (not a time out like you): 
```
026-07-14T11:20:47.110+02:00 INFO ./caller_not_available:0: 2026/07/14 11:20:47 WARNING: [core] [Channel #1 SubChannel #2] grpc:              
  addrConn.createTransport failed to connect to {Addr: "172.17.0.1:8080", ServerName: "172.17.0.1:8080", }. Err: connection error: desc =        
  "transport: Error while dialing: nbnet.NewDialer().DialContext: d.Dialer.DialContext: dial tcp 172.17.0.1:8080: connect: no route to host" 
```

This does not fail with I run the debian vm with netbird in daemon mode, so we clearly do not recover properly in foreground mode. 

I tried again with this PR, but it did not solve the issue. We need to enable advanced routing for the client to bypass the `fwmark` rule (during management dial) by running `Init()` located in `client/net/env_linux.go`: 

```

diff --git a/client/cmd/up.go b/client/cmd/up.go
index baf059ffd..56bfbbb8a 100644
--- a/client/cmd/up.go
+++ b/client/cmd/up.go
@@ -21,6 +21,7 @@ import (
        "github.com/netbirdio/netbird/client/internal"
        "github.com/netbirdio/netbird/client/internal/peer"
        "github.com/netbirdio/netbird/client/internal/profilemanager"
+       nbnet "github.com/netbirdio/netbird/client/net"
        "github.com/netbirdio/netbird/client/proto"
        "github.com/netbirdio/netbird/client/server"
        "github.com/netbirdio/netbird/client/system"
@@ -241,6 +242,8 @@ func runInForegroundMode(ctx context.Context, cmd *cobra.Command, activeProf *pr
                log.Warnf("failed to restore residual state: %v", err)
        }
 
+       nbnet.Init()
+
        err = foregroundLogin(ctx, cmd, config, providedSetupKey, activeProf.ID)
        if err != nil {
                return fmt.Errorf("foreground login failed: %v", err)
```

With the diff above, the problem that I initially observed was fixed. The routes are cleaned and reset upon engine start after successfully dialing the management server.

- Are you able to consistently re-create the time-out behavior you mention, could you share more details? I got a different error than you, but I was running vms and not in docker.
- Does the diff above fix the problem in your environment? It is still with your `RestoreResidualState` addition, this is valid to call from foreground mode also.

We also call `foregroundLogin` in `doForegroundLogin` (`client/cmd/login.go`), so we should apply the fix here too for completeness.

dfry:
Thanks for the thorough reproduction and the `nbnet.Init()` finding — both requests applied:

- `nbnet.Init()` added after `RestoreResidualState` in `runInForegroundMode` (your diff, kept with the residual-state restore as you confirmed)
- both calls mirrored into `doForegroundLogin` in `login.go`

**Consistent reproduction of our timeout variant** — container harness (a persistent netns, i.e. the pod analogue of our field incident):

1. `netbird up -F` (control build) in a `debian:bookworm-slim` container with `NET_ADMIN` + `/dev/net/tun`, enrolled against a real management server; `ip rule` shows `not from all fwmark 0x1bd00 lookup netbird`, and **netbird's resolv.conf takeover is in place** (`# Generated by NetBird`).
2. `kill -9` the process — container/netns survives: fwmark rule, `wt0` (still UP), **and the hijacked resolv.conf** all persist.
3. Retry with the control build → fails after retries with `context deadline exceeded`. The detail that explains the timeout-vs-`no route to host` difference vs your VM repro: the retry can't even *resolve* the management hostname — `lookup <mgmt-host> on 100.82.118.57:53: read: connection refused` (the stale resolv.conf points at the dead peer's own NetBird DNS). Blackholed lookups → deadline exceeded, no route error ever surfaces.
4. Same stale state, this branch's build → `statemanager: cleaning up state dns_state` (resolv restored), management resolves, relay connects, login succeeds.

So in container environments the residual-state restore is load-bearing for the **DNS** takeover specifically, and `nbnet.Init()` covers the fwmark shunt — the combination handles both variants. Happy to add the harness as a CI-able test if you'd find it useful, though it needs a live management endpoint as written.

client/cmd/up.go
```diff
@@ -229,6 +231,22 @@ func runInForegroundMode(ctx context.Context, cmd *cobra.Command, activeProf *pr
 
 	_, _ = profilemanager.UpdateOldManagementURL(ctx, config, configFilePath)
 
+	// Restore residual state left by a previous run that did not shut down
+	// cleanly, mirroring what the daemon does before connecting. Without this,
+	// leftover routing rules can shunt the management connection into a stale
+	// routing table and every subsequent login times out. Foreground mode is
+	// particularly exposed in containers: a crashed container restarts inside
+	// the same (pod) network namespace, so the stale rules survive while the
+	// process state does not.
```
theodorsm: `RestoreResidualState` does not clean up routing rules on Linux, so this is not correct. It recovers DNS, firewall rules, ssh configs and legacy routing. Cleanup of routes happen at engine start.

theodorsm (changes requested):
Thanks for updating the PR. I have feedback about the `RestoreResidualState` comment.

client/cmd/up.go
dfry: Corrected in 2d08be3 — the comment now attributes the recovery accurately (DNS/firewall/ssh/legacy-routing; route cleanup at engine start; `nbnet.Init()` for the fwmark bypass on the management dial), and the login.go mirror comment likewise. Matches the container repro, where the decisive recovery was the `dns_state` cleanup.

theodorsm (approved):

PR #6723, opened 2026-07-10
[management] deduplicate JWT group claims

### LINKED ISSUE #6618: getJWTGroupsChanges does not dedupe group names → duplicate JWT `groups` entries create phantom duplicate groups
## Summary

When the JWT `groups` claim contains the same group name more than once, NetBird's JWT group-sync creates a **separate group row per occurrence** instead of treating the duplicates as one group. The result is multiple distinct group IDs sharing an identical name (e.g. two `netbird-admins` rows with different IDs), and the affected user is auto-joined to all of them. The duplication is not detected or collapsed, so it persists in the store. The root cause is that `getJWTGroupsChanges` iterates the claimed group names without first de-duplicating them, so each repeated name is processed as if it were a new, distinct group.

## Environment

- NetBird self-hosted, embedded IdP (Dex) brokering an upstream OIDC connector.
- JWT group-sync enabled (`jwt_groups_enabled=1`, group claim = `groups`).

## Repro

1. Configure the IdP so that **two** property-mappings/claims both emit the `groups` claim for the same user (a common accident, e.g. an explicit "groups" mapping plus a default profile mapping that also emits the user's groups). The user's resulting token then carries the group names duplicated, e.g.
   `"groups": ["Admins", "netbird-admins", "Admins", "netbird-admins"]`.
2. Log that user in so JWT group-sync runs.
3. Inspect the groups store.

**Expected:** one group row per distinct name; the user is a member of each named group once.

**Actual:** a separate group row is created for each duplicate occurrence (e.g. **two** `netbird-admins` rows with different IDs and **two** `Admins` rows), and the user is auto-joined to all of them. The duplicates are not collapsed.

## Suggested fix

De-duplicate the claimed group names in `getJWTGroupsChanges` **before** the create/diff step: collapse the incoming `groupNames` to a set (preserving the existing case/normalisation semantics) so a duplicated claim is **idempotent**. This makes group-sync robust to misconfigured or multi-mapping IdPs without requiring every operator to guarantee exactly one emitter upstream.

## Impact

- **Integrity / split-brain:** two distinct group IDs share one name. A later policy that binds the group **by name-scoped selection** may resolve to only **one** of the duplicate IDs, while the user is a member of **both**, so the policy silently applies to a subset of the intended membership, or the "same" group means different things in different places. This is a latent authorization correctness hazard, not merely cosmetic clutter.
- It is also easy to hit by accident (any IdP where a default mapping and an explicit mapping both carry groups), so a single dedupe at the sync boundary removes a whole class of operator foot-guns.

_(Related observation: #5399, the embedded Dex connector dropping the `groups` scope on create, is a separate groups-handling issue in the same area.)_

CoderSufiyan:
## Description

JWT group claims can contain repeated names when multiple identity-provider mappings emit the same group. The synchronization path preserved those duplicates, creating multiple JWT groups with the same name and assigning every generated group ID to the user.

This deduplicates claimed group names before calculating membership changes while preserving their original order and case.

## Changes

- Deduplicate JWT group names at the synchronization boundary
- Extend the existing JWT group sync test with repeated claim values
- Verify both group records and user memberships remain unique

## Testing

- `go test ./management/server -run '^TestDefaultAccountManager_SyncUserJWTGroups$' -count=1`
- `go test ./management/server/util -count=1`
- `go build ./management/...`
- `go vet ./management/server ./management/server/util`

Fixes #6618

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate group assignments when incoming group claims contain repeated entries.
  * Group synchronization now produces only unique auto-group memberships.

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -136,6 +136,17 @@ func isUniqueConstraintError(err error) bool {
 // Returns a bool indicating if there are changes in the JWT group membership, the updated user AutoGroups,
 // newly groups to create and an error if any occurred.
 func (am *DefaultAccountManager) getJWTGroupsChanges(user *types.User, groups []*types.Group, groupNames []string) (bool, []string, []*types.Group, error) {
+	uniqueGroupNames := make([]string, 0, len(groupNames))
+	seenGroupNames := make(map[string]struct{}, len(groupNames))
+	for _, name := range groupNames {
+		if _, seen := seenGroupNames[name]; seen {
+			continue
+		}
+		seenGroupNames[name] = struct{}{}
+		uniqueGroupNames = append(uniqueGroupNames, name)
+	}
+	groupNames = uniqueGroupNames
+
 	existedGroupsByName := make(map[string]*types.Group)
 	for _, group := range groups {
 		existedGroupsByName[group.Name] = group
--- a/management/server/account_test.go
+++ b/management/server/account_test.go
@@ -680,7 +680,7 @@ func TestDefaultAccountManager_SyncUserJWTGroups(t *testing.T) {
 		Domain:         domain,
 		UserId:         userId,
 		DomainCategory: "test-category",
-		Groups:         []string{"group1", "group2"},
+		Groups:         []string{"group1", "group2", "group1", "group2"},
 	}
 	t.Run("JWT groups disabled", func(t *testing.T) {
 		err := manager.SyncUserJWTGroups(context.Background(), claims)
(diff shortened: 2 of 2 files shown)


0XFF-96 (approved):

CoderSufiyan:
This PR is mergeable and has been approved by @0XFF-96. When convenient, could a maintainer provide the required review? Thanks.

PR #6744, opened 2026-07-12
[client] Bring the connection up in Go after SSO login

mlsmaycon:
## Describe your changes

Fixes the GUI client sometimes not connecting for a long time after SSO login —
until the user interacts with the app (e.g. hovering the tray icon).

**Root cause:** the login flow orchestrates `Login → WaitSSOLogin → Up` in the
frontend (`client/ui/frontend/src/lib/connection.ts`). `WaitSSOLogin` blocks in
the daemon while the user completes SSO in the external browser. During that
time the tray window is hidden and the webview is suspended (macOS App Nap /
Windows hidden-window timer throttling), so the promise continuation that
issues `Up` doesn't run until the webview resumes — which only happens on user
interaction. This is exactly the stall `client/ui/frontend/src/lib/stallwatch.ts`
was added to detect in #6717 ("the WaitSSOLogin → Up handoff"); that PR added
detection only.

**Fix:** move the handoff off the webview's critical path. A new Go method
`Connection.WaitSSOLoginAndUp` performs `WaitSSOLogin` and then `Up` entirely in
Go, so the daemon connects the moment SSO completes, independent of webview
state. The frontend calls this single method on the SSO path and no longer
issues its own `Up`; the non-SSO (token) path is unchanged. Cancellation still
works via the bound call's context cancel, and the method re-checks `ctx.Err()`
before `Up`. The UI catches up via the status subscription when the window
wakes.

- `client/ui/services/connection.go`: add `WaitSSOLoginAndUp`
- `client/ui/frontend/src/lib/connection.ts`: use it on the SSO path; drop the
  separate frontend `Up` there

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

Note: this adds a UI-service method consumed only by the bundled desktop
frontend (not a public/gRPC/CLI surface). No behavior change outside the GUI
login path. Hard to cover with an automated test — it is webview-suspension
timing behind an external SSO round-trip; verified by manual repro (log in,
leave the tray window hidden, confirm the tunnel comes up without interaction).

## Documentation
Select exactly one:

Internal desktop-client login fix with no user-facing configuration or API
surface.

## Summary by CodeRabbit

* **New Features**
  * Added a streamlined connection flow that waits for SSO authentication to complete before bringing the connection online.
  * After success, the authenticated email address is returned as part of the connection process.
* **Bug Fixes**
  * Improved handling of canceled requests by ensuring interruption is detected after SSO completes.
  * Authentication and connection failures now surface as clearer, more consistent errors.

### DIFF
--- a/client/ui/services/connection.go
+++ b/client/ui/services/connection.go
@@ -162,6 +162,27 @@ func (s *Connection) Up(ctx context.Context, p UpParams) error {
 	return nil
 }
 
+// WaitSSOLoginAndUp blocks until the SSO login completes and then brings the
+// connection up, both from the Go side. Keeping the post-login Up here rather
+// than as a frontend continuation is deliberate: during SSO the tray window is
+// hidden and the webview is suspended (macOS App Nap / hidden-window timer
+// throttling), so a frontend-driven Up would not run until the user woke the
+// window (e.g. by hovering the tray icon). Doing it in Go connects the moment
+// the daemon reports SSO success. Returns the authenticated user's email.
+func (s *Connection) WaitSSOLoginAndUp(ctx context.Context, wait WaitSSOParams, up UpParams) (string, error) {
+	email, err := s.WaitSSOLogin(ctx, wait)
+	if err != nil {
+		return "", err
+	}
+	if err := ctx.Err(); err != nil {
+		return "", err
+	}
+	if err := s.Up(ctx, up); err != nil {
+		return "", err
+	}
+	return email, nil
+}
+
 func (s *Connection) Down(ctx context.Context) error {
 	cli, err := s.conn.Client()
 	if err != nil {
--- a/client/ui/services/connection.go
+++ b/client/ui/services/connection.go
@@ -35,7 +35,7 @@ type LoginResult struct {
 	VerificationURIComplete string `json:"verificationUriComplete"`
 }
 
-// WaitSSOParams are the inputs to WaitSSOLogin.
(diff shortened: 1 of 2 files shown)


client/ui/services/connection.go
```diff
@@ -162,6 +162,27 @@ func (s *Connection) Up(ctx context.Context, p UpParams) error {
...
 
+// WaitSSOLoginAndUp blocks until the SSO login completes and then brings the
+// connection up, both from the Go side. Keeping the post-login Up here rather
+// than as a frontend continuation is deliberate: during SSO the tray window is
+// hidden and the webview is suspended (macOS App Nap / hidden-window timer
+// throttling), so a frontend-driven Up would not run until the user woke the
+// window (e.g. by hovering the tray icon). Doing it in Go connects the moment
+// the daemon reports SSO success. Returns the authenticated user's email.
+func (s *Connection) WaitSSOLoginAndUp(ctx context.Context, wait WaitSSOParams, up UpParams) (string, error) {
+	email, err := s.WaitSSOLogin(ctx, wait)
```
pappz: Please clean up the code. WaitSSOLogin is exported unnecessarily (if I see well from ts code). Move its scope here or unexport it, and move the function after the exported functions in the file (keep the Go recommendations).

pappz (changes requested):

client/ui/services/connection.go
mlsmaycon: done

pappz (approved):

PR #6748, opened 2026-07-13
[client] preserve custom service name on upgrades

### LINKED ISSUE #5711: Debian post-install script doesn't respect custom service name
**Describe the problem**

Netbird CLI client allows redefining the service name: `netbird service install --service netbird-custom`.

After execution of this command `/etc/systemd/system/netbird-custom.service` is created.

But when the package becomes updated post-install script executes the following:
```
upgrade() {
    printf "\033[32m Post Install of an upgrade\033[0m\n"
    if [ "${use_systemctl}" = "True" ]; then
      printf "\033[32m Stopping the service\033[0m\n"
      systemctl stop netbird 2> /dev/null || true
    fi
    if [ -e /lib/systemd/system/netbird.service ]; then
      rm -f /lib/systemd/system/netbird.service
      systemctl daemon-reload
    fi
    # will trow an error until everyone upgrade
    /usr/bin/netbird service uninstall 2> /dev/null || true
    /usr/bin/netbird service install
    /usr/bin/netbird service start
}
```

As a result, the service `netbird` isn't present, and the uninstallation of the custom service doesn't happen. Netbird registers a new fresh `/etc/systemd/system/netbird.service` and launches it.

As a result, we get 2 running instances and 2 active services.

**To Reproduce**

1. Make sure you have no netbird installed.
2. Run `sudo apt install netbird=0.66.4`
3. Stop the service after installation with `sudo netbird service stop`
4. Uninstall systemd unit: `sudo netbird service uninstall`
5. Install systemd unit with the new name: `sudo netbird service install --service netbird-custom`
6. Start the service: `sudo netbird service start --service netbird-custom`
7. Make sure there is the proper name of the service and only 1 process is running

```
$ ls -la /etc/systemd/system | grep netbird
-rw-r--r--  1 root root  606 Mar 26 16:26 netbird-custom.service

$ ps aux | grep netbird
root     3869703  2.5  0.7 1276804 57216 ?       Ssl  16:31   0:00 /usr/bin/netbird service run --log-level info --daemon-addr unix:///var/run/netbird.sock --log-file /var/log/netbird/client.log
work     3869731  0.0  0.0   6544  2304 pts/3    S+   16:31   0:00 grep --color=auto netbird
```

8. Upgrade package: `sudo apt install netbird=0.67.1`
9. Check processes and systemd units again:

```
$ ls -la /etc/systemd/system | grep netbird
-rw-r--r--  1 root root  606 Mar 26 16:31 netbird-custom.service
-rw-r--r--  1 root root  578 Mar 26 16:32 netbird.service

$ ps aux | grep netbird
root     3869919  0.3  0.7 1277212 63104 ?       Ssl  16:32   0:00 /usr/bin/netbird service run --log-level info --daemon-addr unix:///var/run/netbird.sock --log-file /var/log/netbird/client.log
root     3869993  0.3  0.7 1277212 62976 ?       Ssl  16:33   0:00 /usr/bin/netbird service run --log-level info --daemon-addr unix:///var/run/netbird.sock --log-file /var/log/netbird/client.log
```

**Expected behavior**

Netbird must respect the unit name or remove `--service` flag to disallow systemd unit names redefining.

No matter

**NetBird version**

All versions

No

**Debug output**

N/A

**Screenshots**

N/A

**Additional context**

CoderSufiyan:
**Problem**
Package upgrades invoke `netbird service uninstall` without the custom `--service` name, so a custom systemd unit remains running and the upgrade installs a second default `netbird` unit.

**Solution**
Persist the service name with the other install parameters and apply saved parameters during uninstall. This makes package upgrades remove and recreate the configured custom unit. Explicit `--service` flags still take precedence.

**Documentation**

This fixes persisted internal service configuration without changing the CLI interface.

**Test**
`go test ./client/cmd -run TestServiceParams`

Fixes #5711

## Summary by CodeRabbit

* **New Features**
  * Saved service settings now retain the selected service name and restore it when reused.
  * Explicitly provided service names continue to take precedence over saved settings.
  * OIDC connections now request the `groups` scope by default, improving group information availability across supported identity providers.

* **Bug Fixes**
  * Uninstall operations now load existing service settings before proceeding, helping preserve the configured service context.

### DIFF
--- a/client/cmd/service_installer.go
+++ b/client/cmd/service_installer.go
@@ -173,6 +173,10 @@ var uninstallCmd = &cobra.Command{
 			return err
 		}
 
+		if err := loadAndApplyServiceParams(cmd); err != nil {
+			cmd.PrintErrf("Warning: failed to load saved service params: %v\n", err)
+		}
+
 		cfg, err := newSVCConfig()
 		if err != nil {
 			return fmt.Errorf("create service config: %w", err)
--- a/client/cmd/service_params.go
+++ b/client/cmd/service_params.go
@@ -21,6 +21,7 @@ const serviceParamsFile = "service.json"
 // serviceParams holds install-time service parameters that persist across
 // uninstall/reinstall cycles. Saved to <stateDir>/service.json.
 type serviceParams struct {
+	ServiceName           string            `json:"service_name,omitempty"`
 	LogLevel              string            `json:"log_level"`
 	DaemonAddr            string            `json:"daemon_addr"`
 	JSONSocket            string            `json:"json_socket"`
@@ -75,6 +76,7 @@ func saveServiceParams(params *serviceParams) error {
 // variables into a serviceParams struct.
 func currentServiceParams() *serviceParams {
 	params := &serviceParams{
+		ServiceName:           serviceName,
 		LogLevel:              logLevel,
 		DaemonAddr:            daemonAddr,
 		JSONSocket:            jsonSocket,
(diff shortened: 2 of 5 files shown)


CoderSufiyan:
Updated the PR body to include the exact docs acknowledgement checkbox required by CI. The docs check now passes.

The only remaining failing check is `codecov/project`; `codecov/patch` passes, so the changed lines are covered and this appears to be project-wide coverage drift rather than an issue with this patch.

PR #6749, opened 2026-07-13
[client] stop retrying invalid setup keys

### LINKED ISSUE #5715: CLI: "netbird up" hangs in a backoff retry loop when using an invalid setup key
**Describe the problem**

When executing `netbird up` with an invalid setup key, the CLI does not exit with an error. Instead, it gets stuck in a retry loop (takes too long with automation). The daemon receives a `NotFound` or `PermissionDenied` error from the Management Service but treats it as a transient network error rather than a fatal authentication failure. This forces the user to manually interrupt the process and run `netbird down` to stop the background loop.

**To Reproduce**

1. Run `netbird up --hostname test-node --setup-key INVALID-KEY-XXXX`
2. Notice that the command hangs and does not return control to the terminal.
3. Check the client logs.
4. See continuous backoff warnings: `operation failed, retrying in ... rpc error: code = NotFound desc = couldn't add peer: setup key is invalid`.

**Expected behavior**

The CLI and daemon should recognize that an invalid setup key is a permanent error. The retry loop should be aborted immediately, and the `netbird up` command should exit, displaying a clear authentication failure message to the user (since an invalid key won't magically become valid after a few minutes of retries).

Applicable to both (the issue is in the client's internal retry logic).

**NetBird version**

`0.67.1`

No

**Debug output**

Here is the log snippet showing the infinite retry loop:

```text
2026-03-27T10:30:36.774Z ERRO shared/management/client/grpc.go:575: failed to login to Management Service: rpc error: code = NotFound desc = couldn't add peer: setup key is invalid
2026-03-27T10:30:36.774Z ERRO client/internal/auth/auth.go:327: failed registering peer rpc error: code = NotFound desc = couldn't add peer: setup key is invalid
2026-03-27T10:30:36.774Z WARN client/internal/auth/auth.go:462: operation failed, retrying in 11.849974516s: rpc error: code = NotFound desc = couldn't add peer: setup key is invalid
2026-03-27T10:30:48.654Z ERRO shared/management/client/grpc.go:575: failed to login to Management Service: rpc error: code = PermissionDenied desc = no peer auth method provided, please use a setup key or interactive SSO login
...
2026-03-27T10:30:48.680Z WARN client/internal/auth/auth.go:462: operation failed, retrying in 11.499352542s: rpc error: code = NotFound desc = couldn't add peer: setup key is invalid
```

**Screenshots**

N/A

**Additional context**

It looks like the backoff implementation in `client/internal/auth/auth.go` (around line 462 based on the logs) is missing a check for fatal gRPC status codes.

Wrapping gRPC errors like `codes.NotFound` and `codes.PermissionDenied` in `backoff.Permanent(err)` during the peer registration/login phase would easily solve this issue and significantly improve the UX.

N/A

CoderSufiyan:
**Problem**
`netbird up --setup-key` can hang in the daemon retry loop when management returns a permanent auth/setup-key error such as `NotFound` for an invalid setup key. The retry loop already treated some auth failures as permanent, but not all relevant management auth statuses.

**Solution**
Classify `NotFound` and `Unauthenticated` as permanent authentication/setup-key errors. This lets invalid setup keys and unusable credentials fail fast while keeping connection errors like `DeadlineExceeded` and `Unavailable` retryable.

**Documentation**

This only changes retry classification for an existing error path.

**Test**
`go test ./client/internal/auth ./client/server`

Fixes #5715

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication error handling during login by expanding which authentication-related responses stop retrying, including “not found” and other common auth/status failures.
* **Tests**
  * Added unit test coverage to verify authentication error classification across expected gRPC status codes and non-gRPC errors.

### DIFF
--- a/client/internal/auth/auth.go
+++ b/client/internal/auth/auth.go
@@ -213,11 +213,11 @@ func (a *Auth) Login(ctx context.Context, setupKey string, jwtToken string) (err
 				if setupKey != "" && jwtToken == "" && isPeerLoginExpired(err) {
 					err = errSetupKeyOnSSOExpiredPeer
 				}
-				isAuthError = isPermissionDenied(err)
+				isAuthError = isAuthenticationError(err)
 				return err
 			}
 		} else if err != nil {
-			isAuthError = isPermissionDenied(err)
+			isAuthError = isAuthenticationError(err)
 			return err
 		}
 
@@ -470,7 +470,7 @@ func (a *Auth) withRetry(ctx context.Context, operation func(client *mgm.GrpcCli
 }
 
 // isAuthenticationError checks if the error is an authentication-related error that should not be retried.
-// Returns true if the error is InvalidArgument or PermissionDenied, indicating that retrying won't help.
+// Returns true if retrying won't help without changing credentials or setup key.
 func isAuthenticationError(err error) bool {
 	if err == nil {
 		return false
@@ -479,7 +479,9 @@ func isAuthenticationError(err error) bool {
 	if !ok {
 		return false
 	}
-	return s.Code() == codes.InvalidArgument || s.Code() == codes.PermissionDenied
+	return s.Code() == codes.InvalidArgument ||
+		s.Code() == codes.PermissionDenied ||
+		s.Code() == codes.NotFound
 }
 
 // isPermissionDenied checks if the error is a PermissionDenied error.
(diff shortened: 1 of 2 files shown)


CoderSufiyan:
Addressed the CodeRabbit auth classification suggestion by adding `codes.Unauthenticated` to the permanent auth errors and extending the table test.

Validation: `go test ./client/internal/auth ./client/server`

PR #6762, opened 2026-07-14
[client, management] Support per-peer lazy connection state and default proxy peers to lazy

lixmal:
## Describe your changes

Lazy connections were an all-or-nothing account setting. This makes laziness controllable per peer and runs the lazy connection manager continuously, so ephemeral proxy peers connect on-demand by default even when the account-wide flag is off.

- Run the lazy connection manager whenever it is supported instead of gating it on the account-wide flag; the account flag and the local override now decide which peers default to lazy versus always-active.
- Add a per-peer lazy override to the network map (default / lazy / eager) so management can steer individual peers in either direction.
- Default ephemeral proxy peers to lazy on both endpoints, so shared proxy infrastructure is not kept permanently connected to every peer.
- Keep a local NB_LAZY_CONN / MDM override authoritative over the per-peer state.
- Reflect the resolved account-wide default in client status and update it when the flag changes at runtime.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

No new user-facing configuration is introduced; the per-peer eager override has no admin-facing setting yet and the proxy-peer default is internal behavior.

## Summary by CodeRabbit

* **New Features**
  * Added per-peer lazy connection control (default/lazy/eager) via management configuration, including embedded proxy–aware lazy behavior.
  * Lazy peer eligibility and exclusions now incorporate ingress forwarding rules end-to-end.
* **Bug Fixes**
  * Remote lazy feature-flag updates no longer restart or reconcile lazy peers unnecessarily.
  * Exclude-list updates avoid redundant recomputation when the effective set hasn’t changed.
* **Tests**
  * Expanded coverage for per-peer lazy resolution and forwarding-rule-based exclusion behavior, plus updated engine lazy test setup.


shared/management/proto/management.proto
```diff
@@ -486,6 +486,22 @@ message RemotePeerConfig {
...
+}
+
+// LazyState is the management per-peer override for lazy connections.
+enum LazyState {
+  // Follow the account-wide lazy connection flag.
+  LazyStateDefault = 0;
+  // Force a lazy (on-demand) connection regardless of the account flag.
+  LazyStateLazy = 1;
+  // Force an always-active connection regardless of the account flag.
+  LazyStateEager = 2;
```
riccardomanfrin, line 519: Is this used?

client/internal/conn_mgr.go
```diff
@@ -59,69 +63,56 @@ func NewConnMgr(engineConfig *EngineConfig, statusRecorder *peer.Status, peerSto
...
 	if e.rosenpassEnabled {
-		log.Warnf("rosenpass connection manager is enabled, lazy connection manager will not be started")
+		log.Warnf("rosenpass is enabled, lazy connection manager will not be started")
 		e.statusRecorder.UpdateLazyConnection(false)
 		return
 	}
 
 	e.initLazyManager(ctx)
-	e.statusRecorder.UpdateLazyConnection(true)
+	e.statusRecorder.UpdateLazyConnection(e.PeerLazyDefault(mgmProto.LazyState_LazyStateDefault))
```
riccardomanfrin, line 104: is this capturing the fact that proxy peers are lazy when account flag is not set?

client/internal/conn_mgr.go
```diff
@@ -59,69 +63,56 @@ func NewConnMgr(engineConfig *EngineConfig, statusRecorder *peer.Status, peerSto
...
-	}
-
 	if e.rosenpassEnabled {
-		log.Warnf("rosenpass connection manager is enabled, lazy connection manager will not be started")
+		log.Warnf("rosenpass is enabled, lazy connection manager will not be started")
 		e.statusRecorder.UpdateLazyConnection(false)
 		return
 	}
 
 	e.initLazyManager(ctx)
```
riccardomanfrin, line 103: We are starting the lazy manager even when account and local lazy flag say to not use lazy. (majority of customers today?)

By doing so, we are running `toExcludedLazyPeers` on every sync update potentially on a long slice of `RemotePeerConfig`s, even though the result will exactly the entire map of all peers.

client/internal/engine.go
riccardomanfrin: GetPeerConfig can be `nil` ( [from line 1393](client/internal/engine.go#L45)  ). If so GetLazyConnectionEnabled is defaulted to `false`. Is it ok?

shared/management/proto/management.proto
lixmal: It's used on the consumer (client), it's not produced by mgmt yet

client/internal/conn_mgr.go
lixmal: I added a check to avoid running the expensive part when nothing has changed

client/internal/conn_mgr.go
lixmal: Nope, right now it reflects the account wide setting

client/internal/engine.go
lixmal: I've hardened it

pappz (approved):

PR #6767, opened 2026-07-14
[client] warm lazy connections from the DNS resolver

mlsmaycon:
## Describe your changes

With lazy connections enabled, a peer is not dialed until on-demand traffic arrives, so the first request to a peer resolved by name (e.g. the agent-network reverse-proxy) races — or loses to — the WireGuard handshake. Activation was previously reactive only: a data-path packet or an inbound signal.

This adds a proactive, DNS-time trigger. When the local resolver answers for an overlay name, it now warms the lazy connection to the peer(s) the answer points at and waits briefly for one to connect before returning the response — so by the time the client sends its first packet the tunnel is already up.

- `dns/local`: new `PeerActivator` capability + `SetPeerActivator` setter (mirrors the existing `PeerConnectivity`/`SetPeerConnectivity` injection). `ServeDNS` warms on the pre-filter answer, so activating a lazily-idle peer also lets it survive the disconnected-peer filter. No-op when no activator is wired (lazy off) or the answer carries no peer IPs. Budget is `NB_DNS_LAZY_WARMUP_TIMEOUT` (default 2s); on timeout the answer is returned anyway (never SERVFAIL).
- `dns.DefaultServer.SetPeerActivator` forwards to the local resolver.
- `client/internal`: a small engine-side adapter (`dnsPeerActivator`) resolves answer IPs to peers via `Status.PeerStateByIP`, activates them through `ConnMgr.ActivatePeer` (HA fan-out included), and polls `PeerStateByIP` until one is connected. `ConnMgr` is not thread-safe (guarded by `syncMsgMux`) and DNS queries run on their own goroutines, so activation is done under that mutex and the wait is not. It is injected in the engine after the connection manager is constructed (it does not exist at DNS-server construction time), via a `*dns.DefaultServer` type assertion to avoid widening the `Server` interface.

Scope is overlay-only for free: the trigger lives in the local resolver, which only answers for NetBird-managed names; public/upstream DNS is unaffected. Already-connected peers short-circuit, so steady-state DNS latency is unchanged.

N/A — follow-up to the lazy-connection rollout; fixes the agent-network cold-start observed in the e2e (proxy peer stuck disconnected until traffic).

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal client behavior; the only knob is the optional `NB_DNS_LAZY_WARMUP_TIMEOUT` tuning env var with a safe default.

## Tests

- `dns/local`: warm-up invokes the activator with the answer's peer IP; no-activator path is unchanged; a query with no answer does not invoke the activator.

## Summary by CodeRabbit

* **New Features**
  * DNS-time lazy-connection warm-up now wakes idle peers based on resolved A/AAAA records, with server-level enable/disable support.
  * Warm-up timeout is configurable via an environment variable.
* **Bug Fixes**
  * Address handling was made consistent for resolver filtering and activator invocation.
  * Improved safe concurrent activation during connection manager lifecycle transitions.
* **Tests**
  * Added unit tests for warm-up behavior, timeout parsing, and address extraction, plus activator timing/edge-case coverage.
  * Updated e2e tests to ensure warm-up is triggered before proxy/peer connectivity assertions.


client/internal/dns_peer_activator.go
```diff
@@ -0,0 +1,71 @@
...
+	for _, ip := range ips {
+		st, ok := a.status.PeerStateByIP(ip)
+		if !ok || st.ConnStatus == peer.StatusConnected {
+			continue
+		}
+		conn, ok := a.peerStore.PeerConn(st.PubKey)
+		if !ok {
+			continue
+		}
+		a.connMgr.ActivatePeer(ctx, conn)
```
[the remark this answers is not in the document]
mlsmaycon: @coderabbitai check again

client/internal/dns/local/local.go
```diff
@@ -39,6 +40,30 @@ type PeerConnectivity interface {
...
 
+// PeerActivator wakes lazy-connection peers on demand. The local resolver calls
+// it with the tunnel IPs an answer points at, so a peer that is idle (lazily
+// disconnected) begins its WireGuard handshake at DNS-resolution time rather
+// than racing the client's first request packet. nil disables warm-up.
+type PeerActivator interface {
+	// ActivatePeersByIP triggers wake-up for the peer(s) owning ips and blocks
+	// until one is connected or ctx (a short per-query budget) expires. It is a
+	// fast no-op for unknown or already-connected IPs.
+	ActivatePeersByIP(ctx context.Context, ips []string)
```
lixmal: Let's get rid of using string for IPs please, there are too many fallpits. It's a pre-existing debt but we keep extending it

client/internal/dns/local/local.go
```diff
@@ -518,6 +557,33 @@ func (d *Resolver) filterDisconnectedPeerAnswers(logger *log.Entry, question dns
 	return kept
 }
 
+// warmLazyPeers triggers lazy-connection wake-up for the peers a resolved
+// answer points at and waits briefly for one to connect, so the caller's first
+// request doesn't race the WireGuard handshake. No-op when no activator is
```
lixmal: Let's not mention wireguard here, this layer should not depend on that.

client/internal/dns/local/local.go
```diff
@@ -518,6 +557,33 @@ func (d *Resolver) filterDisconnectedPeerAnswers(logger *log.Entry, question dns
...
+	for _, rr := range records {
+		if ip := extractRecordIP(rr); ip != "" {
+			ips = append(ips, ip)
+		}
+	}
+	if len(ips) == 0 {
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(d.ctx, lazyWarmupTimeout())
```
lixmal: This will go through the env parsing for each request, let's parse it once in the constructor (or setter)

client/internal/engine.go
```diff
@@ -654,6 +654,19 @@ func (e *Engine) Start(netbirdConfig *mgmProto.NetbirdConfig, mgmtURL *url.URL)
 	e.connMgr = NewConnMgr(e.config, e.statusRecorder, e.peerStore, wgIface)
 	e.connMgr.Start(e.ctx)
 
+	// Wire DNS-time lazy-connection warm-up now that the connection manager
+	// exists (it does not at DNS-server construction time). A DNS answer that
+	// points at an idle peer then wakes it before the client's first request.
+	if ds, ok := e.dnsServer.(*dns.DefaultServer); ok {
```
lixmal: Do we need this assertion?

client/internal/dns/local/local.go
```diff
@@ -122,6 +158,9 @@ func (d *Resolver) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
 	replyMessage.RecursionAvailable = true
 
 	result := d.lookupRecords(logger, question)
+	// Warm before filtering: activation flips a lazily-idle target to connected,
+	// which then lets it survive the disconnected-peer filter below.
+	d.warmLazyPeers(result.records)
```
lixmal: This will run for all peers, not just proxy peers, which is a real cost

client/internal/engine.go
```diff
@@ -654,6 +654,19 @@ func (e *Engine) Start(netbirdConfig *mgmProto.NetbirdConfig, mgmtURL *url.URL)
...
+		ds.SetPeerActivator(&dnsPeerActivator{
+			connMgr:   e.connMgr,
+			peerStore: e.peerStore,
+			status:    e.statusRecorder,
+			mu:        e.syncMsgMux,
```
lixmal: I'm not sure it's a good idea to reuse this mutex, it is locked for every network map update, which can be long. It will block the dns resolution path.

lixmal (changes requested):

client/internal/dns/local/local.go
```diff
@@ -39,6 +40,30 @@ type PeerConnectivity interface {
...
+	ActivatePeersByIP(ctx context.Context, ips []string)
+}
+
+const defaultLazyWarmupTimeout = 2 * time.Second
+
+// lazyWarmupTimeout is the per-query budget for waking a lazy-connection peer a
+// DNS answer points at. Tunable via NB_DNS_LAZY_WARMUP_TIMEOUT (a Go duration).
+func lazyWarmupTimeout() time.Duration {
+	if v := os.Getenv("NB_DNS_LAZY_WARMUP_TIMEOUT"); v != "" {
+		if d, err := time.ParseDuration(v); err == nil && d > 0 {
```
lixmal: Errors should be logged

mlsmaycon:
Superseded by #6854 — same change on the unprefixed branch `dns-lazy-conn-warmup`, with the review feedback from this PR addressed (netip.Addr interfaces, warm-up scoped to match-only zones, env parsed once with error logging, no `syncMsgMux` on the DNS path, `SetPeerActivator` on the `dns.Server` interface) plus unit tests covering the new behavior. Agent-network e2e is green on that code with lazy connections enabled.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #6021, opened 2026-04-28
Relay protocol selection (WS vs QUIC) + per-pair throughput ceiling (~25 Mbps with 76% Go runtime overhead)


dlf-dds:
We've been doing a load-test campaign on a self-hosted netbird deployment (v0.70.0 daemon + v0.70.0 relay, official `netbirdio/relay:latest` image, t3.micro relay VPS in eu-central-1 / il-central-1, two peer-pairs behind separate AWS NAT Gateways forcing relay routing) and surfaced two related observations that we'd like to raise upstream. Both are reproducible.

Original prompt for the investigation: hyper-derp's announcement post (https://hyper-derp.dev/blog/hyper-derp-announcement/), which benchmarks Netbird's relay code as part of the author's preamble. Our findings below partially confirm and partially refine the claims in that post against our own deployment.

## 1. Daemon picks WebSocket exclusively even when QUIC is available

The relay binary listens for **both** WebSocket (TCP) and QUIC (UDP) on the same port (`:33073`):

```
relay/server/listener/quic/listener.go:39: QUIC server listening on address: :33073
relay/server/listener/ws/listener.go:44:   WS server listening address: :33073
```

The netbird daemon binary contains the `quic-go` library (verified via `strings | grep quic-go`). Yet across our entire deployment (38 active peer connections during the observation window) **100% of relay connections are WebSocket; 0% are QUIC**. Every entry in the relay log is `WS client connected from <ip>`, none are `QUIC client connected from <ip>`.

**Questions for the maintainers:**

- Why does the daemon prefer WebSocket over QUIC when both are available?
- Is there a config field, env var, or CLI flag to express a protocol preference? `netbird up --help` shows none; we found `QUIC_GO_DISABLE_GSO` and `QUIC_GO_DISABLE_ECN` (quic-go library tunables), but nothing controlling daemon-side protocol selection.
- Would maintainers consider exposing a relay-protocol-selection interface (env var, config field, or CLI flag) so operators can choose their preferred transport? A `prefer_quic` toggle would let deployments take advantage of QUIC's congestion-control and HOL-blocking-resistance properties where the network supports UDP.

We'd be happy to send a PR if there's appetite — please point us at the protocol-selection codepath if so.

## 2. Per-pair throughput caps at ~25 Mbps with 76% of CPU time in Go runtime overhead

iperf3-over-WireGuard between two peers forced into the relay path consistently caps at ~25-30 Mbps single-flow, regardless of:

[...]

PR #6775, opened 2026-07-15
[client,relay] Reduce per-packet overhead on the relay data path

Silex:
## Describe your changes

Second slice of the relay-throughput work from #6021 (first slice: #6774). Everything in this PR is **wire-compatible** — no protocol change, an old client and a new relay (or vice versa) interoperate unchanged.

Four independent layers, bottom of the stack first:

**1. Client: bounded backpressure on the inbound relay path.** Instead of dropping a relayed message the moment the per-peer channel is full, the reader applies bounded backpressure before giving up, and the channel is sized for the path's bandwidth-delay product rather than a fixed small constant. Drops that do happen are now visible: `RelayConnState` exposes an `InboundMsgDrops` counter instead of losing packets silently.

**2. Relay server: per-peer write queue.** Forwarding used to write synchronously into the destination peer's connection, so one slow receiver could stall the shared forwarding path. Writes now go through a bounded per-peer queue with a drop counter exported as `relay_transport_queue_drops_total` on the metrics endpoint. Unit tests included.

**3. Batch I/O on the data path (opt-in).** Three syscall-batching levers, each gated by an env var and **off by default**:
- `NB_RELAY_BATCH_READ` — drain the local WG proxy socket with `recvmmsg` instead of one read per packet,
- `NB_RELAY_INJECT_BATCH` — inject received packets into WireGuard with `sendmmsg`,
- plus a `ReadBatch` on the relay client connection feeding the inject path.

**4. A benchmark harness** (`TestPipe...` in `shared/relay/client`) that exercises the full client→relay→client data path in-process, so data-path changes can be measured without a lab setup.

Measured effect on the same pair as #6774 (two 1 Gbit Linux hosts, ~5 ms RTT via a dedicated relay, single iperf3 flow, cubic, stock sysctls, receiver-side Mbit/s), with the batch gates enabled and the WebSocket transport (`NB_RELAY_TRANSPORT=ws`):

| | up | down |
|---|---|---|
| #6774 alone | 180 | 150 |
| + this PR | 227 | 184 |

(Stock 0.74.2 baseline was 97.6/96.5.) A side observation from the same runs: once buffers are sized, the WS transport consistently outperformed QUIC on this path (227 vs 183 up) — QUIC datagram queues drop silently under bursts.

The batch gates ship off exactly as they were tested; happy to flip any of them default-on if that's the direction you prefer after your own validation.

Stacked on #6774 — the first commit here is that PR's commit; review the rest.

#6021

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor
- [x] This change does **not** modify the public API, gRPC protocols, functionality behavior, CLI / service flags, or introduce a new feature — **OR** I have discussed it with the NetBird team beforehand (link the issue / Slack thread in the description). See [CONTRIBUTING.md](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTING.md#discuss-changes-with-the-netbird-team-first). — discussion anchor: #6021

## Documentation
Select exactly one:

The default-on parts (backpressure, write queue, metric) need no operator action; the batch env vars are advanced opt-in tuning knobs, consistent with other undocumented `NB_*` tunables. Can add docs for the metric name if desired.

## Summary by CodeRabbit

* **New Features**
  * Added optional packet batching to improve relay throughput and reduce processing overhead.
  * Added configurable UDP socket buffer sizing, including an option to disable it.
  * Added inbound message drop reporting and transport queue drop metrics.
* **Bug Fixes**
  * Improved backpressure handling so slow destinations no longer block message intake.
  * Added bounded queues and safer cleanup to prevent stalls during congestion or connection closure.
* **Performance**
  * Improved bidirectional relay performance with batched reads and writes.
  * Added support for detecting datagram-capable connections.

### DIFF
--- a/client/iface/wgproxy/ebpf/proxy.go
+++ b/client/iface/wgproxy/ebpf/proxy.go
@@ -104,6 +104,7 @@ func (p *WGEBPFProxy) Listen() error {
 		return err
 	}
 	p.conn = conn
+	nbnet.SizeRelaySocketBuffers(conn)
 
 	go p.proxyToRemote()
 	log.Infof("local wg proxy listening on: %d", proxyPort)
--- a/client/iface/wgproxy/ebpf/proxy.go
+++ b/client/iface/wgproxy/ebpf/proxy.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"os"
 	"sync"
 
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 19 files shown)


Silex:
Closing — superseded by the discussion at #7401, where the diagnosis is being validated first. Branch kept.

PR #6776, opened 2026-07-15
[client,relay] Coalesce relayed packets into batch frames (RFC)

Silex:
## Describe your changes

**Draft / RFC** — this is the one slice of the #6021 throughput work that extends the relay wire protocol, so we'd like design guidance before polishing it (questions at the bottom). Stacked on #6775 (which stacks on #6774).

**What it does:** adds `MsgTypeTransportBatch`, a frame that packs several packets destined to the same peer as repeated `[uint16 length][payload]` entries after the normal transport header. The header layout deliberately matches a plain transport message, so the relay server rewrites only the destination peer ID and forwards the whole frame **opaquely** — no per-packet parsing on the server's hot path. The client packs opportunistically: whatever is already queued gets coalesced up to `MaxMessageSize`, so it adds no latency under light load and packs full frames exactly when it matters. Opt-in via `NB_RELAY_COALESCE` (off by default).

Coalescing only engages on stream transports (WebSocket). QUIC datagrams are capped by path MTU, so they can't carry multi-packet frames — one more reason WS wins on relayed throughput once buffers are sized.

Measured effect (same setup as #6774: two 1 Gbit Linux hosts, ~5 ms RTT, single iperf3 flow, cubic, stock sysctls, receiver-side Mbit/s, WS transport):

| | up | down |
|---|---|---|
| #6774 + batch I/O | 227 | 184 |
| + this PR | 288 | ~180 |

Cumulative vs stock 0.74.2: 97.6 → 288 Mbit/s single-flow (~3×), with zero retransmissions on the tunnelled TCP.

**Why draft — the compatibility question.** A batch frame must be understood by **both** the relay server (it accepts the type) and the **receiving peer** (frames are forwarded opaquely, so the receiver unpacks them). The protocol currently has no capability negotiation: the version byte is hard-validated to 1 everywhere, and `AuthResponse` carries only the server address. As-is, the feature is safe purely because it's opt-in — an operator enables it once the relay and all participating peers run a supporting version. That's workable for self-hosted fleets but not a general answer. Options we see:

1. **Keep it operator-coordinated** (this PR as-is): env-gated everywhere, documented constraint. Simplest; no handshake change.
2. **Protocol version bump to 2.** The relay learns each client's version at auth time; it forwards batch frames opaquely to v2 receivers and unpacks them into single transport messages for v1 receivers; senders enable coalescing only when the relay advertises v2. Backward-compatible in both directions, but old relays reject a v2 auth outright, so new clients need a one-shot fallback reconnect at v1, and the relay grows an unpack path for mixed fleets.
3. **Explicit capability flag** in the handshake (new field or message) rather than overloading the version byte.

Which direction would you take? We'll rework the PR accordingly.

#6021

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor
- [x] This change does **not** modify the public API, gRPC protocols, functionality behavior, CLI / service flags, or introduce a new feature — **OR** I have discussed it with the NetBird team beforehand (link the issue / Slack thread in the description). See [CONTRIBUTING.md](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTING.md#discuss-changes-with-the-netbird-team-first). — discussion anchor: #6021; this PR is a draft specifically to have that design discussion.

## Documentation
Select exactly one:

Draft under design discussion; `NB_RELAY_COALESCE` is an opt-in experimental gate. Docs to be added once the negotiation design is settled.

### DIFF
--- a/client/iface/wgproxy/ebpf/proxy.go
+++ b/client/iface/wgproxy/ebpf/proxy.go
@@ -49,7 +49,8 @@ type WGEBPFProxy struct {
 
 	// batchRead drains several relayed packets from the local WireGuard socket
 	// with one recvmmsg and forwards them back-to-back, cutting the per-packet
-	// syscall and wakeup overhead on the hot send path. Gated by
+	// syscall and wakeup overhead on the hot send path and letting the relay
+	// client's frame coalescer (NB_RELAY_COALESCE) pack full frames. Gated by
 	// NB_RELAY_BATCH_READ; the per-packet path stays the default.
 	batchRead bool
 
@@ -219,7 +220,8 @@ func (p *WGEBPFProxy) proxyToRemote() {
 // the single proxy socket is shared by every relayed peer, recvmmsg yields each
 // packet's source address, which selects the destination turn conn. Draining
 // several packets per recvmmsg amortizes the read syscall and wakeup across the
-// batch on the hot send path.
+// batch and forwards them back-to-back so the relay client's coalescer can pack
+// full frames.
 func (p *WGEBPFProxy) proxyToRemoteBatch(pc net.PacketConn) {
 	batchConn := ipv4.NewPacketConn(pc)
 	bufSize := int(p.mtu) + bufsize.WGBufferOverhead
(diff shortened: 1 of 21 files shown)


Silex:
Closing this draft — the protocol question belongs in a discussion, not a 21-file draft. Context and measurements at #7401.

PR #6796, opened 2026-07-16
[client] Disconnect daemon on GUI quit via async Down

pappz:
The tray Quit menu now disconnects the daemon before exiting instead of only tearing down the GUI. A new DownAsync RPC lets the daemon start the teardown and return immediately: beginDown cancels the connection under the mutex (so it cannot reconnect), then finishDown (the retry-goroutine wait and status reset) runs on a background goroutine. handleQuit aborts any in-flight profile switch first (so a queued Up cannot reconnect during teardown) and calls DownAsync so quitting never blocks on the engine shutdown.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

---
<a href="https://app.blacksmith.sh/netbirdio/codesmith/netbird/pr/6796"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786785261&installation_id=146802194&pr_number=6796&repository=netbirdio%2Fnetbird&return_to=https%3A%2F%2Fgithub.com%2Fnetbirdio%2Fnetbird%2Fpull%2F6796&signature=4fc95ebfed320240170c8318a2bc8750acfa5ad131454e3011c43bc8a5a09ed1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown reliability by continuing teardown even when stopping the service fails (stop errors are logged but not returned).
  * Refined connection shutdown to return “service not up” errors directly for clearer, more immediate RPC behavior.
  * Prevented shutdown hangs by making the tray Quit disconnect time-bounded (5 seconds).
  * Ensured any in-flight profile switch is cancelled before exiting, with quit serialized to avoid races.


client/ui/tray.go
```diff
@@ -30,6 +30,8 @@ const (
 
 	statusError = "Error"
 
+	quitDownTimeout = 5 * time.Second
```
lixmal, line 33: The ctx is not used right now, but this seems a little short

client/server/server.go
```diff
@@ -1154,7 +1174,7 @@ func (s *Server) cleanupConnection() error {
 	// making the run loop the sole owner of engine shutdown.
 	if engine != nil {
 		if err := engine.Stop(); err != nil {
-			return err
+			log.Errorf("failed to stop engine during cleanup: %v", err)
```
lixmal, line 1160: Why change this? The caller already logs (and returns an error)

pappz: At some point we should go over the callers of this function and clean up the situations. engine.Stop NEVER returns an error. This condition is dead. But if we return err then the internal state will be inconsistent. The engine is under stop (or whatever) and the `connectClient` never will be resetet. So I think this is the better sitation. But again:
```
	// TODO: consider calling s.connectClient.Stop() instead of engine.Stop().
	// actCancel() lets the run loop stop the engine too, so both stop it
	// concurrently; ConnectClient.Stop cancels and waits for the run loop,
	// making the run loop the sole owner of engine shutdown.
```

client/ui/tray.go
pappz: It is just defensive code to do not block the UI thread for too long time. We want to wait until the grpc lib confirm the message has been sent but the result do not care really if something went wrong.

lixmal (approved):

client/server/server.go
```diff
@@ -1081,7 +1081,10 @@ func (s *Server) Down(ctx context.Context, _ *proto.DownRequest) (*proto.DownRes
 
 	if err := s.cleanupConnection(); err != nil {
 		s.mutex.Unlock()
-		// todo review to update the status in case any type of error
+		if errors.Is(err, ErrServiceNotUp) {
+			log.Debugf("Down called while service not up: %v", err)
+			return nil, err
+		}
 		log.Errorf("failed to shut down properly: %v", err)
 		return nil, err
```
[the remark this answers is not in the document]
pappz: On the caller side, we are using a context with a timeout. So it should block the caller for a maximum of 5 seconds, right?

PR #6802, opened 2026-07-16
[client] Preserve firewall authorization in connection marks

### LINKED ISSUE #6022: NetBird `netbird-acl-forward-filter` drops k3s/kube-router-forwarded VPN traffic due to mark-bit collision with kube-router netpol
# NetBird `netbird-acl-forward-filter` drops k3s/kube-router-forwarded VPN traffic due to mark-bit collision with kube-router netpol

I'm not gonna lie, the robots helped me figure it out but it does seem like k3s and netbird rules trample each other.

Version 0.70.0
 
## TL;DR

NetBird's `netbird-acl-forward-filter` chain identifies "VPN-peer → local-host" traffic by a `meta mark 0x0001bd20` set in `netbird-mangle-prerouting`. **Bit `0x10000` is set inside that mark.** kube-router's per-pod NetworkPolicy chain (`KUBE-POD-FW-*`) **uses bit `0x10000` as its own "policy-permitted" flag** and unconditionally **clears it** at the end of every per-pod evaluation. After kube-router's iptables FORWARD chain runs, the packet's mark is `0x0002bd20` instead of `0x0001bd20`, so NetBird's `meta mark 0x0001bd20 accept` rule no longer matches and the packet falls through to the chain's catch-all `iifname "wt0" drop`.

The collision is silent — no rejects, no logs from either side.

`NB_DISABLE_FIREWALL=true` works around it by removing NetBird's nftables table entirely.

---

## Environment

| Component | Version / detail |
|---|---|
| NetBird agent | `0.70.0` |
| Host OS | Ubuntu (kernel uses `nf_tables` backend; iptables binary is iptables-nft) |
| Kubernetes | k3s with **kube-router** as netpol controller, **flannel** CNI, default **traefik** ingress |
| Traefik | Helm chart `traefik-37.1.0`, image `docker.io/traefik:v3.5.1`, Service `type: ClusterIP`, Deployment with `hostPort: 80 → 8080` and `hostPort: 443 → 8443` declared on the container; reachable from outside via the **CNI portmap plugin** writing `CNI-DN-…` DNAT rules in `nat` table `PREROUTING`/`OUTPUT` |
| NetBird interface | `wt0`, `10.79.25.88/18`, peer subnet `10.79.0.0/18` |
| VPN client | `10.79.43.248`, attempting `curl https://api-prod-tyo3-01:443` |

## Symptom

- VPN client SYN to `<host-wt0-IP>:443` arrives on `wt0` and times out (no SYN-ACK, no RST).
- LAN clients on the host's primary interface and `localhost` succeed.
- After `systemctl restart netbird`, VPN→host:443 **works for ~60s** then fails again. (Believed to coincide with kube-router's periodic iptables sync re-registering its base chains *after* NetBird's, flipping evaluation order at the same hook priority.)
- `NB_DISABLE_FIREWALL=true` resolves the issue permanently.

### 1. Packet enters and is correctly marked + DNAT'd

dmesg LOGs added at every netfilter hook (`iptables -t … -I … 1 -j LOG`) showed:

```
RAW-PRE:        IN=wt0 SRC=10.79.43.248 DST=10.79.25.88 SPT=…  DPT=443  SYN
MANGLE-PRE:     IN=wt0 SRC=10.79.43.248 DST=10.79.25.88 SPT=…  DPT=443  MARK=0x1bd20   ← NetBird sets the mark
NAT-PRE:        IN=wt0 SRC=10.79.43.248 DST=10.79.25.88 SPT=…  DPT=443  MARK=0x1bd20   ← unchanged
MANGLE-FWD-POST:IN=wt0 OUT=cni0 SRC=10.79.43.248 DST=10.42.0.17 SPT=…  DPT=8443 MARK=0x1bd20  ← CNI-DN DNAT'd to traefik pod
FILTER-FWD:     IN=wt0 OUT=cni0 SRC=10.79.43.248 DST=10.42.0.17 SPT=…  DPT=8443 MARK=0x1bd20
```

So far so good. NetBird's mark is the post-DNAT value still `0x1bd20`.

### 2. Packet never reaches the egress interface

`tcpdump -ni any 'host 10.79.43.248 or host 10.42.0.17'` shows the packet on `wt0 In` only — never on `cni0`, never on any `veth*`. The pod never receives the SYN.

`net.ipv4.ip_forward = 1`, `ip route get 10.42.0.17 → dev cni0`, all interface-level forwarding sysctls = 1, `rp_filter = 2` (loose). Routing is fine.

### 3. NetBird's nftables table

```
table ip netbird {
    set nb0000001 {
        type ipv4_addr; flags dynamic
        elements = { 10.79.43.248, … }              # all peers
    }

    chain netbird-mangle-prerouting {
        type filter hook prerouting priority mangle; policy accept;
        iifname "wt0" ct state new ct mark set 0x0001bd10
        iifname "wt0" ip saddr @nb0000001 fib daddr type local meta mark set 0x0001bd20
    }

    chain netbird-acl-forward-filter {
        type filter hook forward priority filter; policy accept;
        meta mark 0x0001bd20 accept                   # the only "permit VPN→local" rule
        iifname "wt0" jump netbird-rt-fwd             # only accepts established,related
        iifname "wt0" drop                            # catch-all
    }
    …
}
```

In isolation, the SYN's mark **is** `0x0001bd20` when entering the FORWARD hook, so `meta mark 0x0001bd20 accept` *should* match.

### 4. Mark collision with kube-router

iptables FORWARD also runs at `hook forward priority filter` (= 0) — same priority as `netbird-acl-forward-filter`. Its first rule is:

```
KUBE-ROUTER-FORWARD  /* kube-router netpol – TEMCG2JMHZYE7H7T */
```

…which dispatches into per-pod chains (`KUBE-POD-FW-…`). The traefik pod's chain ends with the standard kube-router epilogue:

```
NFLOG  ... mark match ! 0x10000/0x10000 ...
REJECT ... mark match ! 0x10000/0x10000 ...
MARK and 0xfffeffff               ← clears bit 0x10000
MARK or  0x20000                  ← sets  bit 0x20000   (kube-router "permitted")
```

**The collision:**

```
NetBird mark    : 0x0001bd20  =  0000 0001 1011 1101 0010 0000
                                          ^
                                          bit 0x10000 is SET as part of NetBird's tag

After KUBE-POD-FW :
  & 0xfffeffff  : 0x0000bd20  (bit 0x10000 cleared by kube-router)
  | 0x20000     : 0x0002bd20  (bit 0x20000 set)
```

So when control returns to `netbird-acl-forward-filter`, the packet's mark is `0x0002bd20`, **not** `0x0001bd20`. NetBird's `meta mark 0x0001bd20 accept` no longer matches; the packet falls through to `iifname "wt0" drop` and is silently dropped.

### 5. Why it works for ~60s after a restart

Both NetBird's `netbird-acl-forward-filter` and the iptables-compat `FORWARD` chain hook `forward` at the same priority. Order of evaluation between two base chains at the same hook+priority is determined by **registration order**.

- Right after `systemctl restart netbird`, NetBird re-registers its chain *after* iptables — so iptables FORWARD evaluates first, kube-router clears bit `0x10000`, and NetBird's mark-accept rule *misses*. It should already be broken.
- However, in the brief window before kube-router/kube-proxy's next iptables reconcile, the chain ordering is such that NetBird's chain may evaluate first (still seeing the original mark and accepting), with the kube-router pass producing a verdict that doesn't override the prior accept.
- On the next kube-router/kube-proxy reconcile (~60s default), kube-router does an `iptables-restore`, **re-registers** its base chain, and the eval order swaps. From that moment on the bit-clearing wins and NetBird drops the SYN.

(That's the most consistent explanation for the timing; the root cause — bit collision — is the real bug, regardless of which chain wins the race.)

## Reproduction

1. Single-node k3s with default flannel CNI + kube-router netpol + traefik (Helm chart with `hostPort` 80/443 enabled).
2. NetBird ≥ 0.70.0 client on the same host, joined to a network with at least one peer.
3. From a peer, `curl https://<host-wt0-IP>:443`.
4. `systemctl restart netbird` → curl works briefly → fails within ~60s.

## Workaround

Disable NetBird's firewall management:

```bash
sudo netbird service reconfigure --service-env NB_DISABLE_FIREWALL=true
sudo systemctl restart netbird
```

Verify: `sudo nft list table ip netbird` → "No such file or directory".

## Suggested fix

The bit collision is the design flaw. Two options:

1. **Don't use bit `0x10000` in NetBird's marks.** It's effectively reserved by kube-router and possibly other CNIs (Calico uses similar bit-ranges). Pick a mark that doesn't share any low-bit space with `0x00010000`/`0x00020000` (e.g. anything in the upper 16 bits). At minimum, mask only the *NetBird* bits when checking — e.g. tag a connection with `0x80000000` and match `meta mark & 0x80000000 == 0x80000000`. I would suggest either picking a different bit or a config option that allows overriding a different bit.

2. **Match on `ct mark` instead of `meta mark`** for the forward-accept rule. NetBird already sets `ct mark 0x0001bd10` on new connections in `netbird-mangle-prerouting`; that mark is stored in the conntrack entry, is not affected by kube-router's per-packet `MARK` ops, and survives DNAT and chain hops. Rewriting the rule as:

    ```
    meta nfproto ipv4 ct mark and 0x0000ff00 == 0x0000bd00 accept    # or similar
    ```

    would make the accept independent of any other component touching the per-packet mark.

The same class of mark-stomp can affect any environment combining NetBird with another netfilter consumer that uses `--set-mark`/`--xset-mark` on the per-packet mark (kube-router, calico-felix, kube-proxy `KUBE-MARK-MASQ`, fwmark-based policy routing daemons, etc.). The conntrack-mark approach defends against all of them.

## Mark allocation reference

| Mark | Set by | Meaning | Cleared by |
|---|---|---|---|
| `0x00010000` | kube-router `KUBE-NWPLCY-DEFAULT` rule 4 | "Network policy evaluated, permitted" | kube-router `MARK and 0xfffeffff` at end of `KUBE-POD-FW-*` |
| `0x00020000` | kube-router `MARK or 0x20000` at end of `KUBE-POD-FW-*` | "Final accept" — matched by FORWARD line `mark 0x20000/0x20000 ACCEPT` | (kept) |
| `0x00004000` | kube-proxy `KUBE-MARK-MASQ` | "Will be masqueraded in POSTROUTING" | (kept) |
| `0x0001bd10` | NetBird `netbird-mangle-prerouting` (ct mark) | "wt0-inbound new connection" | (ct mark – persistent) |
| `0x0001bd11` | NetBird `netbird-mangle-postrouting` (ct mark) | "wt0-outbound new connection" | (ct mark – persistent) |
| **`0x0001bd20`** | NetBird `netbird-mangle-prerouting` (meta mark) | "wt0-inbound peer→local" | **kube-router clears bit `0x10000` of this mark while passing through `KUBE-POD-FW-*`** |
| `0x0001bd21` / `0x0001bd22` | NetBird (meta mark) | masquerade hint for routed traffic | (unused in this report) |

The `0x0001bd20` value contains **bits 0x10000, 0x8000, 0x2000, 0x1000, 0x800, 0x100, 0x20** — the bit-16 overlap with kube-router's reserved bit is the failure mode.

## Files / commands used during diagnosis

- `iptables -L FORWARD -nv --line-numbers` — confirmed kube-router's `KUBE-ROUTER-FORWARD` is line 1 of FORWARD and `ACCEPT in wt0` is line 3
- `iptables -t nat -S CNI-HOSTPORT-DNAT`, `iptables -t nat -S CNI-DN-…` — confirmed traefik's hostPort DNAT is in place and source-agnostic
- `nft list table ip netbird` — produced the rule set above
- `iptables -L KUBE-NWPLCY-DEFAULT -nv` — confirmed final rule is `MARK or 0x10000` for any traffic, i.e. the chain is permissive (no NetworkPolicy resource exists)
- LOG instrumentation at `raw/PREROUTING`, `mangle/PREROUTING`, `nat/PREROUTING`, `mangle/FORWARD`, `filter/FORWARD`, `nat/POSTROUTING`
- `tcpdump -ni any 'host 10.79.43.248 or host 10.42.0.17'`
- `sysctl net.ipv4.{ip_forward,conf.{all,wt0,cni0}.{forwarding,rp_filter}}` — all confirmed correct

## Related upstream issues

- [#2926 — netbird 0.32.0 breaks K3s 1.32.2+k3s1 with flannel due to iptables conflicts](https://github.com/netbirdio/netbird/issues/2926)
- [#3363 — Add an option to force `nftables` usage](https://github.com/netbirdio/netbird/issues/3363)
- [#4484 — Why use iptables in `client/firewall/nftags/router.linux.go`](https://github.com/netbirdio/netbird/issues/4484)
- [#481 — Add return forward rule for network routes](https://github.com/netbirdio/netbird/issues/481)

## Notes

I'm willing to come up with a PR for changing the bits or overriding with a config option if you wish. I could also try to work on the connection mark one if that's the way you wanted to go. That seems possibly a lot more complicated though. Let me know which way you are leaning.

CoderSufiyan:
## Describe your changes
Persist NetBird firewall authorization and routed-network marks in conntrack instead of relying on mutable per-packet marks. The DNAT guard now checks connection marks, so kube-router, kube-proxy, and other netfilter consumers can modify packet marks without causing authorized traffic to be dropped.

Both iptables and nftables backends are updated. Routed marks remain classified correctly by netflow.

Fixes #6022

## Stack
- [x] This PR is independent

## Checklist
- [x] I have performed a self-review of my own code
- [x] I have tested the changes locally
- [x] I have added or updated tests where applicable
- [x] I have checked for breaking changes

## Documentation

## Validation
- Cross-compiled Linux test binaries for `./client/firewall/iptables`
- Cross-compiled Linux test binaries for `./client/firewall/nftables`
- Cross-compiled Linux test binaries for `./client/internal/netflow/conntrack`
- `git diff --check`

Runtime netfilter tests require a Linux environment.

---
<a href="https://app.blacksmith.sh/netbirdio/codesmith/netbird/pr/6802"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786808434&installation_id=146802194&pr_number=6802&repository=netbirdio%2Fnetbird&return_to=https%3A%2F%2Fgithub.com%2Fnetbirdio%2Fnetbird%2Fpull%2F6802&signature=9712e419e1e17cb5d6a34ad77c961ff3629293f35137fc42a143fa54931b1fa6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved firewall rule behavior for redirected, masqueraded, and return traffic by switching NAT/ACL handling to use connection (conntrack) marks consistently.
  * Strengthened DNAT guard logic to drop incorrectly marked redirected traffic.
  * Improved NetFlow direction detection for firewall-managed connections based on these connection marks.
* **Tests**
  * Added/updated Linux-focused tests to verify connection-mark-based rule generation and NAT matching/cleanup behavior.

### DIFF
--- a/client/firewall/iptables/acl_linux.go
+++ b/client/firewall/iptables/acl_linux.go
@@ -98,7 +98,7 @@ func (m *aclManager) AddPeerFiltering(
 	mangleSpecs = append(mangleSpecs,
 		"-i", m.wgIface.Name(),
 		"-m", "addrtype", "--dst-type", "LOCAL",
-		"-j", "MARK", "--set-xmark", fmt.Sprintf("%#x", nbnet.PreroutingFwmarkRedirected),
+		"-j", "CONNMARK", "--set-mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmarkRedirected),
 	)
 
 	specs = append(specs, "-j", actionToStr(action))
@@ -386,7 +386,8 @@ func (m *aclManager) seedInitialEntries() {
 	m.appendToEntries(mangleFwdKey, []string{
 		"-i", m.wgIface.Name(),
 		"-m", "conntrack", "--ctstate", "DNAT",
-		"-m", "mark", "!", "--mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmarkRedirected),
+		"-m", "connmark", "!", "--mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmarkRedirected),
+		"-m", "connmark", "!", "--mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmarkMasquerade),
 		"-j", "DROP",
 	})
 }
@@ -394,7 +395,7 @@ func (m *aclManager) seedInitialEntries() {
 func (m *aclManager) seedInitialOptionalEntries() {
 	m.optionalEntries["FORWARD"] = []entry{
 		{
-			spec:     []string{"-m", "mark", "--mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmarkRedirected), "-j", "ACCEPT"},
+			spec:     []string{"-m", "connmark", "--mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmarkRedirected), "-j", "ACCEPT"},
 			position: 2,
 		},
 	}
--- a/client/firewall/iptables/acl_linux_test.go
+++ b/client/firewall/iptables/acl_linux_test.go
@@ -1,0 +1,35 @@
+//go:build linux && !android
+
+package iptables
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/netbirdio/netbird/client/iface/wgaddr"
+	nbnet "github.com/netbirdio/netbird/client/net"
+)
+
+type aclTestIface struct{}
+
+func (aclTestIface) Name() string            { return "wt0" }
+func (aclTestIface) Address() wgaddr.Address { return wgaddr.Address{} }
+
+func TestSeedInitialEntriesUsesConnectionMarkForRedirectedTraffic(t *testing.T) {
+	m := &aclManager{entries: make(map[string][][]string), optionalEntries: make(map[string][]entry)}
+	m.wgIface = aclTestIface{}
+
+	m.seedInitialEntries()
+	m.seedInitialOptionalEntries()
+
+	mark := fmt.Sprintf("%#x", nbnet.PreroutingFwmarkRedirected)
+	routeMark := fmt.Sprintf("%#x", nbnet.PreroutingFwmarkMasquerade)
+	assert.Contains(t, m.entries[mangleFwdKey], []string{
+		"-i", "wt0", "-m", "conntrack", "--ctstate", "DNAT",
+		"-m", "connmark", "!", "--mark", mark,
+		"-m", "connmark", "!", "--mark", routeMark, "-j", "DROP",
+	})
+	assert.Equal(t, []string{"-m", "connmark", "--mark", mark, "-j", "ACCEPT"}, m.optionalEntries["FORWARD"][0].spec)
+}
--- a/client/firewall/iptables/router_linux.go
+++ b/client/firewall/iptables/router_linux.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"maps"
 	"net/netip"
-	"slices"
 	"strconv"
 	"strings"
 
@@ -537,7 +536,7 @@ func (r *router) cleanupDataPlaneMark() error {
 func (r *router) addPostroutingRules() error {
 	// First rule for outbound masquerade
 	rule1 := []string{
-		"-m", "mark", "--mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmarkMasquerade),
+		"-m", "connmark", "--mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmarkMasquerade),
 		"!", "-o", "lo",
 		"-j", routingFinalNatJump,
 	}
@@ -548,7 +547,7 @@ func (r *router) addPostroutingRules() error {
 
 	// Second rule for return traffic masquerade
 	rule2 := []string{
-		"-m", "mark", "--mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmarkMasqueradeReturn),
+		"-m", "connmark", "--mark", fmt.Sprintf("%#x", nbnet.PreroutingFwmarkMasqueradeReturn),
 		"-o", r.wgIface.Name(),
 		"-j", routingFinalNatJump,
 	}
@@ -663,7 +662,6 @@ func (r *router) cleanJumpRules() error {
 
 func (r *router) addNatRule(pair firewall.RouterPair) error {
 	ruleKey := firewall.GenKey(firewall.NatFormat, pair)
-	connmarkRuleKey := ruleKey + "-connmark"
 
 	if _, exists := r.rules[ruleKey]; exists {
 		if err := r.removeNatRule(pair); err != nil {
@@ -697,10 +695,8 @@ func (r *router) addNatRule(pair firewall.RouterPair) error {
 	rule = append(rule, sourceExp...)
 	rule = append(rule, destExp...)
 	rule = append(rule,
-		"-j", "MARK", "--set-mark", fmt.Sprintf("%#x", markValue),
+		"-j", "CONNMARK", "--set-mark", fmt.Sprintf("%#x", markValue),
 	)
-	connmarkRule := slices.Clone(rule[:len(rule)-4])
-	connmarkRule = append(connmarkRule, "-j", "CONNMARK", "--set-mark", fmt.Sprintf("%#x", markValue))
 
 	// Ensure nat rules come first, so the mark can be overwritten.
 	// Currently overwritten by the dst-type LOCAL rules for redirected traffic.
@@ -708,13 +704,8 @@ func (r *router) addNatRule(pair firewall.RouterPair) error {
 		// TODO: rollback ipset counter
 		return fmt.Errorf("error while adding marking rule for %s: %v", pair.Destination, err)
 	}
-	if err := r.iptablesClient.Insert(tableMangle, chainRTPRE, 2, connmarkRule...); err != nil {
-		_ = r.iptablesClient.DeleteIfExists(tableMangle, chainRTPRE, rule...)
-		return fmt.Errorf("error while adding connection marking rule for %s: %v", pair.Destination, err)
-	}
 
 	r.rules[ruleKey] = rule
-	r.rules[connmarkRuleKey] = connmarkRule
 
 	r.updateState()
 	return nil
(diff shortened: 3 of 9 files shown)


CoderSufiyan:
The failing `FreeBSD Port / Build & Test` job is an infrastructure failure unrelated to this PR. The VM stopped during dependency configuration while building GNU ports, including a `Segmentation fault (core dumped)`, and the job ended without completing or reporting the `Test FreeBSD port` step. This PR only changes Linux firewall files, while the Linux build and all client unit jobs passed. A maintainer rerun is needed: https://github.com/netbirdio/netbird/actions/runs/29512223426/job/87668274604

PR #6825, opened 2026-07-19
[client] lazyconn: preserve routed prefixes on idle→wake to fix subnet black-hole

MichaelUray:
## Describe your changes

When a routing peer goes idle under lazy connections, its WireGuard peer is torn
down, which also drops the routed subnet prefixes the route watcher had installed.
The activity listener then re-arms the peer's wake endpoint with only the overlay
/32 (`PeerConfig.AllowedIPs`), so traffic destined for the routed subnet no longer
matches the peer's AllowedIPs and is black-holed until the peer is woken by other
means (e.g. a ping to its overlay IP).

The asynchronous route-watcher re-add via `AddAllowedIP` uses `UpdateOnly=true`;
when it races into the window where the peer does not exist, wireguard-go treats
update-only on a missing peer as a silent no-op, so no self-heal occurs and the
subnet stays out of AllowedIPs.

## Fix

Capture each routing peer's static routed prefixes in `UpdateRouteHAMap` and merge
them into the peer's AllowedIPs at activity-listener arm time, so the wake endpoint
covers both the overlay /32 and the routed subnets. The merge is applied to a copy
while the stored `PeerConfig` keeps its pristine overlay-only base; computing the
arm-time set fresh from the current route sync prevents stale prefixes from
accumulating across route changes. This matters because WireGuard AllowedIPs are
peer-exclusive: a stale prefix would let an idle peer re-claim a subnet that has
since moved to another HA peer, breaking failover.

Prefixes are gathered for every route, including single-peer (non-HA) routes,
covering both the cold-start and route-change paths. Dynamic (domain) routes are
skipped as they have no fixed subnet to pre-install.

## Testing

Adds `manager_routewake_test.go` (unit): verifies routed prefixes are merged into
the wake endpoint at arm time for HA and single-peer routes, that the stored
PeerConfig stays overlay-only, and that stale prefixes do not persist across route
changes. `go test ./client/internal/lazyconn/... -race` green.

---
<a href="https://app.blacksmith.sh/netbirdio/codesmith/netbird/pr/6825"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787054567&installation_id=146802194&pr_number=6825&repository=netbirdio%2Fnetbird&return_to=https%3A%2F%2Fgithub.com%2Fnetbirdio%2Fnetbird%2Fpull%2F6825&signature=40d0c57db5294410ed644c096386fab78b8576c308fda3dbaf8adb707cb442ad"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wake-up behavior for routed connections by including the appropriate routed subnet information.
  * Updated wake settings correctly when routes change, preventing stale routes from being used.
  * Ensured each connection receives only its own routed networks.
  * Excluded dynamic or domain-based routes from wake-up configuration.
  * Improved cleanup when connections are removed or the manager shuts down.

* **Tests**
  * Added coverage for route changes, multiple routed connections, dynamic routes, and cleanup scenarios.

### DIFF
--- a/client/internal/lazyconn/manager/manager.go
+++ b/client/internal/lazyconn/manager/manager.go
@@ -2,6 +2,8 @@ package manager
 
 import (
 	"context"
+	"net/netip"
+	"sort"
 	"sync"
 	"time"
 
@@ -55,7 +57,15 @@ type Manager struct {
 	// If any peer in the same HA group is active, all peers in that group should prevent going idle
 	peerToHAGroups map[string][]route.HAUniqueID // peer ID -> HA groups they belong to
 	haGroupToPeers map[route.HAUniqueID][]string // HA group -> peer IDs in the group
-	routesMu       sync.RWMutex
+	// peerToRoutePrefixes holds the routed subnets each routing peer serves
+	// (from the management route sync via UpdateRouteHAMap). The activity
+	// listener installs a peer's wake endpoint via WgInterface.UpdatePeer,
+	// and WireGuard only routes a destination to that endpoint when it falls
+	// inside the peer's AllowedIPs. Merging these routed prefixes into the
+	// lazy PeerConfig.AllowedIPs ensures traffic to a routed subnet wakes the
+	// idle routing peer instead of being black-holed.
+	peerToRoutePrefixes map[string][]netip.Prefix // peer ID -> routed prefixes from route sync
+	routesMu            sync.RWMutex
 }
 
 // NewManager creates a new lazy connection manager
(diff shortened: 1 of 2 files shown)


riccardomanfrin:
Hi @MichaelUray , we are closing this one but you can find the fix #6863

Thanks for the support

PR #6829, opened 2026-07-19
[client] Preserve routed allowed IPs across lazy connection switch

pappz:
## Describe your changes

When a routing peer goes idle under lazy connections, the WireGuard peer is removed and re-created as a wake endpoint. The route manager deliberately keeps routed prefixes installed on idle peers, but its asynchronous re-add uses update_only, which is a silent no-op while the peer is absent. When the re-add races into the removal/re-arm window, the routed subnet is left out of the peer's allowed IPs — and since the peer's status does not change again, nothing retries the add: traffic to the subnet is black-holed until the peer is woken by other means (e.g. a ping to its overlay IP).

Split the idle transition from the full close: Conn.Idle tears down transports, cancels pending delayed endpoint updates and updates the status without touching the WireGuard peer. The activity listener then arms the wake endpoint via the new IdlePeerEndpoint operation, which removes and re-creates the peer in a single transaction: handshake state is dropped (keeping the handshake-first wake semantics of packet staging and first-packet capture/reinjection) while the currently installed allowed IPs are preserved. The read-modify-write runs under the interface mutex, serializing it against concurrent allowed IP updates from the route manager — a concurrent add either lands before the read (included in the re-created set) or after the transaction (on the existing peer), so the race window is gone by construction.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

---
<a href="https://app.blacksmith.sh/netbirdio/codesmith/netbird/pr/6829"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787082043&installation_id=146802194&pr_number=6829&repository=netbirdio%2Fnetbird&return_to=https%3A%2F%2Fgithub.com%2Fnetbirdio%2Fnetbird%2Fpull%2F6829&signature=0b721db25e04e4d7d509a430ba02c9a88f3c1f890a68f20b1aecb806633cfdb9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved lazy connection handling by switching peers to a wake (“idle”) endpoint while preserving configured allowed IPs/routes.
  * Automatically recreates peers when they’re missing during lazy connection setup.
  * Idle behavior now retains the existing peer configuration to support faster reconnection.

* **Bug Fixes**
  * Prevents scheduled/queued endpoint updates from applying after a peer becomes idle.
  * Refined close/idle transitions and cleanup to ensure consistent peer endpoint behavior.

### DIFF
--- a/client/iface/configurer/common.go
+++ b/client/iface/configurer/common.go
@@ -3,6 +3,7 @@ package configurer
 import (
 	"net"
 	"net/netip"
+	"time"
 
 	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
 )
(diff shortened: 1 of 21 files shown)


riccardomanfrin (commented):
The allowed IPs currently live in two places:
  - peerCfg.AllowedIPs, owned by Conn
  - allowedIPsRefCounter, owned by the route manager

To fix the allowed IPs when the peer becomes active you need the union of the two. The routed side is already tracked in allowedIPsRefCounter and is authoritative — so instead of reading the current set back  from the device and merging (IdlePeerEndpoint doing IpcGet/getPeer + mergePrefixes), why not do a declarative reconcile when the WireGuard tunnel / peer is activated?

We can keep allowedIPsRefCounter maintained while the peer is idle (tunnel down) and apply the full allowed-IP set as a whole on setup. That removes the device readback and the dependency on remove+add ordering — the desired set is reapplied from the authoritative source at activation, so the race is gone by construction rather than by locking a narrow window.

*** This is orthogonal to dropping the handshake state on idle — that can stay a separate concern; here I'm only talking about how the allowed-IP set is restored.

PR #6912, opened 2026-07-27
[client] Fall back to the handler chain for DNS route query types a peer cannot resolve

lixmal:
## Describe your changes

A DNS route forwarded every query type to its routing peer and returned whatever came back, so a record type the peer's forwarder cannot resolve failed hard instead of being answered elsewhere. Peers running a client that predates non-address record support answer NOTIMP for all of them, which breaks SRV-based service discovery through a route.

- Ask the routing peer only for the record types a forwarder can resolve, and send the rest straight to the remaining DNS handlers
- Fall back to those handlers as well when a peer reports it cannot resolve a type, so a route served by an older peer keeps working
- Answer a routed name whose type we could not resolve with an empty NOERROR carrying no SOA, so it is not cacheable and cannot outlive the route, rather than an NXDOMAIN that would wipe out the addresses the route does serve or a REFUSED that would send the client looking for one
- Record a deferred query and a rewritten verdict on the response log line, and report the rewrite to the client as an Extended DNS Error
- Set the recursion-available bit on DNS forwarder replies

https://github.com/netbirdio/netbird/discussions/6881

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Restores the DNS resolution behaviour users already expect from a DNS route. No configuration, flags or user-facing options change.

## Summary by CodeRabbit

* **New Features**
  * Enhanced DNS interception and routing with qtype-based disposition and clearer handling of capability/probe failures.
  * Extended DNS Error (EDE) now includes richer, stable upstream and softened-negative details; EDNS0-aware behavior is improved and OPT is stripped when not advertised.
  * Forwarder responses now correctly report recursion availability, with typed, deterministic response annotations.
* **Bug Fixes**
  * Prevented unsupported queries from incorrectly surfacing NXDOMAIN by downgrading softened negatives into empty NOERROR responses when appropriate.
  * Deferred queries now preserve the client’s original EDNS0/AD/OPT state, and authoritative downstream responses remain unchanged.
* **Tests**
  * Added coverage for soft-negative behavior, recursion-available responses, and pristine deferred queries.

### DIFF
--- a/client/internal/routemanager/dnsinterceptor/handler.go
+++ b/client/internal/routemanager/dnsinterceptor/handler.go
@@ -22,7 +22,6 @@ import (
 	nbdns "github.com/netbirdio/netbird/client/internal/dns"
 	"github.com/netbirdio/netbird/client/internal/dns/resutil"
 	"github.com/netbirdio/netbird/client/internal/peer"
-	"github.com/netbirdio/netbird/client/internal/peerstore"
 	"github.com/netbirdio/netbird/client/internal/routemanager/common"
 	"github.com/netbirdio/netbird/client/internal/routemanager/fakeip"
 	iface "github.com/netbirdio/netbird/client/internal/routemanager/iface"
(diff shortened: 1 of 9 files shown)


client/internal/routemanager/dnsinterceptor/handler_test.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(dns.TypeToString[tt.qtype], func(t *testing.T) {
+			assert.Equal(t, tt.want, dispositionFor(tt.qtype), "disposition for %s", dns.TypeToString[tt.qtype])
+
+			if tt.want != dispositionPeer {
+				assert.Equal(t, tt.want == dispositionPeerFirst, resutil.SupportedRecordQtype(tt.qtype),
+					"only types a forwarder resolves may be sent to a peer")
+			}
```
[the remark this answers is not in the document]
lixmal: Not applying this one: `SupportedRecordQtype` deliberately excludes A/AAAA. Address queries are resolved by the forwarder's address path (`handleAddressQuery` → `LookupIP`), not by `LookupRecords`, so the function returns false for them by design and `dispositionFor` decides them before it is consulted.

The suggested diff fails:

```
--- FAIL: TestDispositionFor/A
        expected: true
        actual  : false
--- FAIL: TestDispositionFor/AAAA
        expected: true
        actual  : false
```

The underlying point is fair, so the test now asserts the relationship in both directions instead of skipping address types: they must be absent from `SupportedRecordQtype` (proving they take the address path), and every other peer-bound type must be present.

client/internal/routemanager/dnsinterceptor/handler.go
riccardomanfrin, line 297: Might be deferred in the chain as well?

client/internal/routemanager/dnsinterceptor/handler.go
```diff
@@ -246,42 +303,63 @@ func (d *DnsInterceptor) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
 		return
```
riccardomanfrin, line 303: same here: Might be deferred in the chain as well?

client/internal/routemanager/dnsinterceptor/handler.go
```diff
@@ -246,42 +303,63 @@ func (d *DnsInterceptor) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
...
+	query, hadEdns := peerQuery(r)
 
 	upstream := net.JoinHostPort(upstreamIP.String(), strconv.FormatUint(uint64(d.forwarderPort.Load()), 10))
 	ctx, cancel := context.WithTimeout(context.Background(), dnsTimeout)
 	defer cancel()
 
-	reply := d.queryUpstreamDNS(ctx, w, r, upstream, upstreamIP, peerKey, logger)
+	reply := d.queryUpstreamDNS(ctx, w, query, upstream, upstreamIP, peerKey, logger)
 	if reply == nil {
 		return
```
riccardomanfrin, line 315: worth a log?

riccardomanfrin (dismissed):
all good afaik just couple comments to validate

client/internal/routemanager/dnsinterceptor/handler.go
lixmal: Nope, a public record from the chain could poison the client's cache

client/internal/routemanager/dnsinterceptor/handler.go
lixmal: Same as above

client/internal/routemanager/dnsinterceptor/handler.go
lixmal: Every nil return inside `queryUpstreamDNS` already logs before returning (client-creation failure via `writeDNSError`, timeout and exchange failures at error level) and writes SERVFAIL to the client, so a log here would duplicate it. Added a comment at the call site to make that visible: 473d61afc

PR #6976, opened 2026-07-30
[client] Add a Reconnect entry to the tray menu

atj393:
## Describe your changes

When the client goes stale after a sleep/wake, a network roam, or peers reading connected while no traffic flows, the remedy is always Disconnect, wait, Connect. The tray exposes only the two halves, so that is four interactions plus a judgement call about when the state has settled, and it is awkward to talk a non-technical colleague through over the phone.

This adds a Reconnect entry that does the round trip in one click. It shares Disconnect's visibility rules, since a reconnect with no session to tear down is just Connect.

`Connection.Reconnect` owns the ordering. Down is synchronous, so Up follows only once the daemon reports the teardown finished, and a failed Down aborts rather than stacking an Up on a session that is still up. It takes a context so Disconnect and quit can cancel a round trip mid-flight; without that the queued Up lands after the user has explicitly asked to go down, and brings the session back.

The transitional feedback reuses `DaemonFeed.BeginProfileSwitch` instead of adding a "Reconnecting..." state. A profile switch is already a Down then Up, so the machinery exists: an optimistic Connecting, suppression of the stale blink during teardown, and a login-watch that opens browser-login when the Up lands in NeedsLogin. A parallel state would have to fight the daemon's own status pushes. The name is the part that does not fit two callers; `BeginTransition`/`CancelTransition` would read better and I will rename here if you want it.

Two decisions that the diff cannot explain for itself. A click during an in-flight profile switch is ignored rather than cancelling the switch, because cancelling one midway can leave the daemon and the CLI's on-disk profile state disagreeing. And the in-flight guard is a cancel func rather than the menu item's enabled state, because `buildMenu` recreates every item on relayout, and the status pushes a reconnect produces are what trigger those relayouts.

Discussion: https://github.com/netbirdio/netbird/discussions/6943

No agreed issue exists. The discussion has had no reply since 28 July, so this PR does not meet the ticket-first bar and I am not asking for it to be merged as it stands. It is here as a concrete proposal to react to: validate the discussion and I will follow whatever approach you settle on, or close this and I will take no offence.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

Two boxes left unticked deliberately. The local-testing box: `gofmt`, `go build`, `go vet`, `go test` and `golangci-lint run` all pass over `./client/ui/...` on Windows, but the entry itself has never been clicked, since I have no Wails toolchain or running daemon on this machine, so I will not claim to have tested the feature. The agreement box: covered above.

## Documentation
Select exactly one:

A tray entry with no new setting, flag, or API. If the docs enumerate the tray menu somewhere I missed, point me at it and I will open the matching netbirdio/docs PR.

n/a

## Summary by CodeRabbit

- **New Features**
  - Added a **Reconnect** option to the system tray menu.
  - Reconnects the app by safely disconnecting and reconnecting the service.
  - Prevents duplicate reconnect attempts and coordinates reconnection with disconnecting, quitting, and profile switching.
  - Added localized reconnect and reconnection-failure messages across supported languages.

- **Bug Fixes**
  - Improved recovery when reconnection fails, including clearer status updates and error handling.


atj393:
Thanks @coderabbitai, all four are addressed. Details so a reviewer does not have to diff for them:

**Serialising cancellation with a caller's own Down** (`tray.go`, in c057ab7c). This was the real one. Cancelling the context does not unsend an `Up` already inside its RPC, so Disconnect, quit and a profile switch could each have their `Down` overtaken and leave the session up right after the user asked for it to go down. `beginReconnect` now publishes a done channel that `endReconnect` closes, and the three callers wait on it, bounded at two seconds. The close happens before the reconnect goroutine repaints: quit waits on the UI thread, and a relayout needs that same thread, so closing after the repaint would have deadlocked quit until the timeout on every quit.

I did not add the interleaving test. The ordering guarantee lives in the tray, and `client/ui` has no harness that can drive `handleDisconnect` against a fake daemon, so the test would have to assert against `Connection`, which is the wrong layer and would pass whether or not the tray waits. `TestConnectionReconnectCancelDuringDownSkipsUp` still covers the part that does belong there.

**Reconnect in flight when a profile switch starts** (`tray_profiles.go`, in 4d489bd2). Correct, and a genuine asymmetry: the guard only ran the other way round. `switchProfile` now cancels the reconnect and waits for it, so the switch's own Down and Up cannot race a second Up.

**Unbounded status refresh** (`tray.go`, in 4d489bd2). Correct on both counts. Now bounded by a five second context, with the error logged at debug rather than dropped.

**`require` vs `assert` in the reconnect tests** (`connection_test.go`, in 4d489bd2). Correct per the testing guidance. Outcome assertions moved to `assert` with context messages on the call-sequence comparisons.

PR #6978, opened 2026-07-30
[client] Fix daemon lock order inversion between SetConfig and login

lixmal:
## Describe your changes

`SetConfig` and the login path acquired the daemon's server mutex and the guarded-config mutex in opposite orders, so two concurrent daemon requests could wedge each other permanently. `SetConfig` now takes the guarded-config mutex first, matching the login path, which gives the two locks a single order across the daemon.

- Fix the lock order inversion between `SetConfig` and login by acquiring the guarded-config lock before the server lock
- Add a test that fails on the old order

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal locking fix with no user-visible surface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration updates to prevent potential deadlocks during concurrent login and configuration operations.
  * Preserved existing validation and configuration update behavior.

* **Tests**
  * Added coverage to verify safe lock ordering during configuration changes.

### DIFF
--- a/client/server/lock_order_test.go
+++ b/client/server/lock_order_test.go
@@ -1,0 +1,51 @@
+package server
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/netbirdio/netbird/client/proto"
+)
+
+// The daemon takes guardedConfigMu before s.mutex. authorizeAndPrepareLogin
+// takes s.mutex while holding guardedConfigMu, so a SetConfig that grabbed
+// s.mutex first and then waited for guardedConfigMu would deadlock the daemon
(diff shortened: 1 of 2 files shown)


riccardomanfrin (approved):
guardedConfigMu covers also checkUpdateSettingsDisabled()  and loadMDMPolicy() which read from system so login takes lock, SetConfig will take more time.

Test covers only one direction / order (thouth it's the canonical one). might be good to do another with inverted direction.

SetConfig is stale if t.Fatal fires (line 48) but it's terminating anyway so I guess it's fine.. 

This PR probably needs backport on 0.76.1 ?

PR #7011, opened 2026-07-31
[client] Fix expression order in legacy nftables route rules

Optic00:
## Describe your changes

On the retained compatibility path used with management servers older than v0.30.0, routing peers using the legacy nftables route implementation installed a forwarding rule whose expressions were assembled in the wrong order. In `addLegacyRouteRule` the `Counter` and `Verdict{Kind: VerdictAccept}` expressions were appended to the rule **before** the source and destination match expressions produced by `applyNetwork`.

nftables evaluates the expressions of a rule sequentially and stops at the first terminal verdict. With the accept verdict placed first, a packet reaching the generated rule was accepted immediately, and the source and destination matches that followed were never evaluated. This PR is scoped only to correcting that legacy compatibility path.

**Fix**

The expression slice is now built as `source matches -> destination matches -> counter -> accept verdict`, so the verdict is only reached once all match expressions have passed. The assembly was extracted into a small helper, `buildLegacyRouteRuleExpressions`, which keeps `addLegacyRouteRule` unchanged in behavior other than the ordering and makes the ordering directly testable.

**Test**

`TestBuildLegacyRouteRuleExpressions` is a table-driven unit test covering non-empty source and destination expressions as well as the empty source, empty destination, and both-empty cases. It asserts that the caller supplied match expressions come first and in their original order (compared by identity), and that they are followed by exactly a `Counter` and an accept `Verdict`. The test fails against the previous ordering.

The targeted regression test and the complete `client/firewall/nftables` package test suite pass on Linux amd64.

N/A - found during static review. The affected code is the compatibility path for management servers older than v0.30.0.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor
- [x] This change does **not** modify the public API, gRPC protocols, functionality behavior, CLI / service flags, or introduce a new feature - OR I have discussed it with the NetBird team beforehand (link the issue / Slack thread in the description). See [CONTRIBUTING.md](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTING.md#discuss-changes-with-the-netbird-team-first).

## Documentation
Select exactly one:

- [x] Documentation is **not needed** for this change (explain why): the change corrects the internal ordering of nftables rule expressions so the existing behavior is actually applied. No user-facing configuration, CLI or service flags, gRPC protocols, or public APIs change.

Paste the link from https://github.com/netbirdio/docs here:

n/a

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved legacy firewall route rule construction to consistently apply source and destination matches, counters, and accept actions.
  * Added coverage for routes with missing or optional source and destination criteria.

### DIFF
--- a/client/firewall/nftables/legacy_rule_linux_test.go
+++ b/client/firewall/nftables/legacy_rule_linux_test.go
@@ -1,0 +1,60 @@
+package nftables
+
+import (
+	"testing"
+
+	"github.com/google/nftables/expr"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuildLegacyRouteRuleExpressions(t *testing.T) {
+	sourcePayload := &expr.Payload{}
+	sourceCmp := &expr.Cmp{}
+	destinationPayload := &expr.Payload{}
+	destinationCmp := &expr.Cmp{}
+	nilSourceDestination := &expr.Payload{}
+	nilDestinationSource := &expr.Cmp{}
+
+	tests := []struct {
+		name        string
+		source      []expr.Any
+		destination []expr.Any
+		matches     []expr.Any
+	}{
+		{
+			name:        "both non-empty",
+			source:      []expr.Any{sourcePayload, sourceCmp},
+			destination: []expr.Any{destinationPayload, destinationCmp},
+			matches:     []expr.Any{sourcePayload, sourceCmp, destinationPayload, destinationCmp},
+		},
+		{
+			name:        "nil source",
+			destination: []expr.Any{nilSourceDestination},
+			matches:     []expr.Any{nilSourceDestination},
+		},
+		{
+			name:    "nil destination",
+			source:  []expr.Any{nilDestinationSource},
+			matches: []expr.Any{nilDestinationSource},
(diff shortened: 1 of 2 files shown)


lixmal:
Hi @Optic00, thank you for the contribution.

The issue is real, but the code path runs only when a client is connected to a very old management server (< v0.30.0). So I'm curious how you discovered the issue and whether you're running a version that old.

The linked gh issue is completely unrelated (the rule signatures show that these are post-legacy rules)

Optic00:
Thanks for the quick review.

To answer directly: no, I am not running a management server that old. I found this during a static review of the client firewall code with a security and hardening focus, not from a live incident. In the legacy path, the terminal accept verdict is placed before the source and destination match expressions, so the rule accepts more than intended when that path is used. I isolated that and added a focused regression test that asserts the expression order.

You are also right about #6731. The rule signatures there are post-legacy, so my attribution was wrong. I have removed that reference and narrowed the PR description to the retained legacy compatibility path.

lixmal (approved):

PR #7043, opened 2026-08-03
[client] disambiguate the connection_type metric tag

riccardomanfrin:
## Describe your changes

`recordConnectionMetrics` mapped only `conntype.Relay` to `relay` and let a `default` branch
record everything else as `ice`. That silently included `ICETurn` — an ICE connection through
a TURN server, which [`conn.isRelayed`](https://github.com/netbirdio/netbird/blob/main/client/internal/peer/conn.go#L788-L795) itself counts as relayed — and `None`, the transient state set when the relay drops
([conn.go:632](https://github.com/netbirdio/netbird/blob/main/client/internal/peer/conn.go#L632)) or the peer state is reset ([conn.go:757](https://github.com/netbirdio/netbird/blob/main/client/internal/peer/conn.go#L757)).
Both were reported as direct peer-to-peer, so the `ice` share overstated direct connections on
every platform.

The mapping now lists every priority explicitly and emits `ice_p2p`, `ice_turn`, `relay` or
`unknown`. The new values deliberately do not reuse `ice` to avoid ambuguity.

No public issue. Found while reviewing the first production sample of client metrics: 38% of iOS
connection events were tagged `ice` on a platform that forces relay by default, which traced back
to the `default` branch at [client/internal/peer/conn.go#L963-L968](https://github.com/netbirdio/netbird/blob/main/client/internal/peer/conn.go#L963-L968).

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

- [] I added/updated documentation for this change

Internal metrics documentation only, in `client/internal/metrics/infra/README.md`: the four
`connection_type` values with their derivation, and a note that pre-fix `ice` samples are not
comparable with `ice_p2p`. No public API, CLI or configuration change, so no netbirdio/docs PR.

N/A

## Summary by CodeRabbit

* **New Features**
  * Connection metrics now distinguish direct peer-to-peer, TURN-assisted, relay, and unknown connection types.
  * Metrics include clearer connection and peer identification details.

* **Documentation**
  * Updated connection timing metric values, traffic semantics, priority behavior, and historical data guidance.

* **Bug Fixes**
  * Unset or unrecognized connection priorities are no longer incorrectly classified as peer-to-peer.
  * Unknown-transport metrics are skipped to prevent misleading connection data.

### DIFF
--- a/client/internal/peer/conn.go
+++ b/client/internal/peer/conn.go
@@ -959,12 +959,10 @@ func (conn *Conn) recordConnectionMetrics() {
 	priority := conn.currentConnPriority
 	conn.mu.Unlock()
 
-	var connType metrics.ConnectionType
-	switch priority {
-	case conntype.Relay:
-		connType = metrics.ConnectionTypeRelay
-	default:
-		connType = metrics.ConnectionTypeICE
+	connType := metricsConnType(priority)
+	if connType == metrics.ConnectionTypeUnknown {
+		conn.Log.Debugf("skip connection metrics, no transport is active (priority %s)", priority)
+		return
 	}
 
 	// Record metrics with timestamps - duration calculation happens in metrics package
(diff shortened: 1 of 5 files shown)


client/internal/peer/conn.go
```diff
@@ -959,12 +959,10 @@ func (conn *Conn) recordConnectionMetrics() {
...
 
-	var connType metrics.ConnectionType
-	switch priority {
-	case conntype.Relay:
-		connType = metrics.ConnectionTypeRelay
-	default:
-		connType = metrics.ConnectionTypeICE
+	connType := metricsConnType(priority)
+	if connType == metrics.ConnectionTypeUnknown {
+		conn.Log.Debugf("skip connection metrics, no transport is active (priority %s)", priority)
```
pappz: Do not print this out. Just more noise in the debug logs. Fine if we lose sample from the measurement.

pappz (changes requested):

pappz (approved):

PR #7046, opened 2026-08-03
fix(client): Windows dial fallback across route candidate interfaces

siraykut:
## Describe your changes

On dual-homed Windows hosts, advanced routing (`IP_UNICAST_IF`) can pin management/signal gRPC dials to a metric-preferred NIC that has a default route but no working upstream. The dial then times out and never tries another interface that can reach the management server.

This change:
- Adds `GetCandidateInterfaces` (all usable interfaces, best-first)
- Retries Windows advanced-routing dials across candidates with a split timeout budget
- Logs when a fallback interface succeeds

https://github.com/netbirdio/netbird/discussions/7045

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactoring
- [x] Added unit tests for timeout splitting / interface selection helpers

## Documentation
- [x] Not needed

## Summary by CodeRabbit

* **New Features**
  * Improved Windows network routing by selecting and trying multiple suitable interfaces.
  * Added interface-aware IPv4 and IPv6 connection attempts.
  * Added support for resolving destinations before selecting connection interfaces.
  * Preserved fallback behavior when advanced routing is unavailable.
* **Bug Fixes**
  * Improved handling of unavailable, loopback, inactive, excluded, and duplicate interfaces.
  * Preserved connection timeouts, cancellation, and error reporting.
* **Tests**
  * Added coverage for timeout handling, deadlines, and IP version interface selection.

### DIFF
--- a/client/internal/routemanager/systemops/systemops_windows.go
+++ b/client/internal/routemanager/systemops/systemops_windows.go
@@ -25,6 +25,7 @@ import (
 
 func init() {
 	nbnet.GetBestInterfaceFunc = GetBestInterface
+	nbnet.GetCandidateInterfacesFunc = GetCandidateInterfaces
 }
 
 const (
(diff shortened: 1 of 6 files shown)


siraykut:
Addressed review feedback:

- DNS resolution in the fallback path now uses the caller `ctx` + dialer resolver
- Fallback-success Info log no longer includes hostname/address
- `dialAdvanced` stub build tag now includes iOS (`!windows`)
- Timeout split test uses `InDelta`; assertions have context messages

lixmal:
This doesn't seem right. NetBird selects the same interface as a regular route lookup would with a small difference in how we take metrics (route + interface) into account. So I would rather take a look at a debug bundle or the list of routes/interfaces (did you set any route metric manually?) to rule out any bugs there.

siraykut:
Thanks for taking a look.

Just to clarify: when the issue first appeared, **no route/interface metrics were set manually**. At that time both NICs were up with default routes:

- **Ethernet** (`192.168.68.51`) — interface metric **25** (preferred)
- **Wi-Fi** hotspot (`172.20.10.4`) — interface metric **30**

Probes from the same machine:

- `curl --interface <ethernet-ip> https://<mgmt>:443` → **timeout**
- `curl --interface <wifi-ip> https://<mgmt>:443` → **HTTP 200**
- User-space HTTPS / `Find-NetRoute` for the management IP went via **Wi-Fi**
- NetBird (`nbnet` + `IP_UNICAST_IF` / `GetBestInterface`) kept dialing via **Ethernet** and hit `context deadline exceeded`

I only lowered the Wi-Fi metric afterward as a temporary workaround so the client could connect again.

Happy to provide a debug bundle and the full route/interface dump from a repro (with metrics left at defaults). If you’d rather fix this in route selection (why the metric-preferred NIC is chosen when Windows itself routes that destination via the other iface) instead of a dial fallback, I’m fine reshaping the PR that way.

lixmal:
Yes, please provide a bundle (key), this will confirm whether the lookup logic is wrong

PR #7066, opened 2026-08-05
[client] Ask the OS for privileges when a guarded SSH setting is changed

lixmal:
## Describe your changes

The daemon restricts the SSH settings that hand out a shell on the host — running the SSH server, root login, turning SSH authentication off, and repointing a profile at another management server while the SSH server runs — to root/administrator. Until now the desktop app answered that by greying the control out and printing a `sudo netbird down; sudo netbird up …` line for the user to copy. This makes the app ask the operating system for the privileges instead, at the moment the user asks for the change: the UAC consent dialog on Windows, the authentication dialog on macOS, the polkit agent on Linux. Any credentials are collected by the operating system and never pass through NetBird.

- Ask for privileges through the platform's own prompt when an unprivileged user changes a guarded SSH setting, and apply the change if they authorize it
- Recover from a refused change to the management URL the same way, since that one is only privileged while the host runs the SSH server, which no control can know before it is refused
- Fall back to the previous behaviour — the control explains itself and offers the command — wherever there is no prompt to raise, such as a session with no polkit agent or a daemon reachable only over loopback TCP, where it refuses these changes to everyone
- Treat a dismissed prompt as an answer rather than a failure: the control reverts and nothing is reported
- Ship a polkit action so the Linux dialog names NetBird and what it is about instead of a command line, and require authentication for each change rather than caching it
- Refuse to elevate an executable, or one reachable through a directory, that any account which could not answer the prompt itself is able to write
- Ask only for the guarded settings a request actually changes, so one authorization cannot apply a value the request merely restates
- Translate the new strings, and the privilege hints that shipped in 0.76.0, into all shipped languages; the principal the daemon requires is now named by the frontend so it reads naturally in each

The privileges are obtained by the app running itself once, elevated, in a mode that sends the same request the window would have sent. That mode re-implements a little of what `netbird up` does, which is deliberate for now: `netbird up` restarts the connection to apply a config flag, so using it here would drop the tunnel to flip a checkbox, and the CLI has no non-disruptive equivalent yet. Running our own executable also means the elevated code is exactly this build, at a path we know without searching, which matters where the CLI and the app are installed separately and can be different versions. If a non-disruptive `netbird set-config` is added later, pointing the prompt at the CLI instead becomes a small change.

macos
linux
windows

Follow-up to #6967, which introduced the restrictions this makes usable from the GUI.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

The settings and what they require are unchanged; only how the app asks for the privileges to change them differs, and the prompt explains itself.

## Summary by CodeRabbit

* **New Features**
  * Added secure privilege elevation for settings changes across macOS, Linux, and Windows.
  * Added administrator authorization prompts and clear handling for declined or unavailable elevation.
  * Added guarded SSH and management settings updates with retry support.
  * Added PolicyKit authorization for Linux settings changes.
  * Improved identity-aware daemon connections and platform-specific privilege labeling.

* **Bug Fixes**
  * Improved user, group, and shell resolution across supported platforms.
  * Updated desktop branding to consistently display “NetBird.”
  * Added localized messages for elevation and SSH privilege workflows.


client/internal/elevate/run_darwin.go
```diff
@@ -0,0 +1,355 @@
...
+			return out.String(), err
+		}
+		n, err := syscall.Read(fd, buf)
+		if n > 0 {
+			out.Write(buf[:n])
+		}
+		if n <= 0 || err != nil {
+			return out.String(), nil
+		}
+	}
```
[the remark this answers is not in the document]
lixmal: Fixed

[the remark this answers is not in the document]
lixmal: Done in c4d01f0 — `readPipe` now `continue`s on `EINTR` and logs any other read error at debug before ending the loop. The marker check moved into `checkApplied`, which `TestCheckApplied` covers for empty and non-marker output.

client/ui/services/oneshot.go
```diff
@@ -0,0 +1,233 @@
+//go:build !android && !ios && !freebsd && !js
```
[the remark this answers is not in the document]
lixmal: Settled in 1ea9405 by narrowing the package to match its only consumer: `run_unix.go` is now `//go:build linux`, and FreeBSD falls to `run_unsupported.go`. `client/ui` excludes FreeBSD in every file, so the pkexec path there was unreachable. The stub's doc says pkexec is the mechanism to reinstate if a FreeBSD UI ever ships. The PR description no longer claims FreeBSD.

client/internal/elevate/run_windows.go
```diff
@@ -0,0 +1,193 @@
...
+	seeMaskFlagNoUI = 0x00000400
+
+	// swHide: the one-shot has no window to show.
+	swHide = 0
+
+	// sFalse (S_FALSE) answers CoInitializeEx when COM is already up on this
+	// thread in the mode we asked for; rpcChangedMode (RPC_E_CHANGED_MODE) when
+	// it is up in the other one.
+	sFalse         = 1
+	rpcChangedMode = 0x80010106
```
theodorsm: Can be replaced with `RPC_E_CHANGED_MODE` from the `x/sys/windows` package.

pappz (changes requested):
I would like to draw your attention to the fact that we use GoReleaser instead of the Wails Task releaser. Because of this, we do not maintain the Task-related package generation steps, which makes it easier to follow the changes from upstream.

- This why we do not copy netbird.png
- Modify client/ui/build/linux/nfpm/nfpm.yaml have no effects

Maybe would be better to remove these tasks because we are sure now we want to keep the goreleaser.

lixmal:
@pappz thanks for the hint, I reverted the files

client/internal/elevate/run_darwin.go
theodorsm, line 1: The MacOS implementation seems weak currently (given such a long justification in the comment) I would rather not ship this deprecated API with purego. Let's do do it properly by passing `AuthorizationExternalForm` to the daemon in another PR and have macos just give the user the sudo command to run for now. WDYT?

client/internal/elevate/trusted_unix.go
theodorsm, line 1: This group/user resolution seem a bit weak considering would not resolve groups properly (LDAP/AD) without cgo using `LookupGroupId` and `LookupId`. Reading `/etc/group` also bypasses NSS.

We should refactor the getent functions in our `ssh/server` package and offer it as a public package that we can use here, we are getting multiple uses for this code, and it resolves properly.

theodorsm (changes requested):
Left some comments, yet to try out the new ux on different platforms.

theodorsm:
We should also ask to elevate with pkexec when running the regular cli commands on Linux and not just with ui, so we also get prompted to elevate. This falls outside of this PR, but an idea for a new PR.

client/internal/elevate/trusted_unix.go
lixmal: Fixed

client/internal/elevate/run_windows.go
lixmal: Fixed

client/internal/elevate/run_darwin.go
lixmal: ▎ For the time being I'd like to keep the security boundary where it is. The deprecation is already handled at runtime: every symbol is resolved with an error rather than a panic, so a macOS that removes the API falls back to what we have today, showing the sudo command.

client/internal/getent/unix.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+	}
+	return members
+}
+
+// groupMembersFromFile finds the group's member list in a file of /etc/group's
+// format. A group the file does not describe, because it comes from LDAP or
+// another NSS source, is an error rather than an empty list.
+func groupMembersFromFile(path, name string) ([]string, error) {
+	file, err := os.Open(path)
```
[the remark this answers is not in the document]
lixmal: The guideline is about opening a caller-supplied path as root. groupMembersFromFile only ever opens the constant groupFile (/etc/group); the caller-controlled group name is compared against parsed entries, never used as a path. An unreadable file already returns an error, and the caller treats that as unverifiable.

theodorsm:
@lixmal can you resolve the conflicts?

theodorsm (approved):
Tested on Linux and Windows, both with local users and LDAP/AD users. Works as expected.

lixmal:
@pappz I think you need to approve too before I can merge

pappz (approved):

PR #7070, opened 2026-08-05
[client] Localize daemon notifications via stable message keys

lixmal:
## Describe your changes

Notifications the daemon publishes reached the desktop UI as fixed English prose, so
they stayed English no matter which of the ten shipped languages the user had
selected. The daemon has no notion of the user's locale, so it now names the string
instead of writing it: `SystemEvent` carries a stable `messageKey`, its
`messageArgs`, and an optional `titleKey`, and each UI resolves them against its own
translation bundle. `userMessage` stays populated as the English fallback for the
CLI and for UIs that predate the field.

Two notifications already worked around this by hand, with the daemon attaching
structured metadata and the tray reverse-engineering a localized string from it.
Both now go through the same mechanism, which removes the special cases.

- Add `messageKey`, `messageArgs` and `titleKey` to `SystemEvent`, with the shared
  key catalog and English fallbacks in `client/proto`
- Publish keys instead of English text from the DNS health, exit-node, auto-update,
  panic, MDM and session-expiry events
- Resolve keys in the tray against the active locale, falling back to the daemon's
  English when a key is unknown, so a newer daemon degrades instead of showing a
  raw key
- Localize the notification title, which was previously built from the raw severity
  and category enum names
- Drop the hand-rolled MDM and session-warning paths now that both are ordinary
  localized events
- Translate the new keys for all ten shipped locales

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

The change is internal to the client: the same notifications fire at the same times,
only in the user's own language. `client/ui/i18n/TRANSLATING.md`, the brief
translators work from, is updated in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added localized system-event messages with translated titles, severity, categories, and dynamic details.
  * Session-expiration warnings now show remaining time or expiration status in localized notifications.
  * Added localized notifications for DNS, Exit Node, service errors, policy changes, and automatic updates.
  * Added translations across supported languages, with English fallback text when a translation is unavailable.

* **Bug Fixes**
  * Improved event title and message fallback behavior.
  * Refined configuration-change handling to avoid redundant notifications.


pappz (changes requested):
Here have a conceptual mismatch.

In the original concept: 
The daemon does not send text, it describes an event: category, severity, and structured data about what happened (deadline timestamp, rejection reason)  That is it. Anything else the UI's business, because only it knows who is looking and with which settings.

This patter is the expected, other one are deprecated and we should step back into that direction.

In this sultion the times will not be translated well. 

So the proposal is not to move `HumaniseDuration` to the UI, it is to change what travels: an event code plus typed data, rather than a message key plus rendere text. 
In the simple events nothing changes in practice, there the event and the sentence is the key. For the session warning it preserves what the metadata-based path already had: live countdown, locale-correct formatting, and the variant choice on the UI side.

Send event codes instead of localisation keys. Localisation keys will be untraceable in the long term if we use them everywhere in the codebase. Just because have an event it does means in the UI always will be title and texts. Maybe we will change on it. And it should be handled in the UI only.

PR #7073, opened 2026-08-05
[infrastructure] Add Docker subnet override, conflict check, and enterprise domain alias to the self-hosted install scripts

TechHutTV:
## Describe your changes

A few fixes for the self-hosted install scripts: `getting-started.sh`, `getting-started-enterprise.sh`, and `migrate.sh`. All of this came out of reviewing the host-gateway hairpin NAT proposal in netbirdio/docs#626. Nothing is broken today, these prevent potential issues.

**1. `NETBIRD_DOCKER_SUBNET` override.** The compose network subnet was hardcoded to `172.30.0.0/24`, so if you already had something on that range you were stuck. It's overridable now, validated as a unicast IPv4 /24 before any prompts run. `0/8`, `127/8`, `169.254/16` and `224+` get rejected, and so does `100.64/10`. That last one is worth explaining: NetBird hands out overlay peer addresses from `100.64/10` by default, so a Docker bridge sitting there shadows the overlay without any Docker network actually overlapping. The conflict check below can't catch that, so the validator has to.

Traefik's static IP (`.10`) and the gateway (`.1`) are derived from whatever subnet you pick, which keeps everything depending on that address in step. What that covers differs per script:

- `getting-started.sh`: the compose `subnet` and `gateway`, Traefik's `ipv4_address`, the `reverseProxy.trustedHTTPProxies` `/32` pin in `config.yaml`, and `NB_PROXY_TRUSTED_PROXIES` in `proxy.env`.
- `getting-started-enterprise.sh`: the same, plus `reverseProxy.trustedPeers`, passed into compose as `NETBIRD_NETWORK_SUBNET` and `NETBIRD_NETWORK_GATEWAY` in `.env`.
- `migrate.sh`: only the compose `subnet`/`gateway` and Traefik's `ipv4_address`.

The pins stay `/32`, I didn't widen anything.

**2. Fail-fast Docker network conflict check.** Before writing any files, the scripts look at the existing Docker networks and bail with an actionable error if one overlaps the subnet they're about to use. That beats letting `docker compose up` fall over later with `Pool overlaps with other one on this address space`. Existing networks never get touched, and the error points at the override rather than suggesting you remove somebody else's network. Host routes like LANs and VPNs aren't inspected on purpose, since the override handles those without me parsing route tables.

The deployment's own network gets skipped by name so re-runs keep working, and if a leftover own network is sitting on a different subnet the script fails early there too, because compose would quietly reuse it and the generated config wouldn't match. Getting that name right is the fiddly part:

- Enterprise declares `name: netbird` in the compose file, so it's fixed.
- Community and migrate don't, so the network ends up as `<project>_netbird` and the project name has to be derived the way Compose derives it. I checked this against Compose v5.4.0 rather than guessing. It takes the basename of the logical working directory, lowercases it, deletes every character outside `[a-z0-9_-]`, then trims leading `_` and `-`. So `nb.test` becomes `nbtest`, `my nb` becomes `mynb`, and `NetBird-1.0` becomes `netbird-10`. Characters get deleted, not swapped for a `-`.
- Logical matters in `migrate.sh`, which runs `cd "$INSTALL_DIR" && compose up`. A logical `cd` means compose sees the symlink name, so the project name is resolved with a plain `pwd`. Using `pwd -P` resolves the symlink target instead, and then the script flags its own network as a foreign conflict and aborts every re-run under a symlinked install dir.

**3. Enterprise Docker network alias for the public domain.** In the enterprise stack `netbird-server` dials the public domain over HTTPS to deliver traffic flow, and on a host behind NAT that hairpin can fail. The alias resolves the domain to Traefik from inside the compose network, which costs nothing and keeps the request on the box. The community script deliberately doesn't get it: with the proxy enabled its embedded agent learns `stun:<domain>:3478/udp` from management, and an alias is port-agnostic, so it would steer that UDP at Traefik and quietly degrade P2P proxy connections to relay-only.

**4. Community domain validation.** `check_nb_domain` now limits the domain to letters, digits, dots and hyphens, rejects a leading or trailing dot or hyphen, rejects consecutive dots, and caps length at 253 characters. The domain gets embedded in generated YAML and env files, so this is a character-class restriction, not FQDN validation. It's also not parity with the enterprise script, which rejects bare IPs and requires a resolvable FQDN. The community script still accepts IPs and the `use-ip` sentinel on purpose.

## Testing done

I ran a live install on a clean host. Ubuntu 26.04, Docker 29.7.2, Compose v5.4.0, one vCPU, with a domain.

Fresh install with built-in Traefik, proxy on, CrowdSec off:

- All four containers up with `restarts=0`, and zero CrowdSec references in the rendered compose.
- Certificate issued, HTTPS returns 200 with `ssl_verify_result=0`, `/api/users` returns 401 unauthenticated which confirms routing through Traefik, and HTTP 301s to HTTPS.
- Proxy connected to the server from inside the compose network and registered in the cluster.
- Traefik and dashboard logs clean, nothing in the install log.

With `NETBIRD_DOCKER_SUBNET=10.123.45.0/24` the compose subnet and gateway, Traefik's `ipv4_address`, the `config.yaml` pin and `NB_PROXY_TRUSTED_PROXIES` all moved together, the running container's actual IP matched the pin, and no `172.30` reference was left anywhere in the generated files.

I hit all four branches of the conflict check. A foreign overlapping network aborts before anything is written and leaves the existing network and running stack alone. An own network on a mismatched subnet aborts with the leftover error and a `docker network rm` hint. An own network on a matching subnet gets skipped correctly and the run continues to the existing-install guard. No Docker networks at all returns early.

I ran the mismatch case from a directory called `nb.test3` specifically to exercise the project-name normalization. The script worked out `nbtest3_netbird` and hit the own-network branch instead of the overlap branch, which only happens if the normalization matches Compose.

On the validator, `100.64.5.0/24` is rejected with the documented message and writes nothing, `100.63.255.0/24` and `100.128.0.0/24` are accepted, and `172.30.0.0/16`, `172.30.0.1/24`, `10.0.0.0/25`, `010.1.1.0/24`, `256.1.1.0/24` and the reserved ranges are all rejected.

Unit level, on Linux-style bash and macOS bash 3.2: `valid_ipv4_slash24` across 20 CIDRs, `cidrs_overlap` across 6 pairs including `/8` and `/16` supersets and `0.0.0.0/0`, project-name normalization against real Compose output for 11 directory names, and `check_nb_domain` against 21 inputs. `bash -n` passes on all three scripts.

The bash harness lives outside the repo for now. Happy to move it into `infrastructure_files/tests/` if you'd like it in-tree, see the checklist note below.

Originating discussion: https://github.com/netbirdio/docs/pull/626 (hairpin NAT proposal review).

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

- [X] I added/updated documentation for this change

Docs cover `NETBIRD_DOCKER_SUBNET` for all three scripts, the conflict check and its two error variants, and the enterprise domain alias. The migration page also notes that `migrate.sh` carries the trust pins over unchanged.

https://github.com/netbirdio/docs/pull/919

## New Features

* Added configurable Docker `/24` subnet support for setup and migration workflows.
* Automatically derives gateway and Traefik addresses from the selected subnet.
* Displays selected network settings and includes them in generated configuration.
* Supports non-interactive setup using environment variables and defaults.
* Traefik can resolve the configured public domain within the Docker network.

## Bug Fixes

* Prevents conflicts with existing Docker networks and reserved ranges.
* Improves handling and guidance for overlapping networks from existing installations.
* Avoids unnecessary prompts without an interactive terminal.
* Stops setup or migration when Docker is unavailable for network verification.


TechHutTV:
@coderabbitai resume

PR #7056, opened 2026-08-04
[infrastructure] Generate a session cookie encryption key on fresh self-hosted installs

Optic00:
## Describe your changes

The community and enterprise bootstrap scripts now generate a dedicated `server.auth.sessionCookieEncryptionKey` from 32 random bytes for fresh installations.
The key is Base64-encoded, persisted independently from the datastore encryption key, and reused from the generated configuration after restarts.

The community script now also creates `config.yaml` with mode `0600` before writing it, matching the existing enterprise behavior.

The server already supports the session cookie encryption key, but the bootstrap scripts left it unset.
This adds defense-in-depth for newly generated deployments while preserving the existing server-side nonce validation.
Because this only changes newly generated configuration, existing installations and sessions are unchanged.

A focused regression test checks key presence, decoded length, separation from the datastore key, YAML placement, and file mode for both scripts.
It is included in the infrastructure workflow.

## Issue ticket number and link

Closes #7054

## Stack

None.

### Checklist

- [ ] Is it a bug fix
- [ ] Is a typo/documentation fix
- [x] Is a feature enhancement
- [ ] It is a refactor
- [x] Created tests that fail without the change (if possible)
- [x] I ran and tested this change locally - I did not rely on CI to find out whether it works
- [x] This PR has a single purpose (not a fix + refactor + feature in one)
- [x] This change is a trivial fix, **OR** it links an issue the NetBird team agreed on beforehand. Changes to the public API, gRPC protocols, functionality behavior, CLI / service flags, or new features always need that agreement first. See [CONTRIBUTING.md](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTING.md#ticket-first-pr-second).

> By submitting this pull request, you confirm that you have read and agree to the terms of the [Contributor License Agreement](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT.md).

## Documentation

Select exactly one:

- [ ] I added/updated documentation for this change
- [x] Documentation is **not needed** for this change because it only hardens generated secret material and file permissions for fresh bootstrap-script output.

### Docs PR URL (required if "docs added" is checked)

Not applicable.

## Summary by CodeRabbit

* **Security Enhancements**
  * Improved session security by generating dedicated encryption keys during setup.
  * Protected generated configuration files with restricted permissions.
  * Added verification checks to ensure secure session-cookie configuration is applied consistently across setup scripts.

PR #7086, opened 2026-08-07
fix(idp): generate ephemeral session cookie key when none is configured (upgrade-path gap in #7056)

SashaMIT:
## Problem

When neither `auth.sessionCookieEncryptionKey` nor `NB_EMBEDDED_IDP_SESSION_COOKIE_ENCRYPTION_KEY` is set, the embedded IdP runs with an **empty** cookie encryption key. In the dex session layer, an empty key means session cookies are stored as **plain marshaled protobuf** (`userID` + `connectorID` + `nonce`) — encryption is only applied when the key is non-empty (`session.go`: `if len(encryptionKey) > 0`).

On such an install, anyone who can read or predict a user ID can **forge a session cookie** without knowing any secret.

#7056 fixed this for **fresh** installs by generating a key in the bootstrap scripts. Existing installs that upgrade keep their old config with no key, and stay unencrypted — this PR closes that upgrade-path gap.

## Fix

`resolveSessionCookieEncryptionKey` now generates an ephemeral 32-byte random key (base64) when nothing is configured, and logs a prominent warning:

```
embedded IdP: no session cookie encryption key configured; using an ephemeral random key.
Sessions will NOT survive restarts. Set NB_EMBEDDED_IDP_SESSION_COOKIE_ENCRYPTION_KEY
(or auth.sessionCookieEncryptionKey) to a persistent 32-byte key.
```

Trade-off: sessions invalidate on restart until a persistent key is configured. That is strictly better than silently serving forgeable cookies indefinitely. If generation itself fails, the function errors instead of falling back to unencrypted.

## Tests

- The previous `empty key disables encryption` test asserted the vulnerable behavior; it now asserts ephemeral-key generation (32-byte length, fresh randomness per call).
- Full `management/server/idp` package passes.

## How to verify

Run the management server with no key configured: startup logs the warning, and `netbird-session` cookies are AES-encrypted rather than base64 protobuf.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Security**
  * Session cookies are now automatically protected with a freshly generated encryption key when no key is configured.
  * The system reports an error if secure key generation fails, preventing unencrypted session cookies.
  * A warning is logged when an ephemeral key is generated.

### DIFF
--- a/management/server/idp/embedded.go
+++ b/management/server/idp/embedded.go
@@ -316,7 +316,7 @@ func resolveSessionCookieEncryptionKey(configuredKey string) (string, error) {
 			return "", fmt.Errorf("no session cookie encryption key configured and failed to generate an ephemeral one: %w", err)
 		}
 		log.Warn("embedded IdP: no session cookie encryption key configured; using an ephemeral random key. Sessions will NOT survive restarts. Set " + sessionCookieEncryptionKeyEnv + " (or auth.sessionCookieEncryptionKey) to a persistent 32-byte key.")
-		return base64.StdEncoding.EncodeToString(generated), nil
+		return string(generated), nil
 	}
 
 	if validSessionCookieEncryptionKeyLength(len([]byte(key))) {
--- a/management/server/idp/embedded.go
+++ b/management/server/idp/embedded.go
@@ -2,6 +2,7 @@ package idp
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/base64"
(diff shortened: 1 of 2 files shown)


SashaMIT:
coderabbitai inline addressed in fa29ff8: the ephemeral key is now returned as raw 32 bytes (the consumer uses the string directly, same as the configured raw-key path), and the test asserts the raw length instead of base64-decoding.

PR #7093, opened 2026-08-07
[management] prewarm a posture check cache on network map generation

pascal-fischer:
### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved network map generation when posture checks apply across many peers.
  * Reduced repeated posture validation work during account-wide network map updates.
* **Bug Fixes**
  * Ensured precomputed posture validation results match direct evaluation.
  * Preserved correct behavior when no posture checks are configured.

### DIFF
--- a/management/server/types/account_components.go
+++ b/management/server/types/account_components.go
@@ -10,6 +10,8 @@ import (
 	nbdns "github.com/netbirdio/netbird/dns"
 	"github.com/netbirdio/netbird/management/internals/modules/zones"
 	routerTypes "github.com/netbirdio/netbird/management/server/networks/routers/types"
+	nbpeer "github.com/netbirdio/netbird/management/server/peer"
+	"github.com/netbirdio/netbird/management/server/posture"
 	"github.com/netbirdio/netbird/management/server/telemetry"
 	"github.com/netbirdio/netbird/route"
 )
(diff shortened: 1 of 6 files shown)


management/server/types/account_components.go
```diff
@@ -589,21 +602,109 @@ func (a *Account) validatePostureChecksOnPeerGetFailed(ctx context.Context, sour
...
+			if rule.SourceResource.Type == ResourceTypePeer && rule.SourceResource.ID != "" {
+				peerIDs = append(peerIDs, rule.SourceResource.ID)
+			}
+		}
+
+		for _, postureChecksID := range policy.SourcePostureChecks {
+			set := checkPeerIDs[postureChecksID]
+			if set == nil {
+				set = make(map[string]struct{}, len(peerIDs))
+				checkPeerIDs[postureChecksID] = set
```
dmitri-netbird, line 653: could use a slice of peerIds here instead, then no need to convert it to a map. On line 681 then check if there's already a result for a given peerId?

pascal-fischer: As posture checks are not unique per policy (you can have the same check assigned in multiple policies) we would end up with duplicates if the source peers overlap between policies

management/server/types/account_posture_validation_test.go
```diff
@@ -0,0 +1,72 @@
...
+			results[peerID] = result{peers: peerSet, postureFailedPeers: components.PostureFailedPeers}
+		}
+		return results
+	}
+
+	direct := snapshot()
+	account.PrecomputePostureValidation(ctx)
+	memoized := snapshot()
+
+	require.Equal(t, len(direct), len(memoized))
```
dmitri-netbird, line 54: ugh. these checks will also pass when the cache is empty; It's  bit tricky to verify:
 * you could return a flag with the result indicating whether cached result was returned
 * clear PostureChecks field after the call to `PrecomputePostureValidation`

dmitri-netbird (approved):
an issue with a test, otherwise lgtm

PR #7105, opened 2026-08-08
[management] Deny reverse proxy access to pending and blocked users

mlsmaycon:
## Describe your changes

A user in the Pending Approval state could complete SSO and reach any SSO-protected reverse proxy service distributed to a group they belong to, including the All Users group. The reverse proxy authorization path checked the session token signature, that the user exists, that the user's account matches the service's account, and group membership — never the user's account status. The REST API (`permissions/manager.go`) and peer registration both gate on that state, but the proxy gRPC service does not go through the permissions manager, so neither gate applied. A pending user is persisted as blocked and pending approval, so blocked users reached those services the same way.

`ValidateSession` now denies on account status, reporting `pending_approval` or `user_blocked` so the proxy access log and the denied page carry the cause rather than a generic refusal. `GenerateSessionToken` refuses to mint a token for such a user at all, so the browser never receives a session cookie and the OIDC callback can tell the user why instead of showing "Service configuration error". `ValidateUserGroupAccess` and `ValidateTunnelPeer` close the same gap; for the tunnel path this covers a user blocked after their peer was registered, since peer group membership alone kept mesh-origin access open.

A single helper produces both the denied reason for the RPC responses and the sentinel error for the error-returning callers, so the four entry points cannot drift apart. A user the store cannot resolve is denied rather than passed through.

One thing deliberately left out: session cookies are validated locally by the proxy against the service public key with no management round-trip, so a cookie issued before a user is blocked stays valid until it expires (24h by default). That is a revocation-propagation problem rather than this authorization gap, and every option for it (per-request validation with a cache, short-lived tokens with refresh, push-based revocation) changes the proxy hot path or the proxy/management protocol. Worth its own ticket.

closes #7103 #7106

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

- [x] Documentation is **not needed** for this change (explain why) — no configuration, API, or flag surface changes. This restores the documented behavior of the existing approval and blocking model for a path that was skipping it.

n/a

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented pending-approval and blocked accounts from accessing proxies or tunnel peers.
  * Improved handling of missing, unresolved, or cross-account owners.
  * Prevented session tokens from being issued to unauthorized accounts.
  * OAuth sign-in now displays clear denial messages for pending or blocked accounts.
  * Preserved access for eligible unlinked machine peers and restored access when restrictions are removed.

### DIFF
--- a/management/internals/shared/grpc/proxy.go
+++ b/management/internals/shared/grpc/proxy.go
@@ -1623,6 +1623,13 @@ func userStatusDeniedReason(user *types.User) string {
 	return reason
 }
 
+// sameAccount reports whether a user belongs to a service's account. An empty
+// identifier on either side never matches: two unset accounts must not compare
+// equal into a grant.
+func sameAccount(userAccountID, serviceAccountID string) bool {
+	return userAccountID != "" && serviceAccountID != "" && userAccountID == serviceAccountID
+}
+
 // GenerateSessionToken creates a signed session JWT for the given domain and
 // user. The user's group memberships are embedded in the token so policy-aware
 // middlewares on the proxy can authorise without an extra management round-trip.
@@ -1655,7 +1662,7 @@ func (s *ProxyServiceServer) GenerateSessionToken(ctx context.Context, domain, u
 	// with that service's session key. The proxy validates an installed cookie
 	// locally against the service public key, so a token minted for a user of
 	// another account would be honoured without a management round-trip.
-	if user.AccountID != service.AccountID {
+	if !sameAccount(user.AccountID, service.AccountID) {
 		return "", fmt.Errorf("user %s does not belong to the service account", userID)
 	}
 
@@ -1779,7 +1786,7 @@ func (s *ProxyServiceServer) ValidateSession(ctx context.Context, req *proto.Val
 
 	// A user from another account gets a bare response: none of their identity
 	// belongs in an answer to a proxy serving a different account.
-	if user.AccountID != service.AccountID {
+	if !sameAccount(user.AccountID, service.AccountID) {
 		log.WithFields(log.Fields{
 			"domain":          domain,
 			"user_id":         userID,
@@ -1991,9 +1998,10 @@ func (s *ProxyServiceServer) ValidateTunnelPeer(ctx context.Context, req *proto.
 	}
 
 	groupIDs, groupNames := pairGroupIDsAndNames(peerGroups)
-	principalID, displayIdentity := s.getTunnelPeerInfo(ctx, domain, service, peer)
+	owner := s.resolvePeerOwner(ctx, peer)
+	principalID, displayIdentity := s.getTunnelPeerInfo(ctx, domain, service, peer, owner)
 
-	if reason := s.peerOwnerDeniedReason(ctx, peer); reason != "" {
+	if reason := peerOwnerDeniedReason(peer, owner); reason != "" {
 		log.WithFields(log.Fields{"domain": domain, "peer_id": peer.ID, "user_id": peer.UserID, "reason": reason}).Debug("ValidateTunnelPeer: owner status denies access")
 		return &proto.ValidateTunnelPeerResponse{
 			Valid:          false,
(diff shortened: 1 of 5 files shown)


management/internals/shared/grpc/proxy.go
```diff
@@ -1592,25 +1638,25 @@ func (s *ProxyServiceServer) GenerateSessionToken(ctx context.Context, domain, u
...
+	}
+
+	user, userGroups, err := s.usersManager.GetUserWithGroups(ctx, userID)
+	if err != nil {
+		return "", fmt.Errorf("get user %s: %w", userID, err)
+	}
+
+	if _, err := checkUserStatus(user); err != nil {
+		return "", fmt.Errorf("session token for user %s: %w", userID, err)
 	}
```
braginini, line 1672:
```suggestion
// Bind the OIDC identity to the service account before embedding identity data.
  if user.AccountID != service.AccountID {
        return "", fmt.Errorf("user %s does not belong to service account", userID)
  }
```

braginini: making sure the serivice and the user are in the same acocunt

mlsmaycon: Good catch, applied in 9371515.

It matters more than the later `ValidateSession` denial suggests: the token is signed with *this service's* session key, and `forwardWithSessionCookie` validates an installed cookie locally against the service public key with no management round-trip. So the account binding has to happen before signing, not only at validation time.

I also refuse a `nil` user from the store on the same path, and added a case to `TestAuthCallback_UserDeniedByAccountStatus`. Without the check it fails with the minted token in hand — `sub: otherAccountUserId`, `aud: test-proxy.example.com`. The browser-facing message stays the generic "Service configuration error" so the account topology isn't disclosed.

---
_Generated by [Claude Code](https://claude.ai/code)_

management/internals/shared/grpc/proxy.go
```diff
@@ -1909,6 +1981,18 @@ func (s *ProxyServiceServer) ValidateTunnelPeer(ctx context.Context, req *proto.
 	groupIDs, groupNames := pairGroupIDsAndNames(peerGroups)
 	principalID, displayIdentity := s.getTunnelPeerInfo(ctx, domain, service, peer)
 
+	if reason := s.peerOwnerDeniedReason(ctx, peer); reason != "" {
```
braginini: inside these methods management/internals/shared/grpc/proxy.go:2069 and management/internals/shared/grpc/proxy.go:2041 we are calling  `s.usersManager.GetUser(ctx, peer.UserID);`

Basically we call the database twice for a user info. I think that in the context of calling this on every request, that worsens the performance. 

I suggest we call the user once and pass it to the functions mentioned above.

braginini (changes requested):

management/internals/shared/grpc/proxy.go
mlsmaycon: Agreed, fixed in b747558.

The lookup now happens once in `resolvePeerOwner` and the result is passed to both `peerOwnerDeniedReason` (now a plain function, no receiver) and `getTunnelPeerInfo`. An owner that can't be read comes back nil, which the gate already treats as a denial, so the two callers no longer need to interpret the lookup error separately.

`TestValidateTunnelPeerOwnerStatus` now asserts the call count — exactly one lookup for a linked peer, zero for an unlinked one — so the second read can't come back unnoticed.

---
_Generated by [Claude Code](https://claude.ai/code)_

management/internals/shared/grpc/proxy.go
```diff
@@ -1592,25 +1638,37 @@ func (s *ProxyServiceServer) GenerateSessionToken(ctx context.Context, domain, u
...
+		return "", fmt.Errorf("get user %s: %w", userID, errUserUnresolved)
+	}
+
+	// Bind the OIDC identity to the service's account before signing anything
+	// with that service's session key. The proxy validates an installed cookie
+	// locally against the service public key, so a token minted for a user of
+	// another account would be honoured without a management round-trip.
+	if user.AccountID != service.AccountID {
+		return "", fmt.Errorf("user %s does not belong to the service account", userID)
+	}
```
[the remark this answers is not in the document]
mlsmaycon: Taken in b747558, via a `sameAccount(userAccountID, serviceAccountID string) bool` helper applied to both the mint path and `ValidateSession`, so the two can't drift. `TestSameAccount` pins the empty cases.

Both IDs are non-empty in practice — services carry their tenant key and users are FK'd to an account — but comparing two unset strings into a grant is the wrong failure mode for a security check, so the guard is worth its one line.

On the other finding in this review (`checkUserStatus` should handle `UserStatusDisabled` and fail closed on unknown states): that one doesn't apply. `UserStatus` in `types/user.go:22-24` is an API DTO concept — it lives on `UserInfo.Status`, is derived as Active or Invited from IdP data, and `UserStatusDisabled` is never produced by that path. The stored `types.User` has no status field at all; `checkUserStatus` switches over the `PendingApproval` and `Blocked` booleans, which is a complete partition with no unknown value to fall through. There is no enum here to make fail-closed.

---
_Generated by [Claude Code](https://claude.ai/code)_

management/internals/shared/grpc/proxy.go
```diff
@@ -1944,9 +2048,46 @@ func (s *ProxyServiceServer) ValidateTunnelPeer(ctx context.Context, req *proto.
...
+func peerOwnerDeniedReason(peer *peer.Peer, owner *types.User) string {
+	if peer.UserID == "" {
+		return ""
+	}
+
+	if owner == nil {
+		return deniedReasonUserNotFound
+	}
+
+	return userStatusDeniedReason(owner)
```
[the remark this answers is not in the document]
mlsmaycon: Real, and fixed in 7e0fdd4 — thanks. `usersManager.GetUser` looks up by user ID alone with no account scope, so a peer row pointing outside the service's account does resolve a foreign user.

I put the guard one level up, in `resolvePeerOwner`, rather than in the gate. Gating alone would still have let the foreign identity through `getTunnelPeerInfo` into `UserId`/`UserEmail` on the response and into the minted token; leaving the owner unresolved denies the peer *and* keeps the fallback identity on the peer itself.

Verified by reproduction — dropping the guard makes the new `owner in another account denied and not disclosed` case fail three ways: `Valid: true`, empty denied reason, and `foreign@example.com` on the response.

---
_Generated by [Claude Code](https://claude.ai/code)_

braginini (approved):

PR #7144, opened 2026-08-10
[client, android] Handle network changes without restarting the engine

pappz:
## Describe your changes
On network changes the client restarted the whole engine. That is heavy-handed and slow: it tears down working state to recover from a transition the engine could handle itself. This replaces the restart with proper network event handling.

1. **Suspend the retry loops while no network is available.** Instead of burning through backoff intervals against an unreachable network, the reconnection loops park until the OS reports a usable network again.

2. **Reconnect immediately on a network switch.** When the OS hands us a new network, connections bound to the old one are swept and re-dialed right away, rather than waiting for a timeout to notice they are dead.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

## New Features
- Android and iOS clients can report network availability and network changes.
- Connection status now indicates when the device is offline.
- Mobile connection listeners receive standardized connection-state updates.
- Network changes automatically clean up stale connections and active connection attempts.

## Bug Fixes
- Reconnection pauses while offline and resumes when connectivity returns.
- Improved recovery and backoff behavior after temporary network interruptions.
- Client shutdowns and restarts now exit connection retries cleanly.
- Active connection attempts are canceled when network conditions change.
- Network-aware retry behavior improves reconnection across supported clients.


client/internal/connect.go
```diff
@@ -70,25 +72,53 @@ type ConnectClient struct {
...
+func WithSweeper(sweeper *netsweep.Sweeper) ConnectClientOption {
+	return func(c *ConnectClient) { c.sweeper = sweeper }
 }
 
 func NewConnectClient(
 	ctx context.Context,
 	config *profilemanager.Config,
 	statusRecorder *peer.Status,
+	opts ...ConnectClientOption,
 ) *ConnectClient {
```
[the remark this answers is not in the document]
pappz: Out of scope to fix it.

client/netstate/netstate.go
```diff
@@ -0,0 +1,107 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.online == online {
+		return
+	}
+	s.online = online
+	close(s.changed)
+	s.changed = make(chan struct{})
+	log.Infof("OS network availability changed: online=%t", online)
+}
```
theodorsm, line 54: All other functions are safe with nil State, but not this write. I see this is mentioned in the package comment (that read is safe), but we should also document that function is not safe on the function comment, or do the nil check so we don't panic.

theodorsm (changes requested):
We should also consider coderabbit's comment:
https://github.com/netbirdio/netbird/pull/7144/changes#r3776207686

client/netstate/netstate.go
pappz: Then nil check can hide real issue on writer (set side). By desigen on reader side can be nil but on writer side we expect not nil.

pappz: I will update the comments

client/internal/peer/guard/guard.go
theodorsm, line 91: Nitpick: worth a comment about new network change behavior too?

theodorsm (approved):

PR #7176, opened 2026-08-12
[infrastructure] Let a suite outside this repo use the e2e harness

mlsmaycon:
## Describe your changes

`e2e/harness` documents itself as feature-agnostic — *"any suite can ask for a live management server (with an admin PAT minted through the unauthenticated /api/setup bootstrap) and, later, a proxy and client"* — but three details assume the caller lives in this repo. The terraform provider's acceptance suite wants exactly what the package advertises, and today it cannot use it, so it would have to carry a second harness for the same product.

**This does add exported API.** Three option constructors and two variadic parameters, all additive — no existing call site changes, and `StartCombined(ctx)` / `StartClient(ctx, c, key)` compile and behave exactly as before. Nothing outside `e2e/` is touched, and the package is behind the `e2e` build tag so it is absent from normal builds.

| Added | |
| --- | --- |
| `harness.CombinedOption` | option type for `StartCombined` |
| `harness.WithGeolocation()` | keep the GeoLite download |
| `harness.WithServerEnv(map[string]string)` | add/override combined container env |
| `harness.ClientOption` | option type for `StartClient` |
| `harness.WithClientName(string)` | name an agent (alias + hostname) |

### The three problems

**`repoRoot` found any `go.mod`, not this one.** It walked up from the working directory and used the first module root as the Docker build context. From another module that is the caller's own root, where `combined/Dockerfile.multistage` does not exist. It now requires the ancestor to be this module, and falls back to asking the go tool where this module's source is. For a dependent that is the extracted module directory of the version it pins — which is the right context, because the server it tests against is then built from the same revision as the client library it was compiled against.

**Geolocation was off unconditionally**, in both the container environment and `disableGeoliteUpdate`. That is correct for agent-network ingest, which does not use it, but a suite asserting on location-based posture checks needs the database: management evaluates those rules against it, and a rule it cannot evaluate fails rather than passing without having been checked.

```go
srv, err := harness.StartCombined(ctx, harness.WithGeolocation())
srv, err := harness.StartCombined(ctx, harness.WithServerEnv(map[string]string{"NB_LOG_LEVEL": "debug"}))
```

`WithServerEnv` is the general escape hatch, so a suite needing a setting the harness does not model does not have to fork it.

**`StartClient` pinned one network alias and set no hostname.** A second agent cannot start — the alias collides — and the peer's name was whatever the container happened to get. Both matter to a suite whose fixtures address peers by name, since management records the container hostname at registration and that is the name the peer appears under in the API:

```go
c1, _ := harness.StartClient(ctx, srv, key, harness.WithClientName("peer1"))
c2, _ := harness.StartClient(ctx, srv, key, harness.WithClientName("peer2"))
```

Not filed. This is test-only: no product behaviour, no protocol, no CLI or service flag, and no package outside `e2e/`. The exported surface added is listed above and is entirely additive. If the additive-exported-API-on-a-test-helper case still wants a ticket first, say so and I will open one before you spend review time.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

&gt; By submitting this pull request, you confirm that you have read and agree to the terms of the [Contributor License Agreement](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT.md).

## Documentation
Select exactly one:

Test-helper internals with no user-facing surface. The behaviour is documented where it is used — on the exported options and in the package doc comment.

n/a

## How this was tested

`e2e/harness/options_test.go` covers the wiring without Docker, because that is what the options are: a container environment and a config file, both assembled before anything starts. A mistake there would otherwise surface as a puzzling failure minutes into a container run.

`TestModuleDirResolvesUnderVendorMode` earns its place: a caller that vendors puts the go command in automatic vendor mode, where the lookup succeeds with an **empty** `Dir`, and the harness then reports a missing module for a dependency that is present. It fails without `-mod=readonly` with exactly that misleading error, which I checked both ways.

**The full suite is green in CI.** [Agent Network E2E run 31569967715](https://github.com/netbirdio/netbird/actions/runs/31569967715) on this branch: **30 tests, all passing, zero failures.**

```
ok  github.com/netbirdio/netbird/e2e/agentnetwork  699.023s
ok  github.com/netbirdio/netbird/e2e/harness         0.021s
```

22 of those are the existing `e2e/agentnetwork` suite, unchanged by this PR — `TestProvidersMatrix` (356s), the six guardrail tests, the pricing and attribution tests, `TestProviderSkipTLSVerification`, `TestVLLMProvider`, and the settings-bootstrap pair. They call `StartCombined(ctx)` and `StartClient(ctx, c, key)` with no options, so they are what covers the unchanged defaults. The other 8 are the new `options_test.go`.

This run had to be dispatched by hand: `agent-network-e2e.yml` triggers only on `schedule` and `workflow_dispatch`, never on `pull_request`, so a PR touching this suite gets no automated coverage of it.

For the record on what I could and could not run locally: my sandbox cannot pull from Docker Hub (the blob CDN returns 403), so `nginx:alpine` is unavailable and the 12 data-plane tests that need a mock vLLM upstream all failed there — identically on `main`, at the same point, before and after this change. I ran the rest against locally-built component images via `NB_E2E_{COMBINED,CLIENT,PROXY}_IMAGE`, with the per-test outcome identical before and after. The CI run above is what actually clears those 12.

## Who wants this

[netbirdio/terraform-provider-netbird#198](https://github.com/netbirdio/terraform-provider-netbird/pull/198) is the consumer: it drops a compose file and a hand-written SQL seed in favour of this harness. Its acceptance suite is green against a deployment started through these options — **129 tests run, 2 skipped, 0 failed** (the two skips are cloud-only SCIM tests that skip unconditionally), including the peer fixtures that need `WithClientName` and the geolocation posture checks that need `WithGeolocation`.

That PR is open rather than draft, and only its `go.mod` waits on this: it currently pins a pseudo-version of this branch, which it replaces once this merges.

## Summary by CodeRabbit

- **New Features**
  - Added configurable test environments with optional geolocation data downloads.
  - Added support for custom server environment variables.
  - Added configurable client names for container hostnames and network aliases.
  - Improved automatic discovery of the project source directory.

- **Bug Fixes**
  - Kept generated configuration and container settings consistent.
  - Improved validation when project files or module information are missing.
  - Improved handling of cancelled operations.

- **Tests**
  - Added coverage for geolocation, environment overrides, client naming, module detection, vendored dependencies, and repository discovery.

### DIFF
--- a/e2e/harness/Dockerfile.client
+++ b/e2e/harness/Dockerfile.client
@@ -20,5 +20,9 @@ ENV NETBIRD_BIN="/usr/local/bin/netbird" \
     NB_ENABLE_CAPTURE="false" \
     NB_ENTRYPOINT_SERVICE_TIMEOUT="30"
 ENTRYPOINT [ "/usr/local/bin/netbird-entrypoint.sh" ]
-COPY client/netbird-entrypoint.sh /usr/local/bin/netbird-entrypoint.sh
+# --chmod because the build context is not always a git checkout. A suite in
+# another module builds from this module's extracted copy in the module cache,
+# where every file is 0444 — the cache drops the executable bit git records — and
+# a bare COPY then produces an entrypoint the runtime cannot exec.
+COPY --chmod=0755 client/netbird-entrypoint.sh /usr/local/bin/netbird-entrypoint.sh
 COPY --from=builder /out/netbird /usr/local/bin/netbird
--- a/e2e/harness/client.go
+++ b/e2e/harness/client.go
@@ -61,7 +61,7 @@ func StartClient(ctx context.Context, c *Combined, setupKey string, opts ...Clie
 		opt(&o)
 	}
 
-	root, err := repoRoot()
+	root, err := repoRoot(ctx)
 	if err != nil {
 		return nil, err
 	}
--- a/e2e/harness/client.go
+++ b/e2e/harness/client.go
@@ -32,11 +32,35 @@ type Client struct {
 	container testcontainers.Container
 }
 
+// clientOptions is what the ClientOption values assemble.
+type clientOptions struct {
+	name string
+}
+
+// ClientOption adjusts how StartClient runs the agent.
+type ClientOption func(*clientOptions)
+
+// WithClientName names the agent, which sets both its network alias and its
+// container hostname. The hostname matters beyond addressing: the agent reports
+// it to management at registration, so it is the name the peer appears under in
+// the API.
+//
+// Required to run more than one agent against the same server — the default name
+// is shared, and two containers cannot hold the same alias on one network.
+func WithClientName(name string) ClientOption {
+	return func(o *clientOptions) { o.name = name }
+}
+
 // StartClient builds the client image and runs it on the combined server's
 // network, joining via the given setup key. The image entrypoint brings the
 // daemon up automatically; callers wait for connectivity with WaitConnected /
 // WaitProxyPeer.
-func StartClient(ctx context.Context, c *Combined, setupKey string) (*Client, error) {
+func StartClient(ctx context.Context, c *Combined, setupKey string, opts ...ClientOption) (*Client, error) {
+	o := clientOptions{name: clientAlias}
+	for _, opt := range opts {
+		opt(&o)
+	}
+
 	root, err := repoRoot()
 	if err != nil {
 		return nil, err
@@ -47,9 +71,13 @@ func StartClient(ctx context.Context, c *Combined, setupKey string) (*Client, er
 	}
 
 	req := testcontainers.ContainerRequest{
-		Image:          clientImage,
+		Image: clientImage,
+		// The agent reports the container's hostname to management, so this is
+		// the name the peer is addressable by in the API as well as on the
+		// network. The entrypoint takes no hostname flag of its own.
+		Hostname:       o.name,
 		Networks:       []string{c.network.Name},
-		NetworkAliases: map[string][]string{c.network.Name: {clientAlias}},
+		NetworkAliases: map[string][]string{c.network.Name: {o.name}},
 		Env: map[string]string{
 			"NB_MANAGEMENT_URL": combinedExposedURL,
 			"NB_SETUP_KEY":      setupKey,
--- a/e2e/harness/combined.go
+++ b/e2e/harness/combined.go
@@ -61,10 +61,67 @@ type Combined struct {
 	workDir string
 }
 
+// combinedOptions is what the CombinedOption values assemble.
+type combinedOptions struct {
+	geolocation bool
+	env         map[string]string
+}
+
+// CombinedOption adjusts how StartCombined boots the server. The defaults suit a
+// suite that only drives the API; the options exist for the ones that need more
+// of the product than that.
+type CombinedOption func(*combinedOptions)
+
+// WithGeolocation leaves the GeoLite database download enabled. It is off by
+// default because the download adds startup latency that most suites get nothing
+// for. A suite asserting on location-based posture checks needs it: management
+// evaluates those rules against the database, and without it the rule fails
+// instead of passing without having been checked.
+func WithGeolocation() CombinedOption {
+	return func(o *combinedOptions) { o.geolocation = true }
+}
+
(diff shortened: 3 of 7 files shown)


mlsmaycon:
Both findings were real and are fixed in f4d2b56.

**Vendor mode** — confirmed by reproducing it. Without `-mod=readonly` the lookup succeeds with an empty `Dir`, and the harness then reports a missing module for a dependency that is present:

```
locate example.com/dep: the go tool reported no directory; run `go mod download example.com/dep`
```

`TestModuleDirResolvesUnderVendorMode` covers it, and needs no network — a temporary main module with a local replacement, vendored. It fails with exactly the error above before the flag and passes after, which I checked both ways rather than assuming. `moduleDir` now takes the module path as a parameter so the test can name its own module instead of reaching for netbird's.

Not using `go mod download -json` as a fallback, per your note — it can ignore a local replacement and query the remote.

**Context** — `repoRoot` and `moduleDir` now take the caller's context, threaded from `StartCombined`, `StartClient` and `StartProxy`. No bounded child timeout: all three callers already pass a context with a deadline (the agent-network suite's is 15 minutes), so adding one here would only shorten what the caller chose. `TestModuleDirHonoursContext` asserts a cancelled context fails the lookup.

**testify** — adopted throughout `options_test.go`, with `require` for setup and `assert` for the conditions under test.

Re-verified after the change: the 8 harness tests pass, and `e2e/agentnetwork` per-test outcomes are still byte-identical to unmodified `main` in my environment (8 pass / 12 fail / 2 skip, the 12 being `No such image: nginx:alpine` from no registry access here, unchanged from before this PR).

On the description check wanting an issue link: happy to open one if you'd rather agree the shape first. The exported surface added is three option constructors and two variadic parameters, all additive — no existing call site changes — but it is your call whether that needs a ticket ahead of review.

---

mlsmaycon:
## Agent Network E2E is green on this branch

[Run 31569967715](https://github.com/netbirdio/netbird/actions/runs/31569967715) on `e2e/harness-options` (2614433), dispatched manually: **30 tests, all passing, zero failures.**

```
ok  github.com/netbirdio/netbird/e2e/agentnetwork  699.023s
ok  github.com/netbirdio/netbird/e2e/harness         0.021s
```

- **22 tests in `e2e/agentnetwork`** — the existing suite, unchanged by this PR. `TestProvidersMatrix` (356s), the six guardrail tests, the pricing and attribution tests, `TestProviderSkipTLSVerification`, `TestVLLMProvider`, and the settings-bootstrap pair.
- **8 tests in `e2e/harness`** — the new `options_test.go`, including `TestModuleDirResolvesUnderVendorMode` and `TestModuleDirHonoursContext`, the two regression tests for the review findings.

Worth stating why this run matters rather than just that it is green: `agent-network-e2e.yml` triggers only on `schedule` and `workflow_dispatch`, never on `pull_request`. So this branch had no automated coverage of the suite it changes — every check on the PR so far was build and lint. A dispatch was the only way to find out.

The refactor touches what this suite runs on, so a break would have shown up here:

- **`StartCombined` is now variadic** (`CombinedOption`), and the config template gained a verb. Every one of these 22 tests goes through it — `TestCombinedBootstrap` first. A mis-ordered `Sprintf` argument list would have stopped the server from starting at all.
- **`StartClient` is now variadic** and sets `Hostname` plus per-client `NetworkAliases`. The log shows `netbird-client:e2e` containers created, started and ready throughout, so the default path still resolves the alias.
- **`Dockerfile.client` now uses `COPY --chmod=0755`.** This is the one I most wanted a real run for: the Go module cache extracts files `0444`, dropping git's `100755`, so the entrypoint was executable in-repo and not executable to a dependent module. Those client containers coming up ready confirms the explicit mode did not break the in-repo case it was already fine for.
- **`repoRoot` and `moduleDir` now take a context** and pass `-mod=readonly`. `TestRepoRootFindsThisRepo` asserts the in-repo walk still wins, so the module lookup stays a fallback and an in-repo run never depends on the cache.

Both reverse-proxy and client containers reach ready and terminate cleanly across the run, and the buildx layer cache saved on the primary key, so a rerun will not pay the cold build again.

---

braginini (approved):

PR #7186, opened 2026-08-12
[client] Fall back to an available OAuth flow instead of hard-coding device code

lixmal:
## Describe your changes

Clients on Linux hosts pick the device code flow whenever no desktop environment is detected, and stay on it even when the server or the identity provider does not offer that flow, which fails the login instead of using PKCE.

- Treat the device code flow as a preference rather than a fixed choice: fall back to the other flow when a flow is not configured on the server, comes back incomplete, or is refused by the identity provider when the flow is run, and keep failing on the preferred flow for unrelated errors
- Improve graphical session detection by checking the variables that actually decide whether a browser can be opened (`BROWSER`, `DESKTOP_SESSION`, `XDG_CURRENT_DESKTOP`, `DISPLAY`, `WAYLAND_DISPLAY`, `XDG_SESSION_TYPE`), and let the desktop UI report a graphical session outright instead of inferring one
- Fix `netbird login --extend` and SSH authentication always choosing the device code flow on Linux: both run in the daemon, which does not inherit the user's session environment, so they now take the answer from the caller, and their errors no longer suggest setup keys, which cannot extend a session or authenticate SSH

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

The flow selection is internal to the client; no configuration or user-facing option changes.

## Summary by CodeRabbit

* **New Features**
  * Authentication now selects the most suitable available sign-in method, with automatic fallback when a provider does not support the preferred option.
  * Login and session extensions better recognize whether a graphical session is available.
  * Added clearer setup-key guidance when single sign-on is unavailable.

* **Bug Fixes**
  * Improved handling of unsupported authentication configurations and preserved actionable errors during sign-in.

### DIFF
--- a/client/cmd/login.go
+++ b/client/cmd/login.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"os/user"
-	"runtime"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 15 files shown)


lixmal:
Superseded by #7187 (session extension / SSH auth flow selection) and #7188 (fallback between authorization flows), which split this change in two. The review findings from this PR are addressed in those.

PR #7268, opened 2026-08-20
[management] use peer.Status.LastSeen field to detect stale peer.Status.Connected values

dmitri-netbird:
## Describe your changes
It's currently possible for peer.Status.Connected field to become stale, should the management server crash before marking the peer disconnected. This change updates peer.Status.LastSeen field on every update sent to the peer, and every minute if there are no updates. `peer.Status.Connected` is true when the persisted value is true and the last-seen timestamp is less than 5mins old. 

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved peer connectivity status accuracy by refreshing activity timestamps during update processing.
  - Peers not seen for more than five minutes are now correctly marked as disconnected.
  - Recently active peers retain their connected status.

- **Reliability**
  - Peer activity timestamps are refreshed periodically to better reflect ongoing availability.
  - Failures while refreshing peer activity are now handled consistently.


management/internals/shared/grpc/server.go
```diff
@@ -453,6 +462,18 @@ func (s *Server) handleUpdates(ctx context.Context, accountID string, peerKey wg
...
+				log.WithContext(ctx).Debugf("error updating peer's last seen timestamp %s: %v", peerKey.String(), err)
+				return err
+			}
+
+		case <-ticker.C:
+			if err := s.accountManager.RefreshPeerLastSeen(ctx, accountID, peer.ID); err != nil {
+				log.WithContext(ctx).Debugf("error updating peer's last seen timestamp %s: %v", peerKey.String(), err)
+				return err
+			}
+
```
pascal-fischer, line 476: if we exit on an error we should run the same `s.cancelPeerRoutines(ctx, accountID, peer, streamStartTime)` to cleanup the peer and do not leave a stale

management/server/store/sql_store.go
```diff
@@ -1962,6 +1962,9 @@ func (s *SqlStore) getPeers(ctx context.Context, accountID string) ([]nbpeer.Pee
 			}
 			if peerStatusConnected.Valid {
 				p.Status.Connected = peerStatusConnected.Bool
+				if peerStatusLastSeen.Valid {
+					p.Status.Connected = p.Status.Connected && peerStatusLastSeen.Time.After(time.Now().Add(-5*time.Minute))
+				}
```
pascal-fischer, line 1967: This is only the old getAccount path which this does not play a part. The DeletePeers call will need it in GetPeerByID to be able to bypass the connected state. Probably we need to pass this on all calls that read a peer

management/internals/shared/grpc/server.go
```diff
@@ -439,6 +442,12 @@ func (s *Server) handleUpdates(ctx context.Context, accountID string, peerKey wg
 				}
 			}
 
+			ticker.Reset(tickerInterval)
+			if err := s.accountManager.RefreshPeerLastSeen(ctx, accountID, peer.ID); err != nil {
+				log.WithContext(ctx).Debugf("error updating peer's last seen timestamp %s: %v", peerKey.String(), err)
+				return err
+			}
```
pascal-fischer, line 449: Idk if writing per peer is a good idea on the nmap update path. If we consider large accounts with thousands of peers affected this will be thousands of db calls in high frequency accounts

pascal-fischer (changes requested):

PR #7270, opened 2026-08-20
[client] Port the ipset fallback to the unified filter rules

lixmal:
## Describe your changes

#6332 added a fallback to per-IP ACL rules for kernels that lack the ipset hash modules (see discussion #6125). The unified filter path in #6322 predates that merge and deletes `acl_linux.go`, so the fallback would be lost when the rewrite lands.

This ports the fallback to the unified path:

- `family` probes ipset support once at init, same probe as on main.
- When the probe fails, a multi-source filter rule expands into one iptables rule per source prefix instead of a single `--match-set` rule. Since peer and route ACLs share `AddFilterRule`, route ACLs get the same fallback.
- `Rule` now tracks the expanded specs so delete removes every per-prefix rule, and a failed expansion rolls back the rules already installed. Without the rollback, a partial install would keep matching without being tracked.

Destination sets (domain routes) are not covered here: their prefixes are only known after DNS resolution, so there is nothing to expand. That and falling back at the point of failure instead of a boot-time probe are handled in #7184.

Discussion #6125, ports #6332 onto #6322.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Restores existing behavior that #6332 already shipped without documentation changes; there is no new configuration or user-facing behavior on standard kernels.

### DIFF
--- a/client/firewall/iptables/family_linux.go
+++ b/client/firewall/iptables/family_linux.go
@@ -112,6 +112,10 @@ type family struct {
 	// AddFilterRule writes here; DeleteFilterRule looks up by id.
 	filters      map[nbid.RuleID]*Rule
 	ipsetCounter *ipsetCounter
+	// ipsetSupported records whether the kernel can create the hash:net
+	// sets the source matches rely on; probed once at init. When false,
+	// multi-source rules expand to one rule per source prefix.
+	ipsetSupported bool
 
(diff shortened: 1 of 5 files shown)


lixmal:
The port landed on the base branch as part of resolving #6322's conflicts with main (merge c7f5e3507 + b721a2dbb), so this standalone PR is no longer needed.

PR #7286, opened 2026-08-21
[client] Tolerate a still-locked updater binary when cleaning up after an update

lixmal:
## Describe your changes

Post-update cleanup on Windows logged an error and left `updater.exe` behind in the temp install directory. The daemon runs the cleanup at startup, but the updater is what started it, so the updater's image is often still locked at that moment and Windows refuses to delete a locked executable.

- Retry removing the updater binary while it is still locked, and schedule the deletion for the next reboot if it stays locked
- Stop treating a locked updater as a cleanup failure, so a successful update no longer logs an error
- Skip the updater binary in the installer extension sweep, which reported the same file a second time

### How it happens

```mermaid
sequenceDiagram
    participant U as updater.exe
    participant SCM as Service manager
    participant D as netbird daemon
    U->>SCM: start netbird service
    SCM->>D: start
    Note over U: Still running: restarts the UI, writes the result file
    D->>D: CleanUpInstallerFiles()
    D--xU: delete tmp-install\updater.exe
    Note over D: Access is denied, the image is still locked
    Note over D: Cleanup error logged, updater.exe left behind
    U->>U: exits
```

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal cleanup behavior after an auto-update, no user-facing surface changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows updater cleanup by retrying briefly when the file is temporarily locked.
  * Leaves persistently locked updater files in place for the next update to overwrite, without reporting an installation failure.
  * Treats already-removed updater files as successful cleanup.
  * Improved cleanup behavior across Windows and macOS while preserving logs and result files.

### DIFF
--- a/client/internal/updater/installer/remove_updater_windows.go
+++ b/client/internal/updater/installer/remove_updater_windows.go
@@ -1,0 +1,58 @@
+package installer
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/sys/windows"
+)
+
+const (
+	// The updater is the process that restarted the daemon, so when the daemon
+	// cleans up at startup the updater is often still exiting and Windows refuses
+	// to delete its locked image. These bound how long cleanup waits for it.
+	updaterRemoveAttempts = 5
+	updaterRemoveDelay    = 200 * time.Millisecond
+)
+
+// removeUpdaterBinary deletes the updater copy left in the temp dir, retrying
+// while the still-exiting updater process holds its image. If it stays locked for
+// the whole window the file is scheduled for deletion on the next reboot, so a
+// locked updater is never reported as a cleanup failure.
+func removeUpdaterBinary(path string) error {
+	for attempt := 0; attempt < updaterRemoveAttempts; attempt++ {
+		if attempt > 0 {
+			time.Sleep(updaterRemoveDelay)
+		}
+
+		err := os.Remove(path)
+		if err == nil || os.IsNotExist(err) {
+			return nil
+		}
+		if !isFileLocked(err) {
+			return err
+		}
+	}
+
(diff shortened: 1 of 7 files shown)


client/internal/updater/installer/remove_updater_windows.go
```diff
@@ -0,0 +1,58 @@
...
+
+	log.Debugf("updater binary %s is still locked, scheduling removal on next reboot", path)
+	return scheduleDeleteOnReboot(path)
+}
+
+func isFileLocked(err error) bool {
+	return errors.Is(err, windows.ERROR_ACCESS_DENIED) || errors.Is(err, windows.ERROR_SHARING_VIOLATION)
+}
+
+func scheduleDeleteOnReboot(path string) error {
```
pappz: Why is so big pain if we left behind an updater binary? With the registry operations we risk another virus detection.

client/internal/updater/installer/remove_updater_windows.go
```diff
@@ -0,0 +1,58 @@
...
+		err := os.Remove(path)
+		if err == nil || os.IsNotExist(err) {
+			return nil
+		}
+		if !isFileLocked(err) {
+			return err
+		}
+	}
+
+	log.Debugf("updater binary %s is still locked, scheduling removal on next reboot", path)
```
pappz: I think it should be Info or error level. Without it we can never discover an permanent issue.

pappz (changes requested):

client/internal/updater/installer/remove_updater_windows.go
lixmal: Fixed

client/internal/updater/installer/remove_updater_windows.go
lixmal: It's mostly cosmetic, good point about the registry, fixed

pappz (approved):

PR #7305, opened 2026-08-24
[management, client] add protobuf breaking changes check

dmitri-netbird:
## Describe your changes
This PR adds a github workflow to catch protobuf breaking changes. See https://github.com/netbirdio/netbird/actions/runs/32748152216/job/97498490569 for an example of a failing check.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

## Chores

* API schema validation now runs on pushes to the main and release branches.
* Pull requests that modify protobuf definitions automatically trigger validation.
* Validation uses read-only permissions and pinned tooling for improved security and consistency.
* Automated checks detect potentially breaking schema changes.
* Validation workflows no longer push schema updates automatically.

### DIFF
--- a/.github/workflows/buf.yml
+++ b/.github/workflows/buf.yml
@@ -1,8 +1,8 @@
 name: protobuf checks
 on:
   push:
-    paths:
-      - '**.proto'
+#    paths:
+#      - '**.proto'
 permissions:
   contents: read
   pull-requests: read
--- a/.github/workflows/buf.yml
+++ b/.github/workflows/buf.yml
@@ -16,7 +16,4 @@ jobs:
           persist-credentials: false
       - uses: bufbuild/buf-action@8c6a16e16f12ba20b6470afa9c2ba9b5ba8c97c3 # v1.5.0
         with:
-          format: false
-          lint: false
-          push: false
-          archive: false
+          breaking: true
--- a/.github/workflows/buf.yml
+++ b/.github/workflows/buf.yml
@@ -1,8 +1,12 @@
 name: protobuf checks
 on:
   push:
+    branches:
+      - main
+      - "release-*"
+  pull_request:
     paths:
-      - '**.proto'
+      - "**.proto"
 permissions:
   contents: read
   pull-requests: read
--- a/.github/workflows/buf.yml
+++ b/.github/workflows/buf.yml
@@ -6,7 +6,7 @@ on:
       - "release-*"
   pull_request:
     paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/buf.yml
```diff
@@ -0,0 +1,20 @@
+name: protobuf checks
+on:
+  push:
+    paths:
+      - '**.proto'
```
bison: Not sure what exactly the intent is, but you probably want something like:

```yaml
on:
  push:
    branches:
      - main
      - "release-*"
  pull_request:
    paths:
      - "**.proto"
```

Always run the test on `main` and release branches and on pull requests that affect proto files.

.github/workflows/buf.yml
```diff
@@ -0,0 +1,24 @@
+name: protobuf checks
+on:
+  push:
+    branches:
+      - main
+      - "release-*"
+  pull_request:
+    paths:
+      - "**.proto"
```
bison, line 13: Sorry, I forgot to mention in my original comment that this should probably also run when this file changes, which also let us see it pass here:
```yaml
    - ".github/workflows/buf.yaml"
```
We could also add the buf files if we wanted, even if they don't exist yet...
```yaml
    - "**/buf.yaml"
    - "**/buf.lock"
    - "**/buf.gen.yaml"
```
But maybe that's silly, I don't know.

dmitri-netbird: Nah, I figured we can add buf config-related once we switch to buf; I didn't think it's useful atm? The one at the top makes sense, will update.

bison: Sounds good.

dmitri-netbird: I'll add them too, b/c why not.

.github/workflows/buf.yml
```diff
@@ -0,0 +1,28 @@
+name: protobuf checks
+on:
+  push:
+    branches:
+      - main
+      - "release-*"
+  pull_request:
+    paths:
+      - ".github/workflows/buf.yaml"
```
bison: This was a typo on my part. It should be `.yml` — I hate that 😭

bison (changes requested):

.github/workflows/buf.yml
dmitri-netbird: ugh, I missed it too, "yaml" is what I'd expect, but our naming convention is different. Fixed.

bison (approved):

PR #7306, opened 2026-08-24
[management,signal,proxy,relay,misc] Unify service configuration loading

jnfrati:
## Describe your changes

Service entry points previously loaded defaults, files, environment variables, and flags through different paths, making precedence inconsistent and leaving Relay, Signal, and Proxy without config-file support.

This adds a shared Viper-backed loader with `flags > environment > file > defaults` precedence and migrates Combined, Management, Relay/STUN, Signal, and Proxy to it. Management keeps its existing environment-template expansion, while existing environment and flag aliases remain compatible.

Maintainer-initiated change; no issue.

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Validated example configuration files are included alongside the Relay, Signal, and Proxy binaries, keeping the complete schemas next to their implementations.

## Summary by CodeRabbit

* **New Features**
  * Added consistent configuration loading from files, environment variables, and command-line flags across services.
  * Added example configuration files for proxy, relay, and signal.
  * Added environment-template expansion and optional configuration file support.
  * Added trusted-proxy text configuration parsing.
* **Bug Fixes**
  * Preserved legacy configuration formats, defaults, flag behavior, and environment-variable compatibility.
  * Improved configuration error handling and validation for invalid environment values.
  * Corrected signal port selection and management configuration overrides.


management/cmd/config_test.go
```diff
@@ -0,0 +1,59 @@
...
+func TestLoadManagementConfigUsesDefaultDataDirForEmptyValue(t *testing.T) {
+	configPath := filepath.Join(t.TempDir(), "management.json")
+	require.NoError(t, os.WriteFile(configPath, []byte(`{"Datadir":""}`), 0o600))
+
+	cfg, err := loadManagementConfig(configPath)
+	require.NoError(t, err)
+	assert.Equal(t, defaultMgmtDataDir, cfg.Datadir, "Empty legacy values should use the default data directory")
+}
+
+func TestLoadManagementConfigSources(t *testing.T) {
```
bison: I think we need more exhaustive testing here. For example I think viper lowercases map keys, so the `ExtraConfig` in the IDP config gets messed up. Since management writes back to its own config (😭) that also gets persisted to disk.

Also, one of the main reasons to use environment variables is to keep secrets out of the on-disk config, but management now persists every `NB_*` environment variable back to disk.

management/cmd/management.go
```diff
@@ -211,14 +212,18 @@ func LoadMgmtConfig(ctx context.Context, mgmtConfigPath string) (*nbconfig.Confi
...
+	if (flags.Changed("letsencrypt-domain") || hasCertOverride) && cfg.HttpConfig == nil {
+		cfg.HttpConfig = &nbconfig.HttpServerConfig{}
+	}
+	if flags.Changed("letsencrypt-domain") {
 		cfg.HttpConfig.LetsEncryptDomain = mgmtLetsencryptDomain
 	}
-	if mgmtDataDir != "" {
+	if flags.Changed("datadir") {
 		cfg.Datadir = mgmtDataDir
 	}
```
bison: This is scary because the way it has been working is super confusing. The flag has a default, so this was basically never going to be `""` and the value in the config was probably never used. So if we change this behavior, people might suddenly get a new empty data dir for the server.

util/config/loader.go
```diff
@@ -0,0 +1,429 @@
...
+		}
+	}
+
+	v := viper.New()
+	if configFormat != "" {
+		v.SetConfigType(configFormat)
+	}
+	v.SetEnvPrefix(envPrefix)
+	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_", "-", "_"))
+	v.AllowEmptyEnv(true)
```
bison, line 130: This is problematic for non-string fields. For example: Previously `NB_METRICS_PORT` being empty meant the env var would essentially be ignored. Now it gets set to zero.

proxy/cmd/proxy/cmd/root.go
```diff
@@ -145,118 +138,68 @@ func SetVersionInfo(version, commit, buildDate, goVersion string) {
...
+	if cfg.ProxyToken == "" {
+		return fmt.Errorf("proxy token is required: set proxyToken or %s", envProxyToken)
 	}
 
-	level := logLevel
-	if debugLogs {
+	level := cfg.LogLevel
+	if debugLogs && (!cmd.Flags().Changed("log-level") || cmd.Flags().Changed("debug")) {
 		level = "debug"
 	}
```
bison, line 158: This might be what we want since debug is deprecated, but this is really confusing and seems like a behavior change. Previously if `NB_PROXY_DEBUG_LOGS` was true, you got debug logs. Now I guess log-level can override that?

bison (changes requested):
I really like the idea of this in general, but it's a risky change. I think we need to be really careful here and make sure we have exhaustive tests.

PR #7307, opened 2026-08-24
[management] Guard network map component encoding when the target peer is absent

bcmmbaga:
## Describe your changes

When network map components are computed for a peer that is no longer present in the account snapshot, the degraded components could carry an empty peer entry that the envelope encoder then tried to serialize.

The account-side degrade path now returns no peer entry instead of an empty one, and the encoder omits the target peer from the envelope when it is not available. Clients already resolve their own peer identity from `PeerConfig` and
tolerate a peers list without it, so the receiving side is unaffected.

### Checklist
- [x] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:


dmitri-netbird:
Since the peer has been deleted, we should be able to skip sending an empty update to it altogether, right? If so, we could return an error from `EncodeNetworkMapEnvelope`, something like ErrPeerIsGone, and skip calling `SendUpdate` in this case...

Issue #4470, opened 2025-09-09
Connectivity issues with Netbird on hosts running Kubernetes with Calico/Flannel CNI


MichalisDBA:
Description:

There is a connectivity issue when defining Access Control Policies on a host running Kubernetes with Calico or Flannel CNI plugins.

Steps to Reproduce:

1. Install the Netbird client on a host machine running Kubernetes (not in a container).
2. Ensure a CNI plugin (Calico or Flannel) is installed on the host, either manually or via Rancher.
3. Set up the Netbird client—it should connect to the management panel successfully.
4. Configure an Access Control Policy to allow another Netbird peer to reach this host (e.g., on port 22).
5. Attempt to connect from the other peer.

Observed Behavior:

- The connection fails despite the Access Control Policy being configured.
- After inspecting `iptables -t nat -L -n -v --line-numbers`, the issue appears to originate from the `CNI-HOSTPORT-MASQ` chain.

Details:

The CNI plugin marks all packets from Netbird originating from `127.0.0.1` and the loopback interface with `0x2000/0x2000`. These packets then hit the `POSTROUTING` rule in the `CNI-HOSTPORT-MASQ` chain, which MASQUERADEs them. This breaks connectivity from other Netbird peers.

Example:

```
Chain CNI-HOSTPORT-MASQ (1 references)
num   pkts bytes target     prot opt in     out     source               destination
1        0     0 MASQUERADE  all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x2000/0x2000
```

For debugging, adding a rule to bypass MASQUERADE for local Netbird packets restores connectivity:

`iptables -t nat -I CNI-HOSTPORT-MASQ 1 -s 127.0.0.1/32 -m mark --mark 0x2000/0x2000 -j RETURN`

Resulting chain:

```
Chain CNI-HOSTPORT-MASQ (1 references)
num   pkts bytes target     prot opt in     out     source               destination
1        0     0 RETURN     all  --  *      *       127.0.0.1            0.0.0.0/0            mark match 0x2000/0x2000
2        0     0 MASQUERADE  all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x2000/0x2000
```

Conclusion:

- All Netbird packets are being marked by the CNI plugin and subsequently MASQUERADEd.
- This prevents other Netbird peers from connecting, even though Access Control Policies are correctly set.
[...]

PR #7314, opened 2026-08-25
[client] Keep NetBird traffic out of third-party fwmark rules

lixmal:
## Describe your changes

Container network plugins, CNIs and other VPNs claim bits of the netfilter mark space for
themselves, and our marks share bits with several of them, so their rules act on our traffic.
This stops the proxy from marking packets that do not need it, and lets the range be moved
when a host has a conflicting rule.

- Stop marking the WireGuard proxy's sender raw sockets. They are bound to the loopback device and only send to the local WireGuard instance, which the local routing table resolves without a mark, while a rule matching on marks that rewrites their source makes WireGuard adopt the rewritten address as the peer endpoint
- Let the mark range be moved with `NB_FWMARK_BASE`, so a host running software that claims a bit we use can put our marks elsewhere. The value is the base of the range and the low byte identifies the individual mark, so `NB_FWMARK_BASE=0x11000` puts them at `0x11000`-`0x110FF`
- Add tests for the range parsing and for the sender sockets staying unmarked

The socket change covers both the eBPF proxy and the userspace UDP proxy, which share the
socket helper. The bind proxy used with userspace WireGuard never puts these packets on the
wire, so it is unaffected.

Changing the range while the client is not shut down gracefully leaves the routing rule of the
previous range behind, and it has to be removed by hand.

Related to #4470

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

- [x] I added/updated documentation for this change: https://github.com/netbirdio/docs/pull/951 documents the env var and the leftover routing rule
- [ ] Documentation is **not needed** for this change

## Summary by CodeRabbit

- **New Features**
  - Added configurable firewall-mark ranges through the `NB_FWMARK_BASE` environment setting.
  - Added support for identifying data-plane traffic marks.

- **Bug Fixes**
  - Raw sender sockets no longer receive firewall marks, improving routing behavior.
  - Invalid firewall-mark configuration values now safely fall back to defaults with a warning.

- **Tests**
  - Added coverage for firewall-mark parsing, range validation, default values, and socket mark behavior.

### DIFF
--- a/client/net/net_linux.go
+++ b/client/net/net_linux.go
@@ -21,15 +21,6 @@ func SetSocketMark(conn syscall.Conn) error {
 	return setRawSocketMark(sysconn)
 }
 
-// SetSocketOpt sets the SO_MARK option on the given file descriptor
-func SetSocketOpt(fd int) error {
-	if !AdvancedRouting() {
-		return nil
-	}
-
-	return setSocketOptInt(fd)
-}
-
 func setRawSocketMark(conn syscall.RawConn) error {
 	var setErr error
 
--- a/client/net/net_linux.go
+++ b/client/net/net_linux.go
@@ -51,5 +51,5 @@ func setRawSocketMark(conn syscall.RawConn) error {
 }
 
 func setSocketOptInt(fd int) error {
-	return syscall.SetsockoptInt(fd, syscall.SOL_SOCKET, syscall.SO_MARK, ControlPlaneMark)
+	return syscall.SetsockoptInt(fd, syscall.SOL_SOCKET, syscall.SO_MARK, int(ControlPlaneMark))
 }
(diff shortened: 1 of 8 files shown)


pappz (dismissed):

theodorsm:
@lixmal We should create docs for this env var and highlight the clean up: 
> Changing the range while the client is not shut down gracefully leaves the routing rule of the
previous range behind, and it has to be removed by hand.

client/net/net_linux.go
theodorsm: This function is dead now, perhaps we should remove it?

theodorsm (changes requested):

client/net/net_linux.go
lixmal: Fixed

theodorsm (approved):

PR #7319, opened 2026-08-25
[client] Move MDM enforcement logic into a shared Go layer

pappz:
## Describe your changes

The mobile bridges only carried the policy fetcher, leaving every enforcement decision to the native apps: the desktop derived its UI restrictions in the Wails service layer, the daemon kept the conflict machinery in the server package, and both mobile bridges duplicated the JSON fetch adapter. Anything the native side had to reimplement was a place for iOS and Android to drift apart.

Enforcement now lives in client/mdm and is consumed identically by all three platforms:

- conflicts.go holds the value-aware conflict checks lifted out of the daemon, so the same normalization (canonical URLs, PSK sentinel echo) applies wherever a config change is validated.
- restrictions.go derives the UI enforcement snapshot from a policy and renders it in the JSON shape the desktop frontend already consumes. The service-layer types become aliases, keeping one source of truth.
- jsonloader.go replaces the adapter that was copy-pasted into both bridges.
- changedetector.go moves change detection off the native side: the caller forwards the OS notification and asks whether the managed configuration actually changed, instead of diffing dictionaries itself.

The mobile bridges gain the enforcement the daemon already had. The Preferences getters resolve managed keys from the policy, so a naive UI shows the enforced value; Commit rejects a staged change that diverges from a managed key; NewAuth resolves the managed management URL before persisting the config and overlays the policy on it, so a login can no longer run against a URL the policy forbids. Android's profile mutations fail closed when disableProfiles is set.

NewAuth takes the fetcher as a required argument rather than keeping a policy-blind overload: the apps consume this code as a submodule, so a compile error at the bump is the point. The mobile PSK getter is replaced by a presence check — the key has no reason to cross the bridge, and not returning it means the native side needs no redaction sentinel of its own.

### Checklist
- [ ] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

### DIFF
--- a/client/android/client.go
+++ b/client/android/client.go
@@ -101,7 +101,8 @@ type Client struct {
 	// passes this loader to the resolved Config so applyMDMPolicy
 	// picks up the active overlay. Nil means "MDM enforcement off
 	// for this Client".
-	mdmLoader *mdm.Loader
+	mdmLoader   *mdm.Loader
+	mdmDetector *mdm.ChangeDetector
 
 	// Identifies the running profile for the SSO login hint; see profile_state.go.
 	cfgPath string
--- a/client/android/client_mdm.go
+++ b/client/android/client_mdm.go
@@ -1,18 +1,11 @@
-//go:build ios
+//go:build android
 
-package NetBirdSDK
+package android
 
 import (
 	"github.com/netbirdio/netbird/client/internal/profilemanager"
 	"github.com/netbirdio/netbird/client/mdm"
 )
-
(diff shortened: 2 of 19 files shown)


riccardomanfrin (approved):
mobile `Preferences` getters now each call `mdmLoader.Load()`,
which loads from disk, parses JNI/ObjC and log-Infofs "MDM enrolled" on every read.

Not sure if there's a way around this to reduce the churn

PR #7329, opened 2026-08-26
[client, ios] Fix context cancellation during restart

pappz:
## Describe your changes

A user switching from wifi to cellular lost all Internet traffic until they turned NetBird off. An iOS debug bundle caught it: the network-change restart calls `Stop()`, which cancelled a shared context and returned immediately instead of waiting for the run loop. The teardown then ran long (`failed to remove WireGuard interface utun6: timeout`, ~5s), so the restart's `start` landed on the dying context and failed with `failed creating connection to Management Service: context canceled` — 2ms after the old run finally exited. The tunnel stayed up with no engine behind it, black-holing traffic.

- `Stop()` now drives `ConnectClient.Stop()`, which cancels and waits for the run loop to exit.
- `Run()` owns its cancel locally, so a concurrent call can't cancel it through the shared field.
- `IsLoginRequired()` / `LoginForMobile()` stop overwriting the shared cancel — `restartClient` calls into that path. `LoginForMobile`'s cancel moves into the goroutine that outlives the call so the OAuth wait isn't cut short.
- Android gets the same change **preventively**: same structural defect, but no auto-restart on network change and no interface-removal timeout, so it isn't reachable there today.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal lifecycle fix, no user-facing surface changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved run start and stop handling for more reliable cancellation.
  * Reduced the risk of inconsistent lifecycle states when starting or stopping operations.
  * Ensured cancellation actions are applied atomically during run transitions.

### DIFF
--- a/client/android/client.go
+++ b/client/android/client.go
@@ -41,6 +41,8 @@ const (
 	AnonymizeLevelStrict  = nbAnonymize.LevelStrictString
 )
 
+const stopRunWaitTimeout = 20 * time.Second
+
 // TunAdapter export internal TunAdapter for mobile
 type TunAdapter interface {
 	device.TunAdapter
@@ -89,6 +91,8 @@ type Client struct {
 	stateMu       sync.RWMutex
 	connectClient *internal.ConnectClient
 	config        *profilemanager.Config
+	runGeneration uint64
+	runDone       chan struct{}
 	cacheDir      string
 	// Identifies the running profile for the SSO login hint; see profile_state.go.
 	cfgPath string
@@ -112,13 +116,33 @@ type Client struct {
 	extendCancel context.CancelFunc
 }
 
-func (c *Client) setState(cfg *profilemanager.Config, cacheDir string, cfgPath string, cc *internal.ConnectClient) {
+func (c *Client) beginRun(cancel context.CancelFunc) (uint64, chan struct{}) {
+	done := make(chan struct{})
+
+	c.stateMu.Lock()
+	c.runGeneration++
+	generation := c.runGeneration
+	c.runDone = done
+	c.stateMu.Unlock()
+
+	c.ctxCancelLock.Lock()
+	c.ctxCancel = cancel
+	c.ctxCancelLock.Unlock()
+
+	return generation, done
+}
+
+func (c *Client) publishState(generation uint64, cfg *profilemanager.Config, cacheDir string, cfgPath string, cc *internal.ConnectClient) bool {
 	c.stateMu.Lock()
 	defer c.stateMu.Unlock()
+	if c.runGeneration != generation {
+		return false
+	}
 	c.config = cfg
 	c.cacheDir = cacheDir
 	c.cfgPath = cfgPath
 	c.connectClient = cc
+	return true
 }
 
 func (c *Client) stateSnapshot() (*profilemanager.Config, string, *internal.ConnectClient) {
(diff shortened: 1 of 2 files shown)


client/android/client.go
```diff
@@ -232,31 +259,56 @@ func (c *Client) RunWithoutLogin(platformFiles PlatformFiles, dns *DNSList, dnsR
...
+	c.ctxCancel = nil
+	c.stateMu.Unlock()
+
+	if cancel != nil {
+		cancel()
 	}
 
-	c.ctxCancel()
+	if cc != nil {
+		if err := cc.Stop(); err != nil {
```
riccardomanfrin: This is not time bounded.. if `<-c.runExited` stalls we wait forever

client/android/client.go
```diff
@@ -112,13 +119,30 @@ type Client struct {
 	extendCancel context.CancelFunc
 }
 
-func (c *Client) setState(cfg *profilemanager.Config, cacheDir string, cfgPath string, cc *internal.ConnectClient) {
+func (c *Client) beginRun(cancel context.CancelFunc) (uint64, chan struct{}) {
```
riccardomanfrin: Although it's not possible with today paths (Swift/Kotlin do Stop before any Start), we are allowing multiple `Run` / `RunWithoutLogin` to mess up on the existing engine context/state without interleaving a Stop in between, so this is safe because we disallow multiple calls to the `Run` api rather than because it self protects against it.

client/android/client.go
```diff
@@ -232,31 +259,56 @@ func (c *Client) RunWithoutLogin(platformFiles PlatformFiles, dns *DNSList, dnsR
...
+	ctx := runCtx
 
 	// todo do not throw error in case of cancelled context
 	ctx = internal.CtxInitState(ctx)
 	connectClient := internal.NewConnectClient(ctx, cfg, c.recorder,
 		internal.WithNetEvents(c.netMgr))
-	c.setState(cfg, cacheDir, cfgFile, connectClient)
+	if !c.publishState(generation, cfg, cacheDir, cfgFile, connectClient) {
+		log.Infof("run: superseded by a newer run, aborting startup")
+		return nil
```
riccardomanfrin: We are reporting a clean exit for a run that never started.

client/android/client.go
```diff
@@ -232,31 +259,56 @@ func (c *Client) RunWithoutLogin(platformFiles PlatformFiles, dns *DNSList, dnsR
...
+	}
 	return connectClient.RunOnAndroid(c.tunAdapter, c.iFaceDiscover, c.networkChangeListener, slices.Clone(dns.items), dnsReadyListener, stateFile, cacheDir)
 }
 
 // Stop the internal client and free the resources
 func (c *Client) Stop() {
```
riccardomanfrin, line 252: Multiple Stops if ever occur while a first one is flying, will return immediately reporting that we are stopped whereas we are not

client/android/client.go
```diff
@@ -232,31 +259,56 @@ func (c *Client) RunWithoutLogin(platformFiles PlatformFiles, dns *DNSList, dnsR
...
-	c.ctxCancelLock.Lock()
-	defer c.ctxCancelLock.Unlock()
-	if c.ctxCancel == nil {
-		return
+	c.stateMu.Lock()
+	c.runGeneration++
+	cc := c.connectClient
+	done := c.runDone
+	cancel := c.ctxCancel
+	c.connectClient = nil
```
riccardomanfrin: We are cleaning the state connectClient/ctxCancel/runDone when Stop is called, but we don't do it when `RunOnAndroid`/`Run` return because of errors. 

Not sure how this might affect calls to  `DebugBundle` (`cfg, cacheDir, cc := c.stateSnapshot()`) or others..

riccardomanfrin (commented):
Haven't checked on 7330.. so not sure if it addresses all these concerns and if we can live with them for now... if so let me know and I can stamp.

riccardomanfrin (dismissed):

riccardomanfrin (approved):

PR #7340, opened 2026-08-27
[management,infrastructure] Add the Autonoma test-data endpoint

Marcos-Craviotto:
## Describe your changes

Autonoma is an agentic end-to-end testing platform: it builds a preview
environment for every pull request and has an agent exercise the app in a real
browser. Its Environment Factory is what makes a run repeatable - before each
run it asks the application to create fresh, isolated test data, and afterwards
to remove it. This endpoint is NetBird's side of that contract.

Here that means a whole account: an owner who can sign in, members, peers,
groups, policies, routes, DNS, a proxied service, an agent-network gateway.
Driving it through the REST API takes dozens of ordered calls per run and
nothing guarantees the teardown finishes, so one failed run leaves the
deployment dirty for the next.

`POST /api/autonoma` mounts itself only when both `AUTONOMA_SHARED_SECRET` and
`AUTONOMA_SIGNING_SECRET` are set - a deployment that does not opt in never
registers the route, which is the production guard. Every request carries an
HMAC of its exact body, and the teardown token is signed with the second
secret, so a `down` can only ever remove what its own `up` created.

The factories go through the same managers the REST handlers call, so `AddPeer`
allocates the overlay IP and the DNS label, `SavePolicy` bumps the network
serial, and the embedded IdP gets a real user the agent can sign in as. A raw
insert would skip all of that, and the suite would end up asserting against a
shape the product never produces. Two factories cannot use their manager - a
job needs a live gRPC stream, a proxy needs a process sending heartbeats - and
the package comment names both.

Teardown deletes the account and lets the graph go with it, including rows a
test created mid-run that no recipe named. The few tables that hang off an
account without a GORM association get scoped, idempotent deletes in
`sql_store_testdata.go`, deliberately not on the `Store` interface.

<details>
<summary>How a run uses this endpoint</summary>

https://github.com/user-attachments/assets/2f974921-524c-41bb-830c-5aafdf506aeb

</details>

**Tested:** the full recipe seeded and torn down against a deployed preview of
this branch. The unit tests added here cover the secrets guard - the route is
not registered unless both variables are set, and the two must differ - and the
signature rejections: unsigned, signed with the wrong secret, and a body changed
after it was signed.

None yet 

### Checklist
- [ ] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

The endpoint is not a user-facing feature: it does not exist unless an operator
sets two environment variables, and whoever sets them is driving it from a test
runner rather than reading the manual. What a maintainer needs - the secrets
guard, what teardown does and does not cascade into, and the two places a
factory deliberately does not use its manager - is in the package comment, next
to the code it describes.

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7340?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added support for preview deployments using a containerized combined server.
  - Preview instances can be configured through environment variables, including public URL, database, logging, authentication, and encryption settings.
  - Added a secure endpoint for creating and removing comprehensive test data across accounts, users, networking, DNS, proxy, and agent-network features.
  - Test-data requests require distinct shared and signing secrets and support automatic cleanup.
  - Preview deployments no longer use default bootstrap owner credentials; owner access must be explicitly configured.


jnfrati:
@Marcos-Craviotto could you move this one to a draft for now? Thanks!

PR #7380, opened 2026-09-01
[management] Keep embedded IdP deployments on a single account

bcmmbaga:
## Describe your changes

The embedded IdP only supports single-account deployments. This change configures combined deployments with the default single-account domain, rejects embedded IdP configurations without that domain, keeps the configured domain when the existing account has none, and stops idp-migrate from running against an instance that holds more than one account while giving the remaining account the domain attributes single-account mode resolves against.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7380?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Bug Fixes
- Embedded identity provider startup now reports an error when single-account mode lacks a configured domain.
- Single-account migrations reject environments with multiple accounts and conflicting or unusable domains.
- Existing valid account domains are preserved during migration.
- Migration now validates domain settings before making changes.

## New Features
- Added single-account domain configuration through the migration CLI flag or environment variable.
- Migration tools normalize configured domains and assign a default self-hosted domain when no account domain exists.

## Documentation
- Documented single-account domain configuration and conflict behavior.

### DIFF
--- a/management/server/account.go
+++ b/management/server/account.go
@@ -15,10 +15,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/netbirdio/netbird/management/internals/modules/reverseproxy/service"
-	"github.com/netbirdio/netbird/management/server/job"
-	"github.com/netbirdio/netbird/shared/auth"
-
 	cacheStore "github.com/eko/gocache/lib/v4/store"
 	"github.com/eko/gocache/store/redis/v4"
 	"github.com/rs/xid"
@@ -30,6 +26,7 @@ import (
 	"github.com/netbirdio/netbird/formatter/hook"
 	"github.com/netbirdio/netbird/idp/dex"
 	"github.com/netbirdio/netbird/management/internals/controllers/network_map"
+	"github.com/netbirdio/netbird/management/internals/modules/reverseproxy/service"
 	nbconfig "github.com/netbirdio/netbird/management/internals/server/config"
 	"github.com/netbirdio/netbird/management/server/account"
 	"github.com/netbirdio/netbird/management/server/activity"
@@ -40,6 +37,7 @@ import (
 	"github.com/netbirdio/netbird/management/server/idp"
 	"github.com/netbirdio/netbird/management/server/integrations/integrated_validator"
 	"github.com/netbirdio/netbird/management/server/integrations/port_forwarding"
+	"github.com/netbirdio/netbird/management/server/job"
 	nbpeer "github.com/netbirdio/netbird/management/server/peer"
 	"github.com/netbirdio/netbird/management/server/permissions"
 	"github.com/netbirdio/netbird/management/server/permissions/modules"
@@ -51,6 +49,7 @@ import (
 	"github.com/netbirdio/netbird/management/server/types"
 	"github.com/netbirdio/netbird/management/server/util"
 	"github.com/netbirdio/netbird/route"
+	"github.com/netbirdio/netbird/shared/auth"
 	nbdomain "github.com/netbirdio/netbird/shared/management/domain"
 	"github.com/netbirdio/netbird/shared/management/networkmap/nmdata"
 	"github.com/netbirdio/netbird/shared/management/status"
@@ -237,6 +236,10 @@ func BuildManager(
 	accountsCounter, err := store.GetAccountsCounter(ctx)
 	if err != nil {
 		log.WithContext(ctx).Error(err)
+	}
+
+	if IsEmbeddedIdp(idpManager) && accountsCounter > 1 {
+		log.WithContext(ctx).Warnf("embedded IdP requires a single account, found %d", accountsCounter)
 	}
 
 	// enable single account mode only if configured by user and number of existing accounts is not grater than 1
@@ -1593,7 +1596,10 @@ func (am *DefaultAccountManager) updateUserAuthWithSingleMode(ctx context.Contex
 	if err != nil {
 		return err
 	}
-	userAuth.Domain = domain
+	// Keep the configured single account domain when the existing account has none
+	if domain != "" {
+		userAuth.Domain = domain
+	}
 
 	log.WithContext(ctx).Debugf("overriding JWT Domain and DomainCategory claims since single account mode is enabled")
 	return nil
@@ -1838,6 +1844,7 @@ func (am *DefaultAccountManager) getAccountIDWithAuthorizationClaims(ctx context
 
 	return am.addNewPrivateAccount(ctx, domainAccountID, userAuth)
 }
+
 func (am *DefaultAccountManager) getPrivateDomainWithGlobalLock(ctx context.Context, domain string) (string, context.CancelFunc, error) {
 	domainAccountID, err := am.Store.GetAccountIDByPrivateDomain(ctx, store.LockingStrengthNone, domain)
 	if handleNotFound(err) != nil {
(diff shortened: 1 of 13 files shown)


management/server/idp/migration/migration.go
```diff
@@ -233,3 +237,97 @@ func PopulateUserInfo(s Server, idpManager idp.Manager, dryRun bool) error {
...
+		return fmt.Errorf("failed to count accounts: %w", err)
+	}
+	// The count is checked again here: it is read long after RequireSingleAccount, and marking an
+	// arbitrary account as the primary one for the domain would be wrong.
+	switch {
+	case accountsCounter == 0:
+		log.Info("no accounts yet, nothing to prepare for single account mode")
+		return nil
+	case accountsCounter > 1:
+		return errMultipleAccounts(accountsCounter)
```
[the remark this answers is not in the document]
bcmmbaga: This is an offline migration, management server must not be running while it executes

management/server/idp/migration/migration.go
```diff
@@ -233,3 +237,97 @@ func PopulateUserInfo(s Server, idpManager idp.Manager, dryRun bool) error {
...
+	}
+
+	if accountsCounter > 1 {
+		return errMultipleAccounts(accountsCounter)
+	}
+
+	return nil
+}
+
+func errMultipleAccounts(accountsCounter int64) error {
```
dmitri-netbird, line 265: nit (ok to ignore here, or fix in a follow-up): it's a bit more useful to have a common error defined, and then embed it and add details when returning errors. This would allow us to use `errors.Is()` and `errors.As()` when we need to handle errors in a particular way (or verify error handling in tests).

management/server/idp/migration/migration.go
```diff
@@ -233,3 +237,97 @@ func PopulateUserInfo(s Server, idpManager idp.Manager, dryRun bool) error {
...
+		return fmt.Errorf("failed to get the existing account: %w", err)
+	}
+
+	isPrimary, accountDomain, err := s.Store().IsPrimaryAccount(ctx, accountID)
+	if err != nil {
+		return fmt.Errorf("failed to read domain attributes of account %s: %w", accountID, err)
+	}
+
+	// Keep the account's own domain only when it can be resolved.
+	domain := strings.ToLower(accountDomain)
```
dmitri-netbird: I'm not sure I understand the logic: we drop the domain name the user specified on the command line in favour of the one on the account here, right?

I think there's a couple of issues here:
 * this is done silently
 * this might be (likely even) a desired behaviour if a default single account domain was used, but I don't think this is something the user would expect if they passed the domain name on the command line. I think a proper way to handle this case would be to error out with a message describing the conflict in domain names.

dmitri-netbird: oh, and I also think we should bail in case the domain name on the account is invalid instead of silently replacing it with the name passed on the cli.

dmitri-netbird: In short, I think we either should use the domain name passed in the cli (or the default name), or use the domain name on the account. Mixing them is confusing and would be hard to explain too.

management/server/idp/migration/store.go
```diff
@@ -60,6 +60,20 @@ type Store interface {
 	// CheckSchema verifies that all tables and columns required by the migration
 	// exist in the database. Returns a list of problems; an empty slice means OK.
 	CheckSchema(checks []SchemaCheck) []SchemaError
+
+	// GetAccountsCounter returns the total number of accounts in the store.
+	GetAccountsCounter(ctx context.Context) (int64, error)
```
dmitri-netbird, line 65: `GetAnyAccountID` is kinda weird/non-deterministic and is only really useful when only a single account exists. I think a better approach would be to add a call to return all existing account ids; it can be used to verify the number of accounts in the system *and* then to retrieve an account by id.

management/server/account.go
```diff
@@ -1593,7 +1596,10 @@ func (am *DefaultAccountManager) updateUserAuthWithSingleMode(ctx context.Contex
 	if err != nil {
 		return err
 	}
-	userAuth.Domain = domain
+	// Keep the configured single account domain when the existing account has none
+	if domain != "" {
```
dmitri-netbird, line 1600: tangential: this seems weird; is that even valid when an embedded idp is used?

management/server/idp/migration/store.go
bcmmbaga: It's only reached when the count is exactly one. `EnsureSingleAccountDomain` returns early on zero and errors on more than one just above the call

management/server/account.go
bcmmbaga: Yes is valid. Old accounts from external IdPs can have an empty domain which would overwrite the configured single-account domain. That then fails validation and sends the next login down the new-account path creating a second account.

management/server/idp/migration/migration.go
bcmmbaga: Agreed, the precedence is confusing and the flag name doesn't help, it reads as authoritative but is really only a fallback

An empty account domain isn't a conflict.  It's the main case here to fix as erroring would block the migrations we're trying to enable.

So the three cases:
- account domain empty → use the CLI value or the default, nothing is being overridden
- account domain set and resolvable → keep it, if a domain was explicitly passed and differs, error and describe the conflict
- account domain set but unresolvable (e.g. a single label like `corp`) → error rather than silently replacing it as you suggest

wdyt?

dmitri-netbird: sgtm

management/server/idp/migration/store.go
dmitri-netbird: it's an api on the interface, and we can't know how it's going to be used in future. Api that have narrow use are hard to document and use and lead to bugs down the road. It's easier (esp. in this case) to introduce a clearer api then to deal with changing and fixing api later.

bcmmbaga: I understand the concern, but it’s used in many places for almost the same purpose so changing it would require updating all references.

dmitri-netbird: I didn't realize, I thought a new api call is being added. sg then.

management/server/idp/migration/migration.go
```diff
@@ -270,64 +276,124 @@ func NormalizeSingleAccountDomain(singleAccountDomain string) (string, error) {
...
+	isPrimary     bool
+	skip          bool
+}
+
+// planSingleAccountDomain decides what the account's domain attributes should become. It reads
+// only, so it can run both as a preflight and as the first half of the update.
+func planSingleAccountDomain(s Server, singleAccountDomain string) (singleAccountDomainPlan, error) {
 	ctx := context.Background()
 
+	// An empty value means the operator did not pick a domain, so the default is only a fallback.
```
dmitri-netbird, line 356: nit: I found the comment confusing: maybe "An empty value means the operator did not pick a domain, `DefaultSingleAccountDomain` will only be used as a fallback."

dmitri-netbird (approved):

PR #7398, opened 2026-09-02
[client] Gate settings updates on value, not on field presence

riccardomanfrin:
## Describe your changes

The update-settings kill switch (`--disable-update-settings` / `NB_DISABLE_UPDATE_SETTINGS` / MDM `DisableUpdateSettings`) decided what counted as a "change" by whether a field was *present* in the request. The CLI fills the whole config surface of `SetConfigRequest`/`LoginRequest` from its flags and env on every `netbird up`, so a client configured by environment restates its own configuration on every start and tripped the gate every time — and since the same gate guards `Login` inside the CLI's backoff loop, `netbird up` never completed at all. Reproduced against a daemon started with the switch on: seven `retrying Login ... update settings are disabled` lines and exit 1 after ~30s, for a request that asked for no change.

**1. The fix.** Both gates compare values. `Config.WouldChange` runs the same diff logic as a real update (`Config.apply`) against a copy of the stored config, so a request restating what the profile already holds is allowed, one that diverges is refused exactly as before, and an unevaluable dry run fails closed. The decision is re-taken under `guardedConfigMu` before the write, like the privileged-change one, so a login that was a no-op when checked cannot be written after a concurrent writer moved the profile.

**2. Behaviour changes worth reviewing.** These are deliberate and visible:

- The refusal now travels as `codes.FailedPrecondition` instead of   `codes.Unavailable`, and `netbird up` **fails** on it instead of logging a  warning and carrying on. Unavailable means "the daemon cannot serve this  call", which is why the CLI downgraded it and why `client/ui/services`  classifies it as an unreachable daemon — both wrong for a daemon that  answered and refused. It also matches what the MDM gate already returns for a  managed field. Before this, `netbird up --enable-rosenpass` against a gated  daemon connected while silently dropping the flag.
- Config reads no longer write. `readConfig` used to persist whatever `apply()`  had filled in, and to create the file outright when absent, which made every  reader a writer — including a gate deciding whether to *refuse* a request.  Identity provisioning is now an explicit step (`Config.EnsureIdentity`) and  the callers that provision persist the result themselves.
- The two readers are renamed for what they do: `ReadConfig` →  `ReadOrGenerateConfig`, `GetConfig` → `GetExistingConfig`. Both are in  `client/internal/profilemanager`, so no consumer outside `client/` is  affected; the gomobile-bound surface is unchanged (no exported signature in  `client/android` or `client/ios` moved), so the android-client and ios-client  repos need no code change, only a regenerated `.aar` / `.xcframework`.

**3. Pre-existing bugs this uncovered.** Each has a test that fails without it:

- `apply` compared URLs as raw strings, so a trailing slash, a different host  case, an implicit `:443` or a zero-padded `:0443` read as a new value. This  one is load-bearing for the reported case: `NB_MANAGEMENT_URL` with a
  trailing slash reproduces the deadlock on its own.
- The admin URL is opened in a browser rather than dialed, so its path is part  of its identity; it gets the endpoint-plus-path comparison, the management URL  the endpoint-only one.
- `UpdateConfig` did not collapse the redacted pre-shared key, unlike its two  sibling functions, so a UI round-trip of the mask replaced the stored key with  asterisks.
- `apply` wrote the incoming sync message version through the stored pointer  without checking it was there — a nil dereference reachable from a request  handler.
- The mTLS certificate paths were assigned and marked as changed without being  compared.
- Logging out a profile with no identity failed with `incorrect key size: 0`  (the second logout of the same profile: a logout clears the keys in place),  and was refused with `PermissionDenied` for an unprivileged caller on any  profile whose `ServerSSHAllowed` is unset — which is every legacy profile.
- The mobile logout read the config with the generating reader right after  checking the file exists; the two are not atomic, so a profile removed in  between was resolved from the defaults and recreated by the write that follows.
- `netbird login` kept retrying a final refusal for 30s: the list of terminal  gRPC codes was duplicated between `up` and `login` and had already diverged.

**4. Deduplication.** Three notions of "same management URL" lived in one request path — the config layer, the privileged-change gate and the MDM conflict check — and only one was right; they now share `profilemanager.SameServiceURL`. The terminal-code list is now `terminalLoginError`, once, next to `WithBackOff`.

**Manual verification** against a throwaway daemon with the switch on: a request restating the stored config passes in every URL spelling; a real change is refused in 0.05s (not after the 30s backoff) and leaves the config file
byte-identical; `--enable-rosenpass` is refused too, so the gate judges every field and not just the URL; with the switch off the same change is written; and a self-hosted container whose config was provisioned out of band starts with the switch on.

## Tested on Android (emulator, real app)

Built the gomobile binding and the APK from this branch and exercised the profile
lifecycle through the app's own UI, checking the stored config after every step.

image

**Provenance**

```
$ git -C netbird rev-parse --short HEAD
9f0d68d12
$ ./build-android-lib.sh          # gomobile bind -> gomobile/netbird.aar (107.8 MB)
$ ./gradlew assembleDebug         # app/build/outputs/apk/debug/app-debug.apk
$ adb install -r app/build/outputs/apk/debug/app-debug.apk
Success
```

**1. A profile created from the app carries its identity** (the review's first point:
AddProfile used the bare constructor and wrote a profile with no keys)

```
$ adb shell run-as io.netbird.client ls files/profiles
0050de9378c328e9ab677a691690a063.json

$ adb shell run-as io.netbird.client cat files/profiles/0050de93....json
Name:           prtesti
PrivateKey set: True (len 44)
SSHKey set:     True (len 119)
ManagementURL:  api.netbird.io:443
```

**2. No optional bool is stored as null** — the invariant this PR introduces, as it
lands in a real app profile:

```
"EnableSSHRoot": false,                  "NetworkMonitor": false,
"EnableSSHSFTP": false,                  "ServerSSHAllowed": false,
"EnableSSHLocalPortForwarding": false,   "RemoteJobsAllowed": false,
"EnableSSHRemotePortForwarding": false,  "DisableNotifications": true,
"DisableSSHAuth": false,                 "SSHJWTCacheTTL": 0

nulls in the whole file: ['SyncMessageVersion', 'DNSLabels', 'NATExternalIPs']
```

`SyncMessageVersion` stays three-state on purpose (absent means the client pins no
version); the other two are slices.

**3. Logout clears the identity, and the profile still loads** (the review's second
point: a logged-out profile must keep deserializing)

```
BEFORE logout: PrivateKey len: 44 | SSHKey len: 119 | MgmtURL: api.netbird.io:443
AFTER  logout: PrivateKey len: 0  | SSHKey len: 0   | MgmtURL: api.netbird.io:443
nulls: ['SyncMessageVersion', 'DNSLabels', 'NATExternalIPs']
```

Switching onto the logged-out profile then works, the app shows it as Disconnected,
and — the point of "reads are pure" — loading it does **not** mint keys:

```
$ adb shell run-as io.netbird.client cat files/profiles/0050de93....json
PrivateKey len: 0 | SSHKey len: 0
$ adb logcat -d | grep -iE "exception|fatal|error" | grep -iE "netbird|profile|config"
(no output)
```

**4. The round trip leaves the other profile untouched.** Switched default → prtesti →
default → prtesti → default, logged out prtesti in between, then reconnected:

```
100.115.57.64/16   Connected   default
PrivateKey len: 44 | SSHKey len: 119
```

Same overlay address as before the round trip, so the peer identity survived and
nothing re-registered.

Not covered here: a fresh SSO login on the logged-out profile, which would enroll a
new peer in a real account.

No public issue. Reported by a customer running `netbirdio/kubernetes-operator`
with a read-only rootfs: the daemon answered every login attempt with
`codes.Unavailable`. The presence-based gates on the merge base:
https://github.com/netbirdio/netbird/blob/c3cf7c0c3726f65dfed2fb3e95a7732ba2be2f83/client/server/server.go#L477
and
https://github.com/netbirdio/netbird/blob/c3cf7c0c3726f65dfed2fb3e95a7732ba2be2f83/client/server/server.go#L628

### Checklist
- [x] Is it a bug fix
- [x] It is a refactor

## Documentation
Select exactly one:

The documented meaning of the kill switch is unchanged: settings cannot be changed while it is on. What changes is not documented anywhere — the gRPC status code the local daemon socket returns for a refusal, and the fact that `netbird up` now exits non-zero on it instead of warning and continuing. No flag, environment variable or config field is added, removed or renamed, and the public API surface is untouched. Worth flagging for review rather than for the docs: a GUI that wants a localised message for a refusal needs `FailedPrecondition` added to the classifier in `client/ui/services/errors.go`, which is not in this PR.

N/A


pappz (changes requested):
1. ServiceManager.AddProfile was left out of the provisioning migration (client/internal/profilemanager/service.go). It still calls createNewConfig, which under the new rules no longer generates keys, so it writes a config to disk with no identity 

2. ConfigFromJSON now fails with ErrConfigWithoutIdentity when PrivateKey or SSHKey is empty. The mobile LogoutProfile clears exactly those two fields in place , so on iOS/tvOS, the JSON of a logged-out profile now makes SetConfigFromJSON / Auth.SetConfigFromJSON / copyConfig error out where they used to silently generate a new key. 

5. FailedPrecondition is not classified in client/ui/services/errors.go — the author flags this himself, but it isn't in the PR, so the GUI will show a generic error.

6. Minor: profilemanager.SameServiceURL compares the scheme with ==, while util.SameServiceURL uses EqualFold. No practical impact (url.Parse lowercases the scheme), but two identically named functions with two different rules is a bit of a trap.


Please test this PR on Android and iOS too! This can cause broken profile switch, logout, login.

PR #7408, opened 2026-09-03
[client] Fix browser login popup show from go

pappz:
## Describe your changes

On macOS the SSO browser-login popup could stay invisible and the browser never opened: the popup was created hidden and relied on its own webview to show itself and launch the browser, but a hidden WKWebView can be throttled or suspended before that code runs. The session-expiration dialog then stayed disabled until the PKCE flow timed out.

Dialog windows are now shown from Go once the frontend reports it has painted, with a fallback timer armed at creation, and the browser is launched from Go after the popup is visible. On the fallback path the dialog appears blank until the webview wakes up; the webview background fix for that lives in the Wails fork and comes as a follow-up.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7408?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when displaying browser login, session expiration, installation progress, welcome, and error windows.
  * Auxiliary windows now appear centered on the appropriate screen and are less likely to open before they are ready.
  * Improved window cleanup, event handling, and restoration when multiple auxiliary windows are open.
  * The SSO browser now opens after the browser login window is displayed.
  * Closing installation progress now correctly restores previously hidden windows.
  * Improved consistency of window appearance across supported themes and operating systems.

### DIFF
--- a/client/ui/services/windowmanager.go
+++ b/client/ui/services/windowmanager.go
@@ -32,7 +32,7 @@ const EventSettingsOpen = "netbird:settings:open"
 
 const EventWindowPainted = "netbird:window-painted"
 
-const paintedFallback = 2 * time.Second
+const paintedFallback = 3 * time.Second
 
 const headlessTeardownDelay = 2 * time.Second
 
@@ -123,6 +123,7 @@ type WindowManager struct {
 	pendingTab     map[uint]string
 	pendingEmits   map[uint][]string
 	fallbackTimers map[uint]*time.Timer
+	afterShow      map[uint]func()
 	headlessMain   bool
 	headlessTimer  *time.Timer
 	// recenterOnShow is set only on the minimal-WM/XEmbed path, where the WM neither centers nor
@@ -142,6 +143,7 @@ func NewWindowManager(app *application.App, mainWindow *application.WebviewWindo
 		pendingTab:     map[uint]string{},
 		pendingEmits:   map[uint][]string{},
 		fallbackTimers: map[uint]*time.Timer{},
+		afterShow:      map[uint]func(){},
 	}
 	s.watchPainted()
 	s.watchTriggerLogin()
@@ -215,7 +217,6 @@ func (s *WindowManager) OpenSettings(tab string) {
 // OpenBrowserLogin shows the SSO popup, creating it on first use.
 func (s *WindowManager) OpenBrowserLogin(uri string) {
 	s.mu.Lock()
-	defer s.mu.Unlock()
 	if s.browserLogin == nil {
 		startURL := "/#/dialog/browser-login"
 		if uri != "" {
@@ -228,8 +229,8 @@ func (s *WindowManager) OpenBrowserLogin(uri string) {
 		opts.InitialPosition = application.WindowCentered
 		// Open on the active (where users cursor is) display, like the session-expiration dialog.
 		opts.Screen = s.getScreenBasedOnCursorPosition()
-		s.browserLogin = s.app.Window.NewWithOptions(opts)
-		bl := s.browserLogin
+		bl := s.app.Window.NewWithOptions(opts)
+		s.browserLogin = bl
 		bl.OnWindowEvent(events.Common.WindowClosing, func(_ *application.WindowEvent) {
 			s.mu.Lock()
 			// Only a live user red-X still has this registered; programmatic closers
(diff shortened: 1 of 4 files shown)


client/ui/services/windowmanager.go
```diff
@@ -240,20 +241,48 @@ func (s *WindowManager) OpenBrowserLogin(uri string) {
...
+
+func (s *WindowManager) openBrowser(uri string) {
+	if uri == "" {
+		return
 	}
-	s.centerOnCursorScreen(s.browserLogin)
-	s.browserLogin.Show()
-	s.browserLogin.Focus()
+	go func() {
+		if err := openURL(uri); err != nil {
```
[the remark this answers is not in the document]
pappz: It come from PKCE flow, from an authenticated server

client/ui/services/windowmanager.go
```diff
@@ -362,36 +398,50 @@ func (s *WindowManager) CloseRenewFlow() {
...
 
 func (s *WindowManager) CloseInstallProgress() {
 	s.mu.Lock()
 	w := s.installProgress
 	s.installProgress = nil
+	// The guarded WindowClosing handler no-ops on a programmatic close, so restore
+	// here — but only if a popup was actually open, since hiddenForLogin is shared
+	// with OpenBrowserLogin.
+	if w != nil {
+		s.restoreHiddenWindowsLocked()
```
riccardomanfrin, line 443: this might restore windows of another owner.

`hiddenForLogin` doesn't have one and `w != nil` only checks there was an install in progress, but not who it belongs to. 

To fix tag/check the owner on the entry

client/ui/services/windowmanager.go
```diff
@@ -543,15 +602,24 @@ func (s *WindowManager) armReady(w *application.WebviewWindow) {
...
-			s.markReady(w)
-		})
+	id := w.ID()
+	timer := time.AfterFunc(paintedFallback, func() {
 		s.mu.Lock()
-		s.fallbackTimers[w.ID()] = timer
+		ready := s.ready[id]
 		s.mu.Unlock()
+		if ready {
+			return
```
riccardomanfrin, line 611: Different meaning for ready.
The problem arises from line 641 where ready is defined as `ready := w != nil && s.ready[w.ID()]`

So in 611 it means "it's painted", whereas in 641 it means "frontend mounted and listening".

The problem is that in 614 markReady collapses the two.

client/ui/services/windowmanager.go
```diff
@@ -543,15 +602,24 @@ func (s *WindowManager) armReady(w *application.WebviewWindow) {
 	if w == nil {
 		return
 	}
-	w.RegisterHook(events.Common.WindowRuntimeReady, func(_ *application.WindowEvent) {
-		timer := time.AfterFunc(paintedFallback, func() {
-			log.Warnf("window %q never reported a first render, showing it anyway", w.Name())
-			s.markReady(w)
-		})
+	id := w.ID()
+	timer := time.AfterFunc(paintedFallback, func() {
```
riccardomanfrin, line 606: ready might not make it in time with the 3 seconds budget of the armReady, with the webview boot time and main time budget for ReadySignal:

It might be a good move to rearm or extend it it when WindowRuntimeReady fire, so to give main it's own budget

client/ui/services/windowmanager.go
```diff
@@ -668,6 +737,16 @@ func (s *WindowManager) windowByName(name string) *application.WebviewWindow {
 		return s.mainWindow
 	case "settings":
 		return s.settings
+	case "browser-login":
```
riccardomanfrin, line 740: We should emit and correlate events events by IDs here ( `Events.Emit(EVENT_WINDOW_PAINTED, windowID)`)

Otherwise they get aliased, so a painted report from a closed instance can mark it's replacement as ready.

client/ui/services/windowmanager.go
pappz: fixed

client/ui/services/windowmanager.go
pappz: fixed

riccardomanfrin (dismissed):
I'm approving, the two cubic are not blockers I think..

1. 4070361037 (P1, conf 7) su useAutoSizeWindow.ts:24 .. the event can land before the browser paints it. Suggests to send the report to a post-paint callback.. Tecnically sound: apply() emits after the layout / SetSize. But anchoring the emit to a post-paint callback on a throttled webview would mean to never emit.. cause it loops in a non drawn window since it is hidden.. tautological.. would fallback to the 3 seconds.. which is worse than now.

2. 4070361062 (P3, conf 5) su [useAutoSizeWindow.ts:40](vscode-webview://0f88p8sm96vod1lv16mbfd09ubuo14c8253kl1gd8237cvdtag66/client/ui/frontend/src/hooks/useAutoSizeWindow.ts#L40) valid mechanism but if JS runs wails is running to, so it cannot trigger.

Double check these in case

PR #7409, opened 2026-09-03
[client] Gather fresh system info on every management sync stream connect

pappz:
## Describe your changes

The engine collected the peer meta once at start and reused the same Info for every Sync stream reconnect, so a mobile network switch that redials management kept reporting the old local network addresses. The peer network range posture check was then evaluated against stale data until the client restarted.

Sync now takes a gatherer that runs at each stream connect. The gatherer is cheap: GetInfo plus the cached posture check file results, kept in the new system.InfoSource, which the engine refreshes whenever the checks list changes. No process enumeration runs on the reconnect path.

Also fix the management mock server calling itself instead of SyncFunc.

https://github.com/netbirdio/android-client/issues/251

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7409?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * System information refreshes with posture checks before management synchronization.
  * Management connections gather current system details each time they connect.
  * Current system snapshots reuse refreshed file data and exclude selected network addresses.
  * Initial synchronization includes configured file login-check information.

* **Bug Fixes**
  * Improved synchronization reliability with up-to-date system information.
  * Failed synchronization metadata updates now retain the previous state and retry later.
  * Refresh timeouts are now logged instead of proceeding silently.

### DIFF
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -1243,8 +1243,6 @@ func (e *Engine) updateChecksIfNew(checks []*mgmProto.Checks) error {
 	if isChecksEqual(e.checks, checks) {
 		return nil
 	}
-	e.checks = checks
-
 	info, ok := e.infoSource.Refresh(e.ctx, systemInfoTimeout, checks, e.overlayAddresses()...)
 	if !ok {
 		// Gathering timed out; skip the meta sync this cycle rather than blocking the
@@ -1256,6 +1254,7 @@ func (e *Engine) updateChecksIfNew(checks []*mgmProto.Checks) error {
 	if err := e.mgmClient.SyncMeta(info); err != nil {
 		return fmt.Errorf("could not sync meta: error %s", err)
 	}
+	e.checks = checks
 	return nil
 }
 
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -1480,7 +1480,9 @@ func (e *Engine) receiveManagementEvents() {
 	e.shutdownWg.Add(1)
 	go func() {
 		defer e.shutdownWg.Done()
-		e.infoSource.Refresh(e.ctx, systemInfoTimeout, e.checks, e.overlayAddresses()...)
+		if _, ok := e.infoSource.Refresh(e.ctx, systemInfoTimeout, e.checks, e.overlayAddresses()...); !ok {
+			log.Warnf("posture checks not refreshed before the sync connect, sending the previous results")
+		}
 		err := e.mgmClient.Sync(e.ctx, e.currentSystemInfo, e.handleSync)
 		if err != nil {
 			// happens if management is unavailable for a long time.
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -265,6 +265,8 @@ type Engine struct {
 	// checks are the client-applied posture checks that need to be evaluated on the client
 	checks []*mgmProto.Checks
 
+	infoSource system.InfoSource
+
 	relayManager       *relayClient.Manager
 	stateManager       *statemanager.Manager
 	portForwardManager *portforward.Manager
@@ -1243,7 +1245,7 @@ func (e *Engine) updateChecksIfNew(checks []*mgmProto.Checks) error {
 	}
 	e.checks = checks
 
-	info, ok := system.GetInfoWithChecksTimeout(e.ctx, systemInfoTimeout, checks, e.overlayAddresses()...)
+	info, ok := e.infoSource.Refresh(e.ctx, systemInfoTimeout, checks, e.overlayAddresses()...)
 	if !ok {
 		// Gathering timed out; skip the meta sync this cycle rather than blocking the
 		// sync loop (and syncMsgMux) on a stuck system call. A later sync will retry.
@@ -1278,6 +1280,12 @@ func (e *Engine) applyInfoFlags(info *system.Info) {
 		e.config.DisableSSHAuth,
 		&e.config.RemoteJobsAllowed,
 	)
+}
+
+func (e *Engine) currentSystemInfo(ctx context.Context) *system.Info {
+	info := e.infoSource.Current(ctx, e.overlayAddresses()...)
+	e.applyInfoFlags(info)
+	return info
 }
 
 // overlayAddresses returns our own WireGuard overlay address (v4 and v6) so it
@@ -1473,15 +1481,7 @@ func (e *Engine) receiveManagementEvents() {
 	e.shutdownWg.Add(1)
 	go func() {
 		defer e.shutdownWg.Done()
-		info, ok := system.GetInfoWithChecksTimeout(e.ctx, systemInfoTimeout, e.checks, e.overlayAddresses()...)
-		if !ok {
-			// Gathering timed out; connect the stream with base info so management
-			// connectivity still comes up rather than blocking here.
-			info = system.GetInfo(e.ctx)
-		}
-		e.applyInfoFlags(info)
-
-		err := e.mgmClient.Sync(e.ctx, info, e.handleSync)
+		err := e.mgmClient.Sync(e.ctx, e.currentSystemInfo, e.handleSync)
 		if err != nil {
 			// happens if management is unavailable for a long time.
 			// We want to cancel the operation of the whole client
--- a/client/internal/engine.go
+++ b/client/internal/engine.go
@@ -1481,6 +1481,7 @@ func (e *Engine) receiveManagementEvents() {
 	e.shutdownWg.Add(1)
 	go func() {
 		defer e.shutdownWg.Done()
+		e.infoSource.Refresh(e.ctx, systemInfoTimeout, e.checks, e.overlayAddresses()...)
 		err := e.mgmClient.Sync(e.ctx, e.currentSystemInfo, e.handleSync)
 		if err != nil {
 			// happens if management is unavailable for a long time.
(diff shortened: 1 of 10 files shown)


client/internal/engine.go
```diff
@@ -1280,6 +1282,12 @@ func (e *Engine) applyInfoFlags(info *system.Info) {
 	)
 }
 
+func (e *Engine) currentSystemInfo(ctx context.Context) *system.Info {
+	info := e.infoSource.Current(ctx, e.overlayAddresses()...)
+	e.applyInfoFlags(info)
+	return info
+}
+
```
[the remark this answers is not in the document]
pappz: Not applying this one. The callback runs only the base `GetInfo` plus the cached posture results; the process enumeration that the 15 s bound exists for stays in `Refresh`. On Linux and Windows `GetInfo` reads the once-computed static info and `net.Interfaces`, neither of which blocks, so a timeout here would guard nothing.

Returning a cached full snapshot on timeout would also reintroduce the stale network addresses this PR removes: the point of gathering at connect time is that the addresses reflect the network the stream is being opened on.

The Android and darwin `GetInfo` implementations do fork `getprop`/`uname` and `sw_vers` for values that are static for the process lifetime. Caching those behind a `sync.Once`, as `static_info.go` does for desktop, is the right improvement, and a separate one.

d375e2a did not address this comment; it seeded the cache with the login response checks.

client/system/info_source.go
```diff
@@ -0,0 +1,38 @@
...
+	files atomic.Pointer[[]File]
+}
+
+// Refresh gathers the info with the posture checks evaluated, bounded by timeout.
+func (s *InfoSource) Refresh(ctx context.Context, timeout time.Duration, checks []*proto.Checks, excludeIPs ...netip.Addr) (*Info, bool) {
+	info, ok := GetInfoWithChecksTimeout(ctx, timeout, checks, excludeIPs...)
+	if !ok {
+		return nil, false
+	}
+	files := slices.Clone(info.Files)
```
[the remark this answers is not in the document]
pappz: Not applying this one. Keeping the last successful snapshot after a failed check collection would make posture fail open: a peer whose process scan keeps failing would go on passing a process check on results that may no longer be true. Reporting no files on that path is the fail-closed outcome, and it is what `main` already sends through `SyncMeta` on the same fallback; the server-side state is unchanged by this PR.

Distinguishing the fallback from a successful collection would also need a wider return contract on `GetInfoWithChecksTimeout`. If we ever want a different failure policy for posture data, that is a separate decision, not part of this fix.

client/system/info_source.go
```diff
@@ -0,0 +1,38 @@
...
+	s.files.Store(&files)
+	return info, true
+}
+
+// Current gathers the info without evaluating the checks, reusing the last Refresh results.
+func (s *InfoSource) Current(ctx context.Context, excludeIPs ...netip.Addr) *Info {
```
riccardomanfrin, line 31: From the Sync (`handleSyncStream`) we call it without timeout.
Before in `receiveManagementEvents` the gather was within `systemInfoTimeout` (15s) and `GetInfo` was just a fallback after timeout. Now it's standard path and on Darwin it forks `sw_vers`, while on Android it calls `uname / getprop / serial `.. 
An appended `exec`-ution of it blocks connection to mgmt indefinitely.

shared/management/client/grpc.go
```diff
@@ -424,10 +424,14 @@ func (c *GrpcClient) sendJobResponse(
...
 }
 
-func (c *GrpcClient) handleSyncStream(ctx context.Context, serverPubKey wgtypes.Key, sysInfo *system.Info, msgHandler func(msg *proto.SyncResponse) error, backOff backoff.BackOff) error {
+func (c *GrpcClient) handleSyncStream(ctx context.Context, serverPubKey wgtypes.Key, getInfo func(ctx context.Context) *system.Info, msgHandler func(msg *proto.SyncResponse) error, backOff backoff.BackOff) error {
 	ctx, cancelStream := context.WithCancel(ctx)
 	defer cancelStream()
 
+	var sysInfo *system.Info
+	if getInfo != nil {
+		sysInfo = getInfo(ctx)
```
riccardomanfrin: we getInfo before connectToSyncStream, so it's possible we do it even though the conn will fail (so we don't have a need for the collected data. postpone after conn success or can drive attempted forked commands on a client that is not even connected.

client/internal/engine.go
```diff
@@ -1473,15 +1480,8 @@ func (e *Engine) receiveManagementEvents() {
...
-		if !ok {
-			// Gathering timed out; connect the stream with base info so management
-			// connectivity still comes up rather than blocking here.
-			info = system.GetInfo(e.ctx)
-		}
-		e.applyInfoFlags(info)
-
-		err := e.mgmClient.Sync(e.ctx, info, e.handleSync)
+		e.infoSource.Refresh(e.ctx, systemInfoTimeout, e.checks, e.overlayAddresses()...)
+		err := e.mgmClient.Sync(e.ctx, e.currentSystemInfo, e.handleSync)
```
riccardomanfrin: we do GetInfo twice back to back here, once in Refresh and once in Sync

client/internal/engine.go
```diff
@@ -1473,15 +1480,8 @@ func (e *Engine) receiveManagementEvents() {
...
-		info, ok := system.GetInfoWithChecksTimeout(e.ctx, systemInfoTimeout, e.checks, e.overlayAddresses()...)
-		if !ok {
-			// Gathering timed out; connect the stream with base info so management
-			// connectivity still comes up rather than blocking here.
-			info = system.GetInfo(e.ctx)
-		}
-		e.applyInfoFlags(info)
-
-		err := e.mgmClient.Sync(e.ctx, info, e.handleSync)
+		e.infoSource.Refresh(e.ctx, systemInfoTimeout, e.checks, e.overlayAddresses()...)
```
riccardomanfrin: worth logging on `!ok` (?)

shared/management/client/grpc.go
```diff
@@ -424,10 +424,14 @@ func (c *GrpcClient) sendJobResponse(
 	return nil
 }
 
-func (c *GrpcClient) handleSyncStream(ctx context.Context, serverPubKey wgtypes.Key, sysInfo *system.Info, msgHandler func(msg *proto.SyncResponse) error, backOff backoff.BackOff) error {
+func (c *GrpcClient) handleSyncStream(ctx context.Context, serverPubKey wgtypes.Key, getInfo func(ctx context.Context) *system.Info, msgHandler func(msg *proto.SyncResponse) error, backOff backoff.BackOff) error {
 	ctx, cancelStream := context.WithCancel(ctx)
 	defer cancelStream()
 
+	var sysInfo *system.Info
+	if getInfo != nil {
```
riccardomanfrin: is this guard needed?
grpc.go:432 is the only caller and with `nil` we would continue with `infoToMetaData(nil)` which is not ok

shared/management/client/client_test.go
```diff
@@ -397,6 +399,56 @@ func wgKeyFromBytes(raw []byte) string {
 	return k.String()
 }
 
+func TestClient_SyncGathersInfoOnEveryConnect(t *testing.T) {
```
riccardomanfrin, line 402: Might be flacky?
gathers increments on every attempt, not only when connect succeeds. If a connect fails before reachin the server (conn not Ready, `WaitForStateChange`), server will see `host-3` while assert on `host-2` fails.

It also doesn't wait for the Sync goroutine so it's at risk of log after test.

management/server/mock_server/management_server_mock.go
```diff
@@ -30,7 +30,8 @@ func (m ManagementServiceServerMock) Login(ctx context.Context, req *proto.Encry
 
 func (m ManagementServiceServerMock) Sync(msg *proto.EncryptedMessage, sync proto.ManagementService_SyncServer) error {
 	if m.SyncFunc != nil {
-		return m.Sync(msg, sync)
+		m.SyncFunc(msg, sync)
+		return nil
```
riccardomanfrin: Ok for recusion fix but returning nil prevents mocking an error coming from the server

client/system/info_source.go
pappz: On Darwin have a known bug that hang forewer. On Android the uname, getprop, etc the risks are minimal. The fix is comoplex (extra go threads). I do not think so make sense to risk a new bug. Is it ok?

shared/management/client/grpc.go
pappz: SyncRequest, whose only field is the meta, and then receives the network map
stream. So the gathered info is the content of the stream-opening message itself, there is no later point on this stream to send it. Before handleSyncStream runs, withMgmtStream has already waited for network
availability. Alternative way if we send cached data but in this case the message will be duplcated and wast network messages. By the way the server just reject the sync only in edge cases.

riccardomanfrin (approved):

PR #7426, opened 2026-09-04
[client] Return the context error when the SSH handshake fails with it

pappz:
## Describe your changes

The handshake mapped the context deadline onto the socket but returned the raw socket error. Which error surfaces depends on a race between the x/crypto ssh readLoop and kexLoop goroutines: the kexLoop write fails with i/o timeout and closes the conn, and the readLoop then reports use of closed network connection. Callers checking errors.Is(err, context.DeadlineExceeded) never matched, and TestSSHClient_ContextCancellation flaked on the FreeBSD job.

Handshake now wraps the context error when the context is done or its deadline has passed. The deadline comparison is needed because the socket deadline and the context timer fire independently, so ctx.Err() can still be nil when the deadline-triggered socket error arrives.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7426?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * SSH handshakes now stop promptly when their context is canceled or reaches its deadline.
  * Context cancellation and deadline errors are reported clearly while preserving unrelated connection errors.
  * Connections are properly closed when a handshake is interrupted or completes after cancellation.

* **Tests**
  * Added coverage for canceled contexts, expired deadlines, and non-context handshake errors.

### DIFF
--- a/client/ssh/handshake.go
+++ b/client/ssh/handshake.go
@@ -27,7 +27,7 @@ func Handshake(ctx context.Context, conn net.Conn, addr string, config *ssh.Clie
 	sshConn, chans, reqs, err := ssh.NewClientConn(conn, addr, config)
 	if err != nil {
 		closeHandshake(conn, "conn after handshake error")
-		return nil, fmt.Errorf("ssh handshake: %w", err)
+		return nil, handshakeError(ctx, err)
 	}
 
 	if err := conn.SetDeadline(time.Time{}); err != nil {
@@ -42,4 +42,14 @@ func closeHandshake(c io.Closer, label string) {
 	if err := c.Close(); err != nil {
 		log.Debugf("ssh: close %s: %v", label, err)
 	}
+}
+
+func handshakeError(ctx context.Context, err error) error {
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		return fmt.Errorf("ssh handshake: %w: %w", ctxErr, err)
+	}
+	if deadline, ok := ctx.Deadline(); ok && !time.Now().Before(deadline) {
+		return fmt.Errorf("ssh handshake: %w: %w", context.DeadlineExceeded, err)
+	}
+	return fmt.Errorf("ssh handshake: %w", err)
 }
--- a/client/ssh/handshake.go
+++ b/client/ssh/handshake.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"net"
-	"time"
 
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/crypto/ssh"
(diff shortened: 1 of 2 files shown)


client/ssh/handshake.go
```diff
@@ -43,3 +43,13 @@ func closeHandshake(c io.Closer, label string) {
 		log.Debugf("ssh: close %s: %v", label, err)
 	}
 }
+
+func handshakeError(ctx context.Context, err error) error {
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		return fmt.Errorf("ssh handshake: %w: %w", ctxErr, err)
+	}
+	if deadline, ok := ctx.Deadline(); ok && !time.Now().Before(deadline) {
```
riccardomanfrin, line 51: as asked by cubic add `errors.Is(err, os.ErrDeadlineExceeded)` to the condition

client/ssh/handshake.go
riccardomanfrin, line 23: wrap in handshakeError here too

riccardomanfrin, line 35: wrap in handshakeError here too

client/ssh/handshake_test.go
```diff
@@ -0,0 +1,76 @@
...
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = listener.Close() })
+
+	go func() {
+		c, err := listener.Accept()
+		if err != nil {
+			return
+		}
+		t.Cleanup(func() { _ = c.Close() })
```
riccardomanfrin, line 68: If test ends before the goroutine registers the function it might append after runCleanup has emptied the list. No panic but accepted conn is not closed.

I'd use `defer c.Close()` to fix

riccardomanfrin (changes requested):
It looks like ctx.Done() is not observed.

One way to fix it is to

```go
defer context.AfterFunc(ctx, func() { _ = conn.Close() })()
```

riccardomanfrin (approved):
In PR description:

> "the deadline comparison is needed because the socket deadline and the context timer fire independently"

This was changed by 3rd commit

There is a low risk with `handshake.go:30` on `!stop()` branch where we close `sshConn` but don't drain `chans`/`reqs`. Not sure if this is a real problem

PR #7435, opened 2026-09-04
[management] Harden OIDC issuer validation and discovery

bcmmbaga:
## Describe your changes

This PR hardens OIDC issuer validation and discovery handling. Issuers are now limited to `http`/`https` URLs without user info, query strings, or fragments.

Discovery requests no longer follow redirects, and cap responses at 1 MiB. 

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7435?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Bug Fixes

- Identity provider issuer validation now requires HTTPS URLs and rejects unsafe formats, including query strings, fragments, embedded user information, and unsupported schemes.
- OpenID Connect discovery validation now rejects redirects, oversized responses, malformed documents, and valid documents followed by trailing content, providing more consistent error handling.

### DIFF
--- a/management/server/types/identity_provider.go
+++ b/management/server/types/identity_provider.go
@@ -99,7 +99,13 @@ func (idp *IdentityProvider) Validate() error {
 	}
 	if idp.Issuer != "" {
 		parsedURL, err := url.Parse(idp.Issuer)
-		if err != nil || parsedURL.Scheme == "" || parsedURL.Host == "" {
+		if err != nil || parsedURL.Host == "" {
+			return ErrIdentityProviderIssuerInvalid
+		}
+		if parsedURL.Scheme != "http" && parsedURL.Scheme != "https" {
+			return ErrIdentityProviderIssuerInvalid
+		}
+		if parsedURL.User != nil || parsedURL.RawQuery != "" || parsedURL.Fragment != "" {
 			return ErrIdentityProviderIssuerInvalid
 		}
 	}
--- a/management/server/types/identity_provider.go
+++ b/management/server/types/identity_provider.go
@@ -3,6 +3,7 @@ package types
 import (
 	"errors"
 	"net/url"
+	"strings"
 )
 
 // Identity provider validation errors
@@ -105,7 +106,7 @@ func (idp *IdentityProvider) Validate() error {
 		if parsedURL.Scheme != "http" && parsedURL.Scheme != "https" {
 			return ErrIdentityProviderIssuerInvalid
 		}
-		if parsedURL.User != nil || parsedURL.RawQuery != "" || parsedURL.Fragment != "" {
+		if parsedURL.User != nil || strings.ContainsAny(idp.Issuer, "?#") {
 			return ErrIdentityProviderIssuerInvalid
 		}
 	}
--- a/management/server/types/identity_provider.go
+++ b/management/server/types/identity_provider.go
@@ -103,10 +103,13 @@ func (idp *IdentityProvider) Validate() error {
 		if err != nil || parsedURL.Host == "" {
 			return ErrIdentityProviderIssuerInvalid
 		}
-		if parsedURL.Scheme != "http" && parsedURL.Scheme != "https" {
+		if parsedURL.Scheme != "https" {
 			return ErrIdentityProviderIssuerInvalid
 		}
-		if parsedURL.User != nil || strings.ContainsAny(idp.Issuer, "?#") {
+		if parsedURL.User != nil {
+			return ErrIdentityProviderIssuerInvalid
+		}
+		if strings.ContainsAny(idp.Issuer, "?#") {
 			return ErrIdentityProviderIssuerInvalid
 		}
 	}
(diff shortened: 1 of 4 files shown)


management/server/types/identity_provider.go
```diff
@@ -99,7 +100,13 @@ func (idp *IdentityProvider) Validate() error {
 	}
 	if idp.Issuer != "" {
 		parsedURL, err := url.Parse(idp.Issuer)
-		if err != nil || parsedURL.Scheme == "" || parsedURL.Host == "" {
+		if err != nil || parsedURL.Host == "" {
```
dmitri-netbird, line 103: I checked the rfc (https://openid.net/specs/openid-connect-discovery-1_0.html):
```
issuer
    REQUIRED. URL using the https scheme with no query or fragment components that the OP asserts as its Issuer Identifier. If Issuer discovery is supported (see [Section 2](https://openid.net/specs/openid-connect-discovery-1_0.html#IssuerDiscovery)), this value MUST be identical to the issuer value returned by WebFinger. This also MUST be identical to the iss Claim value in ID Tokens issued from this Issuer.
```

A couple of things standout:
 * query or fragments should be unconditionally not present. A correct check, I think,  then would be to verify that `parsedURL.Fragment` and `parsedURL.RawQuery` both are empty. Not sure why the check is coupled with `parsedURL.User`?
* Issuer is explicitly "https" only (which makes sense, as it's used for jwks retrieval). Why are we allowing "http" here?

dmitri-netbird: Also: I don't think there's anything in the rfc re: issuer host to be the same as the discovery one. The only requirement is that jwt iss claim to be the same as the issuer

bcmmbaga: - `Fragment`/`RawQuery`: The check has false negative
```go
url.Parse("https://x/p#")  → Fragment="", RawQuery=""                  → passes
url.Parse("https://x/p?")  → Fragment="", RawQuery="", ForceQuery=true → passes
```
The check accepts input it should reject thus why was dropped and not used here

bcmmbaga: - Userinfo isn't one of the components the spec allows in an Issuer Identifier, and Go's `http.Client` turns it into a `Basic Authorization` header so `https://user:pass@host` makes the server send caller-chosen credentials to a caller-chosen host hence the separate condition
-  There's no host-comparison check anywhere

dmitri-netbird: oh wow, sry I sent you on a wild goose chase re: fragment & query, should've confirmed my assumptions beforehand.

dmitri-netbird: the comment re: host was re: this line: https://github.com/netbirdio/netbird/pull/7435/changes#diff-25a5a5108d535e7638ca8c18654970b20555728182d84675db573805dccfc846R72

management/server/types/identity_provider.go
```diff
@@ -99,7 +100,16 @@ func (idp *IdentityProvider) Validate() error {
 	}
 	if idp.Issuer != "" {
```
dmitri-netbird, line 101: is issuer an optional parameter (asking b/c it's a required field in a return from oidc discovery end-point)?

dmitri-netbird (approved):

management/server/types/identity_provider.go
bcmmbaga: That check compares the issuer provided by the user with the `issuer` value returned in the well-known configuration. It’s not comparing the issuer with the discovery itself

management/server/types/identity_provider.go
bcmmbaga: Optional only for Google and Entra. For other IdPs, it’s still required

here.https://github.com/netbirdio/netbird/blob/269bf1e5e085afe48631292998e2dd51335bdd4d/management/server/types/identity_provider.go#L98-L100

management/server/types/identity_provider.go
dmitri-netbird: the issue is that we use the `issuer` parameter to construct a discovery url. And then validate that the issuer field returned from the discovery end-point matches the one specified by the user. This enforces that the discovery and issuer hosts are the same, which isn't mandated by the rfc, and will probably cause issues down the road.

bcmmbaga: That is mandated by the OIDC Discovery spec. See https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationValidation

dmitri-netbird: ah, I was wrong then, sry about all the noise.

PR #7440, opened 2026-09-05
[client] Support arbitrary UIDs in rootless image

jnfrati:
## Describe your changes

A process can have a valid non-root UID without an entry in `/etc/passwd`. When current-user lookup fails, profile commands cannot resolve that user's profile namespace.

Resolve these invocations using the numeric kernel UID and GID. Keep existing root and sudo lookup failures fail-closed, and retain tests for numeric profile paths and those failure cases.

This PR now contains only `invoking_user.go` and its tests. The Dockerfile changes and container smoke test are on `jnfrati/rootless-image-permissions` for separate review.

N/A — agreed internally with the NetBird team and authorized by a maintainer before implementation.

## Stack

Targets `main`. The `jnfrati/rootless-image-permissions` branch builds on this branch.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

This fixes user resolution without introducing flags or configuration requirements. Container image and deployment changes are outside this PR.


client/Dockerfile-rootless
```diff
@@ -29,5 +32,5 @@ ENV \
 ENTRYPOINT [ "/usr/local/bin/netbird-entrypoint.sh" ]
 ARG TARGETPLATFORM
 ARG NETBIRD_BINARY=$TARGETPLATFORM/netbird
-COPY client/netbird-entrypoint.sh /usr/local/bin/netbird-entrypoint.sh
-COPY "${NETBIRD_BINARY}"  /usr/local/bin/netbird
+COPY --chown=1000:0 --chmod=0750 client/netbird-entrypoint.sh /usr/local/bin/netbird-entrypoint.sh
+COPY --chown=1000:0 --chmod=0750 "${NETBIRD_BINARY}" /usr/local/bin/netbird
```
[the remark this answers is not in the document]
jnfrati: Fixed in 68ece96a0. Both executables are now owned by root with mode 0750; the runtime UID retains group-0 execute access but cannot modify them. The arbitrary-UID container smoke test passes with this ownership.

client/Dockerfile-rootless
```diff
@@ -29,5 +32,5 @@ ENV \
 ENTRYPOINT [ "/usr/local/bin/netbird-entrypoint.sh" ]
 ARG TARGETPLATFORM
 ARG NETBIRD_BINARY=$TARGETPLATFORM/netbird
-COPY client/netbird-entrypoint.sh /usr/local/bin/netbird-entrypoint.sh
-COPY "${NETBIRD_BINARY}"  /usr/local/bin/netbird
+COPY --chown=0:0 --chmod=0750 client/netbird-entrypoint.sh /usr/local/bin/netbird-entrypoint.sh
+COPY --chown=0:0 --chmod=0750 "${NETBIRD_BINARY}" /usr/local/bin/netbird
```
riccardomanfrin: we change from 775 (world exec) to 750. So it will require `CAP_DAC_OVERRIDE` or group 0.. is this ok?

client/internal/profilemanager/invoking_user.go
```diff
@@ -35,7 +40,24 @@ func InvokingUser() (*user.User, error) {
...
+		return nil, err
+	}
+
+	log.Debugf("current user lookup for UID %d: %v; using numeric UID", uid, err)
+	uidString := strconv.Itoa(uid)
+	return &user.User{
+		Username: uidString,
+		Uid:      uidString,
+		Gid:      strconv.Itoa(getegid()),
+		HomeDir:  os.Getenv("HOME"),
```
riccardomanfrin: is this homedir used? I think it read through .Username, userBaseConfigDir by code around

riccardomanfrin (commented):
P1 by cubic is lgtm: persisted state is not usable by another user: 0600 + chmod -R g=u cancels the setgid that was enforced by adduser.

UID A writes , UID B fails to read (`FATL … active_profile.json: permission denied`)

client/Dockerfile-rootless
jnfrati: Yeah, the change originally was made so that we preserve the integrity of the binaries, but I agree that it's a risk for people currently using the image, out of nowhere it will stop working and there's going to be 0 signal around it.
I'll make sure that 0755 is compatible with openshift and revert this one

client/internal/profilemanager/invoking_user.go
jnfrati: Removed!

client/Dockerfile-rootless
```diff
@@ -1,20 +1,23 @@
...
 
 FROM alpine:3.24
 
 RUN apk add --no-cache \
       bash \
       ca-certificates \
-    && adduser -D -h /var/lib/netbird netbird
+    && adduser -D -u 1000 -h /var/lib/netbird netbird \
+    && chgrp -R 0 /var/lib/netbird \
+    && chmod -R g=u /var/lib/netbird
```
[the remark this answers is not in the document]
jnfrati: @riccardomanfrin Agreed that a different runtime UID cannot generally reuse the owner-only state. Commit 6cff57002 keeps this PR scoped to an arbitrary UID chosen at first use, with the same UID retained for a persistent volume. It does not add cross-UID migration or broaden permissions in the shared file writers.

The container smoke test now recreates the container on the same volume and verifies that its saved profile survives and remains writable. I also tested the image on OpenShift under restricted-v2 with all capabilities dropped: startup and bidirectional overlay traffic passed, and same-UID pod recreation on Ceph RBD preserved the NetBird public key/IP and saved profile and reconnected successfully.

One deployment caveat: default fsGroup handling widened existing 0600 files to 0660 on remount. OnRootMismatch preserved correctly permissioned files on the tested volume; the Dockerfile guidance now qualifies this behavior.

Is the stable-UID-per-volume contract acceptable for this PR? Leaving this thread open pending that scope agreement.

client/Dockerfile-rootless
```diff
@@ -1,20 +1,28 @@
...
```
lixmal: This clears the setgid bit, intentional? This is also world readable

jnfrati: This PR no longer covers the Dockerfile-rootless so I think we should be good there.

riccardomanfrin (commented):
`TestInvokingUserFallsBackToProcessUser` was the only test covering the happy path I think.. I added it back and it passes. is there a reason why you removed it? Otherwise if you can restore it it would be better

riccardomanfrin (approved):

PR #7096, opened 2026-08-07
[client] Create GUI windows on demand and destroy them on close

pappz:
## Describe your changes

The main and Settings windows were created at startup and kept alive hidden on close, so an idle tray held two webview processes for surfaces the user may never open. Both are now built on first show and destroyed on close, which takes the idle footprint on macOS from ~160 MB to ~74 MB.

The WindowManager owns creation: it rebuilds the main window on the next show and hands out live pointers, since a stored one goes stale. Every show is deferred until the frontend reports it has rendered, so a freshly created window is never on screen empty, with a timeout so a frontend that never reports cannot strand a window hidden.

### Checklist
- [x] Is it a bug fix
- [ ] Is a typo/documentation fix
- [ ] Is a feature enhancement
- [ ] It is a refactor
- [ ] Created tests that fail without the change (if possible)
- [ ] I ran and tested this change locally — I did not rely on CI to find out whether it works
- [ ] This PR has a single purpose (not a fix + refactor + feature in one)
- [ ] This change is a trivial fix, **OR** it links an issue the NetBird team agreed on beforehand. Changes to the public API, gRPC protocols, functionality behavior, CLI / service flags, or new features always need that agreement first. See [CONTRIBUTING.md](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTING.md#ticket-first-pr-second).

> By submitting this pull request, you confirm that you have read and agree to the terms of the [Contributor License Agreement](https://github.com/netbirdio/netbird/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT.md).

## Documentation
Select exactly one:

- [ ] I added/updated documentation for this change
- [x] Documentation is **not needed** for this change (explain why)

### Docs PR URL (required if "docs added" is checked)
Paste the PR link from https://github.com/netbirdio/docs here:

https://github.com/netbirdio/docs/pull/__

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved main-window reopening from the system tray and macOS Dock.
  * Windows now appear only after they are ready to render, reducing blank or incomplete displays.
  * Closing the main or settings window now resets it for a clean next launch.
  * Session expiry and update flows reliably bring the main window forward and navigate appropriately.
  * Closing the last window no longer quits the application on Windows and Linux.

PR #7449, opened 2026-09-07
[client] Fix the Windows tray deadlock on re-entrant window creation

pappz:
## Describe your changes

**Symptom.** On Windows with 0.78.1 the desktop UI stops reacting: left and right click on the tray icon do nothing, the tray menu never opens, the icon never changes state and the main window cannot be opened, while the daemon stays healthy and the CLI keeps working. The process is alive, there is no crash and nothing is logged.

**Reproduction.**
1. Kill `netbird-ui.exe` and start it again, or start it tray-only, so the main window does not exist yet (since #7096 it is created on demand).
2. Double-click the tray icon. The UI hangs for good.
3. Doing step 2 as right-click and "Open NetBird" instead opens the window normally. Closing it with the X and double-clicking again hangs the UI again.

A single left click during startup with an expired session hits the same path, because the session-expired transition starts creating the main window from the status feed goroutine at the moment the user clicks.

**Root cause.** A full memory dump of a frozen process shows the main thread on one goroutine stack with two nested tray click handlers:

```
wndProc → bindTrayClick → Tray.ShowWindow → WindowManager.ShowMain → ensureMain → ensureWindow (factory(), createMu held)
  → newMainWindow → NewWithOptions → Run → InvokeSync → run → setupChromium → Embed → pumpUntilInited → DispatchMessageW
    → wndProc → bindTrayClick → Tray.ShowWindow → ShowMain → ensureMain → ensureWindow → Mutex.lockSlow → semacquire → gopark
```

The Wails Windows systray calls the OnClick handler synchronously inside the tray window procedure. Creating a window on a running app runs synchronously on the main thread, and the WebView2 embed step pumps a nested message loop until the controller reports ready. `ensureWindow` held the non-reentrant `createMu` across that creation, so the second `WM_LBUTTONUP` of the double click was dispatched from inside the pump, re-entered `ShowWindow` on the same thread and blocked on the lock the outer frame still held. The pump never returned, and the message loop, the menu and the icon updates died with it. The same inversion existed between a goroutine holding `createMu` while the main thread pumped, and in the `Open*` dialog functions that held `mu` across `NewWithOptions`, `Show`, `Hide` and the `raiseToForeground` InvokeSync. Upstream Wails through v3.0.0-beta.17 has the same synchronous handler and nested pump, so the fix belongs in the WindowManager.

**Fix.** WindowManager serialises creation with a per-slot creating flag and queues the callers' operations until the window exists, so a re-entrant or concurrent caller returns immediately and its show is applied once the window is ready. No Wails call runs while `mu` is held any more: the dialog windows are built in factories outside the lock, and `hideOtherWindows`/`restoreHiddenWindows` snapshot their lists under the lock and call Wails after releasing it. The tray click and second-instance handlers call `ShowWindow` off the message loop, matching how Wails already dispatches menu callbacks. `windowmanager_test.go` covers the re-entrant and concurrent creation paths; the re-entrant test deadlocks without the change.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7449?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved responsiveness when reopening the application from the system tray or another launch attempt.
  * Improved reliability when opening multiple application windows concurrently.
  * Ensured window operations are processed in the correct order during simultaneous actions.
  * Prevented unnecessary window restoration when no login popup was open.

* **Refactor**
  * Streamlined window creation and operation handling for more consistent behavior.

### DIFF
--- a/client/ui/services/windowmanager.go
+++ b/client/ui/services/windowmanager.go
@@ -131,6 +131,7 @@ type WindowManager struct {
 	newMain        func(startURL string) *application.WebviewWindow
 	creating       map[string]bool
 	pendingOps     map[string][]windowOp
+	restoreGen     uint64
 	ready          map[uint]bool
 	showPending    map[uint]bool
 	pendingTab     map[uint]string
@@ -567,6 +568,11 @@ func (s *WindowManager) ensureMain(startURL string, op windowOp) {
 
 func (s *WindowManager) withWindow(name string, slot **application.WebviewWindow, factory func() *application.WebviewWindow, op windowOp) {
 	s.mu.Lock()
+	if s.creating[name] {
+		s.pendingOps[name] = append(s.pendingOps[name], op)
+		s.mu.Unlock()
+		return
+	}
 	if w := *slot; w != nil {
 		s.mu.Unlock()
 		op(w, false)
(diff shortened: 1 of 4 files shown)


client/ui/services/windowmanager.go
```diff
@@ -868,39 +969,61 @@ func (s *WindowManager) retitleAll() {
...
 		w.Hide()
-		s.hiddenForLogin = append(s.hiddenForLogin, w)
+		hidden = append(hidden, w)
+	}
+	if len(hidden) == 0 {
+		return
+	}
+
+	s.mu.Lock()
+	restored := s.restoreGen != gen
```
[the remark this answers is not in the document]
pappz: Won't fix. This needs two different modals to open while a restore lands inside the first one's few-ms Hide() loop. Worst case is cosmetic: one window stays visible behind the second modal, nothing gets stuck hidden and the second modal's close still restores what it hid. Not worth the extra bookkeeping.

gralin:
We started using latest (0.78.1) NetBird client on Windows and immediately started having this issues on multiple machines without knowing about this bug and how to trigger it. I just replaced netbird-ui.exe with local build from this PR and it fixes the problem. Thank you @pappz and looking forward to the merge :)

client/ui/services/windowmanager.go
```diff
@@ -565,65 +609,127 @@ func (s *WindowManager) OpenMain() {
...
+		delete(s.pendingOps, name)
+		s.mu.Unlock()
+	}()
 
+	w := factory()
+	if w == nil {
+		return nil
+	}
 	s.mu.Lock()
 	*slot = w
```
riccardomanfrin, line 668: factory() does the init of WebView2 which takes time. and we publish `*slot = w` only here.
Now every close during this period of time finds a nil slot and is treated as no-op.

riccardomanfrin (commented):
Not in scope, but  as of today [windowmanager_test.go](https://github.com/netbirdio/netbird/pull/7449/changes#diff-7f51ded2ac3fff1b996ed03bc8d5d623d01d177ad6e2c7659e4925a7466acad8) tests are not run in CI because of  golang-test-linux.yml:170, golang-test-darwin.yml:57, golang-test-windows.yml:77 — `grep -v -e /client/ui` statements

We could probably just exclude the specific folder `grep -e '/client/ui$'` and let `services` and `preferences` tests run in CI. gtk4/webkitgtk is already installed on linux (golang-test-linux.yml:57)

client/ui/services/windowmanager.go
pappz: `Close*` now goes through `closeWindow`. If the slot is still being created, it stores a closer in `pendingClose` instead of returning. `finishCreation` checks that map before running any op, so the window is closed right after the factory returns and is never shown.

Two details:
- Ops queued behind a close are dropped. `windowOp` carries no factory, so they cannot be replayed into a new creation. The frontend callers re-issue on the next state change.
- `CloseBrowserLogin` and `CloseRenewFlow` use the same restoring closer for the popup slot, because the popup's `WindowClosing` hook only restores on a user close. If two closes race one creation, the first one wins.

Tests cover the deferred close, the drain, the panic and nil-factory paths, and the `CloseRenewFlow` restore.

riccardomanfrin (approved):

PR #7485, opened 2026-09-09
[client] Report the remote jobs key in the MDM UI snapshot

pappz:
## Describe your changes

The `allowRemoteJobs` policy key was already enforced end to end: `GetRemoteJobsAllowed` returns the managed value over the local one, and `Commit` rejects a divergent write through `CheckMDMConflicts`. What was missing is the UI-facing half. `BuildRestrictions` never reported the key, so `mdm.remoteJobsAllowed` was absent from the snapshot every frontend reads.

A client that cannot see the key cannot lock the control. On mobile the result is a toggle that looks editable, flips, fails to save, and silently snaps back with no explanation. Desktop is affected too: its `MDMFields` is a type alias for this same `mdm.Fields` (`client/ui/services/settings.go`), and `applyMDMRestrictions` fills it by matching the `mDMManagedFields` keys from `GetConfigResponse` against the struct's JSON tags.

This adds the missing field to `mdm.Fields` and sets it from `policy.HasKey(KeyRemoteJobsAllowed)`, matching every other managed boolean in the builder. No enforcement behavior changes; this only lets a UI render what the policy already enforces.

That exposed a second, related defect. The field's JSON tag was `remoteJobsAllowed` while the policy key is `allowRemoteJobs`. Since `mDMManagedFields` reports the raw policy keys and the desktop matches them against JSON tags, the field would never have turned true. Every other field in `mdm.Fields` already uses its policy key verbatim as the JSON tag, so this was the only divergence; the second commit aligns it.

The field is additive and mirrors the existing convention, so a frontend that does not know the key keeps decoding the snapshot unchanged. The iOS client consumes it in a companion change.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7485?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added support for reporting whether remote jobs are allowed through device management restrictions.
  - Included the remote jobs setting in serialized restriction data.

### DIFF
--- a/client/mdm/restrictions.go
+++ b/client/mdm/restrictions.go
@@ -20,6 +20,7 @@ type Fields struct {
 	DisableMetricsCollection bool   `json:"disableMetricsCollection"`
 	SplitTunnelMode          bool   `json:"splitTunnelMode"`
 	SplitTunnelApps          bool   `json:"splitTunnelApps"`
+	RemoteJobsAllowed        bool   `json:"remoteJobsAllowed"`
 	DisableAdvancedView      *bool  `json:"disableAdvancedView"`
 }
 
@@ -60,6 +61,7 @@ func BuildRestrictions(policy *Policy) Restrictions {
 	r.MDM.DisableMetricsCollection = policy.HasKey(KeyDisableMetricsCollection)
 	r.MDM.SplitTunnelMode = policy.HasKey(KeySplitTunnelMode)
 	r.MDM.SplitTunnelApps = policy.HasKey(KeySplitTunnelApps)
+	r.MDM.RemoteJobsAllowed = policy.HasKey(KeyRemoteJobsAllowed)
 	if v, ok := policy.GetBool(KeyAllowServerSSH); ok {
 		r.MDM.AllowServerSSH = &v
 	}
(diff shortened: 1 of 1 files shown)


riccardomanfrin (commented):
can we align between `allowRemoteJobs` and `remoteJobsAllowed`?
It's the same argument as cubic finding

riccardomanfrin (approved):
```bash
$ ag -i remoteJobsAllowed |wc -l
  117

$ ag -i allowRemoteJobs 
  mdm/policy.go
  69:	KeyRemoteJobsAllowed = "allowRemoteJobs"

  mdm/restrictions.go
  23:	RemoteJobsAllowed        bool   `json:"allowRemoteJobs"`

  internal/profilemanager/config_test.go
  316:		assert.True(t, *cfg.RemoteJobsAllowed, "MDM allowRemoteJobs must enable the flag")

```

We are allowing the divergence between JSON / MDM key with code and protobuff vocabulary.

I'm approving but let's consider if it makes sense to also align the code and protobuf in a later PR (?)

PR #7487, opened 2026-09-09
[client] Stop stale inactivity events from idling a re-added peer

mnaza:
### The bug

An inactivity event carried only a peer key, so it could not say which connection the peer was found idle on.

A peer can be removed and re-added between the check and the moment the consumer reads the event. The re-added peer is a **different connection**, and the old event says nothing about it — but `onPeerInactivityTimedOut` looked the peer up by key, found the new connection, saw `expectedWatcher == watcherInactivity`, and idled it. A connection that had just been established was torn down.

Separately, `interestedPeers` was read by the one-minute ticker goroutine while `AddPeer` / `RemovePeer` wrote it. Go answers a concurrent map iterate-and-write with an unrecoverable `fatal error`, not a recoverable panic, so this takes the whole client down. `-race` reproduces it.

### The fix

- `InactivePeersChan` now carries `map[string]peerid.ConnID` — for each idle peer, the connection it was sampled on.
- `onPeerInactivityTimedOut` drops an event whose `ConnID` is not the peer's live one.
- `interestedPeers` is guarded, and `checkStats` walks a snapshot so the lock is not held while logging.

The guard alone only silenced the race detector; the identity check is what stops a stale event reaching a new connection.

### Tests

In `client/internal/lazyconn/inactivity`:
- the event names the connection it was found on;
- a removed and re-added peer is reported on its **new** connection, not the replaced one;
- concurrent add/remove/check (fails with `fatal error: concurrent map iteration and map write` without the guard).

Verified on amd64, under `-race`, and with `GOARCH=386`.

- [x] Documentation is **not needed** for this change (explain why) — an internal concurrency and ordering fix with no user-facing surface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when peer activity is updated and checked concurrently.
  * Prevented stale inactivity notifications from affecting peers after their connections change.
  * Inactivity handling now tracks the specific connection associated with an idle peer.
  * Prevented potential race conditions during inactivity monitoring and statistics collection.
* **Tests**
  * Added coverage for concurrent peer additions, removals, and inactivity checks.
  * Added validation that inactivity events apply only to the connection where inactivity was detected.

### DIFF
--- a/client/internal/lazyconn/inactivity/manager.go
+++ b/client/internal/lazyconn/inactivity/manager.go
@@ -3,6 +3,7 @@ package inactivity
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -26,8 +27,10 @@ type Manager struct {
 	inactivePeersChan chan map[string]struct{}
 
 	iface               WgInterface
-	interestedPeers     map[string]*lazyconn.PeerConfig
 	inactivityThreshold time.Duration
+
+	mu              sync.RWMutex
+	interestedPeers map[string]*lazyconn.PeerConfig
 }
 
 func NewManager(iface WgInterface, configuredThreshold *time.Duration) *Manager {
@@ -60,6 +63,9 @@ func (m *Manager) AddPeer(peerCfg *lazyconn.PeerConfig) {
 		return
 	}
 
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
 	if _, exists := m.interestedPeers[peerCfg.PublicKey]; exists {
 		return
 	}
@@ -72,6 +78,9 @@ func (m *Manager) RemovePeer(peer string) {
 	if m == nil {
 		return
 	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
 
 	pi, ok := m.interestedPeers[peer]
 	if !ok {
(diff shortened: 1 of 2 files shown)


mnaza:
You were right, and thank you for saying it plainly — the lock was a symptom fix.

The map access was genuinely unsynchronised (Go answers a concurrent iterate-and-write with an unrecoverable fatal error, which is why `-race` caught it), but that is not the bug that reaches users. The ordering one is: the event carried only a peer key, so it could not say which connection the peer was idle on. A peer removed and re-added between the check and the read is a different connection; the consumer looked it up by key, found the new one, saw it was waiting for an inactivity event, and idled a connection that had just been established.

Pushed `8079852`: the event now carries the `ConnID` it was sampled on, and `onPeerInactivityTimedOut` drops an event that does not name the live connection. Two tests cover it in the inactivity package — the event names the connection it was found on, and a removed/re-added peer is reported on its new connection. Green on amd64, under `-race`, and on `GOARCH=386`.

Happy to adjust the shape if you would rather carry the identity differently.

mnaza:
Both review points were fair, and the first was a real miss on my side — pushed `8aecb44`.

**HA groups.** `checkHaGroupActivity` reads the same event map and tested only whether the peer key was present. An entry naming a connection that has since been replaced says nothing about the one live now, so a re-added group peer counted as idle and the group stopped deferring — the same ordering mistake the consumer had just stopped making, one function over. It now compares the stored `ConnID` with the peer's live one, the same rule as `onPeerInactivityTimedOut`.

**Test ticker.** The new tests swapped the package-global `newTicker` without restoring it, so a later test could inherit a ticker that never fires and sit until its own timeout. They now install and restore it via `t.Cleanup`. I left the two older tests in that file alone rather than widen the diff; happy to convert them too if you would prefer the file consistent.

Verified on amd64, with `inactivity` under `-race`, and on `GOARCH=386`.

One note so it is not read as mine: `TestManager_BindMode_MultiplePeers` in the sibling `activity` package fails under `-race` on a pristine checkout of `main` as well — I re-checked in a clean worktree today. Untouched by this PR.

PR #7488, opened 2026-09-09
[client] Locate the UDP header through the IHL field in the STUN filter

mnaza:
## Describe your changes

`sharedsock` exists so STUN can share the WireGuard port. Its cBPF filter decides what the
raw socket delivers. The ipv4 socket is opened as `AF_INET, SOCK_RAW, IPPROTO_UDP`, so the
packet arrives with its IP header. The filter read the destination port at a fixed offset
22 and the magic cookie at 32. Both are correct only when IHL is 5.

A datagram carrying IP options has a longer header. The two loads then read the wrong
bytes, the cookie test fails, and the kernel drops the packet. `ReadFrom` never sees it.
The asymmetry is the part worth noting. `ReadFrom` parses with
`gopacket.NewDecodingLayerParser(layers.LayerTypeIPv4, …)`, which is IHL-aware. So the
filter discards exactly what the parser behind it was written to handle. The symptom is a
peer that never learns its server-reflexive candidate on the shared port. It falls back to
a relay.

IP options are rare, and middleboxes often strip them. This is not a common path. It is an
uncommon one, and those are what this code exists for.

The fix puts the header length into X with `LoadMemShift`. Both fields are then read
relative to it. The ipv6 path is unchanged in effect and now has its own function. The
kernel strips the IPv6 header and any extension headers there, so the fixed offsets 2 and
12 are right for it.

The test assembles the real filter and runs it in `bpf.NewVM`. It needs no privileges and
no socket. Against the previous filter:

```
--- FAIL: TestIncomingSTUNFilterIPv4/one_option_word
        match = false, want true
--- FAIL: TestIncomingSTUNFilterIPv4/three_option_words
        match = false, want true
```

Trivial fix: the filter is corrected to accept the packets the decoder behind it already
handles. No API, protocol, flag or behavior change beyond that.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

- [x] Documentation is **not needed** for this change (explain why) — an internal packet
  filter offset fix with no user-visible surface: no setting, flag, or documented behavior
  changes.

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7488?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved STUN packet filtering for IPv4 packets with optional IP header fields.
  * Ensured valid IPv4 and IPv6 STUN traffic is recognized only on the configured port.
  * Prevented packets with incorrect ports or invalid STUN markers from matching the filter.

* **Tests**
  * Added coverage for IPv4 and IPv6 filtering, including optional IPv4 headers, incorrect ports, and invalid packets.

### DIFF
--- a/sharedsock/stun_filter_linux.go
+++ b/sharedsock/stun_filter_linux.go
@@ -16,27 +16,52 @@ func NewIncomingSTUNFilter() BPFFilter {
 
 // GetInstructions returns raw BPF instructions for ipv4 and ipv6 that filter out anything but STUN packets
 func (filter *IncomingSTUNFilter) GetInstructions(dstPort uint32) (raw4 []bpf.RawInstruction, raw6 []bpf.RawInstruction, err error) {
-	raw4, err = rawInstructions(22, 32, dstPort)
+	raw4, err = rawInstructions4(dstPort)
 	if err != nil {
 		return nil, nil, err
 	}
-	raw6, err = rawInstructions(2, 12, dstPort)
+	raw6, err = rawInstructions6(dstPort)
 	if err != nil {
 		return nil, nil, err
 	}
 	return raw4, raw6, nil
 }
 
-func rawInstructions(dstPortOff, cookieOff, dstPort uint32) ([]bpf.RawInstruction, error) {
-	// UDP raw socket for ipv4 receives the rcvdPacket with IP headers
-	// UDP raw socket for ipv6 receives the rcvdPacket with UDP headers
+// rawInstructions4 filters an ipv4 raw socket, which delivers the IP header along
+// with the packet. The IP header length varies, so the UDP header is located
+// through the IHL field rather than at a fixed offset. Reading the destination
+// port at a fixed 22 would miss every datagram carrying IP options.
+func rawInstructions4(dstPort uint32) ([]bpf.RawInstruction, error) {
 	instructions := []bpf.Instruction{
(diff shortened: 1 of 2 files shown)


pappz (changes requested):
Please fix the broken unit test.

mnaza:
Fixed. The test compared the filter result with `> 0`, but a BPF program returns the number of bytes to accept, and the conventional accept-everything value `0xffffffff` comes back as `-1` once the VM returns it as an `int` on 32-bit platforms. So every matching packet read as a miss on 386 while amd64 passed. It now compares against zero, which is what the semantics actually are. Verified on both amd64 and GOARCH=386; the filter itself is unchanged.

mnaza:
@pappz The test fix is pushed (`1eb3a63`) — the reasoning is in my comment above — and all checks are green, including the 386 job that was failing. The review still shows as changes-requested and I can't clear that from my side; could you take another look when you have a moment?

PR #7490, opened 2026-09-09
[client] Fix peers not being notified when the relay connection drops

pappz:
### The bug

`Manager.AddCloseListener` deduplicated registrations by comparing
`reflect.ValueOf(callback).Pointer()`. For a method value that pointer is the
compiler-generated wrapper, not the receiver, so
`peerA.onRelayClientDisconnected` and `peerB.onRelayClientDisconnected` compare
equal. All peers on the home relay register under the same `connectionURL` key,
so every registration after the first was dropped. Foreign relays had the same
defect, scoped to the peers sharing that server.

On a relay drop only one peer was notified. The rest kept `statusRelay` at
connected and never fired `guard.SetRelayedConnDisconnected()`, so no reconnect
was attempted. The conn itself was closed by `closeAllConns` — nothing leaked,
the state machine just never learned about it.

Peers whose proxy read loop is not running had no fallback at all: the
non-controller side (`Work()` is controller-only) and any peer holding relay
behind an active ICE connection. For those the manager listener was the only
channel.

Repro: three peers on one home relay, all relayed. Open A→B and A→C, then kill
A's relay connection. One peer reconnects, the other keeps reporting connected
over relay with no traffic passing. The common real-world shape is relay held
as backup behind working P2P: it dies unnoticed and only surfaces later when
P2P breaks and the fallback isn't there.

`TestCloseListenerPerReceiver` covers this; on the old code it fails with
`not every listener was notified, got: map[peer-a:1]`.

### Callbacks → context

The peer-level listener registry is removed from `Manager`. A relayed `Conn`
now exposes `Context()`, cancelled on teardown, built with
`context.WithCancelCause` so `context.Cause` names the reason
(`ErrServerDisconnected`, `ErrPeerDisconnected`, `net.ErrClosed`). `ctx.Err()`
is unchanged, so this is additive.

This is what quic-go does — `Conn`, `Stream` and `SendStream` expose
`Context()` and there is no close-listener API — and grpc's
`GetState`/`WaitForStateChange` is the same idea: consumers wait, they don't
register. Putting the signal on the resource removes the whole class of
problems the bug came from: listener identity, lifetime, deregistration, and
the window between registering and the resource dying.

`WorkerRelay` watches that context in a goroutine whose lifetime is the
connection's, so there is nothing extra to clean up.

Stale signals are filtered where the teardown happens. `Conn` records which
relayed connection the current proxy was built from, and `onRelayDisconnected`
takes the connection the signal belongs to, dropping it under `conn.mu` when it
is no longer current. Check and effect share one critical section, so a signal
for a superseded connection cannot tear down its replacement. This also covers
the proxy read loops, whose disconnect listener previously took no argument and
had the same defect.

### Also here

- `onRelayConnectionIsReady` checks the conn context: a relay dying between
  `OpenConn` and the handoff otherwise left the peer building a WireGuard
  endpoint over a dead connection.
- `OpenConn` returns `*Conn` instead of `net.Conn` down the chain. Every path
  already produced a `*Conn`; the interface only forced callers to assert back.
- `Client.SetOnDisconnectListener` is untouched — server-level, drives the
  reconnect guard and foreign relay eviction.

### Tests

- `TestCloseListenerPerReceiver` — three receivers on one relay, each notified
  exactly once. Fails without the change.
- `TestConnContextCancelledOnServerDisconnect` / `TestConnContextCauseOnLocalClose`
  — the cancellation cause distinguishes a relay drop from a local close.
- `TestNotifierDoubleAdd` removed: it covered the deleted mechanism.
- `TestForeignAutoClose` asserted nothing (both branches logged). It now waits
  for the relay to leave the client map and fails if it doesn't.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal relay client API and peer connection handling; no user-facing surface changes.

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7490?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/client/internal/peer/conn.go
+++ b/client/internal/peer/conn.go
@@ -560,7 +560,7 @@ func (conn *Conn) onRelayConnectionIsReady(rci RelayConnInfo) {
 	conn.mu.Lock()
 	defer conn.mu.Unlock()
 
-	if conn.ctx.Err() != nil {
+	if conn.ctx.Err() != nil || rci.relayedConn.Context().Err() != nil {
 		if err := rci.relayedConn.Close(); err != nil {
 			conn.Log.Warnf("failed to close unnecessary relayed connection: %v", err)
 		}
--- a/client/internal/peer/conn.go
+++ b/client/internal/peer/conn.go
@@ -135,9 +135,10 @@ type Conn struct {
 	// used to store the remote Rosenpass key for Relayed connection in case of connection update from ice
 	rosenpassRemoteKey []byte
 
-	wgProxyICE   wgproxy.Proxy
-	wgProxyRelay wgproxy.Proxy
-	handshaker   *Handshaker
+	wgProxyICE     wgproxy.Proxy
+	wgProxyRelay   wgproxy.Proxy
+	relayedConnRef *relayClient.Conn
+	handshaker     *Handshaker
 
 	guard *guard.Guard
 	wg    sync.WaitGroup
@@ -568,6 +569,7 @@ func (conn *Conn) onRelayConnectionIsReady(rci RelayConnInfo) {
 	}
 
 	conn.dumpState.RelayConnected()
+	conn.relayedConnRef = rci.relayedConn
 	conn.Log.Debugf("Relay connection has been established, setup the WireGuard")
 
 	wgProxy, err := conn.newProxy(rci.relayedConn)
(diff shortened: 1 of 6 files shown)


theodorsm (approved):
Tested and works well, reconnects happen faster too.

PR #7497, opened 2026-09-10
[management] Expire unvalidated custom domain registrations

mlsmaycon:
## Describe your changes

Unvalidated custom domains reserve their names indefinitely. Give pending registrations 48 hours to validate, then remove expired registrations at management startup and every 60 minutes. Removal releases the name and records `CustomDomainValidationExpired` for the original account.

Validation and deletion use conditional database writes so cleanup preserves a registration that became valid, and late validation cannot restore a deleted registration. Existing pending rows receive a full 48-hour window on upgrade. Legacy registrations with dependent services are retained for operator review. Cleanup waits for its activity write during graceful shutdown.

New custom domain registrations reuse `domain.FromString` for lowercase punycode conversion and the shared non-wildcard validator for syntax checks. A trailing dot is removed before availability checks, CNAME lookups, and storage. Cleanup retains literal SQL `LIKE` matching for existing rows and underscores accepted by the shared validator.

Local validation: `make lint`, `make lint-all`, and `TMPDIR=/tmp GOFLAGS=-short make test-unit` passed. Focused race tests passed for domain cleanup and validation, expiry event persistence during shutdown, server lifecycle, store behavior, and migration, including SQLite, PostgreSQL, and MySQL coverage. Event API integration tests passed 30 repetitions under the race detector after synchronizing the test with asynchronous audit persistence. Fake-clock tests use SQLite; boundary and concurrency tests exercise the selected database engine. The initial full `make test-unit` run failed existing macOS DNS integration tests (`TestDarwinDNSUncleanShutdownCleanup` and `TestMatchDomainBatching`) that depend on host DNS state and permissions. Short mode skips those tests; `/tmp` avoids macOS Unix socket path limits.

Follow-up to https://github.com/netbirdio/netbird/pull/7341, tracked in the same internal NetBird work item. No public issue.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

https://github.com/netbirdio/docs/pull/969

### DIFF
--- a/management/internals/modules/reverseproxy/domain/manager/expiration.go
+++ b/management/internals/modules/reverseproxy/domain/manager/expiration.go
@@ -1,0 +1,73 @@
+package manager
+
+import (
+	"context"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/netbirdio/netbird/management/internals/modules/reverseproxy/domain"
+	"github.com/netbirdio/netbird/management/server/activity"
+)
+
+const (
+	validationCleanupInterval = 60 * time.Minute
+	validationCleanupBatch    = 100
+)
+
+// RunValidationCleanup removes expired registrations on startup and hourly until cancellation.
+func (m Manager) RunValidationCleanup(ctx context.Context) {
+	ticker := time.NewTicker(validationCleanupInterval)
+	defer ticker.Stop()
+	for {
+		m.cleanupExpiredDomains(ctx, time.Now().UTC())
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+		}
+	}
+}
+
+func (m Manager) cleanupExpiredDomains(ctx context.Context, now time.Time) {
+	var afterID domain.ID
+	for ctx.Err() == nil {
+		domains, err := m.store.GetExpiredCustomDomains(ctx, now, afterID, validationCleanupBatch)
+		if err != nil {
+			if ctx.Err() == nil {
+				log.WithContext(ctx).WithError(err).Error("list expired custom domain registrations")
+			}
+			return
+		}
+		for _, d := range domains {
+			if ctx.Err() != nil {
+				return
+			}
+			m.deleteExpiredDomain(ctx, d, now)
+			afterID = domain.ID(d.ID)
+		}
+		if len(domains) < validationCleanupBatch {
+			return
+		}
+	}
+}
+
+func (m Manager) deleteExpiredDomain(ctx context.Context, d *domain.Domain, now time.Time) {
+	deleted, err := m.store.DeleteExpiredCustomDomain(ctx, d, now)
+	if err != nil {
+		if ctx.Err() == nil {
+			log.WithContext(ctx).WithFields(log.Fields{"accountID": d.AccountID, "domainID": d.ID}).
+				WithError(err).Warn("could not expire custom domain registration")
+		}
+		return
+	}
+	if !deleted {
+		return
+	}
+	meta := d.EventMeta()
+	if d.ValidationExpiresAt != nil {
+		meta["validation_expires_at"] = d.ValidationExpiresAt.UTC().Format(time.RFC3339)
+	}
+	m.accountManager.StoreEvent(ctx, activity.SystemInitiator, d.ID, d.AccountID,
+		activity.CustomDomainValidationExpired, meta)
+}
--- a/management/internals/modules/reverseproxy/domain/manager/manager.go
+++ b/management/internals/modules/reverseproxy/domain/manager/manager.go
@@ -277,11 +277,15 @@ func (m Manager) ValidateDomain(ctx context.Context, accountID, userID, domainID
 	if m.validator.IsValid(context.Background(), d.Domain, []string{targetCluster}) {
 		d.Validated = true
 		if _, err := m.store.UpdateCustomDomain(context.Background(), accountID, d); err != nil {
-			log.WithFields(log.Fields{
+			entry := log.WithFields(log.Fields{
 				"accountID": accountID,
 				"domainID":  domainID,
-				"domain":    d.Domain,
-			}).WithError(err).Error("update custom domain in store")
+			}).WithError(err)
+			if sErr, ok := status.FromError(err); ok && sErr.Type() == status.PreconditionFailed {
+				entry.Debug("custom domain registration is no longer pending validation")
+				return
+			}
+			entry.Error("update custom domain in store")
 			return
 		}
 		log.WithFields(log.Fields{"accountID": accountID, "domainID": domainID}).
--- a/management/internals/modules/reverseproxy/domain/manager/manager.go
+++ b/management/internals/modules/reverseproxy/domain/manager/manager.go
@@ -19,6 +19,7 @@ import (
 	"github.com/netbirdio/netbird/management/server/permissions/operations"
 	nbstore "github.com/netbirdio/netbird/management/server/store"
 	"github.com/netbirdio/netbird/management/server/types"
+	nbdomain "github.com/netbirdio/netbird/shared/management/domain"
 	"github.com/netbirdio/netbird/shared/management/status"
 )
 
@@ -130,6 +131,7 @@ func (m Manager) GetDomains(ctx context.Context, accountID, userID string) ([]*d
 	return ret, nil
 }
 
+// CreateDomain registers a normalized custom domain and attempts DNS validation.
 func (m Manager) CreateDomain(ctx context.Context, accountID, userID, domainName, targetCluster string) (*domain.Domain, error) {
 	ok, ctx, err := m.permissionsManager.ValidateUserPermissions(ctx, accountID, userID, modules.Services, operations.Create)
 	if err != nil {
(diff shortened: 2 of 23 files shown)


management/internals/server/server.go
```diff
@@ -236,14 +237,35 @@ func (s *BaseServer) Start(ctx context.Context) error {
...
+		mgr.RunValidationCleanup(ctx)
+	}()
+}
+
 // Stop attempts a graceful shutdown, waiting up to 5 seconds for active connections to finish
 func (s *BaseServer) Stop() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
+	if s.domainCleanupStop != nil {
+		s.domainCleanupStop()
```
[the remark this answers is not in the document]
mlsmaycon: Fixed in 1dd61dbbb. CustomDomainValidationExpired is persisted synchronously before the cleanup worker returns, using a bounded context that survives worker cancellation. A real SQLite event-store regression test cancels cleanup, records the event, closes and reopens the store, and verifies the event remains. It failed before the fix and now passes under the race detector.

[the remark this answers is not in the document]
mlsmaycon: The five-second context governs graceful connection shutdown; Stop also joins owned workers before closing their stores. Every cleanup query uses WithContext(ctx), so worker cancellation reaches the database operation, and the loop checks cancellation before further work. Keeping the join prevents store closure while cleanup is still using it. The expiration event now completes synchronously with its own five-second bound. A driver that ignores cancellation would also block database closure, so dropping the join would not establish a bounded shutdown. Keeping the join here.

management/internals/modules/reverseproxy/domain/manager/manager.go
```diff
@@ -243,6 +247,14 @@ func (m Manager) ValidateDomain(ctx context.Context, accountID, userID, domainID
 		}).WithError(err).Error("get custom domain from store")
 		return
 	}
+	if d.Validated {
+		return
+	}
+	if d.ValidationExpiresAt == nil || !time.Now().Before(*d.ValidationExpiresAt) {
```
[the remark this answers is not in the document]
mlsmaycon: Fixed in 1dd61dbbb. Expected PreconditionFailed results are logged at debug level, while database errors remain at error level. The log fields contain account and registration IDs. Deadline-crossing tests still verify that rejected writes emit no DomainValidated event.

mlsmaycon:
The new exported production functions have doc comments. Private helpers remain self-descriptive, following AGENTS.md's comments guidance to explain hidden constraints and avoid comments that repeat the code. The docstring-percentage warning does not justify adding redundant comments to private helpers or tests; lint passes without changing any thresholds or exclusions.

management/server/event.go
```diff
@@ -61,23 +61,33 @@ func (am *DefaultAccountManager) GetEvents(ctx context.Context, accountID, userI
...
+			AccountID:   accountID,
+			Meta:        meta,
+		})
+		if err != nil {
+			log.WithContext(ctx).Errorf("received an error while storing an activity event, error: %s", err)
+		}
+	}
+	if activityID == activity.CustomDomainValidationExpired {
+		// The domain is already deleted; shutdown must allow its audit write to finish.
+		ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)
```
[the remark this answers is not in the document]
mlsmaycon: Fixed in 6edadad5b. Store.Save now applies its context to the activity insert and deleted-user metadata write. A real SQLite connection-pool test holds the only connection, verifies the write returns DeadlineExceeded, and checks that releasing the connection cannot persist the timed-out event. The test failed before the fix and now passes under the race detector. Ordinary asynchronous audit writes detach request cancellation so completed operations still retain their events.

management/server/event.go
```diff
@@ -61,23 +61,34 @@ func (am *DefaultAccountManager) GetEvents(ctx context.Context, accountID, userI
...
+		defer cancel()
+		save(ctx)
+		return
 	}
+	// Request cancellation must not discard the audit record of a completed operation.
+	go save(context.WithoutCancel(ctx))
```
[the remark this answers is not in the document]
mlsmaycon: This preserves the existing asynchronous audit behavior: before this PR, Store.Save ignored its context, so those writes already had no deadline. Once Save honors contexts, detaching the request is necessary to avoid losing successful-operation audit records when HTTP requests finish; TestStoreEvent_CanceledContext covers that. CustomDomainValidationExpired has the new bounded synchronous write required by this cleanup worker. A global timeout/drop policy for all other audit events would change behavior beyond domain expiry, so I am leaving that as a separate improvement.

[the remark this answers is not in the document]
mlsmaycon: This preserves the existing asynchronous audit behavior: before this PR, Store.Save ignored its context, so those writes already had no deadline. Once Save honors contexts, detaching the request is necessary to avoid losing successful-operation audit records when HTTP requests finish; TestStoreEvent_CanceledContext covers that. CustomDomainValidationExpired has the new bounded synchronous write required by this cleanup worker. A global timeout/drop policy for all other audit events would change behavior beyond domain expiry, so I am leaving that as a separate improvement.

management/server/store/sql_store_domain_expiration.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"gorm.io/gorm"
+
+	"github.com/netbirdio/netbird/management/internals/modules/reverseproxy/domain"
+	rpservice "github.com/netbirdio/netbird/management/internals/modules/reverseproxy/service"
+	"github.com/netbirdio/netbird/shared/management/status"
+)
+
+// GetExpiredCustomDomains lists pending registrations in stable batches across accounts.
+func (s *SqlStore) GetExpiredCustomDomains(ctx context.Context, now time.Time, afterID domain.ID, limit int) ([]*domain.Domain, error) {
```
dmitri-netbird, line 17: drive-by comment: is this used or is planned to be used anywhere else? If not, why not delete expired domains here, that would save queries and reduce traffic...

dmitri-netbird: ah, I see, it's more complicated than I thought and involves a services check.

management/server/store/sql_store_domain_expiration.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	if count > 0 {
+		return false, status.Errorf(status.PreconditionFailed, "expired custom domain still has dependent services")
+	}
+	return false, nil
+}
+
+func customDomainServices(db *gorm.DB, d *domain.Domain) *gorm.DB {
+	name := strings.ToLower(strings.TrimSuffix(d.Domain, "."))
+	escaped := strings.NewReplacer("!", "!!", "%", "!%", "_", "!_").Replace(name)
```
dmitri-netbird, line 55: bordering on a nit: I assume we validate domain names elsewhere, and none of these characters are valid for domain names. As domains are coming from a trusted source, I think this is excessive?

mlsmaycon: Good catch. I am adding a normalization when registering the custom domain. But the escape logic will stay to handle SQL constraints.

management/internals/modules/reverseproxy/domain/manager/expiration.go
```diff
@@ -0,0 +1,73 @@
...
+	validationCleanupInterval = 60 * time.Minute
+	validationCleanupBatch    = 100
+)
+
+// RunValidationCleanup removes expired registrations on startup and hourly until cancellation.
+func (m Manager) RunValidationCleanup(ctx context.Context) {
+	ticker := time.NewTicker(validationCleanupInterval)
+	defer ticker.Stop()
+	for {
+		m.cleanupExpiredDomains(ctx, time.Now().UTC())
```
dmitri-netbird, line 23: May want to delay cleanup on startup (like jitter + sync.Once + time.After? Or just reset the ticker after the first tick?); It's probably not a lot of domains to cleanup, but things are busy, and deleting a domain involves a rather expensive query.

management/server/store/sql_store.go
```diff
@@ -5733,12 +5737,21 @@ func (s *SqlStore) CreateCustomDomain(ctx context.Context, accountID string, dom
...
 }
 
+// UpdateCustomDomain completes validation only while the original registration is pending.
 func (s *SqlStore) UpdateCustomDomain(ctx context.Context, accountID string, d *domain.Domain) (*domain.Domain, error) {
-	d.AccountID = accountID
-	result := s.db.Select("*").Save(d)
+	if !d.Validated {
+		return nil, status.Errorf(status.InvalidArgument, "custom domain update must complete validation")
+	}
+	result := s.db.WithContext(ctx).Model(&domain.Domain{}).
```
dmitri-netbird, line 5745: Pascal is telling me that using db.WithContext when the context is coming from an api request can result in unexpected context cancellations.

dmitri-netbird (approved):
One more comment, but lgtm

management/server/store/sql_store.go
mlsmaycon: will check on different environment and fix it if I confirm it again.

PR #7500, opened 2026-09-10
[client] Continue Kubernetes discovery after reverse-DNS failures

### LINKED ISSUE #7437: `netbird kubernetes list/write-kubeconfig` fails with "lookup …in-addr.arpa: no such host" — reverse lookups go to the system resolver instead of NetBird's
**Environment**
- NetBird client 0.77.1 and 0.78.1 (macOS 15, Apple Silicon); self-hosted management
- Kubernetes operator 0.8.0 with a working `ClusterProxy` (peers connected, `Ready=True`)
- Client DNS management enabled; no nameserver groups configured initially

**Steps**
1. Have a ClusterProxy registered and reachable.
2. Run `netbird kubernetes list` (or `netbird kubernetes write-kubeconfig <cluster>`).

**Actual**
```
Error: lookup z.y.x.100.in-addr.arpa. on <upstream resolver>:53: no such host
```
The IP in the error is an arbitrary *other* peer in the network (an offline tablet, a laptop), not the proxy. The command aborts on the first peer whose PTR the resolver cannot answer. Nothing is listed.

**Analysis**
- The lookup is performed against the address in `/etc/resolv.conf` (the upstream resolver), not against NetBird's embedded resolver, and it is fatal.
- On macOS, NetBird installs a scoped resolver only for the account domain (`*.nb.<domain>`), so `in-addr.arpa` queries never reach it. Adding a nameserver group matching the mesh `/16` reverse zone makes the OS route PTRs correctly (`dig -x <peer-ip> @<NetBird resolver IP>` answers), but the daemon still queries the upstream resolver directly — `GODEBUG=netdns=cgo` makes no difference — so that fix has no effect on the CLI.
- The client already holds every peer's IP→FQDN mapping; a DNS round-trip is unnecessary to display it.

**Workaround** (works, but changes system DNS for every affected peer): a **primary** nameserver group for the operator peers, so NetBird's resolver becomes the system resolver and the daemon's PTR queries land on it. After that, `list` and `write-kubeconfig` work.

**Suggested fix**
1. Resolve peer display names from the daemon's own peer store (or its embedded resolver), never via the system resolver.
2. Make PTR failures non-fatal: show the IP and continue — one unresolvable peer should not hide every cluster.

Devesh36:
## Describe your changes

netbird kubernetes list and write-kubeconfig currently abort when any peer has no PTR record. A single unrelated offline peer can therefore hide reachable Kubernetes clusters.

Reverse-DNS failures are now logged at debug level and skipped so discovery continues. A regression test covers the failure path.

Focused tests pass:
go test ./client/cmd -run 'Test(GetKubernetesClustersSkipsReverseLookupFailures|FingerprintClusters|WriteKubeconfig|ResolveKubeconfigPath)$'

The full client/cmd package run is currently affected by unrelated Unix-socket bind failures in the macOS sandbox.

Closes #7437

### Checklist
- [x] Is it a bug fix
- [ ] Is it a feature enhancement
- [ ] It is a refactor

## Documentation
Select exactly one:

- [x] Documentation is **not needed** for this change: this only changes failure handling in an existing CLI discovery flow and adds no new user-facing option.

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7500?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Kubernetes cluster discovery now continues when reverse-DNS lookups fail, rather than stopping the entire discovery process.
  * Unresolvable or unrelated peer hostnames are skipped without preventing other discovery results from being processed.
  * Discovery now uses peer-provided fully qualified domain names when available and still respects canceled requests.

* **Tests**
  * Added coverage for reverse-DNS failures, stored peer hostnames, and cancellation handling.

### DIFF
--- a/client/cmd/kubernetes.go
+++ b/client/cmd/kubernetes.go
@@ -119,6 +119,10 @@ type kubernetesCluster struct {
 	version string
 }
 
+type reverseResolver interface {
+	LookupAddr(context.Context, string) ([]string, error)
+}
+
 func getKubernetesClusters(ctx context.Context, peers []*proto.PeerState, nameFilter string) ([]kubernetesCluster, error) {
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 	transport.TLSClientConfig = &tls.Config{
@@ -132,13 +136,23 @@ func getKubernetesClusters(ctx context.Context, peers []*proto.PeerState, nameFi
 		// https://github.com/golang/go/issues/17093
 		PreferGo: true,
 	}
+	return getKubernetesClustersWithResolver(ctx, peers, nameFilter, &resolver, httpClient)
+}
 
+func getKubernetesClustersWithResolver(
+	ctx context.Context,
+	peers []*proto.PeerState,
+	nameFilter string,
+	resolver reverseResolver,
+	httpClient *http.Client,
+) ([]kubernetesCluster, error) {
 	kcs := []kubernetesCluster{}
 	attempted := map[string]struct{}{}
 	for _, peer := range peers {
 		fqdns, err := resolver.LookupAddr(ctx, peer.IP)
 		if err != nil {
-			return nil, err
+			log.Debugf("could not reverse-resolve peer %s: %v", peer.IP, err)
+			continue
 		}
 		for _, fqdn := range fqdns {
 			if _, ok := attempted[fqdn]; ok {
--- a/client/cmd/kubernetes.go
+++ b/client/cmd/kubernetes.go
@@ -123,6 +123,7 @@ type reverseResolver interface {
 	LookupAddr(context.Context, string) ([]string, error)
 }
 
+// getKubernetesClusters discovers Kubernetes clusters from the supplied peer states.
 func getKubernetesClusters(ctx context.Context, peers []*proto.PeerState, nameFilter string) ([]kubernetesCluster, error) {
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 	transport.TLSClientConfig = &tls.Config{
@@ -139,6 +140,7 @@ func getKubernetesClusters(ctx context.Context, peers []*proto.PeerState, nameFi
 	return getKubernetesClustersWithResolver(ctx, peers, nameFilter, &resolver, httpClient)
 }
 
+// getKubernetesClustersWithResolver discovers clusters using the provided dependencies.
 func getKubernetesClustersWithResolver(
 	ctx context.Context,
 	peers []*proto.PeerState,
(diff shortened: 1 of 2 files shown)


Devesh36:
Addressed the review feedback in follow-up commit 10ee3b30: use the daemon-provided peer FQDN before attempting reverse DNS; preserve context cancellation and deadline errors; continue after ordinary reverse-DNS failures; assert both peer IPs are processed in the regression test; and add comments for touched functions. Focused client/cmd tests pass. The branch also includes the remote main merge already present on the PR branch; no history was rewritten.

g-roliveira:
This change stops discovery from working rather than fixing it.

`peerFQDNs` now returns `[]string{peer.GetFqdn()}` whenever the FQDN is non-empty, and with a DNS domain configured it is non-empty for every peer in the network map (`shared/management/networkmap/encode.go` sets `Fqdn: rPeer.FQDN(dnsName)`). `LookupAddr` is therefore never reached.

That matters because the cluster alias is not in `Fqdn`. `Fqdn` is `<dns_label>.<dns_domain>` (`shared/management/networkmap/nmdata/peer.go`), while `netbird-kubeapi-proxy` arrives as an *extra* DNS label: the proxy sets `DNSLabels: [clusterName + "." + cmd.KubernetesDNSSuffix]`, which becomes an extra A record in the peer custom zone (`shared/management/networkmap/peers_custom_zone.go`) and therefore an extra PTR record (`client/internal/dns.go`, `collectPTRRecords`).

A proxy peer consequently answers a reverse lookup with two names:

```
$ dig -x <proxy-overlay-ip> @<daemon-dns-address>
<ip>.in-addr.arpa. 300 IN PTR <pod-name>.<dns-domain>.
<ip>.in-addr.arpa. 300 IN PTR prod.netbird-kubeapi-proxy.netbird.cloud.
```

`Fqdn` is only the first of the two. With this patch, `comps[1]` in `getKubernetesClustersWithResolver` is the first label of the account DNS domain, so every peer is skipped and `netbird kubernetes list` prints `No Kubernetes clusters available.` instead of failing. A silently empty list is worse than the current error, and it no longer matches the output documented in netbirdio/docs (`use-cases/kubernetes/api-server-proxy.mdx`).

The test added here does not catch this: the fixture `Fqdn: "cluster.netbird-kubeapi-proxy"` has no DNS domain, which is not a shape any peer has.

Failing soft is the right half of this change and worth keeping. The lookup itself just has to keep running, because the alias exists only in reverse DNS. Pointing that lookup at the daemon's embedded resolver instead of the system resolver clears the original abort without losing the alias.

PR #7515, opened 2026-09-11
[management] Prevent deleting custom domains used by services

mlsmaycon:
## Describe your changes

Deleting a custom domain must not release its name while services still reference it. Return HTTP 412 when any service uses the domain or a subdomain, including disabled services, while preserving the registration and avoiding a deletion activity event.

Service creation and updates recheck custom-domain authorization inside their write transaction and hold shared locks on matching registrations until commit. Deletion takes an exclusive lock, preventing stale authorization without serializing service writes. Free cluster domains keep their existing path. The API schema and user documentation describe how to remove or move dependencies before deleting a domain.

Validation: reproduced the API regression before the fix; both affected manager packages pass with the race detector, and store dependency/concurrency tests pass on SQLite, PostgreSQL, and MySQL. Changed-code and full-repository lint pass. API types and the store mock were regenerated with the repository generators.

The host-safe unit suite passed with `TMPDIR=/tmp GOFLAGS=-short make test-unit`. Short mode avoids the previously identified host-dependent macOS DNS tests; the temporary path avoids macOS Unix socket length limits.

Follow-up to https://github.com/netbirdio/netbird/pull/7497 and https://github.com/netbirdio/netbird/pull/7341, tracked in the same agreed internal NetBird work item. No public issue.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

https://github.com/netbirdio/docs/pull/972

## Summary by CodeRabbit

* **New Features**
  * Service creation and updates now verify that custom domains remain valid and target the correct cluster before changes are saved.
* **Bug Fixes**
  * Prevented deletion of custom domains still used by services or subdomains, including disabled services.
  * Blocked deletion returns HTTP 412 without changing the domain or its services.
  * Improved consistency when domain deletion occurs during service creation or updates.
* **Documentation**
  * Updated API documentation to describe domain usage requirements and the 412 response.

### DIFF
--- a/management/server/store/sql_store_domain_deletion.go
+++ b/management/server/store/sql_store_domain_deletion.go
@@ -1,0 +1,49 @@
+package store
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+
+	"github.com/netbirdio/netbird/management/internals/modules/reverseproxy/domain"
+	"github.com/netbirdio/netbird/shared/management/status"
+)
+
+// LockCustomDomains locks an account's registrations until the caller's transaction ends.
+func (s *SqlStore) LockCustomDomains(ctx context.Context, accountID string) ([]*domain.Domain, error) {
+	var domains []*domain.Domain
+	if err := s.db.WithContext(ctx).Clauses(clause.Locking{Strength: string(LockingStrengthUpdate)}).
+		Where(accountIDCondition, accountID).Order("id").Find(&domains).Error; err != nil {
+		return nil, fmt.Errorf("lock custom domains: %w", err)
+	}
+	return domains, nil
+}
+
+// DeleteCustomDomain removes a registration only when no service uses its namespace.
+func (s *SqlStore) DeleteCustomDomain(ctx context.Context, accountID string, domainID string) error {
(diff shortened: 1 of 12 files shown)


management/server/store/sql_store_domain_deletion.go
```diff
@@ -0,0 +1,58 @@
...
+		if err := tx.Clauses(clause.Locking{Strength: string(LockingStrengthUpdate)}).
+			Take(&d, accountAndIDQueryCondition, accountID, domainID).Error; err != nil {
+			if errors.Is(err, gorm.ErrRecordNotFound) {
+				return status.Errorf(status.NotFound, "custom domain not found")
+			}
+			return fmt.Errorf("lock custom domain for deletion: %w", err)
+		}
+
+		result := tx.Where(accountAndIDQueryCondition, accountID, domainID).
+			Where("NOT EXISTS (?)", customDomainServices(tx, &d).Select("1")).Delete(&domain.Domain{})
```
bcmmbaga, line 49: `accountAndIDQueryCondition` scopes the domain deletion but `customDomainServices` searches services across all accounts.

Eg. account A can register `example.com` while account B registers `team.example.com`. B’s authorized service at `app.team.example.com` would then incorrectly prevent A from deleting `example.com`


The dependency lookup should also be scoped to the domain’s account

bcmmbaga (changes requested):

PR #7528, opened 2026-09-14
[client,android] Generate debug bundle to file

pappz:
## Describe your changes

Adds a file export path for the Android debug bundle and makes the bundle complete on Android.

- `DebugBundleFile` returns the path of the generated zip instead of uploading it, so the app can hand the file to the user. `DebugBundle` keeps uploading, and both now share one implementation. The exported zip is renamed to `netbird.debug-file.*.zip`, so the stale cleanup below never removes a caller-owned file.
- An interrupted generation or upload leaves a zip behind in the cache directory. `RemoveStaleBundles` drops leftovers older than an hour before a new run; younger ones survive, because a concurrent remote debug job may still be writing them.
- The Android client never enabled sync response persistence, so `network_map.json` was silently missing from every Android bundle. Both start paths now enable it, matching iOS and the desktop daemon.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

- [x] Documentation is **not needed** for this change (explain why): the new API is internal to the Android client bridge; the user-facing behavior is documented on the Android app side and no public CLI, flag, or protocol changes.

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7528?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added an option to generate and retrieve Android debug bundles without uploading them.
  - Enabled synchronization responses to persist during Android runs.

- **Bug Fixes**
  - Improved debug bundle cleanup by removing outdated generated bundles while preserving caller-owned exported bundles.
  - Continued preserving recent bundles and unrelated files during cleanup.


client/android/client.go
```diff
@@ -256,6 +246,7 @@ func (c *Client) RunWithoutLogin(platformFiles PlatformFiles, dns *DNSList, dnsR
 	connectClient := internal.NewConnectClient(ctx, cfg, c.recorder,
 		internal.WithNetEvents(c.netMgr))
 	c.setState(cfg, cacheDir, cfgFile, connectClient)
+	connectClient.SetSyncResponsePersistence(true)
```
riccardomanfrin: This on iOS calls `NewDiskStore(dir)` while on android calls `NewMemoryStore()`

client/internal/debug/debug.go
```diff
@@ -379,9 +379,52 @@ func NewBundleGenerator(deps GeneratorDependencies, cfg BundleConfig) *BundleGen
...
+	if err := os.Rename(path, exported); err != nil {
+		return "", fmt.Errorf("export debug bundle: %w", err)
+	}
+	return exported, nil
+}
+
 // Generate creates a debug bundle and returns the location.
 func (g *BundleGenerator) Generate() (resp string, err error) {
-	bundlePath, err := os.CreateTemp(g.tempDir, "netbird.debug.*.zip")
+	bundlePath, err := os.CreateTemp(g.tempDir, bundleFilePattern)
```
riccardomanfrin: here we use the tempDir while elsewhere (e.g. `RemoveStaleBundles`) I see we use the cacheDir.. is it ok /wanted?

client/internal/debug/debug.go
```diff
@@ -379,9 +379,52 @@ func NewBundleGenerator(deps GeneratorDependencies, cfg BundleConfig) *BundleGen
...
+	for _, path := range matches {
+		info, err := os.Stat(path)
+		if err != nil || info.ModTime().After(cutoff) {
+			continue
+		}
+		if err := os.Remove(path); err != nil {
+			log.Debugf("remove stale debug bundle %s: %v", path, err)
+			continue
+		}
+		log.Infof("removed stale debug bundle %s", path)
```
riccardomanfrin: We log Info here, but we log Debug just above for error

client/android/client.go
```diff
@@ -327,6 +318,21 @@ func (c *Client) NotifyNetworkChange() {
...
+// the cache directory instead of uploading it, so the app can hand the file to
+// the user for inspection. The caller owns the file and removes it once done;
+// the stale-bundle cleanup of later runs never touches it.
+// anonymize and anonymizeLevel behave as in DebugBundle.
+func (c *Client) DebugBundleFile(platformFiles PlatformFiles, anonymize bool, anonymizeLevel string) (string, error) {
+	return c.debugBundle(platformFiles, anonymize, anonymizeLevel, false)
+}
+
+// debugBundle builds the bundle zip and either uploads it, returning the upload
+// key and removing the file, or leaves the file in place and returns its path.
```
riccardomanfrin: "leaves the file".. it renames it if it's the "file" kind

riccardomanfrin (commented):
1. Can we test that `network_map.json` is actually exported?
2. While we cleanup the legacy bundles (netbird.debug.*.zip) we don't cleanup the "file" stale ones (pr says "the caller owns the file and removes it", but on android we use the cache dir (`context.getCacheDir()`) so Android OS itself might cope with it (?)

client/android/client.go
pappz: It's a fact. What are you trying to say with it? I haven't changed this behavior in this PR.

client/internal/debug/debug.go
pappz: Same dir. Andoird populate it from cache dir.

pappz:
> 1. Can we test that `network_map.json` is actually exported?
> 2. While we cleanup the legacy bundles (netbird.debug.*.zip) we don't cleanup the "file" stale ones (pr says "the caller owns the file and removes it", but on android we use the cache dir (`context.getCacheDir()`) so Android OS itself might cope with it (?)

done

PR #7548, opened 2026-09-15
[client] Skip expiration session warnings that fire after their window

pappz:
## Describe your changes

The warning timers run on the monotonic clock, which does not advance while device is suspended. A timer armed for T-10 or T-2 can therefore fire long after the window it was armed for, delivering a "session expires soon" notification once that window is already gone.

Gate both callbacks on the wall clock at fire time: the T-10 warning is skipped once the final-warning window has been reached, and the final warning is skipped once the deadline itself has passed. Both set their edge guard before returning so a skipped warning cannot fire again for the same deadline.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7548?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session warning behavior after Android devices resume from suspension.
  * Prevented outdated interactive and final warnings from appearing after their scheduled time has passed.
  * Ensured disabled final-warning settings do not trigger post-deadline warnings.


client/internal/auth/sessionwatch/watcher.go
```diff
@@ -303,6 +303,12 @@ func (w *Watcher) fire(armedFor time.Time) {
 		w.mu.Unlock()
 		return
 	}
+	if w.lateLocked(armedFor, max(w.finalLead, 0)) {
+		w.firedAt = armedFor
```
riccardomanfrin: Is this correct? we didn't deliver a warning.. probably harmless if we are about to re-instrument the watcher but I'm not sure.
If the device hasn't synced with NTP and has drifted time and callback did rerun it would skip anyway.

client/internal/auth/sessionwatch/watcher.go
```diff
@@ -303,6 +303,12 @@ func (w *Watcher) fire(armedFor time.Time) {
...
+		w.mu.Unlock()
+		log.Infof("auth session expiry soon warning skipped (final-warning window already passed)")
```
riccardomanfrin: Does it make sense to add now and cutoff here? is it better info or warn? (same doubt for below Infof)

riccardomanfrin (commented):
**1. Test coverage**

The package already has `fakeRecorder` and `NewWithLeads`, but the new guard reads `time.Now()` internally, so the cases that matter are hard to pin down: the late paths are reachable only by feeding a deadline that is already in the past, and the boundary case (resume inside the [T-10, T-2] window, which must still warn) depends on real time. Extracting the comparison as a free function and injecting the clock covers both sides deterministically, and lets the existing timing-based tests drop their sleeps:

```go
func isLate(now, armedFor time.Time, cutoffLead time.Duration) bool {
	return !now.Round(0).Before(armedFor.Add(-cutoffLead).Round(0))
}
```

with `nowFn func() time.Time` on the Watcher, defaulting to `time.Now`. It also drops the `Locked` suffix from a helper that reads no shared state.

Cases worth covering: skipped past the final window, skipped past the deadline, and a resume at T-3 that must still fire.

**2. A resume before the deadline now goes silent**

Suspend at T-60, resume at T-5: the monotonic timer fires late and the guard discards it, but the user still has 5 minutes to extend. Before this change they got a late warning; now they get nothing and walk into the expiry without knowing they had time.

As far as I can tell the only thing that re-arms that warning is an engine restart: the watcher is engine-scoped, so a fresh one runs `Update` and recomputes the delay against the current wall clock. `Update` with the same deadline value is a no-op (`watcher.go:169`), so a post-resume sync on its own does not repair it. Does a resume on Android reliably restart the engine? If it does, worth stating in the commit message. If it does not, the change trades a late warning for no warning at all.

**Open question, not about this PR**

This watcher, like others we have, works on a human scale (minutes). Management delivers the absolute time at which the session expires, and we need to warn the user early enough (several minutes) to extend. Arming a relative timer for a point in the future has two problems:

a. the device clock may not be synced yet (NTP has not kicked in);
b. suspension time-warps past the armed point.

A periodic wall clock check does not make (a) go away: if the clock is wrong, the check is wrong too. The difference is recovery. A poll re-evaluates and warns as soon as NTP corrects, while the one-shot timer has already fired and burned its edge guard, and the same-value `Update` never re-arms it. For (b), the first tick after resume sees the true remaining time and decides correctly.

We already do this elsewhere: `client/internal/session.go:29` polls the wall clock every 2s for the expired-session path (`IsLoginRequired` -> `onExpireListener`), wired in `server.go:308`. Same process, same concern, opposite mechanism.

With a 10s tick against minute-scale leads the granularity is irrelevant, and a fair amount of machinery collapses: the two timers, `armOneShotLocked`, the `finalLead <= 0` special case and the late guards all become one comparison against `time.Now`, with the decision a pure function of (deadline, now, guards) that tests can table without sleeping. The ticker only needs to run while a deadline is tracked.

If we go that way, the edge-guard write on the skip path has to go with it, otherwise the first tick after a resume finds the warning already marked as delivered and stays quiet.

PR #3459, opened 2025-03-07
[client] Fix DNS Nrpt policies

webalexeu:
## Describe your changes
- The registry key that is currently used to check if the client is in GPO mode or not is not correct. You can have DNSClient key present if you define other DNS settings by GPO (Ex: Disable Netbios, disable mutlicast, ...) but not using GPO to configure Nrpt policies.
- When client is in GPO mode, there is no need to push the local policy configuration as this entry will be discared. The way of Microsoft to check if there is NRPT settings configured by GPO is to check the DnsPolicyConfig key is present (That's why it's coherent to check that key and not the DNSClient key)

> If any NRPT settings are configured in domain Group Policy, then all local Group Policy NRPT settings are ignored
https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/dn593632(v=ws.11)

## Issue ticket number and link
Partial fix https://github.com/netbirdio/netbird/issues/3332

### Checklist
- [x] Is it a bug fix
- [ ] Is a typo/documentation fix
- [ ] Is a feature enhancement
- [ ] It is a refactor
- [ ] Created tests that fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

Issue #2895, opened 2024-11-15
Broken DNS resolution of *.our.domain on Windows client


tomashora:
**Describe the problem**

The resolution of all subdomains under *our.domain does not work for certain application (for example any web browser or ping tool, however nslookup resolves IP correctly). This used to happen in the past when netbird was shut down incorrectly, as [discussed on Slack](https://netbirdio.slack.com/archives/C02KHAE8VLZ/p1705270559847019). Now it seems to happen the same way - hard laptop shutdown, system boots up, dns not resolved.

This results that the clients cannot connect.

**To Reproduce**

TBD

**Expected behavior**

All DNS records should be resolved correctly.

**Are you using NetBird Cloud?**

Self-hosted (v0.31.1 incl. relay as well as coturn)

**NetBird version**

0.31.1

**NetBird status -dA output:**

X

**Do you face any (non-mobile) client issues?**

2024-11-15T08:27:16+01:00 ERRO util/grpc/dialer.go:38: Failed to dial: dial: dial tcp: lookup netbird.our.domain: no such host

**Screenshots**

X

**Additional context**

**Output of Resolve-DnsName -Name  www.our.domain**
`Resolve-DnsName: www.unipi.technology : Daná operace se vrátila, protože vypršel časový limit. //Time exceeded`

**Output of: ping www.our.domain**
`Ping request could not find host www.our.domain. Please check the name and try again.`

**Output of: nslookup www.our.domain**
```
Server:  dns.google
Address:  8.8.8.8

Non-authoritative answer:
Name:    our.domain
Address:  correct IP address
Aliases:  www.our.domain
```

**Output of Get-DnsClientNrptPolicy**
```
Namespace                        : .ourdomain.local
QueryPolicy                      :
SecureNameQueryFallback          :
DirectAccessIPsecCARestriction   :
DirectAccessProxyName            :
DirectAccessDnsServers           :
DirectAccessEnabled              :
DirectAccessProxyType            : NoProxy
DirectAccessQueryIPsecEncryption :
DirectAccessQueryIPsecRequired   : False
NameServers                      : 10.220.255.254
DnsSecIPsecCARestriction         :
DnsSecQueryIPsecEncryption       :
DnsSecQueryIPsecRequired         : False
DnsSecValidationRequired         : False
NameEncoding                     : Utf8WithoutMapping

Namespace                        : .our.domain
QueryPolicy                      :
SecureNameQueryFallback          :
DirectAccessIPsecCARestriction   :
[...]

PR #7563, opened 2026-09-16
[client] Remove the empty GPO DNS policy store on Windows teardown

lixmal:
## Describe your changes

The group policy `DnsPolicyConfig` key survives the deletion of the last NRPT rule it held, and the client treats its presence as "group policy configures the NRPT", so an empty one left behind makes every later run write its rules there. Rules in that store only reach the resolver when the policy engine next applies DNS client policy, and a rule the client writes belongs to no GPO, so nothing schedules that application: both adding and removing a rule are held up by a minute or more, and for a removal that is a catch-all rule resolving every name over an interface that no longer exists. With the store absent the local one is authoritative and a change applies at once.

Older versions created that key on any machine with a DNS client policy, even when the NRPT was not managed by policy, so machines still carry one today. The state shows up as an empty rule list from `Get-DnsClientNrptRule` while `Get-DnsClientNrptPolicy` still returns the client's namespaces, alongside "detected GPO DNS policy configuration, using policy store" in the client log.

- Remove an empty group policy DNS policy store before deciding which store the rules go to, so a leftover one does not send this run's rules there
- Remove it again on teardown, once the rules of the session are gone
- Leave a store alone while it still holds a rule, ours or anybody else's
- Treat an absent interface value, or an absent interface key, as deleted, so a teardown that failed part way can be rerun

The order on startup is what decides the session: clearing the store first leaves the local one authoritative from the first rule onwards. Clearing it during a session would not, since the next rule written would recreate it.

Updates #5579. Follow-up to #3459, which narrowed the store detection but left the key itself behind.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

Internal registry cleanup on Windows with no user-facing configuration or behavior to document.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows DNS restoration by removing empty Group Policy DNS stores after managed rules are cleared.
  * Preserves policy stores containing remaining settings or third-party rules.
  * Logs cleanup failures without interrupting DNS restoration.
  * Repeated cleanup operations now complete successfully when registry values or interface keys are already absent.


ev5unleash:
This appears similar to a field case we are seeing, although our trigger is an unclean/hard shutdown rather than a normal disconnect.

Environment: Windows 10 Pro 25H2, build 26200, 64-bit; NetBird client 0.78.1; self-hosted deployment.

After a hard shutdown or sleep/resume:
- Resolve-DnsName, ping, browsers, Outlook, and mapped drives fail or time out.
- nslookup succeeds when querying an external resolver.
- Get-DnsClientNrptPolicy -Effective shows a catch-all Namespace "." pointing to 10.131.255.254, the NetBird local DNS proxy.
- Rebooting clears the condition.

This also appears related to #2895. Could you confirm whether #7563 covers stale NRPT state after an unclean shutdown, not only clean disconnect? A regression test covering hard-shutdown recovery with missing persisted rule state would be valuable.

theodorsm (approved):

PR #7597, opened 2026-09-20
[management] Trigger peer network-map update on DNS zone creation

### LINKED ISSUE #7594: Creating a DNS zone does not trigger a network-map update — connected peers don't receive it until their next full sync
### Describe the problem

On self-hosted 0.78.2, creating a DNS zone (with records) via `POST /api/dns/zones` succeeds, but already-connected peers in the zone's distribution groups never receive it: no DNS config update reaches the client — no new match domain is installed, and queries for the zone's names fall through to upstream resolvers. A `netbird down && netbird up` on a client delivers the zone immediately, and freshly enrolling peers get it in their first network map. Editing or deleting a zone propagates live to connected peers — only creation is silent.

### Cause (from source)

In `management/internals/modules/zones/manager/manager.go`, `UpdateZone` and `DeleteZone` both call `go m.accountManager.UpdateAccountPeers(...)` after saving, but `CreateZone` has no such call, so no network-map push happens for the create. (`zones/records/manager/manager.go` does push on record create/update/delete, but records embedded in the zone-create payload go through `CreateZone`.)

### To Reproduce

1. Self-hosted 0.78.2 (combined image); a connected peer P in group G, client DNS enabled.
2. `POST /api/dns/zones` with a zone distributed to G and one A record (zone domain == the record name, though the shape doesn't matter).
3. On P: the name does not resolve through the mesh, and the client log shows no DNS config change (no "added N Match domains").
4. Restart P's client (`netbird down && netbird up`) → the zone arrives and resolves correctly.

### Expected behavior

Zone creation pushes an account-peers update, exactly like zone update and delete do.

### Environment

- Self-hosted management 0.78.2 (combined server image)
- Clients 0.78.2 (macOS and Linux), client DNS enabled
- Not NetBird Cloud

Tyagiquamar:
## Describe your changes

When a DNS zone is created via \CreateZone\, \UpdateAccountPeers\ was not called, unlike in \UpdateZone\ and \DeleteZone\ (and sibling record management in \
ecords/manager\). As a result, connected peers did not receive network map updates reflecting the newly created DNS zone until their next full reconnect/sync.

This change adds \go m.accountManager.UpdateAccountPeers(ctx, accountID, types.UpdateReason{Resource: types.UpdateResourceZone, Operation: types.UpdateOperationCreate})\ upon successful creation of a DNS zone, matching \UpdateZone\, \DeleteZone\, and DNS records manager behavior.

Fixes #7594

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor
- [x] I ran and tested this change locally (I did not rely on CI to find out whether it works)

## Documentation
- [x] Documentation is **not needed** for this change (internal network map update propagation fix for newly created DNS zones, no user facing API or CLI change)

<a href="https://cubic.dev/pr/netbirdio/netbird/pull/7597?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Zone peer information updates for create, update, and delete operations now continue reliably even if the originating request is canceled.
  * Improved reliability of account peer updates following zone changes.


Rohmilchkaese:
Thanks for this. One pointer from a reproduction on v0.79.0 (details in #7594): the same `go m.accountManager.UpdateAccountPeers(ctx, …)` pattern on the request context sits in `management/internals/modules/zones/records/manager/manager.go` (create l.99, update l.158, delete l.205) and in `management/server/account.go:461`, and record create/update/delete fail to propagate for exactly the reason this PR fixes for zones. Either applying `context.WithoutCancel` there too, or detaching once inside `Controller.UpdateAccountPeers`, would close the whole family in one change.

Tyagiquamar:
Updated the implementation in ea12436 to address review feedback:
- Detached cancellation across all asynchronous \go UpdateAccountPeers(...)\ call sites (\zones/records/manager/manager.go\ and \management/server/account.go\) using \context.WithoutCancel(ctx)\.
- Preserved standard context cancellation semantics for synchronous callers while protecting fire-and-forget background peer updates from HTTP request termination.
- Added cancellation regression tests in \manager_test.go\ verifying record lifecycle operations (Create, Update, Delete) continue peer updates even after request context cancellation.

Tyagiquamar:
Pushed commit 2e28114 with the following updates:

1. Test Synchronization:
- Updated the permission-validation mocks across zone manager and record manager tests to return the actual request-derived cancellable context.
- Captured the context passed to `UpdateAccountPeers` in `updateCall` rather than capturing `callCtx.Err()` during callback dispatch.
- Deterministically synchronized the assertion flow: wait for the `UpdateAccountPeers` invocation to be received, cancel the request context via `reqCancel()`, and assert that the captured context remains active and not canceled.
- Validated cleanly with `go test -v -race` in Docker on both packages.

2. Regarding unbounded goroutine concerns with `context.WithoutCancel`:
- The background dispatch of `UpdateAccountPeers` is decoupled from the originating HTTP request lifetime by design. Across NetBird (`peer.go:1559`, `account.go:1762`, `controller.go:1063`), background peer updates consistently use `context.WithoutCancel(ctx)` to avoid aborting network-map distributions when the client HTTP connection closes.
- Inside `sendUpdateAccountPeers`, `GetAccountWithBackpressure` uses `requestbuffer.Buffer`, which coalesces requests and performs its store fetch under the server lifecycle context rather than request-derived deadlines. Adding arbitrary timeouts here would deviate from existing conventions.

PR #7601, opened 2026-09-21
[client] Fix tray left-click on macOS 26+ by routing through OpenMenu

### LINKED ISSUE #7312: Tray icon requires right-click on macOS 27 GoldenGate beta — left-click works on older macOS with same version
**Describe the problem**

On macOS 27 GoldenGate (beta, Darwin 27), the NetBird tray icon requires a **right-click** (or Ctrl+click) to open the menu. On other machines running the exact same NetBird version but on older macOS versions, the menu opens on **left-click** as expected.

This issue was also present on macOS 26.0.

Reinstalling NetBird does not fix the behavior.

**To Reproduce**

1. Install NetBird 0.77.1 on macOS 26.0 or macOS 27 GoldenGate beta
2. Click the tray icon with a regular left-click
3. Nothing happens — menu only opens on right-click / Ctrl+click

**Expected behavior**

Left-clicking the tray icon should open the NetBird menu, consistent with behavior on macOS 15 and earlier.

**Environment**

- NetBird version: 0.77.1
- macOS: 27 GoldenGate (beta), Darwin 27.0.0, Apple Silicon (arm64)
- Also reproduced on: macOS 26.0
- Confirmed working correctly (left-click) on same NetBird version on macOS 15

**Likely cause**

NetBird uses the [Fyne](https://fyne.io) framework for its UI. Apple changed how `NSStatusItem` handles click events starting in macOS 26 — the tray icon's click handler behavior may no longer match Fyne's current implementation, causing the menu to only respond to secondary (right) clicks.

This may require a Fyne dependency update or a macOS 26/27-specific fix in the tray icon implementation.

gautamkrishnar:
## Describe your changes

On macOS 27 (GoldenGate), left-clicking the tray icon did nothing while right-click still opened the menu. macOS 26 was unaffected.

Two issues combined to cause this:

**1. Deprecated `NSStatusItem` target/action stopped firing on macOS 27 (Wails bug)**
`NSStatusItem.setTarget`/`setAction` (deprecated since macOS 10.14) stopped working in macOS 27, so the action handler `statusItemClicked:` never fired on left-click. A fix has been submitted upstream: [wailsapp/wails#6148](https://github.com/wailsapp/wails/pull/6148). The netbirdio/wails fork will need to pull this fix and a dependency bump will follow.

**2. Broken event-monitor native tracking path for left-click (this PR)**
Even with the action handler firing, the Wails event-monitor was intercepting left-click, temporarily setting `statusItem.menu`, and relying on native `NSStatusBarButton` tracking to open the menu — a path that is also broken in macOS 27. This PR bypasses it by wiring an explicit `OnClick → OpenMenu()` handler in a new `tray_click_darwin.go`.

With `clickHandler` set, `systrayPreClickCallback` returns 0 for left-click (skipping the broken event-monitor path). The action handler fires and routes through `OpenMenu() → showMenu()`, which synthesizes a `mouseDown` to open the menu — the same path right-click already uses successfully on all macOS versions.

**Behavior on older macOS is unchanged**: the menu still opens on left-click via the same `showMenu` path that right-click uses, with no visual regression.

**History**: commit `c77e5ce` previously reverted an `OnClick → OpenMenu` attempt because `showMenu` then blocked the main GCD queue synchronously. The netbirdio/wails fork's `showMenu` now uses `dispatch_async`, so the action handler returns before the blocking `mouseDown` call — making this approach safe.

Fixes #7312

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor

## Documentation
Select exactly one:

No user-facing behavior change — left-click still opens the menu, same as before.


gautamkrishnar:
The docstring coverage warning flags `bindTrayClick` across the three changed files. Per the repository's AGENTS.md, function comments are mandatory only for **exported** functions — `bindTrayClick` is unexported. The new `tray_click_darwin.go` file has a full block comment explaining the why (the non-obvious constraint), which is the project's convention for private functions. No docstrings are needed here.

pappz:
@gautamkrishnar Thank you for the work on this issue! Because the fix depends on the Wails library, I recommend that you speed up the things. If you have motivation, create a PR in our forked repository. https://github.com/netbirdio/wails

gautamkrishnar:
@pappz i already opend a PR with them https://github.com/wailsapp/wails/pull/6148 waiting for their approval

gautamkrishnar:
Update, it looks like they already have a fix in their beta version. I am now checking locallly if the beta version fixes the issue with netbird.

pappz:
@gautamkrishnar Good to know! Could you send me the link regarding to the patch PR in Wails repo? I would like to sync the repos.

gautamkrishnar:
@pappz 

the fix by wailsapp tea is mentioned in https://github.com/wailsapp/wails/issues/6147#issuecomment-5768104022 I will test the latest version of the wailsapp locally and let you know.

If its working with netbird you can sync the fork to get the latest versons.

Issue #7334, opened 2026-08-27
Management omits persisted nameserver groups from peer network maps


Dzyubuk:
## Summary
In a self-hosted deployment, Management 0.77.1 persists an enabled conditional nameserver group and its peer-group assignments, but sends affected Windows peers a network map with an empty DNSConfig.
## Environment
- Management: netbirdio/management 0.77.1
- Client: Windows 0.77.1
- Self-hosted Management and Dashboard
## Configuration
- One enabled nameserver group with one resolver at port 53
- One match domain
- Distribution groups include the built-in All group and a custom group
- The test peer is confirmed in both distribution groups in the Management SQLite datastore
## Expected behavior
The test peer receives the configured nameserver group in its network map and resolves the configured match domain through that resolver.
## Actual behavior
After restarting the NetBird service, the peer receives a fresh network map but the DNS section is empty:
```json
"DNSConfig": {
  "ServiceEnable": false,
  "NameServerGroups": [],
  "CustomZones": []
}
```
The client reports dnsServers as an empty array. Its tunnel adapter has no DNS server and no NRPT rule. The configured match domain resolves through the local DNS provider and returns NXDOMAIN. A packet capture on the configured resolver receives no query from the peer.
## Persistence verification
The Management SQLite datastore confirms the nameserver group is enabled, has the expected resolver and match domain, and references the expected distribution-group IDs. The group_peers table confirms the test peer is a member of the built-in All group and the custom distribution group.
## Additional notes
- Recreating only the Management container after updating 0.77.0 to 0.77.1 did not change the result.
- Dashboard displays the nameserver group, its match domain, and its distribution groups correctly.
- A sanitized debug bundle is available for maintainers through a private channel; it is not attached publicly because it contains internal metadata.
## Request
Could maintainers identify why Management omits a persisted nameserver group from the generated peer network map, and advise any required server-side configuration or migration?

Dzyubuk:
Additional observation from a separate macOS peer on client 0.77.1:
- netbird status reports the current enabled conditional nameserver group.
- macOS system resolver state still contains duplicate supplemental resolvers for the same match domain pointing at the previous resolver address.
- This indicates stale OS DNS reconciliation after a nameserver-group resolver change, while the Windows peer in the report receives an empty DNSConfig from Management.
This may be a separate client-side cleanup issue, but it demonstrates inconsistent DNS-state delivery/application across peers.

hfzndev:
[...]

PR #7615, opened 2026-09-22
[management] preserve group references without public ids

Amilliox:
## Describe your changes

Component network maps use a group's public ID as the wire identifier. On upgraded accounts, a group can temporarily have an empty `public_id`. When more than one such group is present, they are encoded with the same empty wire ID and collapse into one entry during decode.

That can drop group membership references used by nameserver distribution groups, so a persisted nameserver group may disappear from the peer's DNS configuration even though the underlying group assignments are still present.

This change falls back to the stable internal group ID only when `PublicID` is empty. The same fallback is used both when encoding groups and when encoding references to those groups, keeping the wire representation consistent and unique.

A regression test covers two groups with missing public IDs and verifies that the target peer keeps its nameserver group after the component-envelope round trip.

Related to #7334.

The reported symptoms are consistent with this failure mode. I have not inspected the reporter's private SQLite database, so this PR intentionally limits the claim to the reproduced collision case.

### Checklist
- [x] Is it a bug fix
- [ ] It is a refactor
- [x] I ran and tested this change locally / in the fork CI
- [x] This change is a trivial fix, **OR** it links an issue the NetBird team agreed on beforehand. Changes to the public API, gRPC protocols, functionality behavior, CLI / service flags, or new features always need that agreement first.

> By submitting this pull request, you confirm that you have read and agree to the terms of the Contributor License Agreement.

## Documentation

Select exactly one:

- [x] Documentation is **not needed** for this change (internal bug fix; no user-facing API or configuration change)

### Docs PR URL

N/A


dmitri-netbird:
Thank you for the PR. Could you describe the circumstances when the issue (missing public_ids on groups)  has occurred, or provide steps to reproduce the issue?

